PR #4499, opened 2022-02-14
rt: make `JoinSet` unstable

carllerche:
There are still some open questions related to the API. Until they are
resolved, the type should be marked as unstable so releases are not
blocked.

Specifically, how can tasks spawned onto the `JoinSet` be canceled?
My inclination would be to return an `AbortHandle` from `JoinSet::spawn`.

### DIFF
--- a/tokio/src/macros/cfg.rs
+++ b/tokio/src/macros/cfg.rs
@@ -373,6 +373,16 @@ macro_rules! cfg_trace {
     };
 }
 
+macro_rules! cfg_unstable {
+    ($($item:item)*) => {
+        $(
+            #[cfg(tokio_unstable)]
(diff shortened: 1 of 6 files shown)


Darksonn (approved):

hawkw (approved):
We should probably also add a note in the `JoinSet` docs that it's unstable, and link to the documentation on unstable features, like the way we do for `task::Builder` etc:
https://github.com/tokio-rs/tokio/blob/8fb15da8f83278a51deb5514f998062dd3d21afe/tokio/src/task/builder.rs#L7-L9

Once that's added, this looks good to me!

hawkw:
@carllerche mind making the docs change i suggested so we can merge this? thanks!

Noah-Kennedy (approved):

PR #4500, opened 2022-02-15
deprecate reexports from std

nipunn1313:
Looking a bit into the history, they appear motivated in an earlier
less stable version of tokio when the types may have been changing
or weren't in std yet.

## Motivation

Reexports for the sake of it are not helpful (confusing actually)
from std. For new developers, it's not obvious that
tokio::time::Duration is the same as std::time::Duration. This indirection
may be useful for re-exporting a particular version of a dependent crate,
but for `std` it appears to be mainly confusing.

## Solution

Deprecate the re-exports of std. Upon the next major release, remove them.

### DIFF
--- a/tokio/src/macros/support.rs
+++ b/tokio/src/macros/support.rs
@@ -4,6 +4,10 @@ cfg_macros! {
     pub use crate::util::thread_rng_n;
 }
 
+// Re-export not needed since it's pulling from std
+#[deprecated(note = "Prefer importing std::future::Future directly")]
 pub use std::future::Future;
+#[deprecated(note = "Prefer importing std::pin::Pin directly")]
 pub use std::pin::Pin;
+#[deprecated(note = "Prefer importing std::task::Poll directly")]
 pub use std::task::Poll;
--- a/tokio/src/time/mod.rs
+++ b/tokio/src/time/mod.rs
@@ -109,6 +109,7 @@ pub use timeout::{timeout, timeout_at, Timeout};
 #[cfg(not(loom))]
 mod tests;
 
-// Re-export for convenience
+// Re-export not needed since it's pulling from std
+#[deprecated(note = "Prefer importing std::time::Duration directly")]
 #[doc(no_inline)]
 pub use std::time::Duration;


taiki-e:
This is intentional. 
https://github.com/tokio-rs/tokio/blob/9a3ce91ef54210d418318696dccdf0d14fe2b5fd/tokio/src/io/mod.rs#L202-L205

And the #[doc(no_inline)] makes it clear that it is a re-export from std. (At least for users reading the documentation)

time

tokio/src/macros/support.rs
```diff
@@ -4,6 +4,10 @@ cfg_macros! {
     pub use crate::util::thread_rng_n;
 }
 
+// Re-export not needed since it's pulling from std
+#[deprecated(note = "Prefer importing std::future::Future directly")]
 pub use std::future::Future;
+#[deprecated(note = "Prefer importing std::pin::Pin directly")]
 pub use std::pin::Pin;
+#[deprecated(note = "Prefer importing std::task::Poll directly")]
 pub use std::task::Poll;
```
taiki-e, line 13: Note that this module is not a public API

https://github.com/tokio-rs/tokio/blob/9a3ce91ef54210d418318696dccdf0d14fe2b5fd/tokio/src/lib.rs#L365-L371

hawkw:
The re-exports exist in modules that are named the same as modules in the standard library, such as `io` and `time`. Because these modules have the same name, they cannot both be imported --- for example, you cannot write
```rust
use tokio::io;
use std::io;
```
and then reference a name such as 
```rust
io::Result
```

The re-exports are intended to solve this issue. If you import only `tokio::io`, you can still refer to `io::Result` prefixed with `io`, because `tokio::io` re-exports it from `std`. 

This is the intention behind providing these re-exports, and why we only provide them in modules with names that collide with the names of modules in `std`. They were not intended to export a particular version of the standard library, because that isn't really possible.

nipunn1313:
Cool. Appreciate the explanation. It sounds like the motive here is to avoid having to rename imports when using code that uses both std::time::Duration and tokio::time

I had some coworkers find the reexport of Duration confusing as it seems like there is a separate implementation of Duration. I tend to agree. Renaming the conflicting import explicitly would avoid that confusion.

hopefully the no_inline is enough to help people figure it out

Noah-Kennedy:
It's also important to note that removing this would be a breaking change, and I'd argue deprecation would be as well.

nipunn1313:
Yep. Something like this would definitely have to be executed across a major release.

hawkw:
> It's also important to note that removing this would be a breaking change, and I'd argue deprecation would be as well.

deprecations shouldn't be considered a breaking change, i think; if you are denying deprecation warnings, you are essentially opting in to having your build broken in semver-compatible releases. actually *removing* the deprecated APIs would be breaking, though.

Noah-Kennedy:
Good point. I tend to think of these warnings primarily as an indicator of future removal though, rather than a discouragement tool.

Issue #4073, opened 2021-08-26
meta: Runtime metrics stabilization


carllerche:
Tracks the stabilization of runtime statistics.

RFC: #3845
PRs: #4043

## Roadmap

* [x] Release current implementation as unstable to crates.io (#4083)
* [x] Polish docs (remove TODO)
* [x] Add more counters
* [x] Write `tokio-metrics` providing a higher level api to be consumed that is easier to understand.
* [ ] Resolve open questions.
* [x] Validate design by evaluating users' experience reports.
    * [x] @Matthias247: [graph](https://media.discordapp.net/attachments/500336346770964480/893251797920129054/unknown.png?width=1272&height=866)

## Open questions

* Naming: Should the type be named `Metrics`, `Stats`, or `PerfCounters`.
* Should `RuntimeStats::workers()` return `&[WorkerStats]` or an iterator.
* Should there be a feature flag to enable stats explicitly?
  * Should there be a runtime Builder option to enable / disable stats.
* Builder API for more complex configurations, like histograms (#5685).
* `inc_budget_forced_yield_count` should become a per-worker metric.
* Should some current counters be lowered to internal counters?
  * steal_count
  * steal_operations
  * overflow_count

## Additional counters

* Duration between last two polls.
    *  There are open questions related to how this should be used.
* Worker queue depth
    * It is unclear how this should be tracked.

carllerche:
@LucioFranco @Matthias247 any opportunities to try using the metrics?

Matthias247:
I was on vacation and after that mostly block on other things. But I might be able to try this out this week.

I will however say upfront that I'll expect mostly to report back on the general accessor APIs and how integration into an application will look like. I think that putting `poll_count`/`steal_count`/etc on a dashboard will not be super useful for most people, because the numbers in itself have no significant meaning. They don't necessarily indicate that something is right or wrong. The not-yet-implemented timing metrics are more interesting, because they would indicate issues with code in tasks blocking too long. I will nevertheless check and see how the other metrics would look like.

carllerche:
We should add a counter tracking the number of "false-positive" runtime wakeups. This would be incremented when a worker wakes up without having any work to do.

eadventurous:
> The not-yet-implemented timing metrics are more interesting, because they would indicate issues with code in tasks blocking too long.

[...]

PR #4507, opened 2022-02-16
tokio: Add initial io driver metrics

LucioFranco:
### Work in progress

This adds the initial IO driver metrics.

Ref: https://github.com/tokio-rs/tokio/issues/4073

### Todo

- [x] tests
- [x] docs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 target
 Cargo.lock
+
+.cargo/config.toml
--- a/tokio/src/io/driver/mod.rs
+++ b/tokio/src/io/driver/mod.rs
@@ -14,9 +14,13 @@ pub(crate) use registration::Registration;
 mod scheduled_io;
 use scheduled_io::ScheduledIo;
 
+mod metrics;
+
 use crate::park::{Park, Unpark};
 use crate::util::slab::{self, Slab};
 use crate::{loom::sync::Mutex, util::bit};
+
+use metrics::IoDriverMetrics;
 
 use std::fmt;
 use std::io;
@@ -74,6 +78,8 @@ pub(super) struct Inner {
 
     /// Used to wake up the reactor from a call to `turn`.
     waker: mio::Waker,
+
+    metrics: IoDriverMetrics,
 }
 
 #[derive(Debug, Eq, PartialEq, Clone, Copy)]
@@ -130,6 +136,7 @@ impl Driver {
                 registry,
                 io_dispatch: allocator,
                 waker,
+                metrics: IoDriverMetrics::default(),
             }),
         })
     }
@@ -167,13 +174,17 @@ impl Driver {
         }
 
         // Process all the events that came in, dispatching appropriately
+        let mut ready_count = 0;
         for event in events.iter() {
             let token = event.token();
 
             if token != TOKEN_WAKEUP {
                 self.dispatch(token, Ready::from_mio(event));
+                ready_count += 1;
             }
         }
+
+        self.inner.metrics.incr_ready_count_by(ready_count);
 
         self.events = Some(events);
 
@@ -279,6 +290,25 @@ cfg_not_rt! {
     }
 }
 
+cfg_metrics! {
+   impl Handle {
+        // TODO: Remove this when handle contains `Arc<Inner>` so that we can return
+        // &IoDriverMetrics instead of using a closure.
+        //
+        // Related issue: https://github.com/tokio-rs/tokio/issues/4509
+        pub(crate) fn with_io_driver_metrics<F, R>(&self, f: F) -> Option<R>
+        where
+            F: Fn(&IoDriverMetrics) -> R,
+        {
+            if let Some(inner) = self.inner() {
+                Some(f(&inner.metrics))
+            } else {
+                None
+            }
+        }
+   }
+}
+
 impl Handle {
     /// Forces a reactor blocked in a call to `turn` to wakeup, or otherwise
     /// makes the next call to `turn` return immediately.
@@ -335,12 +365,18 @@ impl Inner {
         self.registry
             .register(source, mio::Token(token), interest.to_mio())?;
 
+        self.metrics.incr_fd_count();
+
         Ok(shared)
     }
 
     /// Deregisters an I/O resource from the reactor.
     pub(super) fn deregister_source(&self, source: &mut impl mio::event::Source) -> io::Result<()> {
-        self.registry.deregister(source)
+        self.registry.deregister(source)?;
+
+        self.metrics.dec_fd_count();
+
+        Ok(())
     }
 }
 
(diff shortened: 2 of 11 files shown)


tokio/src/io/driver/mod.rs
```diff
@@ -197,6 +201,8 @@ impl Driver {
             return;
         }
 
+        self.inner.metrics.incr_ready_count();
```
LucioFranco, line 204: Do we want to do this at the top of this method to capture invalid tokens as well? or do we want this to represent only valid wakeups?

tokio/src/runtime/metrics/runtime.rs
```diff
@@ -446,4 +446,14 @@ impl RuntimeMetrics {
     pub fn worker_local_queue_depth(&self, worker: usize) -> usize {
         self.handle.spawner.worker_local_queue_depth(worker)
     }
+
+    pub fn io_driver_fd_count(&self) -> Option<u64> {
```
LucioFranco, line 450: These return options do we want to do this if the io driver has dropped? Or should we just panic?

.gitignore
```diff
@@ -1,2 +1,4 @@
 target
 Cargo.lock
+
+.cargo/config.toml
```
carllerche, line 4:
```suggestion
.cargo/config.toml

```

tokio/src/io/driver/mod.rs
carllerche, line 204: We probably want to batch this. Increment the number of ready per tick.

tokio/src/io/driver/mod.rs
```diff
@@ -298,6 +304,17 @@ impl Handle {
     pub(super) fn inner(&self) -> Option<Arc<Inner>> {
         self.inner.upgrade()
     }
+
+    pub(crate) fn with_io_driver_metrics<F, R>(&self, f: F) -> Option<R>
```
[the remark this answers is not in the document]
carllerche: Why not return the `Arc`? An `upgrade()` already does a ref count bump. Also, we probably should separately get rid of the weak ref inc.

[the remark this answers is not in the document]
carllerche: At least leave a TODO, maybe open an issue to track getting rid of `Weak`?

tokio/src/runtime/metrics/io.rs
```diff
@@ -0,0 +1,21 @@
...
+
+#[derive(Default)]
+pub(crate) struct IoDriverMetrics {
+    pub(super) fd_count: AtomicU64,
+    pub(super) ready_count: AtomicU64,
+}
+
+impl IoDriverMetrics {
+    pub(crate) fn incr_fd_count(&self) {
+        self.fd_count.fetch_add(1, Relaxed);
```
carllerche, line 11: Since there is only one "producer" of metrics instead of `fetch_add`, i would do a load / store combo. It is cheaper.

tokio/src/runtime/metrics/runtime.rs
carllerche: Default to zero. If it is `None`, the runtime is shutdown so there are no more registered FDs.

carllerche: Oh, I see the issue is that the counter should always increment...yeah, we don't want `Option` in the final API...

carllerche: I would start w/ `unwrap_or(0)` for now, but track a TODO to fix before stabilizing the API.

tokio/src/io/driver/mod.rs
```diff
@@ -279,6 +290,25 @@ cfg_not_rt! {
     }
 }
 
+cfg_metrics! {
+   impl Handle {
+        // TODO: Remove this when handle contains `Arc<Inner>` so that we can return
+        // &IoDriverMetrics instead of using a closure.
+        //
+        // Related issue: https://github.com/tokio-rs/tokio/issues/4509
+        pub(crate) fn with_io_driver_metrics<F, R>(&self, f: F) -> Option<R>
```
carllerche, line 299: Fwiw, calling `self.inner()` does an arc bump anyway. I probably wouldn't bother w/ the closure, but it isn't a bit deal.

tokio/src/lib.rs
```diff
@@ -483,7 +483,7 @@ pub(crate) use self::doc::winapi;
 
 #[cfg(all(not(docsrs), windows, feature = "net"))]
 #[allow(unused)]
-pub(crate) use ::winapi;
+pub(crate) use winapi;
```
carllerche, line 486: what is this change?

LucioFranco: I think rustfmt did it? I didn't manually change it but I don't think it affects anything since this got updated in 2018 edition iirc

tokio/src/io/driver/mod.rs
LucioFranco: Yeah, plan to refactor with removal of weak from io driver handle.

tokio/src/runtime/metrics/io.rs
```diff
@@ -0,0 +1,29 @@
+#![cfg_attr(not(feature = "net"), allow(dead_code))]
+
+use std::sync::atomic::{AtomicU64, Ordering::Relaxed};
+
+#[derive(Default)]
+pub(crate) struct IoDriverMetrics {
+    pub(super) fd_count: AtomicU64,
```
carllerche, line 7: Is it worth tracking registered & deregistered separately? Right now, if an FD is registered then immediately deregistered, this will show up as the counter not moving. What is the value in tracking outstanding FDs? If registered & deregistered ops are tracked separately, churn will be detectable.

Separately, do we want to track `fd_count` or punt that to `tracing`? I am not sure. Thoughts @hawkw

carllerche (commented):
The structure of the PR looks good. I left some comments inline.

tokio/src/runtime/metrics/io.rs
LucioFranco: Wouldn't this catch things like leaked fds etc? I think there is value in tracking this internally to the driver in addition to what ever tracing can bring?

carllerche: tracking register & deregister separately would be a super set. Total fds == registers - deregisters.

tokio/src/runtime/metrics/runtime.rs
```diff
@@ -447,3 +447,65 @@ impl RuntimeMetrics {
...
+        ///
+        /// #[tokio::main]
+        /// async fn main() {
+        ///     let metrics = Handle::current().metrics();
+        ///
+        ///     let n = metrics.io_driver_fd_count();
+        ///     println!("{} fds currently registered with the runtime's I/O driver.", n);
+        /// }
+        /// ```
+        pub fn io_driver_fd_count(&self) -> u64 {
```
carllerche, line 469: Instead of returning the count here, i would just expose the counter values. So: `fd_registered_count` and `fd_deregistered_count`.

LucioFranco: Updated

carllerche (approved):
lgtm thanks.

PR #4511, opened 2022-02-16
Include more documentation for thread_pool/worker

rcmgleite:
## Motivation

After spending some time reading the internals of the runtime I thought that having more comments on the code around specific behaviors would be helpful to others.


Darksonn (approved):
Thanks.

PR #4525, opened 2022-02-22
remove `example`'s `Cargo.toml` dependencies path,make example works.

zys864:
## Motivation

The poll request [#4521](https://github.com/tokio-rs/tokio/runs/5273189299?check_suite_focus=true) CI `cargo test --workspace --all-features` failed,
## Solution
remove `examples/Cargo.toml`  path dependencies


djc:
Thanks!

PR #4538, opened 2022-02-24
task: implement `JoinMap`

hawkw:
Depends on #4530.

## Motivation

In many cases, it is desirable to spawn a set of tasks associated with
keys, with the ability to cancel them by key. As an example use case for
this sort of thing, see Tower's [`ReadyCache` type][1].

Now that PR #4530 adds a way of cancelling tasks in a
`tokio::task::JoinSet`, we can implement a map-like API based on the
same `IdleNotifiedSet` primitive.

## Solution

This PR adds an implementation of a `JoinMap` type to `tokio::task`,
using the `IdleNotifiedSet` that backs `JoinMap`, and a `HashMap` of the
`AbortHandle`s added in #4530. Keys are stored in both the `HashMap`
*and* alongside the task's `JoinHandle` in the `IdleNotifiedSet` in
order to identify tasks as they complete. This allows returning the key
in `join_one` regardless of whether a task was cancelled, panicked, or
completed successfully, and removing the corresponding `AbortHandle`
from the `HashMap`. Of course, this requires keys to be `Clone`.

 Overall, I think the way this works is pretty straightforward; much of
 this PR is just API boilerplate to implement the union of applicable
 APIs from `JoinSet` and `HashMap`.

Individual tasks can be aborted by key using the `JoinMap::abort`
method, and a set of tasks whose key match a given predicate can be
aborted using `JoinMap::abort_matching`.

When tasks complete, `JoinMap::join_one` returns their associated key
alongside the output from the spawned future, or the key and the
`JoinError` if the task did not complete successfully.

## Bikesheds & Future Work

* I used a mixture of `HashMap` and `JoinSet` naming. Tasks are
`spawn`ed rather than `insert`ed, and `abort`ed rather than `remove`d,
but other methods use map-inspired naming. Open to changing this, of
course!

* We also could change the `join_one` method to only return the future's
output, and not the task's key, but I thought it was nicer to do this
rather than not. In particular, this allows the user to determine
_which_ task panicked or was cancelled when a `JoinError` is returned,
which seems kind of important to me.

[1]: https://github.com/tower-rs/tower/blob/master/tower/src/ready_cache/cache.rs

### DIFF
--- a/tokio-util/src/task/join_map.rs
+++ b/tokio-util/src/task/join_map.rs
@@ -1,0 +1,257 @@
+use std::{
+    borrow::Borrow,
+    collections::hash_map::{HashMap, RandomState},
+    fmt,
+    future::Future,
+    hash::{BuildHasher, Hash},
+};
+use tokio::{
+    runtime::Handle,
+    task::{AbortHandle, JoinError, JoinSet, LocalSet},
+};
+
+pub struct JoinMap<K, V, S = RandomState> {
+    aborts: HashMap<K, AbortHandle, S>,
+    joins: JoinSet<(K, V)>,
+}
+
+impl<K, V> JoinMap<K, V> {
+    /// Create a new `JoinMap`.
+    pub fn new() -> Self {
+        Self::with_hasher(RandomState::new())
+    }
+}
+
+impl<K, V, S> JoinMap<K, V, S> {
+    /// Creates an empty `JoinMap` which will use the given hash builder to hash
+    /// keys.
+    ///
+    /// The created map has the default initial capacity.
+    ///
+    /// Warning: `hash_builder` is normally randomly generated, and
+    /// is designed to allow HashMaps to be resistant to attacks that
+    /// cause many collisions and very poor performance. Setting it
+    /// manually using this function can expose a DoS attack vector.
+    ///
+    /// The `hash_builder` passed should implement the [`BuildHasher`] trait for
+    /// the HashMap to be useful, see its documentation for details.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use std::collections::HashMap;
+    /// use std::collections::hash_map::RandomState;
+    ///
+    /// let s = RandomState::new();
+    /// let mut map = HashMap::with_hasher(s);
+    /// map.insert(1, 2);
+    /// ```
+    #[inline]
+    pub fn with_hasher(hash_builder: S) -> Self {
+        Self {
+            aborts: HashMap::with_hasher(hash_builder),
+            joins: JoinSet::new(),
+        }
+    }
+
+    /// Returns the number of tasks currently in the `JoinMap`.
+    pub fn len(&self) -> usize {
+        self.joins.len()
+    }
+
+    /// Returns whether the `JoinMap` is empty.
+    pub fn is_empty(&self) -> bool {
+        self.joins.is_empty()
+    }
+}
+
+impl<K, V, S> JoinMap<K, V, S>
+where
+    K: Hash + Eq + Clone + 'static,
+    V: 'static,
+    S: BuildHasher,
+{
+    /// Spawn the provided task on the `JoinMap`, returning an [`AbortHandle`]
+    /// that can be used to remotely cancel the task.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if called outside of a Tokio runtime.
+    ///
+    /// [`AbortHandle`]: crate::task::AbortHandle
+    pub fn spawn<F>(&mut self, key: K, task: F)
+    where
+        F: Future<Output = V>,
+        F: Send + 'static,
+        K: Send,
+        V: Send,
+    {
+        let task = self.joins.spawn(mk_task(&key, task));
+        self.insert(key, task)
+    }
+
+    /// Spawn the provided task on the provided runtime and store it in this
+    /// `JoinMap` returning an [`AbortHandle`] that can be used to remotely
+    /// cancel the task.
+    ///
+    /// [`AbortHandle`]: crate::task::AbortHandle
+    pub fn spawn_on<F>(&mut self, key: K, task: F, handle: &Handle)
+    where
+        F: Future<Output = V>,
+        F: Send + 'static,
+        K: Send,
+        V: Send,
+    {
+        let task = self.joins.spawn_on(mk_task(&key, task), handle);
+        self.insert(key, task)
+    }
+
+    /// Spawn the provided task on the current [`LocalSet`] and store it in this
+    /// `JoinMap`, returning an [`AbortHandle`]  that can be used to remotely
+    /// cancel the task.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if it is called outside of a `LocalSet`.
+    ///
+    /// [`LocalSet`]: crate::task::LocalSet
+    /// [`AbortHandle`]: crate::task::AbortHandle
+    pub fn spawn_local<F>(&mut self, key: K, task: F)
+    where
+        F: Future<Output = V>,
+        F: 'static,
+    {
+        let task = self.joins.spawn_local(mk_task(&key, task));
+        self.insert(key, task);
+    }
+
+    /// Spawn the provided task on the provided [`LocalSet`] and store it in
+    /// this `JoinMap`, returning an [`AbortHandle`] that can be used to
+    /// remotely cancel the task.
+    ///
+    /// [`LocalSet`]: crate::task::LocalSet
+    /// [`AbortHandle`]: crate::task::AbortHandle
+    pub fn spawn_local_on<F>(&mut self, key: K, task: F, local_set: &LocalSet)
+    where
+        F: Future<Output = V>,
+        F: 'static,
+    {
+        let task = self.joins.spawn_local_on(mk_task(&key, task), local_set);
+        self.insert(key, task)
+    }
+
+    fn insert(&mut self, key: K, abort: AbortHandle) {
+        if let Some(prev) = self.aborts.insert(key, abort) {
+            prev.abort();
+        }
+    }
+
+    /// Waits until one of the tasks in the set completes and returns its output.
+    ///
+    /// Returns `None` if the set is empty.
+    ///
+    /// # Cancel Safety
+    ///
+    /// This method is cancel safe. If `join_one` is used as the event in a `tokio::select!`
+    /// statement and some other branch completes first, it is guaranteed that no tasks were
+    /// removed from this `JoinMap`.
+    pub async fn join_one(&mut self) -> Result<Option<(K, V)>, JoinError> {
+        match self.joins.join_one().await {
+            Ok(Some((key, val))) => {
+                self.aborts.remove(&key);
+                Ok(Some((key, val)))
+            }
+            // XXX(eliza): should have a way to clean up "dead" abort handles
+            // when a task panics or is cancelled by a runtime dropping, etc...
+            res => res,
+        }
+    }
+
+    /// Aborts all tasks and waits for them to finish shutting down.
+    ///
+    /// Calling this method is equivalent to calling [`abort_all`] and then calling [`join_one`] in
+    /// a loop until it returns `Ok(None)`.
+    ///
+    /// This method ignores any panics in the tasks shutting down. When this call returns, the
+    /// `JoinMap` will be empty.
+    ///
+    /// [`abort_all`]: fn@Self::abort_all
+    /// [`join_one`]: fn@Self::join_one
+    pub async fn shutdown(&mut self) {
+        self.abort_all();
+        while self.join_one().await.transpose().is_some() {}
+    }
+
+    pub fn abort<Q: ?Sized>(&mut self, key: &Q) -> bool
+    where
+        Q: Hash + Eq,
+        K: Borrow<Q>,
+    {
+        match self.aborts.remove(key) {
+            Some(task) => {
+                task.abort();
+                true
+            }
+            None => false,
+        }
+    }
+
+    pub fn contains_task<Q: ?Sized>(&mut self, key: &Q) -> bool
+    where
+        Q: Hash + Eq,
+        K: Borrow<Q>,
+    {
+        self.aborts.contains_key(key)
+    }
+}
+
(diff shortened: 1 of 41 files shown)


hawkw:
Note that I still need to add some tests + more examples, but I wanted to go ahead and put the PR up so we could agree on the API.

tokio-util/src/task/join_map.rs
```diff
@@ -0,0 +1,499 @@
...
+                // nice if we could just look up the individiual task and remove
+                // *it* from the map. but, we don't have the key in this case.
+                // perhaps we could instead add the ability to compare
+                // `AbortHandle`s and `JoinHandle`s to see if they correspond to
+                // the same task, and change to some kind of map type allowing
+                // inverse lookups...but that would also mean adding `Hash` and `Eq`
+                // commitments to the `JoinHandle`/`AbortHandle` APIs, which i'm not
+                // sure if we want to do...
+                self.aborts.retain(|_, task| task.is_active());
+                Err(e)
```
Darksonn, line 280: I don't like this either.

hawkw:
Okay, after thinking about it a bit, I've rewritten the implementation to avoid the O(_n_) `HashMap::retain` call when tasks are cancelled. The new implementation moves `JoinMap` into `tokio::task`, and changes it to use the same `IdleNotifiedSet` primitive as `JoinSet` directly, rather than wrapping a `JoinSet`. This way, we can just store the keys in the `IdleNotifiedSet`, along with the `JoinHandle`s, rather than having to move them into the task. This means that we now *always* get a key back when a task completes, regardless of whether that task was cancelled, panicked, or completed successfully.

There are a couple good implications of always getting a key back:

- It lets us change the API to always return the key in `join_one`. This way, the user can determine _which_ task completed, even if is cancelled or panics. This was not previously possible with the old API.

- Because we always get the key back, we can always perform an O(1) `HashMap::remove` to clear out abort handles for dead tasks, rather than an O(_n_) `HashMap::retain`. This significantly improves performance when tasks panic or are cancelled.

- We no longer need to expose task status APIs on `AbortHandle`, which we may not want to commit to as part of implementing `JoinMap`.

The tradeoff is that this code now has to live in `tokio` rather than `tokio-util`. I think that's probably acceptable, though.

tokio-util/src/task/join_map.rs
hawkw: c4f6b7ff263d5681f82241f69f0360de74eb1602 gets rid of this.

tokio/src/task/join_map.rs
```diff
@@ -0,0 +1,566 @@
...
+    /// A hash map of keys to `AbortHandle`s, used to cancel tasks by key.
+    ///
+    /// Note: this currently stores an `Option<AbortHandle>`, in order to
+    /// implement the `abort_matching` method without adding a temporary
+    /// allocation. Because we need to take the `AbortHandle`s by value to abort
+    /// them, we have to take them out of the map somehow. We cannot do that
+    /// while iterating without also removing the tasks that are _not_ being
+    /// aborted, which would then require that we store those K-V pairs someplace
+    /// and then put them back afterwards (requiring an allocation). This seems
+    /// not great.
```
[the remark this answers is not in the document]
hawkw: we certainly can; I just thought it was better for the `AbortHandle` API to take `self`, because --- unlike a `JoinHandle` --- there's nothing else an `AbortHandle` can do once the task has been aborted. it might be better to just change it to take `&self`, though.

tokio/src/task/join_map.rs
```diff
@@ -0,0 +1,566 @@
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use tokio_util::task::JoinMap;
+    /// let mut map: JoinMap<&str, i32> = JoinMap::with_capacity(10);
+    /// ```
+    #[inline]
+    #[must_use]
+    pub fn with_capacity(capacity: usize) -> Self {
```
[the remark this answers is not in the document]
hawkw: because i forgot :) i'll add it.

i noticed that std marks all collection constructors with `#[must_use]` and thought we should consider doing that as well.

tokio/src/task/join_map.rs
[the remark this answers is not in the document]
Darksonn: Taking `&self` is more convenient for the caller. E.g. it allows using it from the destructor of a struct without wrapping it in an `Option`.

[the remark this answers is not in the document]
hawkw: Yeah, I've changed it to do that. I think it's the right call.

hawkw:
I still need to write some tests for this, btw.

tokio/src/task/join_map.rs
```diff
@@ -0,0 +1,540 @@
...
+        self.joins.for_each(|(_, jh)| jh.abort());
+    }
+
+    /// Removes all tasks from this `JoinMap` without aborting them.
+    ///
+    /// The tasks removed by this call will continue to run in the background even if the `JoinMap`
+    /// is dropped. They may still be aborted by key.
+    pub fn detach_all(&mut self) {
+        self.joins.drain(drop);
+    }
```
Darksonn, line 511: This should also clear the map.

tokio/src/task/join_map.rs
```diff
@@ -0,0 +1,540 @@
...
+    /// ```
+    #[inline]
+    pub fn capacity(&self) -> usize {
+        self.aborts.capacity()
+    }
+}
+
+impl<K, V, S> JoinMap<K, V, S>
+where
+    K: Hash + Eq + Clone + 'static,
```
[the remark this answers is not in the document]
hawkw: I considered something like that, but I wasn't sure if it was worth introducing the additional allocation for the `Arc` --- in the current approach, if there's already capacity in the map, the only allocation is for actually spawning the future. I think which approach is better really depends on how expensive the keys are to clone. If they're relatively cheap, like an `SocketAddr` or something, we may as well just copy the bytes. If keys are `String`s or something, the `Arc` is probably worth it. 

We could also just say something in the docs that keys will always be cloned, and encourage users whose keys are expensive to introduce an `Arc`? I could be convinced either way here, though.

tokio/src/task/join_map.rs
```diff
@@ -0,0 +1,540 @@
...
+    /// # Panics
+    ///
+    /// This method panics if called outside of a Tokio runtime.
+    ///
+    /// [`AbortHandle`]: crate::task::AbortHandle
+    pub fn spawn<F>(&mut self, key: K, task: F)
+    where
+        F: Future<Output = V>,
+        F: Send + 'static,
+        K: Send,
```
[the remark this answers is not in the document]
hawkw: ah, you're right --- that was an artifact of the previous implementation where keys were moved into the spawned future.

tokio/src/task/join_map.rs
hawkw: I wasn't sure --- this way, the user can still abort those tasks by key, even if they have been detached. Is this something we want, or not?

tokio/src/task/join_map.rs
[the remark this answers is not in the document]
hawkw: Oh, I see, you meant re-using the `Arc` created by `IdleNotifiedSet`. We could do that, but I'm a little uncomfortable making changes to the internal `IdleNotifiedSet` API that expose the ability to violate its invariants just for this. Hmm.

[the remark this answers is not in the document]
hawkw: > The `IdleNotifiedSet` would indeed require some changes, and I'm not sure exactly what they should be.

Yeah, I'm experimenting with this now to see how practical it is. I'm pretty sure there isn't a way to do it that doesn't require _some_ unsafe code in `JoinMap`, unless we modify `IdleNotifiedSet` enough that its invariants change...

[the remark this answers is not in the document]
hawkw: okay, i have a working --- although quite ugly --- implementation of this idea: https://github.com/tokio-rs/tokio/compare/eliza/join-map...eliza/join-map-arc?expand=1

if we want to move forward with that, i can clean it up and merge it into this PR.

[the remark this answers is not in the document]
Darksonn: I don't think the thing I described earlier requires direct use of hashbrown. If we are ok with that, then you can do [this][1] instead, storing the key in the map only, and not in the `IdleNotifiedSet`.

[1]: https://users.rust-lang.org/t/caching-a-hash-value/72466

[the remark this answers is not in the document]
hawkw: > I don't think the thing I described earlier requires direct use of hashbrown. If we are ok with that, then you can do [this](https://users.rust-lang.org/t/caching-a-hash-value/72466) instead, storing the key in the map only, and not in the `IdleNotifiedSet`.

@Darksonn If I understand what you're describing correctly, I believe any scheme where entries in a map are looked up using a stored hash value will require a direct dependency on `hashbrown` --- looking up map entries by hash requires the `RawEntry` API. `std::collections::HashMap` exposes this only on nightly, but it's a stable part of the `hashbrown` API.

hawkw:
Okay, I've rewritten the implementation again based on Alice's comments in https://github.com/tokio-rs/tokio/pull/4538#discussion_r817633617. Would love another look!

tokio/src/task/join_map.rs
```diff
@@ -0,0 +1,782 @@
...
+
+        false
+    }
+
+    /// keys all tasks with keys matching `predicate`.
+    ///
+    /// `predicate` is a function called with a reference to each key in the
+    /// map. If it returns `true` for a given key, the corresponding task will
+    /// be cancelled.
+    // XXX(eliza): do we want to consider counting the number of tasks aborted?
```
Noah-Kennedy, line 424: We should probably clean this comment up a bit

Noah-Kennedy (changes requested):
I think one comment needs cleaned up a bit, otherwise this looks good imo

tokio/src/task/join_map.rs
hawkw: just to clarify, do you just mean removing the `// XXX` comment, or are you suggesting changes to the RustDoc comment as well?

tokio/src/task/join_map.rs
```diff
@@ -0,0 +1,782 @@
...
+    ///    the key associated with the task.
+    ///  * `Some((key, Err(err))` if one of the tasks in this JoinMap` has
+    ///    panicked or been aborted. `key` is the key associated  with the task
+    ///    that panicked or was aborted.
+    ///  * `None` if the `JoinMap` is empty.
+    pub async fn join_one(&mut self) -> Option<(K, Result<V, JoinError>)> {
+        crate::future::poll_fn(|cx| self.poll_join_one(cx)).await
+    }
+
+    /// keys all tasks and waits for them to finish shutting down.
```
nylonicious, line 377:
```suggestion
    /// Aborts all tasks and waits for them to finish shutting down.
```

tokio/src/task/join_map.rs
```diff
@@ -0,0 +1,782 @@
...
+            entry.with_value_and_context(|(_actual_hash, jh), _| {
+                jh.abort();
+            });
+            return true;
+        }
+
+        false
+    }
+
+    /// keys all tasks with keys matching `predicate`.
```
nylonicious, line 419:
```suggestion
    /// Aborts all tasks with keys matching `predicate`.
```

tokio/src/task/join_map.rs
```diff
@@ -0,0 +1,782 @@
...
+        hasher.finish()
+    }
+}
+
+impl<K, V, S> JoinMap<K, V, S>
+where
+    K: 'static,
+    V: 'static,
+{
+    /// keys all tasks on this `JoinMap`.
```
nylonicious, line 617:
```suggestion
    /// Aborts all tasks on this `JoinMap`.
```

tokio/src/task/join_map.rs
Noah-Kennedy: The XXX

nylonicious (changes requested):
3 small typos

tokio/src/task/join_map.rs
Noah-Kennedy: This was intended for PR reviewers and to be removed before merge if I understand correctly?

hawkw: correct. after thinking about it a bit, I don't think it's really necessary for us to do this, as the user-provided predicate can always count aborted tasks if desired...

hawkw:
@nylonicious thanks for catching those --- that was a find-and-replace mistake, I definitely wouldn't have noticed it otherwise! :)

Noah-Kennedy (approved):

hawkw:
Hmm, the Hyper CI job is failing, on Windows only: https://github.com/tokio-rs/tokio/runs/5514991976?check_suite_focus=true. I'm gonna restart it.

carllerche:
The patch itself looks good. Others have commented there.

I have already expressed my hesitation with adding `JoinMap` to Tokio proper in Discord. I will restate some here.

First, adding an explicit `JoinMap` requires us to match a "map" API. Which one? Should we copy the stdlib `HashMap`? That would be the obvious API to copy, but it has limitations. First, `K: Borrow<Q>` has limitations (I won't rehash here, but this is one reason why `IndexMap` moved away from this`. What happens when the user hits these limitations or they are missing "Map" apis that we did not implement on `JoinMap` or they want to use something like `IndexMap` anyway? How can they solve these?

Separetely, having two types means the user has to pick between them. Do they pick `JoinSet` or `JoinMap`? It adds a new decision for them to make and the docs need to have a clear explanation of when to use each one.

If `JoinSet` generates an "identifier" with each spawned task. We can expose the identifier on the AbortHandle. We also can provide the identifier when a value is joined on `JoinSet` (e.g. add a `join_one_with_id()` method). These changes would let the user implement `JoinMap` logic themselves using whatever map structure they want, either std::HashMap or `IndexMap`.

My vote is to hold back on `JoinMap` and add the id concept to `JoinSet`. We can provide `JoinMap` in `tokio-util`.

That said, if everyone else thinks I am wrong, I wont force the issue. Worst case is, we deprecate `JoinMap` at a later time if we think it sin't the right option post stabilization.

nylonicious (approved):

PR #4541, opened 2022-02-25
Improve basic_scheduler block_on loop

jon-chuang:
## Motivation
Firstly, if result is available from the thread-local block_on, we should return `Ready(out)`, not assign it to Core even if it is available. That's a performance optimization

## Solution
Improve


jon-chuang:
Actually this is a mistake. If a thread is notified, it must take control of core, so that it can notify the next thread in queue. Else, we get deadlock.

Issue #4554, opened 2022-03-02
tokio can't listen for signals between SIGRTMIN and SIGRTMAX


weisbrja:
**Version**
`tokio v1.17.0`

**Platform**
`Linux thinkpad 5.16.11-arch1-1 #1 SMP PREEMPT Thu, 24 Feb 2022 02:18:20 +0000 x86_64 GNU/Linux`

**Description**
I'm trying to listen to signals between `libc::SIGRTMIN()` and `libc::SIGRTMAX()`, which should be possible according to the POSIX standard (and it worked with the `signal-hook` crate).

I tried this code:

```rust
use tokio::signal::unix::{signal, SignalKind};

fn main() {
    let _signal = signal(SignalKind::from_raw(libc::SIGRTMIN() + 3)).unwrap();
}
```

And it panicked with "signal too large".

I tracked the error down with `gdb`, and the reason it fails, is because of this check in `tokio/src/signal/unix.rs`:

```rust
fn signal_enable(signal: SignalKind, handle: &Handle) -> io::Result<()> {
    // ...

    let globals = globals();
    let siginfo = match globals.storage().get(signal as EventId) {
        Some(slot) => slot,
        None => return Err(io::Error::new(io::ErrorKind::Other, "signal too large")),
    };

    // ...
}
```

weisbrja:
I dug a bit in the source code and I think I found the issue.
The number of unix signals is defined in `tokio/src/signal/unix.rs`:
```rust
// Number of different unix signals
// (FreeBSD has 33)
const SIGNUM: usize = 33;
```
However, that is simply not true, because the POSIX standard allows for signals up to `SIGRTMAX`, which is greater than 33.

My proposed solution would be just replacing this:

```rust
// Number of different unix signals
// (FreeBSD has 33)
const SIGNUM: usize = 33;

impl Init for OsStorage {
    fn init() -> Self {
        (0..SIGNUM).map(|_| SignalInfo::default()).collect()
    }
}
```

With this:

```rust
impl Init for OsStorage {
    fn init() -> Self {
        (0..=libc::SIGRTMAX()).map(|_| SignalInfo::default()).collect()
    }
}
```

Any thoughts on this?

PR #4555, opened 2022-03-03
tokio: add support for signals up to SIGRTMAX

weisbrja:
The POSIX standard not only supports "reliable signals", but also
"real-time signals". This commit adds support for the latter.

Fixes: #4554

### DIFF
--- a/tokio/src/signal/registry.rs
+++ b/tokio/src/signal/registry.rs
@@ -237,7 +237,7 @@ mod tests {
     #[test]
     fn record_invalid_event_does_nothing() {
         let registry = Registry::new(vec![EventInfo::default()]);
-        registry.record_event(42);
+        registry.record_event(1302);
     }
 
     #[test]
--- a/tokio/src/signal/unix.rs
+++ b/tokio/src/signal/unix.rs
@@ -32,9 +32,7 @@ impl Init for OsStorage {
         #[cfg(target_os = "linux")]
         let possible = 0..=libc::SIGRTMAX();
 
-        possible
-            .map(|_| SignalInfo::default())
-            .collect()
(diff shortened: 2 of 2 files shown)


weisbrja:
I don't know, why the `libc` crate doesn't seem to have `SIGRTMAX()` on FreeBSD, as it works on my Linux machine.

ipetkov:
Not sure why the BSD equivalent isn't present in libc, probably because no one has needed it or bothered to contribute the change upstream...

Given that we're currently missing support for real-time signals on all platforms, I'd be okay with adding support for Linux systems now and keeping the BSDs cfg'd to the current value. We can open a separate follow up issue to add support in the future

weisbrja:
I tried to implement conditional compilation.

weisbrja:
And I also noticed that it is wrong to include `SIGRTMAX`, because the range starts at 0 and there is no such signal.

ipetkov:
> And I also noticed that it is wrong to include `SIGRTMAX`, because the range starts at 0 and there is no such signal.

Is SIGRTMAX supposed to include the highest such signal? My reading of the manual is that it is:

> The range SIGRTMIN through SIGRTMAX inclusive shall include at least {RTSIG_MAX} signal numbers.

ipetkov:
Yep, that sounds correct. Tokio doesn't shift the numbers around to match the index on the vector, the 0th entry is just not used

weisbrja:
But isn't it possible to just create a signal 0 then?

weisbrja:
And I guessed that signal 33 has to be included too, right?

ipetkov (approved):
> And I guessed that signal 33 has to be included too, right?

Yep, good eye, thanks for fixing this too!

PR #4558, opened 2022-03-08
chore: fix a bunch of annoying clippy lints

hawkw:
## Motivation

Recent Clippy releases have added some new lints that trigger on some
code in Tokio. These aren't a big deal, but seeing them in my editor is
mildly annoying.

## Solution

This branch fixes the following issues flagged by Clippy:

* manual `Option::map` implementation
* use of `.map(...).flatten(...)` that could be replaced with
  `.and_then(...)`
* manual implementation of saturating arithmetic on `Duration`s
* simplify some boolean expressions in assertions (`!res.is_ok()` can be
`res.is_err()`)
* fix redundant field names in initializers
* ~~allow the `clippy::unnecessary_to_owned` lint in `ToSocketAddrs` impl
  for `&[SocketAddr]` --- the "unnecessary" `to_vec()` call is actually
  necessary to ensure the iterator is `'static`~~
* replace an unnecessary cast to `usize` with an explicitly typed
  integer literal

### DIFF
--- a/tokio/src/io/driver/mod.rs
+++ b/tokio/src/io/driver/mod.rs
@@ -300,11 +300,7 @@ cfg_metrics! {
         where
             F: Fn(&IoDriverMetrics) -> R,
         {
-            if let Some(inner) = self.inner() {
-                Some(f(&inner.metrics))
-            } else {
-                None
-            }
+            self.inner().map(|inner| f(&inner.metrics))
         }
    }
 }
--- a/tokio/src/net/addr.rs
+++ b/tokio/src/net/addr.rs
@@ -141,7 +141,10 @@ impl sealed::ToSocketAddrsPriv for &[SocketAddr] {
         // in order to ensure that the returned iterator is valid for the
         // `'static` lifetime, which the borrowed `slice::Iter` iterator would
         // not be.
-        #[allow(clippy::unnecessary_to_owned)]
+        // Note that we can't actually add an `allow` attribute for
+        // `clippy::unnecessary_to_owned` here, as Tokio's CI runs clippy lints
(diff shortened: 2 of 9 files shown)


Noah-Kennedy:
Looks like clippy is failing.

hawkw:
> Looks like clippy is failing.

Hmm, it looks like it's failing because [the `clippy::unnecessary_to_owned` lint doesn't exist](https://github.com/tokio-rs/tokio/runs/5468521904?check_suite_focus=true#step:5:49). This seems to be because the Clippy CI workflow [is running on Rust 1.52.0 for some reason](https://github.com/tokio-rs/tokio/runs/5468521904?check_suite_focus=true#step:3:37) and that lint wasn't added to Clippy [until Rust 1.58](https://rust-lang.github.io/rust-clippy/master/#unnecessary_to_owned).

I'm not sure why the Clippy CI job isn't running on the latest stable Rust, but we seem to have [explicitly pinned it to 1.52.0 on purpose](https://github.com/tokio-rs/tokio/blob/2f944dfa1be67411f54f625674fb8f5f14927825/.github/workflows/ci.yml#L15)...

I guess the simplest solution is to just remove the `allow` attribute for that lint, so that the clippy CI build passes; users with newer Rust toolchains can just accept that they'll get warnings there, I guess.

hawkw:
Oh, it looks like the reason Clippy is pinned to 1.52 is to avoid breaking LTS releases when new lints are added: https://github.com/tokio-rs/tokio/commit/2cee1db20c2cdd54fe0edb6cd717fb9ba89f09a3

I'll just go ahead and remove the `allow` attributes.

Darksonn (approved):

PR #4562, opened 2022-03-11
ci: run doctests for unstable APIs

hawkw:
It turns out that the CI job for testing `tokio_unstable` features isn't
actually running doctests for `tokio_unstable`, just lib and integration
tests. This is because RustDoc is responsible for running doctests, and
it needs the unstable cfg passed to it separately from `RUSTFLAGS`.

This means that if the examples for unstable APIs are broken, CI won't
catch this, which is not great!

This commit changes the `test-unstable` CI job to pass `--cfg
tokio_unstable` in `RUSTDOCFLAGS` as well as `RUSTFLAGS`. This way,
doctests for unstable APIs should actually run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -174,6 +174,9 @@ jobs:
         working-directory: tokio
         env:
           RUSTFLAGS: --cfg tokio_unstable -Dwarnings
+          # in order to run doctests for unstable features, we must also pass
+          # the unstable cfg to RustDoc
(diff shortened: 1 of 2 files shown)


Noah-Kennedy (approved):

hawkw:
cool, found a typo in the runtime metrics doctests --- the system works! https://github.com/tokio-rs/tokio/runs/5515810817?check_suite_focus=true

PR #4576, opened 2022-03-21
Ignore VS Code

huntc:
Ignores VS Code files as they provide personal settings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ target
 Cargo.lock
 
 .cargo/config.toml
+
+.vscode


taiki-e:
Thanks for the PR! For editor-specific settings, it is recommended to add to a global gitignore file (`~/.config/git/ignore`).
See also: https://docs.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

huntc:
I'm unsure that this should be a global setting though given that I have other repositories requiring VSC settings...

taiki-e:
Sorry, I'm not sure what the best practice is in such cases. In my previous case where I needed to commit VS Code settings, I explicitly included the setting files like the [angular does](https://github.com/angular/angular/blob/a049840e0db4941e382b6b7046d333e58fccf688/aio/.gitignore#L25-L29). (In fact some VS Code extensions dumped a lot of logs to the subdirectory of `.vscode`, so that seemed to be the best way in my case.)

PR #4577, opened 2022-03-21
A Consistent Overhead Byte Stuffing (COBS) codec.

huntc:
From [Wikipedia](https://en.wikipedia.org/wiki/Consistent_Overhead_Byte_Stuffing):

_Consistent Overhead Byte Stuffing (COBS) is an algorithm for encoding data bytes that results in efficient, reliable, unambiguous packet framing regardless of packet content, thus making it easy for receiving applications to recover from malformed packets. It employs a particular byte value, typically zero, to serve as a packet delimiter (a special value that indicates the boundary between packets). When zero is used as a delimiter, the algorithm replaces each zero data byte with a non-zero value so that no zero data bytes will appear in the packet and thus be misinterpreted as packet boundaries._

I've taken particular care here to see that there is a maximum buffer size in order to reduce attack surface vectors. While a default buffer size of `usize::MAX` could be provided, I prefer to have to always think about what that dimension should be.

TODO:

* [x] write tests
* [x] flesh out the encoder
* [ ] flesh out the decoder
* [ ] try in my own app


tokio-util/src/codec/cobs.rs
```diff
@@ -0,0 +1,115 @@
...
+        let encoded_remaining_len = src.len() % CHUNK_LEN;
+        let encoded_len = if encoded_remaining_len > 0 {
+            encoded_len + encoded_remaining_len + MAX_BYTE_OVERHEAD
+        } else {
+            encoded_len
+        };
+        dst.reserve(encoded_len);
+        for (i, byte) in src.iter().enumerate() {
+            if i % CHUNK_LEN == 0 {
+                dst.put_u8(0);
```
luben, line 71: Should we not put the delimiter here, not 0?

luben:
The example C implementation in the wikipedia article is a single-pass. Your implementation adds second pass. Is it faster than single pass?

tokio-util/src/codec/cobs.rs
huntc: That's actually a count of bytes until we reach the delimiter, so no.

huntc:
> The example C implementation in the wikipedia article is a single-pass. Your implementation adds second pass. Is it faster than single pass?

I found the C implementation difficult to comprehend and wasn't convinced that it was correct either. I couldn't see how the first offset byte was being written. I should take another look though and perhaps work toward a single pass.

huntc:
I'm going to close this for now. I no longer have an immediate need for it, and as mentioned on the channel forum, it probably belongs outside of Tokio. Thanks.

PR #4581, opened 2022-03-25
sync: add `watch::Receiver::same_channel`

ahrens:
## Motivation

I'd like to determine if two `watch::Receiver`s are part of the same channel (i.e. will receive values from the same `watch::Sender`.

Specifically, I'd like to have a `HashMap<K, watch::Receiver<V>>`, and then later be able to check if *my* entry is still there, or someone else has removed it (and potentially replaced it with a Receiver for a different Sender).

Closes: https://github.com/tokio-rs/tokio/issues/4579

## Solution

Add a `same_channel()` method which uses `Arc::ptr_eq` on `Receiver::shared`

### DIFF
--- a/tokio/src/sync/watch.rs
+++ b/tokio/src/sync/watch.rs
@@ -467,6 +467,22 @@ impl<T> Receiver<T> {
         }
     }
 
+    /// Returns `true` if receivers belong to the same channel.
+    ///
+    /// # Examples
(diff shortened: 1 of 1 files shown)


Darksonn (commented):
Thanks.

Darksonn (approved):

PR #4592, opened 2022-04-01
Import tracing without the default attributes feature

paolobarbolini:
## Motivation

tracing enables the `attributes` feature by default, which we don't need here. I'd like to keep a close eye on the proc macro crates in my dependency tree.

## Solution

Disable default features and enable only the ones we need.


Darksonn (approved):
This seems ok to me. @hawkw any thoughts?

hawkw (approved):
seems good to me!

PR #4595, opened 2022-04-03
Implement Weak version of mpsc::Sender

b-naber:
Attempt to fix https://github.com/tokio-rs/tokio/issues/4023


tokio/src/sync/mpsc/chan.rs
```diff
@@ -25,6 +26,16 @@ impl<T, S: fmt::Debug> fmt::Debug for Tx<T, S> {
     }
 }
 
+pub(crate) struct TxWeak<T, S> {
+    inner: Weak<Chan<T, S>>,
+}
```
Darksonn, line 31: Loom doesn't support `Weak`. Regardless, an ordinary `Arc` should still work I think.

tokio/src/sync/mpsc/chan.rs
```diff
@@ -149,6 +168,50 @@ impl<T, S> Tx<T, S> {
...
+            // If we were able to upgrade, `Chan` is guaranteed to still exist,
+            // even though the channel might have been closed in the meantime.
+            // Need to check here whether the channel was actually closed.
+
+            let mut tx_count = inner.tx_count.load(Relaxed);
+            loop {
+                // FIXME Haven't thought the orderings on the CAS through yet
+                match inner
+                    .tx_count
+                    .compare_exchange(tx_count, tx_count + 1, SeqCst, SeqCst)
```
Darksonn, line 189: I don't think you should increment the counter in the first place if it is already closed.

Darksonn (commented):
Seems like a relatively simple implementation.

tokio/src/sync/mpsc/chan.rs
b-naber: Not entirely sure what you mean here. Do you think the Cas is not necessary here or do you want to check whether `tx_count` is 0 before going into the Cas?

Darksonn: The CAS is only necessary if you check whether `tx_count` is 0 first.

b-naber:
Looks like we can't use loom's `Arc` version either, since it doesn't have a `downgrade` assoc function. Is it a problem using `std::Arc` or shall we add `downgrade` to `loom::Arc`?

Darksonn:
I believe that it should be possible to get this work using the normal `Arc` because we count the number of senders separately in `tx_count`.

Darksonn:
Well, it's true that this means that the `Chan` struct may stay alive until all `Weak` handles are dropped, but its unclear whether that's a problem.

b-naber:
Ok, I thought that would be a no-go. If that's fine then I'll gladly change the PR. 

Probably should mention in the docs for `WeakSender`, that all outstanding instances need to be dropped for the memory of the underlying channel to be freed, shouldn't we?

Darksonn:
I mean, that's true for the weak version of `Arc` too. I think the bigger question is when pending messages are dropped if there are some left in the channel at the end.

b-naber:
> I mean, that's true for the weak version of `Arc` too.

Is it? I think `inner` of the `Arc` is dropped once the strong count reaches 0.

> I think the bigger question is when pending messages are dropped if there are some left in the channel at the end.

hm I think they should. When the last (strong) `Sender` is dropped, we'll put a 'closed' message on the list and any message that is put onto the list after that should be dropped imo. What do you think?

Darksonn:
> Is it? I think inner of the Arc is dropped once the strong count reaches 0.

The value in the `Arc` is dropped, yes, but the allocation containing it stays around.

> hm I think they should. When the last (strong) `Sender` is dropped, we'll put a 'closed' message on the list and any message that is put onto the list after that should be dropped imo. What do you think?

Messages can't be put on the list after dropping the last sender?

Perhaps you mean that we should drop pending messages when dropping the receiver. I'm happy with doing that.

Darksonn:
I believe that we already prevent sending messages after the `Receiver` is dropped. The question is what happens to those sent before the `Receiver` was dropped.
```
let (send, recv) = mpsc::unbounded();
send.send("hello");
drop(recv);
// is "hello" dropped now?
```

tokio/src/sync/mpsc/chan.rs
```diff
@@ -129,6 +130,49 @@ impl<T, S> Tx<T, S> {
...
+    }
+
+    // Returns a boolean that indicates whether the channel is closed.
+    pub(super) fn upgrade(self) -> Option<Self> {
+        let mut tx_count = self.inner.tx_count.load(Acquire);
+
+        if tx_count == 0 {
+            // channel is closed
+            return None;
+        }
```
Darksonn, line 152: You have this `if` twice in this function. If you move it to the top of the loop, then you can remove the duplication.

Darksonn (commented):
Besides these comments, I think this looks good.

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -1009,6 +1066,29 @@ impl<T> fmt::Debug for Sender<T> {
...
+        }
+    }
+}
+
+impl<T> WeakSender<T> {
+    /// Tries to convert a WeakSender into a [`Sender`]. This will return `Some`
+    /// if there are other `Sender` instances alive and the channel wasn't
+    /// previously dropped, otherwise `None` is returned.
+    pub fn upgrade(self) -> Option<Sender<T>> {
+        self.chan.upgrade().map(|tx| Sender::new(tx))
```
Darksonn, line 1082: This triggers clippy:
```suggestion
        self.chan.upgrade().map(Sender::new)
```

tokio/tests/sync_mpsc.rs
```diff
@@ -657,3 +665,213 @@ fn recv_timeout_panic() {
...
+// Tests that no pending messages are put onto the channel after `Rx` was
+// dropped.
+//
+// Note: After the introduction of `WeakSender`, which internally
+// used `Arc` and doesn't call a drop of the channel after the last strong
+// `Sender` was dropped while more than one `WeakSender` remains, we want to
+// ensure that no messages are kept in the channel, which were sent after
+// the receiver was dropped.
+#[tokio::test(start_paused = true)]
+async fn test_msgs_dropped_on_rx_drop() {
```
Darksonn, line 844: One of the `unwrap` calls in test this fails on wasm. I have not investigated why, but you might want to change them to `expect` with a message so you can tell which one failed.

tokio/tests/sync_mpsc.rs
```diff
@@ -657,3 +665,213 @@ fn recv_timeout_panic() {
     let (tx, _rx) = mpsc::channel(5);
     tx.send_timeout(10, Duration::from_secs(1)).now_or_never();
 }
+
+#[tokio::test]
+async fn weak_sender() {
+    let (tx, mut rx) = channel(11);
+    let tx_weak = tx.clone().downgrade();
```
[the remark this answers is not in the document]
b-naber: With regards to the fourth test scenario: We can't actually have any outstanding `Permit`s if there are no `Sender`s left.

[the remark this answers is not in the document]
Darksonn: It should be possible to have an [`tokio::sync::mpsc::OwnedPermit`](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.OwnedPermit.html).

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -22,6 +22,50 @@ pub struct Sender<T> {
     chan: chan::Tx<T, Semaphore>,
 }
 
+/// A Sender that does not influence RAII semantics, i.e. if all [`Sender`]
+/// instances of a channel were dropped and only `WeakSender` instances remain,
+/// the channel is closed.
```
Darksonn, line 27: The first line of the documentation for the struct is visible on the module documentation too. It should be a short sentence that fits on a single line on the module documentation.
```suggestion
/// A sender that does not prevent the channel from being closed.
///
/// If all [`Sender`]
/// instances of a channel were dropped and only `WeakSender` instances remain,
/// the channel is closed.
```
+ reflow text

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -22,6 +22,50 @@ pub struct Sender<T> {
...
+/// `Sender`s were already dropped, otherwise `Some` (at which point it does
+/// influence RAII semantics again).
+///
+/// [`Sender`]: Sender
+/// [`WeakSender::upgrade`]: WeakSender::upgrade
+///
+/// #Examples
+///
+/// ```rust
+/// use tokio;
```
Darksonn, line 40:
```suggestion
/// ```
```

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -22,6 +22,50 @@ pub struct Sender<T> {
     chan: chan::Tx<T, Semaphore>,
 }
 
+/// A Sender that does not influence RAII semantics, i.e. if all [`Sender`]
+/// instances of a channel were dropped and only `WeakSender` instances remain,
+/// the channel is closed.
+///
+/// In order to send messages, the `WeakSender` needs to be upgraded using
+/// [`WeakSender::upgrade`], which returns `Option<Sender>`, `None` if all
```
Darksonn, line 30: This sentence doesn't really make sense at this comma. How about this?

> which returns `Option<Sender>`. It returns `None` if all `Sender`s have been dropped, and otherwise it returns a `Sender`.

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -22,6 +22,50 @@ pub struct Sender<T> {
...
+///             } else if i == 1 {
+///                 // only WeakSender instance remains -> channel is dropped
+///                 assert!(rx.recv().await.is_none());
+///             }
+///         }
+///     })
+///     .await;
+///
+///     assert!(tx_weak.upgrade().is_none());
+/// }
```
Darksonn, line 62: This example ignores panics in the spawned task, and its rather complicated. I suggest replacing it with a much simpler example - e.g. the example could look like this:
```Rust
let (tx, rx) = channel(15);
let tx_weak = tx.clone().downgrade();

// Upgrading will succeed because `tx` still exists.
assert!(tx_weak.clone().upgrade().is_some());

// If we drop `tx`, then it will fail.
drop(tx);
assert!(tx_weak.clone().upgrade().is_none());
```

tokio/src/sync/mpsc/chan.rs
```diff
@@ -129,6 +130,39 @@ impl<T, S> Tx<T, S> {
...
+
+            // Notify the receiver
+            self.wake_rx();
+        }
+
+        self
+    }
+
+    // Returns a boolean that indicates whether the channel is closed.
+    pub(super) fn upgrade(self) -> Option<Self> {
```
Darksonn, line 146:
```suggestion
    /// Returns the upgraded channel, or None if the upgrade failed.
    pub(super) fn upgrade(self) -> Option<Self> {
```

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -22,6 +22,41 @@ pub struct Sender<T> {
...
+///   
+///     // Upgrading will succeed because `tx` still exists.
+///     assert!(tx_weak.clone().upgrade().is_some());
+///   
+///     // If we drop `tx`, then it will fail.
+///     drop(tx);
+///     assert!(tx_weak.clone().upgrade().is_none());
+/// }
+///
+/// ```
```
Darksonn, line 55:
```suggestion
/// }
/// ```
```

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -22,6 +22,41 @@ pub struct Sender<T> {
...
+///     let (tx, _rx) = channel::<i32>(15);
+///     let tx_weak = tx.clone().downgrade();
+///   
+///     // Upgrading will succeed because `tx` still exists.
+///     assert!(tx_weak.clone().upgrade().is_some());
+///   
+///     // If we drop `tx`, then it will fail.
+///     drop(tx);
+///     assert!(tx_weak.clone().upgrade().is_none());
+/// }
```
Darksonn, line 53: This test causes a CI failure for some reason.

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -1009,6 +1057,29 @@ impl<T> fmt::Debug for Sender<T> {
     }
 }
 
+impl<T> Clone for WeakSender<T> {
+    fn clone(&self) -> Self {
+        WeakSender {
+            chan: self.chan.clone(),
+        }
+    }
+}
```
Darksonn, line 1066: This increments the tx count, but it shouldn't.

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -991,6 +1027,19 @@ impl<T> Sender<T> {
...
         self.chan.semaphore().0.available_permits()
     }
+
+    /// Converts the `Sender` to a [`WeakSender`] that does not count
+    /// towards RAII semantics, i.e. if all `Sender` instances of the
+    /// channel were dropped and only `WeakSender` instances remain,
+    /// the channel is closed.
+    pub fn downgrade(&self) -> WeakSender<T> {
+        // Note: If this is the last `Sender` instance we want to close the
+        // channel when downgrading, so it's important to move into `self` here.
```
Darksonn, line 1037: This comment is wrong.

tokio/src/sync/mpsc/chan.rs
```diff
@@ -46,7 +47,7 @@ pub(crate) trait Semaphore {
     fn is_closed(&self) -> bool;
 }
 
-struct Chan<T, S> {
+pub(crate) struct Chan<T, S> {
```
Darksonn, line 50:
```suggestion
pub(super) struct Chan<T, S> {
```

tokio/tests/sync_mpsc.rs
```diff
@@ -657,3 +663,273 @@ fn recv_timeout_panic() {
...
+// Tests that no pending messages are put onto the channel after `Rx` was
+// dropped.
+//
+// Note: After the introduction of `WeakSender`, which internally
+// used `Arc` and doesn't call a drop of the channel after the last strong
+// `Sender` was dropped while more than one `WeakSender` remains, we want to
+// ensure that no messages are kept in the channel, which were sent after
+// the receiver was dropped.
+#[tokio::test(start_paused = true)]
+async fn test_msgs_dropped_on_rx_drop() {
```
Darksonn, line 842: It seems to me that this test could be simplified by not spawning tasks and not using sleep.

tokio/tests/sync_mpsc.rs
```diff
@@ -657,3 +663,273 @@ fn recv_timeout_panic() {
...
+    });
+
+    let (_, _) = join!(rx_handle, tx_handle);
+}
+
+// Tests that a `WeakSender` is upgradeable when other `Sender`s exist.
+#[tokio::test]
+async fn downgrade_upgrade_sender_success() {
+    let (tx, _rx) = mpsc::channel::<i32>(1);
+    let weak_tx = tx.clone().downgrade();
```
Darksonn, line 881: Now that it takes `&self`, we don't need this `.clone()`. I'm sure the other tests also have extra calls to `.clone()`.
```suggestion
    let weak_tx = tx.downgrade();
```

tokio/tests/sync_mpsc.rs
```diff
@@ -657,3 +663,273 @@ fn recv_timeout_panic() {
...
+// Tests that `Clone` of `WeakSender` doesn't decrement `tx_count`.
+#[tokio::test]
+async fn test_weak_sender_clone() {
+    let (tx, _rx) = mpsc::channel::<i32>(1);
+    let tx_weak = tx.downgrade();
+    let tx_weak2 = tx.downgrade();
+    drop(tx);
+
+    assert!(tx_weak.upgrade().is_none() && tx_weak2.upgrade().is_none());
+}
```
Darksonn, line 935: This test doesn't use `Clone` on `WeakSender` at all, so the comment must be wrong.

tokio/src/sync/mpsc/bounded.rs
Darksonn: This one still applies.

Darksonn:
Please click "resolve" on the things you have fixed.

tokio/tests/sync_mpsc.rs
```diff
@@ -657,3 +660,264 @@ fn recv_timeout_panic() {
...
+
+// Tests that a `WeakSender` cannot be upgraded after a `Sender` was dropped,
+// which existed at the time of the `downgrade` call.
+#[tokio::test]
+async fn downgrade_drop_upgrade() {
+    let (tx, _rx) = mpsc::channel::<i32>(1);
+
+    // the cloned `Tx` is dropped right away
+    let weak_tx = tx.clone().downgrade();
```
Darksonn, line 887: You missed this one.
```suggestion
    let weak_tx = tx.downgrade();
```

b-naber: That was intentional, otherwise the test is equivalent to the test `downgrade_upgrade_sender_failure`. Whether this is worth testing for is another question though, I can remove the test if you want.

Darksonn: If it is intentional, then it's fine.

Darksonn (approved):
Looks good to me.

PR #4616, opened 2022-04-13
blocking: #[track_caller] for spawn_blocking

domodwyer:
This PR adds a `#[track_caller]` annotation to [`runtime::blocking::spawn_blocking()`](https://github.com/domodwyer/tokio/blob/eed85e46740cb21449fcfa92e14485d6b6c936e2/tokio/src/runtime/blocking/pool.rs#L108).

Similar to #4483, the driver behind this change is to correctly tag the caller as the spawn location in tracing spans, primarily for consumption with `tokio-console`. 

Prior to this PR, any tasks spawned with `spawn_blocking()` would show a spawn location within tokio:

Screenshot 2022-04-13 at 14 04 38

After this PR, the spawn location is correctly recorded as the `main.rs` file in my reproducer:

Screenshot 2022-04-13 at 14 05 42

I am not sure how to effectively test this, and #4483 didn't test it either - happy to add tests with a bit of guidance!

---

* blocking: #[track_caller] for spawn_blocking (eed85e46)

      Annotates spawn_blocking() with #[track_caller] in order to produce the 
      correct tracing span spawn locations that show the caller as the spawn point.

### DIFF
--- a/tokio/src/runtime/blocking/pool.rs
+++ b/tokio/src/runtime/blocking/pool.rs
@@ -104,6 +104,7 @@ const KEEP_ALIVE: Duration = Duration::from_secs(10);
 /// Runs the provided function on an executor dedicated to blocking operations.
 /// Tasks will be scheduled as non-mandatory, meaning they may not get executed
 /// in case of runtime shutdown.
+#[track_caller]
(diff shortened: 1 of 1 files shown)


Darksonn (approved):
Thanks. You don't need to add any tests because it is essentially impossible to test without a lot of additional setup.

PR #4629, opened 2022-04-20
rt: internally split Handle into two structs

carllerche:
Previously, `runtime::Handle` was a single struct composed of the
internal handles for each runtime component. This patch splits the
`Handle` struct into a `HandleInner` which contains everything
**except** the task scheduler handle. Now, `HandleInner` is passed to
the task scheduler during creation and the task scheduler is responsible
for storing it. `Handle` only  needs to hold the scheduler handle and
can access the rest of the component handles by querying the task
scheduler.

The motivation for this change is it now enables the multi-threaded
scheduler to have direct access to the blocking spawner handle.
Previously, when spawning a new thread, the multi-threaded scheduler had
to access the blocking spawner by accessing a thread-local variable.
Now, in theory, the multi-threaded scheduler can use `HandleInner`
directly. However, this change hasn't been done in this PR yet.

Also, now the `Handle` struct is much smaller.

This change is intended to make it easier for the multi-threaded
scheduler to shutdown idle threads and respawn them on demand.


tokio/src/runtime/blocking/pool.rs
```diff
@@ -220,7 +220,7 @@ impl fmt::Debug for BlockingPool {
 // ===== impl Spawner =====
 
 impl Spawner {
-    pub(crate) fn spawn(&self, task: Task, rt: &Handle) -> Result<(), ()> {
+    pub(crate) fn spawn(&self, task: Task, rt: &dyn ToHandle) -> Result<(), ()> {
```
carllerche, line 223: `dyn` because it isn't really performance critical so :shrug: less code gen.

tokio/src/runtime/handle.rs
```diff
@@ -16,7 +16,11 @@ use std::{error, fmt};
 #[derive(Debug, Clone)]
 pub struct Handle {
     pub(super) spawner: Spawner,
+}
 
+/// All internal handles that are *not* the scheduler's spawner.
+#[derive(Debug)]
+pub(crate) struct HandleInner {
```
hawkw, line 23: nit, take it or leave it: as an alternative to `HandleInner`, what if we called this `Resources` or something, since it's mostly resource drivers? i guess the blocking spawner isn't a resource driver...

tokio/src/runtime/handle.rs
```diff
@@ -196,85 +205,11 @@ impl Handle {
...
-
-        #[cfg(not(all(tokio_unstable, feature = "tracing")))]
-        let _ = name;
-
-        let (task, handle) = task::unowned(fut, NoopSchedule);
-        let spawned = self
-            .blocking_spawner
-            .spawn(blocking::Task::new(task, is_mandatory), self);
-        (handle, spawned.is_ok())
+    pub(crate) fn as_inner(&self) -> &HandleInner {
```
hawkw, line 211: very nitpicky, feel free to ignore: i might just drop the `as_` prefix and call this `inner()` --- it's more like an accessor than a conversion, and dropping the `as_` saves 3 characters? but, i don't really care...

tokio/src/runtime/basic_scheduler.rs
```diff
@@ -397,6 +402,10 @@ impl Spawner {
     pub(crate) fn reset_woken(&self) -> bool {
         self.shared.woken.swap(false, AcqRel)
     }
+
+    pub(crate) fn as_handle_inner(&self) -> &HandleInner {
+        &self.shared.handle_inner
+    }
```
hawkw, line 408: stupid naming nit, take it or leave it: i might have just called this `inner`?

hawkw (approved):
overall, this seems fine to me! i left some nitpicky comments you can feel free to disregard.

i'm not the _biggest_ fan of the `HandleInner` name, but i'm not sure if there's anything better --- i suggested `Resources`, since it's mostly resource drivers, but i'm not sure if that's a big improvement :woman_shrugging:

tokio/src/runtime/handle.rs
carllerche: It isn't really resources, it is `Handle` - task scheduler handle. "Inner" isn't great, but it is an established pattern. I rather stick w/ convention unless there is a reason not to.

tokio/src/runtime/basic_scheduler.rs
carllerche: `inner()` on `BasicScheculer` would mean the basic scheduler's inner.

tokio/src/runtime/handle.rs
carllerche: I'll change this in a follow up since CI takes so long to run and I have more PRs coming.

tokio/src/runtime/basic_scheduler.rs
hawkw: oh, yeah, i missed that. carry on!

Darksonn (approved):
Seems fine.

PR #4652, opened 2022-05-03
Rewrite tokio-util::sync::CancellationToken

Finomnis:
Related to https://github.com/tokio-rs/tokio/issues/4639.

## Motivation
CancellationToken uses a lot of unsafe code and seems to be error-prone.

## Solution

@Darksonn suggested a rewrite.

After a discussion on discord this is my attempt to combine and polish our ideas.

A thorough review will be necessary.

## State

- [x] Architectural discussion on discord
- [x] Rewrite attempt
- [x] Close TODOs in code
- [x] Clippy
- [x] More tests? (Related: https://github.com/tokio-rs/tokio/pull/4645)
- [x] Thorough review of @Darksonn


tokio-util/src/sync/cancellation_token.rs
```diff
@@ -318,561 +193,311 @@ impl<'a> core::fmt::Debug for WaitForCancellationFuture<'a> {
...
-            from_parent: SynchronizedThroughParent {
-                prev_peer: None,
-                next_peer: None,
-            },
-            state: AtomicUsize::new(state.pack()),
-            synchronized: Mutex::new(SynchronizedState::new()),
-        }
-    }
+        let mut dropped_children = vec![];
+        std::mem::swap(&mut node.children, &mut dropped_children);
```
Darksonn, line 313: You can use `mem::take` here.

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -318,561 +193,311 @@ impl<'a> core::fmt::Debug for WaitForCancellationFuture<'a> {
...
+    ///
+    /// To prevent that this problem leaks into the rest of the code, it is abstracted
+    /// in this function.
+    fn with_locked_node_and_parent<
+        F: FnOnce(MutexGuard<'_, Inner>, Option<MutexGuard<'_, Inner>>) -> Ret,
+        Ret,
+    >(
+        node: &Arc<TreeNode>,
+        func: F,
+    ) -> Ret {
```
Darksonn, line 337:
```suggestion
    fn with_locked_node_and_parent<F, Ret>(
        node: &Arc<TreeNode>,
        func: F,
    ) -> Ret
    where
        F: FnOnce(&mut Inner, Option<&mut Inner>) -> Ret,
    {
```

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -318,561 +193,311 @@ impl<'a> core::fmt::Debug for WaitForCancellationFuture<'a> {
...
-                    return next_state;
-                }
-                Err(actual) => {
-                    current_packed_state = actual;
-                    current_state = StateSnapshot::unpack(actual);
+            let potential_parent = {
+                let locked_node = node.inner.lock().unwrap();
+                match locked_node.parent.clone() {
+                    Some(parent) => parent,
+                    // If we locked the node and its parent is `None`, we are in a valid state
```
Darksonn, line 343: If this check fails, you can reuse the parent `actual_parent` from the previous iteration.

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -318,561 +193,311 @@ impl<'a> core::fmt::Debug for WaitForCancellationFuture<'a> {
...
+        {
+            let mut locked_replacement_child = replacement_child.inner.lock().unwrap();
+            locked_replacement_child.parent_idx = pos;
         }
 
-        current_state
+        // Swap the element to remove out of the list
+        std::mem::swap(&mut parent.children[pos], &mut replacement_child);
+
+        replacement_child
```
Darksonn, line 418: Here you can use `mem::replace`.

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -55,30 +50,24 @@ use guard::DropGuard;
...
 pub struct WaitForCancellationFuture<'a> {
     /// The CancellationToken that is associated with this WaitForCancellationFuture
-    cancellation_token: Option<&'a CancellationToken>,
-    /// Node for waiting at the cancellation_token
-    wait_node: ListNode<WaitQueueEntry>,
-    /// Whether this future was registered at the token yet as a waiter
-    is_registered: bool,
+    _cancellation_token: CancellationToken,
+    /// Future, to wait for cancellation
+    future: Option<Pin<Box<tokio::sync::futures::Notified<'a>>>>,
```
Darksonn, line 63: We do not need to heap allocate here.
```suggestion
    future: Option<tokio::sync::futures::Notified<'a>>,
```

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -318,561 +193,311 @@ impl<'a> core::fmt::Debug for WaitForCancellationFuture<'a> {
...
-        // wait queue at the cancellation_token, it must be removed before dropping.
-        // Otherwise the cancellation_token would access invalid memory.
-        if let Some(token) = self.cancellation_token {
-            if self.is_registered {
-                token.unregister(&mut self.wait_node);
-            }
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<()> {
+        match &mut self.future {
+            Some(future) => Pin::new(future).poll(cx),
+            None => Poll::Ready(()),
```
Darksonn, line 199: When the `Notified` returns `Poll::Ready`, I think it would be best to call `get_future` again and only return `Poll::Ready` if you get back `None`.

tokio-util/src/sync/cancellation_token.rs
Finomnis: The `MutexGuard` is in there on purpose, so that `func` can choose to unlock either if required. (See `remove_child`)

tokio-util/src/sync/cancellation_token.rs
Finomnis: I wasn't able to get it to compile without Pin<Box<>>. I'll try again later.

tokio-util/src/sync/cancellation_token.rs
Finomnis: Updated

tokio-util/src/sync/cancellation_token.rs
Finomnis: Updated

tokio-util/src/sync/cancellation_token.rs
Finomnis: Updated

tokio-util/src/sync/cancellation_token.rs
Finomnis: Updated

tokio-util/src/sync/cancellation_token.rs
Finomnis: It seems that I need to implement [projections](https://doc.rust-lang.org/std/pin/#projections-and-structural-pinning), but I've never worked with that before. Is my assumption correct?

Finomnis: After reading into it a little, this seems very advanced and quite error prone. Are you sure we need that?
Is there any situation where the Waker could wake the future without being cancelled?

Finomnis: I think the point of using `Pin<Box<>>` is related. To remove that, we also would need projections.

tokio-util/src/sync/cancellation_token.rs
Finomnis: I believe to solve this we would have to implement [projections](https://doc.rust-lang.org/std/pin/#projections-and-structural-pinning).

Finomnis: I read more about projections and got to the conclusion that it's difficult. I realized we already use a projection crate, `pin-project-lite`. Sadly, this one doesn't work (to my understanding) with the `WaitForCancellationFuture`, because we wrap our nested future in an `Option`. Which brings us again to the same problem that we have a `Pin<&mut Option<Future>>`, but we need a `Pin<&mut Future>` to actually do the polling.
Will read about it some more, but I'm still wondering how problematic a `Pin<Box<>>` would really be.

Finomnis: Never mind. `.as_pin_mut()` does exactly that.

Updated.

tokio-util/src/sync/cancellation_token.rs
Finomnis: Still can't figure out the lifetimes.

Problem is that `Pin<&mut Future>` has the anonymous lifetime, but creating a new future with `.get_future()` creates a different lifetime that cannot be stored in the object.
I have no Idea if I'm simply doing something wrong, though.

If you have any idea, feel free to give me a hint.

Finomnis: Ok finally found the relevant functions. And learned a lot about `Pin`.

Updated.

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -319,560 +190,323 @@ impl<'a> Future for WaitForCancellationFuture<'a> {
...
-                Ordering::SeqCst,
-                Ordering::SeqCst,
-            ) {
-                Ok(_) => {
-                    return next_state;
-                }
-                Err(actual) => {
-                    current_packed_state = actual;
-                    current_state = StateSnapshot::unpack(actual);
+            potential_parent = {
```
Darksonn, line 356: I don't think this scope is useful. Just put
```Rust
potential_parent = actual_parent;
```
at the end.

Darksonn (commented):
Overall this looks good.

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -319,560 +190,323 @@ impl<'a> Future for WaitForCancellationFuture<'a> {
...
+                match parent {
+                    Some(mut parent) => {
+                        move_children_to_parent(&mut node, &mut parent);
+
+                        // Remove the node from the parent
+                        remove_child(&mut parent, node);
+                    }
+                    None => {
+                        disconnect_children(&mut node);
+                    }
```
Darksonn, line 464: If this happens in parallel with a call to cancel on the parent node, after `node` is removed from its parent, but before `cancel` is called on `node`, then the cancel is lost for the children.

Possible solution: Don't destroy the tree when cancelling.

tokio-util/src/sync/cancellation_token.rs
Finomnis: Fixed

tokio-util/src/sync/cancellation_token.rs
Finomnis: Just for understanding: 
- You call "Cancel" on the parent.
- The parent disconnects the node
- On other thread: the last handle gets dropped. Because the node no longer has a parent, "disconnect_children" gets called
- The "cancel" function finds a node with no children and decides that's it

If with "don't destroy the tree when cancelling" you mean, "First cancel everything, then destroy the tree", I agree.
We just gotta be careful about what happens when we get multiple cancels on different nodes simultaneously.

Related question: Do we actually have to destroy the tree when cancelling? Do we gain anything from it? Originally I thought it might reduce the number of nodes, but now that we implemented that we move children to parents when no handles exist any more, I don't think disconnecting the tree gives us anything any more.

We further have to be careful though, that when we iterate through the tree we don't run into the same problem.

The actual problem is that there is a gap in cancel() between locking the children for disconnect and locking the children for propagating the cancellation. I'll attempt to fix that.

Finomnis: I realized while iterating over the tree for cancellation+dropping, we can only iterate over it depth first, because we have to keep the lock of the parent alive during the entire time of iterating through the children, and we cannot lock two children simultaneously. That means we have to descend one child at a time.

Finomnis: Fixed, potentially

Darksonn:
The diff that GitHub generates for this is completely unreadable. Can you move the `implementation` module to a separate file? We already have a folder for sub-modules of the cancellation token.

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -319,560 +190,374 @@ impl<'a> Future for WaitForCancellationFuture<'a> {
...
+        loop {
+            // Deadlock safety:
+            // At this very point, we *assume* that 'potential_parent' is the parent of 'node'.
+            // But both are currently unlocked. Therefore, we cannot rely any more that this is the case.
+            // Even worse, 'potential_parent' might actually become a *sibling* of 'node', which would violate
+            // our 2. rule of deadlock safety.
+            //
+            // However, we can prove that the lock order has to be 'potential_parent' first, then 'node',
+            // because based on invariant #2, 'node' is younger than 'potential_parent', as
+            // 'potential_parent' once was the parent of 'node'.
```
Darksonn, line 374: I'm not a fan of the way this is phrased. You are saying "this doesn't work because X, but actually it works anyway because Y". I would prefer something like "this works because Y. It might seem like X is an issue, but reasons"

Finomnis:
> The diff that GitHub generates for this is completely unreadable. Can you move the `implementation` module to a separate file? We already have a folder for sub-modules of the cancellation token.

Fixed

tokio-util/src/sync/cancellation_token.rs
Finomnis: Makes sense. Will rephrase.

tokio-util/src/sync/cancellation_token.rs
Finomnis: Ok if I close this?

tokio-util/src/sync/cancellation_token.rs
Finomnis: Fixed, potentially

tokio-util/src/sync/cancellation_token/implementation.rs
```diff
@@ -0,0 +1,341 @@
...
+//!     This is important during both cancellation and refcounting.
+//!
+//! 2. If node B *is* or *was* a child of node A, then node B was created *after* node A.
+//!     This is important for deadlock safety, as it is used for lock order.
+//!     Node B can only become the child of node A in two ways:
+//!         - being created with `child_node()`, in which case it is trivially true that
+//!           node A already existed when node B was created
+//!         - being moved A->C->B to A->B because node C was removed in `decrease_handle_refcount()`.
+//!           In this case the invariant still holds, as B was younger than C, and C was younger
+//!           than A, therefore B is also younger than A.
```
Darksonn, line 19: This can also happen in `cancel`.

tokio-util/src/sync/cancellation_token/implementation.rs
```diff
@@ -0,0 +1,341 @@
...
+//! 3. If two nodes are both unlocked and node A is the parent of node B, then node B is a child of node A.
+//!     It is important to always restore that invariant before dropping the lock of a node.
+//!
+//! ** Deadlock safety **
+//!
+//! Principles that provide deadlock safety:
+//!     1. We always lock in the order of creation time. We can prove this through invariant #2.
+//!        Specifically, through invariant #2, we know that we always have to lock a parent before its child.
+//!     2. We never lock two siblings simultaneously, because we cannot establish an order.
+//!        There is one exception in `with_locked_node_and_parent()`, which is described in the function itself.
```
Darksonn, line 30: I would simplify this to just say that we always lock in order of creation, which we can determine through invariant #2. You don't need to talk about siblings here.

tokio-util/src/sync/cancellation_token/implementation.rs
```diff
@@ -0,0 +1,341 @@
...
+//!
+use crate::loom::sync::{Arc, Mutex, MutexGuard};
+
+/// A node of the cancellation tree structure
+///
+/// The actual data it holds is wrapped inside a mutex for synchronization.
+pub(crate) struct TreeNode {
+    inner: Mutex<Inner>,
+    waker: tokio::sync::Notify,
+}
```
Darksonn, line 40: Does it make more sense to name this file `tree_node.rs` instead?

tokio-util/src/sync/cancellation_token/implementation.rs
```diff
@@ -0,0 +1,341 @@
...
+        // This is true even if the potential_parent is no longer the current parent or even its sibling,
+        // as the invariant still holds.
+        let locked_parent = potential_parent.inner.lock().unwrap();
+        let locked_node = node.inner.lock().unwrap();
+
+        let actual_parent = match locked_node.parent.clone() {
+            Some(parent) => parent,
+            // If we locked the node and its parent is `None`, we are in a valid state
+            // and can return.
+            None => return func(locked_node, None),
```
Darksonn, line 164: The potential parent might not match and be `Some` here. We shouldn't keep the previous parent locked in that case.
```suggestion
            // If we locked the node and its parent is `None`, we are in a valid state
            // and can return.
            None => {
                drop(locked_parent);
                return func(locked_node, None);
            }
```

tokio-util/src/sync/cancellation_token/implementation.rs
```diff
@@ -0,0 +1,341 @@
...
+/// We basically have to create a "lock wave" from the parent to children, that nothing else
+/// can bypass. Otherwise, other actions could add/remove children that our cancellation would miss.
+///
+/// Example: Holding a reference of a child does not influence its handle reference count, so
+/// it will happily drop its grandchildren in [decrease_handle_refcount] before we got the chance
+/// to propagate the cancellation to them, if we violate invariant #1.
+/// Unless we keep the child locked during the entire time, from the moment on when we
+/// disconnected it from its parent.
+pub(crate) fn cancel(node: &Arc<TreeNode>) {
+    let waker = &node.waker;
```
Darksonn, line 293: You should wake the wakers after releasing the lock. Otherwise the future might try to lock the mutex while we're still holding it, which is a waste of resources.

tokio-util/src/sync/cancellation_token/implementation.rs
```diff
@@ -0,0 +1,341 @@
...
+        }
+
+        // Disconnect the node from its parent.
+        // Now that it is cancelled, it doesn't need any cancellation from
+        // its parent any more.
+        // This has to happen at the very end, as this unlocks `node`, and therefore
+        // we would validate invariant #1 otherwise.
+        if let Some(mut parent) = parent {
+            remove_child(&mut parent, node);
+        };
```
Darksonn, line 328: You don't need a semicolon here.
```suggestion
        if let Some(mut parent) = parent {
            remove_child(&mut parent, node);
        }
```

tokio-util/src/sync/cancellation_token/implementation.rs
```diff
@@ -0,0 +1,341 @@
...
+
+            // Now the child is cancelled and detached and all its children are adopted.
+            // Just continue until all (including adopted) children are cancelled and detached.
+        }
+
+        // Disconnect the node from its parent.
+        // Now that it is cancelled, it doesn't need any cancellation from
+        // its parent any more.
+        // This has to happen at the very end, as this unlocks `node`, and therefore
+        // we would validate invariant #1 otherwise.
```
Finomnis, line 325: Typo, should be 'invalidate'. Rephrase the entire sentence, as it is confusing to read.

tokio-util/src/sync/cancellation_token/implementation.rs
```diff
@@ -0,0 +1,341 @@
...
+/// it will happily drop its grandchildren in [decrease_handle_refcount] before we got the chance
+/// to propagate the cancellation to them, if we violate invariant #1.
+/// Unless we keep the child locked during the entire time, from the moment on when we
+/// disconnected it from its parent.
+pub(crate) fn cancel(node: &Arc<TreeNode>) {
+    let waker = &node.waker;
+
+    with_locked_node_and_parent(node, |mut node, parent| {
+        // Cancel the subtree
+        node.is_cancelled = true;
```
Darksonn, line 297: If the node is already cancelled, you can just return immediately.

tokio-util/src/sync/cancellation_token/implementation.rs
Darksonn, line 328: There's nothing that really requires that we do this, and not doing it avoids locking the parent for the duration of the cancel call.

tokio-util/src/sync/cancellation_token/implementation.rs
```diff
@@ -0,0 +1,341 @@
...
+            // `node` is already locked, so we can lock the child
+            let mut locked_child = child.inner.lock().unwrap();
+
+            // Detach the child from node
+            // No need to modify node.children, as the child already got removed with `.pop`
+            locked_child.parent = None;
+            locked_child.parent_idx = 0;
+
+            // Adopt children
+            move_children_to_parent(&mut locked_child, &mut node);
```
Darksonn, line 311: Here, we _could_ avoid adopting the grandchildren that have no children. It would be more efficient.

tokio-util/src/sync/cancellation_token/implementation.rs
```diff
@@ -0,0 +1,341 @@
...
+
+        // One by one, adopt grandchildren and then cancel and detach the child
+        while let Some(child) = node.children.pop() {
+            // `node` is already locked, so we can lock the child
+            let mut locked_child = child.inner.lock().unwrap();
+
+            // Detach the child from node
+            // No need to modify node.children, as the child already got removed with `.pop`
+            locked_child.parent = None;
+            locked_child.parent_idx = 0;
```
Darksonn, line 308: If the node is already cancelled, you can just `continue` after these two statements.

tokio-util/src/sync/cancellation_token/implementation.rs
Finomnis: "Return" implies "drop" of all the variables in the context. That's how `DropGuard` works, essentially.

tokio-util/src/sync/cancellation_token/implementation.rs
Finomnis: It invalidates invariant #1. As soon as we detach `node` and then unlock it, `decrease_handle_refcount` might assume that `node` cannot be cancelled any more and detach all of its children.

The solution would be to not drop the lock of `node` in `remove_child`, but we cannot do that as we need to lock a sibling in the process of removing it.

Finomnis: I might be able to split this into two steps, where we first swap `node` to the end of `parent.children`, then lock it and remove it. Then it would be safe to unlock `parent`. But there cannot be a lock gap between removing `node` from `parent` and iterating through `node.children`.

tokio-util/src/sync/cancellation_token/implementation.rs
Finomnis: Would be, is it worth the extra complexity?

tokio-util/src/sync/cancellation_token/implementation.rs
Finomnis: Is this a question about an opinion or the request to change the name?

tokio-util/src/sync/cancellation_token/implementation.rs
Darksonn: Yes, but without the explicit drop, the mutex guard is dropped after calling `func`.

tokio-util/src/sync/cancellation_token/implementation.rs
Darksonn: I have a weak preference for `tree_node.rs`.

tokio-util/src/sync/cancellation_token/implementation.rs
Finomnis: Ouch, my bad.

tokio-util/src/sync/cancellation_token/implementation.rs
Darksonn: Let's do everything else first, then we can think about it.

tokio-util/src/sync/cancellation_token/implementation.rs
Finomnis: Section got removed completely during refactoring

tokio-util/src/sync/cancellation_token/implementation.rs
Finomnis: Fixed

tokio-util/src/sync/cancellation_token/implementation.rs
```diff
@@ -0,0 +1,355 @@
...
+        locked_child.parent = None;
+        locked_child.parent_idx = 0;
+
+        // Cancel or adopt grandchildren
+        while let Some(grandchild) = locked_child.children.pop() {
+            let mut locked_grandchild = grandchild.inner.lock().unwrap();
+
+            // For performance reasons, only adopt grandchildren that have children.
+            // Otherwise, just cancel them right away, no need for another iteration.
+            if locked_grandchild.children.is_empty() {
```
Darksonn, line 313:
```suggestion
            if locked_grandchild.children.is_empty() || locked_grandchild.is_cancelled {
```

tokio-util/src/sync/cancellation_token/implementation.rs
```diff
@@ -0,0 +1,355 @@
...
+                // Cancel the grandchild
+                locked_grandchild.is_cancelled = true;
+                drop(locked_grandchild);
+                grandchild.waker.notify_waiters();
+            } else {
+                // Otherwise, adopt grandchild
+                locked_grandchild.parent = locked_child.parent.clone();
+                locked_grandchild.parent_idx = locked_node.children.len();
+                drop(locked_grandchild);
+                locked_node.children.push(grandchild);
```
Darksonn, line 327: This is the same, but I find it easier to read this way.
```suggestion
                locked_grandchild.parent = node.clone();
                locked_grandchild.parent_idx = locked_node.children.len();
                drop(locked_grandchild);
                locked_node.children.push(grandchild);
```

tokio-util/src/sync/cancellation_token/implementation.rs
```diff
@@ -0,0 +1,355 @@
...
+
+    // One by one, adopt grandchildren and then cancel and detach the child
+    while let Some(child) = locked_node.children.pop() {
+        // `node` is already locked, so we can lock the child
+        let mut locked_child = child.inner.lock().unwrap();
+
+        // Detach the child from node
+        // No need to modify node.children, as the child already got removed with `.pop`
+        locked_child.parent = None;
+        locked_child.parent_idx = 0;
```
Darksonn, line 305: Here you can call `continue` if `locked_child.is_cancelled`. Its children are already cancelled in that case.

tokio-util/src/sync/cancellation_token/implementation.rs
Finomnis: Fixed

tokio-util/src/sync/cancellation_token/implementation.rs
Finomnis: Implemented

tokio-util/src/sync/cancellation_token/implementation.rs
Finomnis: Code got removed anyway

Finomnis: Fixed

tokio-util/src/sync/cancellation_token/implementation.rs
Finomnis: You too fast with reviews :D:D just implemented that

tokio-util/src/sync/cancellation_token/implementation.rs
Finomnis: Not possible, is borrowed

tokio-util/src/sync/cancellation_token/implementation.rs
Finomnis: Same for locked_grandchild, btw

tokio-util/src/sync/cancellation_token/implementation.rs
Finomnis: I implemented it a little different to avoid a second waker call

tokio-util/src/sync/cancellation_token/implementation.rs
Finomnis: I would have preferred your version as well. But it's borrowed in the `locked_node`

Finomnis: Never mind. My bad.

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -55,31 +52,20 @@ use guard::DropGuard;
...
-pub struct WaitForCancellationFuture<'a> {
-    /// The CancellationToken that is associated with this WaitForCancellationFuture
-    cancellation_token: Option<&'a CancellationToken>,
-    /// Node for waiting at the cancellation_token
-    wait_node: ListNode<WaitQueueEntry>,
-    /// Whether this future was registered at the token yet as a waiter
-    is_registered: bool,
+pin_project! {
+    /// A Future that is resolved once the corresponding [`CancellationToken`]
+    /// was cancelled
```
Darksonn, line 60:
```suggestion
    /// A Future that is resolved once the corresponding [`CancellationToken`]
    /// is cancelled.
```

Darksonn:
Please delete the `tokio-util/src/sync/intrusive_double_linked_list.rs` file.

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,351 @@
...
+pub(crate) fn cancel(node: &Arc<TreeNode>) {
+    let mut locked_node = node.inner.lock().unwrap();
+
+    if locked_node.is_cancelled {
+        return;
+    }
+
+    // One by one, adopt grandchildren and then cancel and detach the child
+    while let Some(child) = locked_node.children.pop() {
+        // `node` is already locked, so we can lock the child
```
Darksonn, line 285:
```suggestion
        // This can't deadlock because the mutex we are already
        // holding is the parent of child.
```

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,351 @@
...
+        locked_child.parent_idx = 0;
+
+        // If child is already cancelled, detaching is enough
+        if locked_child.is_cancelled {
+            continue;
+        }
+
+        // Cancel or adopt grandchildren
+        while let Some(grandchild) = locked_child.children.pop() {
+            let mut locked_grandchild = grandchild.inner.lock().unwrap();
```
Darksonn, line 300:
```suggestion
            // This can't deadlock because the two mutexes we are already
            // holding is the parent and grandparent of grandchild.
            let mut locked_grandchild = grandchild.inner.lock().unwrap();
```

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,351 @@
...
+        // Cancel the child
+        locked_child.is_cancelled = true;
+        drop(locked_child);
+        child.waker.notify_waiters();
+
+        // Now the child is cancelled and detached and all its children are adopted.
+        // Just continue until all (including adopted) children are cancelled and detached.
+    }
+
+    // Cancel the subtree
```
Darksonn, line 336:
```suggestion
    // Cancel the node itself.
```

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -261,20 +154,19 @@ impl CancellationToken {
     ///
     /// This will wake up all tasks which are waiting for cancellation.
     pub fn cancel(&self) {
```
Darksonn, line 156:
```suggestion
    ///
    /// Be aware that cancellation is not an atomic operation. It is possible
    /// for another thread running in parallel with a call to `cancel` to first
    /// receive `true` from `is_cancelled` on one child node, and then receive
    /// `false` from `is_cancelled` on another child node. However, once the
    /// call to `cancel` returns, all child nodes have been fully cancelled.
    pub fn cancel(&self) {
```

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -261,20 +154,19 @@ impl CancellationToken {
...
-        self.state().cancel();
+        tree_node::cancel(&self.inner);
     }
 
     /// Returns `true` if the `CancellationToken` had been cancelled
```
Darksonn, line 160:
```suggestion
    /// Returns `true` if the `CancellationToken` has been cancelled.
```

Darksonn (commented):
Some documentation related comments.

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -261,20 +154,19 @@ impl CancellationToken {
...
     pub fn is_cancelled(&self) -> bool {
-        self.state().is_cancelled()
+        tree_node::is_cancelled(&self.inner)
     }
 
     /// Returns a `Future` that gets fulfilled when cancellation is requested.
```
Darksonn, line 165:
```suggestion
    /// Returns a `Future` that gets fulfilled when cancellation is requested.
    ///
    /// The future will complete immediately if the token is already cancelled
    /// when this method is called.
    ///
    /// # Cancel safety
    ///
    /// This method is cancel safe.
```

Darksonn:
I realized that there is an alternate way to implement the future. I think this implementation is slightly simpler, and it has the advantage that it's impossible to first get a `true` from `token.is_cancelled()` followed by getting a `Pending` from polling the future.
```Rust
pin_project! {
    /// A Future that is resolved once the corresponding [`CancellationToken`]
    /// is cancelled.
    pub struct WaitForCancellationFuture<'a> {
        cancellation_token: &'a CancellationToken,
        #[pin]
        future: tokio::sync::futures::Notified<'a>,
    }
}

impl<'a> Future for WaitForCancellationFuture<'a> {
    type Output = ();

    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<()> {
        let mut this = self.project();

        loop {
            if this.cancellation_token.is_cancelled() {
                return Poll::Ready(());
            }

            if this.future.poll(cx).is_pending() {
                return Poll::Pending;
            }

            this.future.set(this.cancellation_token.inner.waker.notified());
        }
    }
}

impl CancellationToken {
    pub fn cancelled(&self) -> WaitForCancellationFuture<'_> {
        WaitForCancellationFuture {
            cancellation_token: self,
            future: self.inner.waker.notified(),
        }
    }
}

// and just delete tree_node::get_future
```

Finomnis:
> I realized that there is an alternate way to implement the future. I think this implementation is slightly simpler, and it has the advantage that it's impossible to first get a `true` from `token.is_cancelled()` followed by getting a `Pending` from polling the future.
> 
> ```rust
> pin_project! {
>     /// A Future that is resolved once the corresponding [`CancellationToken`]
>     /// is cancelled.
>     #[must_use = "futures do nothing unless polled"]
>     pub struct WaitForCancellationFuture<'a> {
>         cancellation_token: &'a CancellationToken,
>         #[pin]
>         future: tokio::sync::futures::Notified<'a>,
>     }
> }
> 
> impl<'a> Future for WaitForCancellationFuture<'a> {
>     type Output = ();
> 
>     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<()> {
>         let mut this = self.project();
> 
>         loop {
>             if this.cancellation_token.is_cancelled() {
>                 return Poll::Ready(());
>             }
> 
>             if this.future.as_mut().poll(cx).is_pending() {
>                 return Poll::Pending;
>             }
> 
>             this.future.set(this.cancellation_token.inner.waker.notified());
>         }
>     }
> }
> 
> impl CancellationToken {
>     pub fn cancelled(&self) -> WaitForCancellationFuture<'_> {
>         WaitForCancellationFuture {
>             cancellation_token: self,
>             future: self.inner.waker.notified(),
>         }
>     }
> }
> 
> // and just delete tree_node::get_future
> ```

Gotta be honest this one was a little hard for me to understand why there is no race condition in there, but I believe your code is correct. Changed.

I only had to implement a getter for the future, as I didn't want to expose the waker outside of the tree_node mod.

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -319,560 +204,18 @@ impl<'a> Future for WaitForCancellationFuture<'a> {
...
-                current_state.pack(),
-                next_state.pack(),
-                Ordering::SeqCst,
-                Ordering::SeqCst,
-            ) {
-                Ok(_) => break next_state,
-                Err(actual) => current_state = StateSnapshot::unpack(actual),
+            if this.future.as_mut().poll(cx).is_pending() {
+                return Poll::Pending;
             }
```
Darksonn, line 215: Since you found it non-obvious, we could add a comment like this one.
```suggestion
            // No wakeups can be lost here because there is always a call to
            // `is_cancelled` between the creation of the future and the call to
            // `poll`.
            if this.future.as_mut().poll(cx).is_pending() {
                return Poll::Pending;
            }
```

Finomnis:
@Darksonn What I kinda dislike about your new implementation of the future is that we will have dangling futures that will never get triggered. (E.g. every WaitForCancellationFuture that gets created after the node was cancelled). But we also never query it because we check for `is_cancelled` first. It just threw me off at first.

Darksonn:
The old implementation also created futures that we never used. They were just dropped inside `get_future` at the `return None`. Additionally, not dropping the future immediately doesn't have any cost in this case because the destructor is a no-op if you haven't polled it yet.

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,347 @@
...
+                locked_grandchild.parent = Some(node.clone());
+                locked_grandchild.parent_idx = locked_node.children.len();
+                drop(locked_grandchild);
+                locked_node.children.push(grandchild);
+            }
+        }
+
+        // Cancel the child
+        locked_child.is_cancelled = true;
+        drop(locked_child);
```
Darksonn, line 336:
```suggestion
        // Cancel the child
        locked_child.is_cancelled = true;
        locked_child.children = Vec::new();
        drop(locked_child);
```

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,347 @@
...
+        child.waker.notify_waiters();
+
+        // Now the child is cancelled and detached and all its children are adopted.
+        // Just continue until all (including adopted) children are cancelled and detached.
+    }
+
+    // Cancel the node itself.
+    locked_node.is_cancelled = true;
+    drop(locked_node);
```
Darksonn, line 345:
```suggestion
    // Cancel the node itself.
    locked_node.is_cancelled = true;
    locked_node.children = Vec::new();
    drop(locked_node);
```

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,347 @@
...
+            if locked_grandchild.is_cancelled {
+                continue;
+            }
+
+            // For performance reasons, only adopt grandchildren that have children.
+            // Otherwise, just cancel them right away, no need for another iteration.
+            if locked_grandchild.children.is_empty() {
+                // Cancel the grandchild
+                locked_grandchild.is_cancelled = true;
+                drop(locked_grandchild);
```
Darksonn, line 323:
```suggestion
                locked_grandchild.is_cancelled = true;
                locked_grandchild.children = Vec::new();
                drop(locked_grandchild);
```

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,347 @@
...
+    if parent.children.len() == pos + 1 {
+        parent.children.pop().unwrap();
+    } else {
+        // If `node` is not the last element in the list, we need to
+        // replace it with the last element
+        let replacement_child = parent.children.pop().unwrap();
+        replacement_child.inner.lock().unwrap().parent_idx = pos;
+        parent.children[pos] = replacement_child;
+    }
+}
```
Darksonn, line 229:
```suggestion
    }

    let len = parent.children.len();
    if 4*len <= parent.children.capacity() {
        parent.children.shrink_to(2*len);
    }
}
```

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,347 @@
...
+///
+/// `parent` MUST have been a parent of the node when they both got locked,
+/// otherwise there is a potential for a deadlock as invariant #2 would be violated.
+///
+/// To aquire the locks for node and parent, use [with_locked_node_and_parent].
+fn move_children_to_parent(node: &mut Inner, parent: &mut Inner) {
+    // Pre-allocate in the parent, for performance
+    parent.children.reserve(node.children.len());
+
+    for child in node.children.drain(..) {
```
Darksonn, line 194:
```suggestion
    for child in std::mem::take(&mut node.children) {
```

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,347 @@
...
+    locked_parent.children.push(child.clone());
+
+    child
+}
+
+/// Disconnects the given parent from all of its children.
+///
+/// Takes a reference to [Inner] to make sure the parent is already locked.
+fn disconnect_children(node: &mut Inner) {
+    for child in node.children.drain(..) {
```
Darksonn, line 114:
```suggestion
    for child in std::mem::take(&mut node.children) {
```

Darksonn (commented):
We should probably make sure to drop the capacity of vectors if it is much larger than the number of children, to avoid a spike in the number of children from consuming lots of memory forever.

tokio-util/src/sync/cancellation_token/tree_node.rs
Finomnis: This one seems unnecessary

Darksonn: Hmm, it might not be possible to have an empty vector with non-zero capacity if you add the others, but I don't think it hurts to include it either.

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,359 @@
...
+            }
+        };
+
+        // Loop until we managed to lock both the node and its parent
+        if Arc::ptr_eq(&actual_parent, &potential_parent) {
+            return func(locked_node, Some(locked_parent));
+        }
+
+        drop(locked_node);
+        drop(locked_parent);
```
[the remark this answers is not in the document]
Finomnis: Dropping `locked_parent` is necessary because `potential_parent` is borrowed in it, I think. The other one is to preserve drop order.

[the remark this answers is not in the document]
hawkw: ah, i see --- i missed that. maybe worth adding a comment?

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,359 @@
...
+                parent_idx: 0,
+                children: vec![],
+                is_cancelled: false,
+                num_handles: 1,
+            }),
+            waker: tokio::sync::Notify::new(),
+        }
+    }
+
+    pub(crate) fn get_notified_future(&self) -> tokio::sync::futures::Notified<'_> {
```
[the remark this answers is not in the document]
Finomnis: Sounds good.

Finomnis:
@hawkw Thanks for the review. I won't get to address them until tomorrow evening, though. Hope that's fine.

tokio-util/src/sync/cancellation_token/tree_node.rs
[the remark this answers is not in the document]
Finomnis: Potentially fixed

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,359 @@
...
+        let locked_node = node.inner.lock().unwrap();
+
+        let actual_parent = match locked_node.parent.clone() {
+            Some(parent) => parent,
+            // If we locked the node and its parent is `None`, we are in a valid state
+            // and can return.
+            None => {
+                // Drop locked_parent as it is not relevant any more
+                drop(locked_parent);
+                return func(locked_node, None);
```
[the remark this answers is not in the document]
Finomnis: Potentially fixed

tokio-util/src/sync/cancellation_token/tree_node.rs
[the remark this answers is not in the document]
Finomnis: Fixed

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,359 @@
+//! GENERAL NOTES
+//! -------------
+//!
+//! ** Invariants **
+//!
+//! Those invariants shall be true at any time, and are required to prove correctness
```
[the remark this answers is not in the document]
Finomnis: Removed the phrase

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,359 @@
+//! GENERAL NOTES
+//! -------------
+//!
+//! ** Invariants **
```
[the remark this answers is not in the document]
Finomnis: Fixed?

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,359 @@
+//! GENERAL NOTES
```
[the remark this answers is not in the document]
Finomnis: Potentially fixed. Feedback needed.

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,369 @@
...
+                return func(locked_node, None);
+            }
+        };
+
+        // Loop until we managed to lock both the node and its parent
+        if Arc::ptr_eq(&actual_parent, &potential_parent) {
+            return func(locked_node, Some(locked_parent));
+        }
+
+        // Drop locked_parent, as potential_parent is borrowed in it
```
hawkw, line 186: nit, maybe:
```suggestion
        // Drop locked_parent before reassigning to potential_parent, 
        // as potential_parent is borrowed in it
```

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,369 @@
+//! This mod provides the logic for the inner tree structure of the CancellationToken.
```
hawkw, line 1: Thanks, the summary is helpful :+1:

hawkw (commented):
Thanks, the comment changes look good to me. One very minor nitpick is that it would be nice if the comments could be wrapped at the same line length as other comments in Tokio, but that's not a hard blocker IMO.

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -0,0 +1,373 @@
...
+//!
+//! CancellationTokens are only light handles with references to TreeNode.
+//! All the logic is actually implemented in the TreeNode.
+//!
+//! A TreeNode is part of the cancellation tree and may have one parent and an arbitrary number of
+//! children.
+//!
+//! A TreeNode can receive the request to perform a cancellation through a CancellationToken.
+//! This cancellation request will cancel the node and all of its children.
```
Darksonn, line 10: Grandchildren are also cancelled.
```suggestion
//! This cancellation request will cancel the node and all of its descendants.
```

Darksonn (approved):

Ten0:
> I realized that there is an alternate way to implement the future. I think this implementation is slightly simpler, and it has the advantage that it's impossible to first get a `true` from `token.is_cancelled()` followed by getting a `Pending` from polling the future.
> 
> ```rust
> pin_project! {
>     /// A Future that is resolved once the corresponding [`CancellationToken`]
>     /// is cancelled.
>     #[must_use = "futures do nothing unless polled"]
>     pub struct WaitForCancellationFuture<'a> {
>         cancellation_token: &'a CancellationToken,
>         #[pin]
>         future: tokio::sync::futures::Notified<'a>,
>     }
> }
> 
> impl<'a> Future for WaitForCancellationFuture<'a> {
>     type Output = ();
> 
>     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<()> {
>         let mut this = self.project();
> 
>         loop {
>             if this.cancellation_token.is_cancelled() {
>                 return Poll::Ready(());
>             }
> 
>             if this.future.as_mut().poll(cx).is_pending() {
>                 return Poll::Pending;
>             }
> 
>             this.future.set(this.cancellation_token.inner.waker.notified());
>         }
>     }
> }
> 
> impl CancellationToken {
>     pub fn cancelled(&self) -> WaitForCancellationFuture<'_> {
>         WaitForCancellationFuture {
>             cancellation_token: self,
>             future: self.inner.waker.notified(),
>         }
>     }
> }
> 
> // and just delete tree_node::get_future
> ```

I'm struggling to understand the point of `this.future.set(this.cancellation_token.inner.waker.notified());`.
I'm not sure why this can't be written as:
```rust
impl<'a> Future for WaitForCancellationFuture<'a> {
    type Output = ();

    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<()> {
        let mut this = self.project();
        if this.cancellation_token.is_cancelled() {
            return Poll::Ready(());
        }

        this.future.as_mut().poll(cx)
    }
}
```

As far as I can tell this would preserve both properties of:
- We can't get a `is_cancelled` then have poll return `Pending`
- We can't get a poll return `Ready` then have `is_cancelled = false` because it's set before

NobodyXu:
@Ten0 Because once future returns Poll::Ready it usually cannot be polled again, otherwise it might panic.

Previous implementation always return Poll::Ready(()) if the token is akready cancelled and changing this might be a breaking change.

Ten0:
Wouldn't property 2 "We can't get a poll return Ready then have is_cancelled = false because it's set before" guarantee that it would still return ready immediately again without re polling the inner future?

NobodyXu:
> Wouldn't property 2 "We can't get a poll return Ready then have is_cancelled = false because it's set before" guarantee that it would still return ready immediately again without re polling the inner future? 

You are right, you can indeed rewrite it and remove the loop.

Finomnis:
I remember having had this discussion already, and there were reasons. @Darksonn might she to pitch in, she explained it to me back then

Finomnis:
I think the reason was that `Notified` is susceptible to spurious wakeups. So getting a `Ready` from it doesn't guarantee that the `CancellationToken` is actually cancelled, so we decided that the only positive return path of this function should be through an `is_cancelled` check. To make it easier to prove correctness.

Finomnis:
Regarding the order: would you mind explaining how it would be possible to first get `is_cancelled` and then afterwards get `Pending`? I can't see this flow path right now...

Ten0:
> I think the reason was that `Notified` is susceptible to spurious wakeups.

I can't see anywhere that it's susceptible to spurious wake ups though (unless those are explicitly triggered by the tree implem).
That being said if indeed `Notified` is susceptible to such (explicit?) spurious wakeups and cant't be re-polled after it has been notified then I understand this.

Thanks! :)

Finomnis:
Ah OK never mind then, I must have misunderstood your original comment then :)

To my understanding `Notified` is meant to signal that some information changed and had to be re-evaluated. It is not meant to actually propagate said information.

Ten0:
> To my understanding Notified is meant to signal that some information changed and had to be re-evaluated. It is not meant to actually propagate said information.

When said information is a boolean that can only go from false to true that might be enough. But I get the will for consistency.

Darksonn:
The implementation of `Notified` can indeed result in spurious wakeups. I don't think the cases where it can happen actually matter for this particular use-case, but I prefer to not rely on that implementation detail.

Ten0:
Alright, thanks! Might be worth mentioning those in the documentation then (did I miss them?).

Issue #4509, opened 2022-02-16
io: Refactor out usage of `Weak` in the io handle


LucioFranco:
Remove the use of `Weak` in the I/O Driver handle similar to how the timer does it. This will allow the `IoHandle` to store `Arc<Inner>` rather than require `Weak::upgrade` to access the inner timer.

The reason for `Weak` was to avoid cycles caused by wakers stored in the driver pointing to the runtime which points back to the driver. The cycle can be broken by making sure nobody can register a new io resources on (shutdown?) and by emptying the driver.

hidva:
Hi, everyone, I have communicated with Lucio and I will try to solve this problem this week.

PR #4656, opened 2022-05-05
io: Refactor out usage of Weak in the io handle

hidva:
## Motivation

Fix https://github.com/tokio-rs/tokio/issues/4509

## Solution

The cycle is broken by making sure nobody can register a new io resources on shutdown and by emptying the driver.

### DIFF
--- a/tokio/src/io/driver/mod.rs
+++ b/tokio/src/io/driver/mod.rs
@@ -18,13 +18,13 @@ mod metrics;
 
 use crate::park::{Park, Unpark};
 use crate::util::slab::{self, Slab};
-use crate::{loom::sync::Mutex, util::bit};
+use crate::{loom::sync::RwLock, util::bit};
 
 use metrics::IoDriverMetrics;
 
 use std::fmt;
 use std::io;
-use std::sync::{Arc, Weak};
+use std::sync::Arc;
 use std::time::Duration;
 
 /// I/O driver, backed by Mio.
@@ -37,10 +37,8 @@ pub(crate) struct Driver {
     events: Option<mio::Events>,
 
     /// Primary slab handle containing the state for each resource registered
-    /// with this driver. During Drop this is moved into the Inner structure, so
-    /// this is an Option to allow it to be vacated (until Drop this is always
-    /// Some).
-    resources: Option<Slab<ScheduledIo>>,
+    /// with this driver.
+    resources: Slab<ScheduledIo>,
 
     /// The system event queue.
     poll: mio::Poll,
@@ -52,7 +50,7 @@ pub(crate) struct Driver {
 /// A reference to an I/O driver.
 #[derive(Clone)]
 pub(crate) struct Handle {
-    inner: Weak<Inner>,
+    pub(super) inner: Arc<Inner>,
 }
 
 #[derive(Debug)]
@@ -61,20 +59,17 @@ pub(crate) struct ReadyEvent {
     pub(crate) ready: Ready,
 }
 
-pub(super) struct Inner {
-    /// Primary slab handle containing the state for each resource registered
-    /// with this driver.
-    ///
-    /// The ownership of this slab is moved into this structure during
-    /// `Driver::drop`, so that `Inner::drop` can notify all outstanding handles
-    /// without risking new ones being registered in the meantime.
-    resources: Mutex<Option<Slab<ScheduledIo>>>,
+struct IoDispatcher {
+    allocator: slab::Allocator<ScheduledIo>,
+    is_shutdown: bool,
+}
 
+pub(super) struct Inner {
     /// Registers I/O resources.
     registry: mio::Registry,
 
     /// Allocates `ScheduledIo` handles when creating new resources.
-    pub(super) io_dispatch: slab::Allocator<ScheduledIo>,
+    io_dispatch: RwLock<IoDispatcher>,
 
     /// Used to wake up the reactor from a call to `turn`.
     waker: mio::Waker,
@@ -130,11 +125,10 @@ impl Driver {
             tick: 0,
             events: Some(mio::Events::with_capacity(1024)),
             poll,
-            resources: Some(slab),
+            resources: slab,
             inner: Arc::new(Inner {
-                resources: Mutex::new(None),
                 registry,
-                io_dispatch: allocator,
+                io_dispatch: RwLock::new(IoDispatcher::new(allocator)),
                 waker,
                 metrics: IoDriverMetrics::default(),
             }),
@@ -149,7 +143,7 @@ impl Driver {
     /// to bind them to this event loop.
     pub(crate) fn handle(&self) -> Handle {
         Handle {
-            inner: Arc::downgrade(&self.inner),
+            inner: Arc::clone(&self.inner),
         }
     }
 
@@ -160,7 +154,7 @@ impl Driver {
         self.tick = self.tick.wrapping_add(1);
 
         if self.tick == COMPACT_INTERVAL {
-            self.resources.as_mut().unwrap().compact()
+            self.resources.compact()
         }
 
         let mut events = self.events.take().expect("i/o driver event store missing");
@@ -194,7 +188,7 @@ impl Driver {
     fn dispatch(&mut self, token: mio::Token, ready: Ready) {
         let addr = slab::Address::from_usize(ADDRESS.unpack(token.0));
 
-        let resources = self.resources.as_mut().unwrap();
+        let resources = &mut self.resources;
 
         let io = match resources.get(addr) {
             Some(io) => io,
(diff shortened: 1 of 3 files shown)


hawkw:
this no longer seems to compile (i think because `inner` is no longer an `Option`?): https://github.com/tokio-rs/tokio/runs/6311895497?check_suite_focus=true

hidva:
@LucioFranco  Please take a look, I have done a benchmark with [monoio-benchmark](https://github.com/monoio-rs/monoio-benchmark), and there is no performance fallback.

LucioFranco:
@carllerche can you take a look at this?

tokio/src/io/driver/mod.rs
```diff
@@ -212,27 +215,6 @@ impl Driver {
     }
 }
 
-impl Drop for Driver {
-    fn drop(&mut self) {
-        (*self.inner.resources.lock()) = self.resources.take();
-    }
-}
-
-impl Drop for Inner {
```
LucioFranco, line 221: I think we still want this impl, to shutdown on drop?

tokio/src/runtime/metrics/runtime.rs
```diff
@@ -529,7 +529,7 @@ cfg_net! {
                 .as_inner()
                 .io_handle
                 .as_ref()
-                .and_then(|h| h.with_io_driver_metrics(f))
+                .map(|h| f(h.metrics()))
```
LucioFranco, line 532: We should update this fn to return the inner metrics type, originally this was a fn because of some of the option stuff but now we should be able to return a handle. There is an open PR for the timer metrics which shows the correct impl of this fn.

LucioFranco (changes requested):
Looks like a good start! I would love some other eyes on this as well though cc @carllerche

tokio/src/io/driver/mod.rs
```diff
@@ -61,20 +59,26 @@ pub(crate) struct ReadyEvent {
...
-    /// The ownership of this slab is moved into this structure during
-    /// `Driver::drop`, so that `Inner::drop` can notify all outstanding handles
-    /// without risking new ones being registered in the meantime.
-    resources: Mutex<Option<Slab<ScheduledIo>>>,
+struct IoDispatcher {
+    allocator: slab::Allocator<ScheduledIo>,
+    is_shutdown: bool,
+}
 
+impl IoDispatcher {
```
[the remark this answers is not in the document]
hidva: done

tokio/src/io/driver/mod.rs
```diff
@@ -316,13 +299,14 @@ impl Handle {
...
+            .waker
+            .wake()
+            .expect("failed to wake I/O driver");
     }
 
-    pub(super) fn inner(&self) -> Option<Arc<Inner>> {
-        self.inner.upgrade()
+    pub(super) fn inner(&self) -> &Arc<Inner> {
+        &self.inner
     }
```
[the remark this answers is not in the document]
hidva: done

tokio/src/io/driver/mod.rs
hidva: done

tokio/src/runtime/metrics/runtime.rs
hidva: I think you're talking about `Handle::with_io_driver_metrics()`? which has been replaced by `Handle::metrics()`.

As for `RuntimeMetrics::with_io_driver_metrics()`, if we delete this function, there will be a lot of duplicate code with pattern `self.handle.as_inner().io_handle.as_ref().map(|h|h.metrics().METRIC_NAME.load(Relaxed)).unwrap_or(0)`.

tokio/src/io/driver/mod.rs
```diff
@@ -374,6 +355,36 @@ impl Inner {
 
         Ok(())
     }
+
+    /// shutdown the dispatcher.
+    fn shutdown(&self) -> bool {
+        let mut io = self.io_dispatch.write().unwrap();
```
renancloudwalk, line 361: forgive me if this is a dumb question:

is it ok for this `unwrap` to cause panics?  Isn't it better to use a `Result`/`io::Error` ?

renancloudwalk (approved):
👍

tokio/src/io/driver/mod.rs
hidva: It is fine, see [RwLock::write()](https://doc.rust-lang.org/stable/std/sync/struct.RwLock.html#method.write)

tokio/src/runtime/metrics/runtime.rs
LucioFranco: I thought we don't need to do the unwrap/map anymore since the handle contains an Arc not a Weak?

hidva: `self.handle.as_inner().io_handle` is still an Option, `Option<crate::io::driver::Handle>`, just as:

```rust
pub fn time_driver_entry_count(&self) -> u64 {
    // unwrap_or here.
    self.handle.time_handle.as_ref().map(|h| h.metrics().entry_count.load(Relaxed)).unwrap_or(0)
}
```

hidva: @LucioFranco ping~

LucioFranco: Ok yes this seems fine to me.

hidva: So what else do I need to do here?

LucioFranco: Nothing here, I think.

tokio/src/io/driver/mod.rs
```diff
@@ -292,18 +280,11 @@ cfg_not_rt! {
...
-                F: Fn(&IoDriverMetrics) -> R,
-            {
-                self.inner().map(|inner| f(&inner.metrics))
-            }
-       }
+        impl Handle {
+                pub(crate) fn metrics(&self) -> &IoDriverMetrics {
+                    &self.inner.metrics
+                }
+        }
```
Darksonn, line 287: The formatting is wonky here.

hidva: It's really a little strange. I've used rustfmt repeatedly.

image

hidva: Could you help me point out where it needs to be adjusted? I didn't see it[sad].

hidva: done

tokio/src/io/driver/mod.rs
```diff
@@ -280,11 +280,11 @@ cfg_not_rt! {
...
-        impl Handle {
-                pub(crate) fn metrics(&self) -> &IoDriverMetrics {
-                    &self.inner.metrics
-                }
-        }
+       impl Handle {
+            pub(crate) fn metrics(&self) -> &IoDriverMetrics {
+                &self.inner.metrics
+            }
+       }
```
Darksonn, line 287: The `impl Handle` is currently indented by seven spaces rather than eight.

tokio/src/io/driver/mod.rs
```diff
@@ -251,7 +230,16 @@ impl Park for Driver {
         Ok(())
     }
 
-    fn shutdown(&mut self) {}
+    fn shutdown(&mut self) {
+        if self.inner.shutdown() {
+            self.resources.for_each(|io| {
```
carllerche, line 235: If this statement is commented out, do tests fail? i.e. if there is a memory leak, is it caught my CI? If not, we should add one.

carllerche (commented):
Thanks for doing this. The general direction looks good to me. I would like to double check that we have CI tests that check for memory leaks (valgrind). Given that a bug here will result in a mem leak, it would be good to check :) I left a note inline where, if commented, I would hope CI would fail.

Thanks!

carllerche:
There are a few places valgrind is used in CI:
* https://github.com/tokio-rs/tokio/blob/master/.github/workflows/ci.yml#L151
* https://github.com/tokio-rs/tokio/blob/master/.github/workflows/stress-test.yml#L41

I wonder if we could run our entire CI suite w/ valgrind...

Darksonn:
Regarding leak checks, see the recently merged #4696. You may want to merge in master to get this step in the PR's CI.

tokio/src/io/driver/mod.rs
hidva: done.

Sorry, because I see the last commit is indented like this.

tokio/src/io/driver/mod.rs
hidva: Yes. 

First, we have to comment out [clear_wakers()](https://github.com/hidva/tokio/blob/43fce8970e483138ebdca57bd9af32e115d007bd/tokio/src/io/driver/registration.rs#L209).

Second, if we commented out `io.shutdown()` here, we'd get a memory leak, otherwise we wouldn't.

```
==756352== LEAK SUMMARY:
==756352==    definitely lost: 96 bytes in 1 blocks
==756352==    indirectly lost: 20,856 bytes in 35 blocks
==756352==      possibly lost: 0 bytes in 0 blocks
==756352==    still reachable: 9,952 bytes in 69 blocks
==756352==         suppressed: 0 bytes in 0 blocks
==756352==
==756352== For lists of detected and suppressed errors, rerun with: -s
==756352== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
```

(It looks like we don't need to call `clear_wakers()` in `Drop for Registration ` anymore.

hidva:
> Regarding leak checks, see the recently merged #4696. You may want to merge in master to get this step in the PR's CI.

done

Darksonn (approved):

Darksonn:
Thanks!

PR #4676, opened 2022-05-10
docs: mention that Clippy must be run with the MSRV

SabrinaJewson:
## Motivation

In #4675 I learnt the hard way that Tokio uses Clippy on its MSRV. 

## Solution

Document this in the contributor's guide.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,10 @@ env:
   # - README.md
   # - tokio/README.md
   # - CONTRIBUTING.md
+  # - tokio/Cargo.toml
+  # - tokio-util/Cargo.toml
+  # - tokio-test/Cargo.toml
+  # - tokio-stream/Cargo.toml
   rust_min: 1.49.0
 
 defaults:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,10 @@ env:
   rust_stable: stable
   rust_nightly: nightly-2022-03-21
   rust_clippy: 1.52.0
+  # When updating this, also update:
+  # - README.md
+  # - tokio/README.md
+  # - CONTRIBUTING.md
   rust_min: 1.49.0
 
 defaults:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -131,6 +131,20 @@ cargo check --all-features
 cargo test --all-features
 ```
 
+Clippy must be run using the MSRV, so Tokio can avoid having to `#[allow]` new
+lints whose fixes would be incompatible with the current MSRV:
+
+<!--
+When updating this, also update:
+- .github/workflows/ci.yml
+- README.md
+- tokio/README.md
+-->
+
+```
+cargo +1.49.0 clippy --all-features
+```
+
 When building documentation normally, the markers that list the features
 required for various parts of Tokio are missing. To build the documentation
 correctly, use this command:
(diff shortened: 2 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -131,6 +131,20 @@ cargo check --all-features
...
 
+Clippy must be run using the MSRV, so Tokio can avoid having to `#[allow]` new
+lints whose fixes would be incompatible with the current MSRV:
+
+<!--
+When updating this, also update:
+- .github/workflows/ci.yml
+- README.md
+- tokio/README.md
+-->
```
hawkw, line 142: it would be really nice to have a script for this or something, or some way to to have a single authoritative definition of the MSRV that is referenced elsewhere. there are now several places where the MSRV will need to be updated.

obviously, this would be best left for a separate PR, but i thought it was worth noting.

.github/workflows/ci.yml
```diff
@@ -13,6 +13,10 @@ env:
   rust_stable: stable
   rust_nightly: nightly-2022-03-21
   rust_clippy: 1.52.0
+  # When updating this, also update:
+  # - README.md
+  # - tokio/README.md
+  # - CONTRIBUTING.md
```
hawkw, line 19: i believe `Cargo.toml`'s `rust-version` keys will also need to be updated (e.g. https://github.com/tokio-rs/tokio/blob/71e18f7b754b29741542bc21bdf99cd30b2fd605/tokio/Cargo.toml#L11)
```suggestion
  # - CONTRIBUTING.md
  # - tokio/Cargo.toml
  # - tokio-util/Cargo.toml
  # - tokio-test/Cargo.toml
  # - tokio-stream/Cargo.toml
```

tokio/README.md
```diff
@@ -163,6 +163,12 @@ several other libraries, including:
 
 ## Supported Rust Versions
 
+<!--
+When updating this, also update:
+- .github/workflows/ci.yml
+- CONTRIBUTING.md
```
hawkw, line 169: - README.md
- tokio/Cargo.toml
- tokio-util/Cargo.toml
- tokio-test/Cargo.toml
- tokio-stream/Cargo.toml

hawkw (approved):
thanks for documenting this, this is definitely good to have written down!

i wonder if we want to add a bash script or makefile or something for running clippy with the correct version, eventually?

i suggested adding some other places where we record the MSRV in the lists of places that need to be updated, but besides that, this change looks good to me!

hawkw:
ugh...i forgot that we have a CI job that checks that the two README files are identical, which is now failing because they list different other locations that need to be updated: https://github.com/tokio-rs/tokio/runs/6374617535?check_suite_focus=true

we should probably just add each README's own path to the list of updated items so that CI job can pass.

README.md
```diff
@@ -163,6 +163,17 @@ several other libraries, including:
 
 ## Supported Rust Versions
 
+<!--
+When updating this, also update:
+- .github/workflows/ci.yml
+- CONTRIBUTING.md
+- tokio/README.md
```
hawkw, line 170: adding this to fix the "both readmes are identical" CI job
```suggestion
- README.md
- tokio/README.md
```

tokio/README.md
```diff
@@ -163,6 +163,17 @@ several other libraries, including:
...
+- README.md
```
hawkw, line 170: adding this to fix the "both readmes are identical" CI job
```suggestion
- README.md
- tokio/README.md
```

PR #4683, opened 2022-05-12
task: add `Builder::{spawn_on, spawn_local_on, spawn_blocking_on}`

hawkw:
## Motivation

`task::JoinSet` currently has both `spawn`/`spawn_local` methods,
and `spawn_on`/`spawn_local_on` variants of these methods that take a
reference to a runtime `Handle` or to a `LocalSet`, and spawn tasks on
the provided runtime/`LocalSet`, rather than the current one. The
`task::Builder` type is _also_ an API type that can spawn tasks, but it
doesn't have `spawn_on` variants of its methods. It occurred to me that
it would be nice to have similar APIs on `task::Builder`.

## Solution

This branch adds `task::Builder::spawn_on`,
`task::Builder::spawn_local_on`, and `task::Builder::spawn_blocking_on`
methods, similar to those on `JoinSet`. In addition, I did some
refactoring of the internal spawning APIs --- there was a bit of
duplicated code that this PR reduces.

### DIFF
--- a/tokio/src/runtime/context.rs
+++ b/tokio/src/runtime/context.rs
@@ -68,6 +68,15 @@ cfg_time! {
     }
 }
 
+cfg_rt! {
+    pub(crate) fn spawn_handle() -> Option<crate::runtime::Spawner> {
+        match CONTEXT.try_with(|ctx| (*ctx.borrow()).as_ref().map(|ctx| ctx.spawner.clone())) {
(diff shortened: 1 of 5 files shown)


Noah-Kennedy (approved):

carllerche:
Was there a consideration to add a builder method for `handle(&mut self, &Handle)` instead of `on` methods for each spawn variant?

PR #4696, opened 2022-05-15
Enable ASAN in CI

Finomnis:
## Motivation

LLVM's ASAN is a very useful tool to find memory leaks, and it is built-in for the rust nightly compier.

When I ran it on my local computer, there were several memory leaks detected in the tests.
According to the LLVM docs, ASAN does not produce false positives.

Sadly, ASAN is not yet enabled in the CI.

## Solution
- [x] Enable ASAN in CI
- [x] Fix leaks


Darksonn (approved):
Seems reasonable.

PR #4716, opened 2022-05-24
Wasm Improvements WiP

rjzak:
Signed-off-by: Richard Zak <richard@profian.com>

## Motivation

Trying to get Tokio working on WebAssembly

## Solution

Working on supporting the `wasm32-wasi` target. The use of Tokio with WebAssembly should be seamless and easy to the extent possible.

## Testing

[rjzak/tokio-echo-test](https://github.com/rjzak/tokio-echo-test) is a simple echo server and provides an example server running with WebAssembly. It currently crashes when the first connection comes in, but works fine as a Linux binary.


carllerche:
Could you include the output you get? Ideally w/ a full backtrace?

rjzak:
I'm not sure how to get a full stack trace, but this is the output where a connection with `nc` is made after "Listening" displays.
```
Listening.
Error: Error { kind: Unsupported, message: "operation not supported on this platform" }
Error: Exited with i32 exit status 1
```

rjzak:
Two things:
1. I set the target toolchain to nightly, and further progress was made. Now the Wasm binary is able to receive connections. However, it doesn't respond.
2. I did add the print statements to track execution, as suggested.
```
Listening.
Tokio TcpListener::accept() /home/rjzak/CLionProjects/tokio/tokio/src/net/tcp/listener.rs:161
Tokio Registration::async_io() /home/rjzak/CLionProjects/tokio/tokio/src/io/driver/registration.rs:250
Tokio Registration::async_io() self.readiness(READABLE) = ReadyEvent { tick: 1, ready: Ready { is_readable: true, is_writable: false, is_read_closed: true, is_write_closed: false } } returned @ /home/rjzak/CLionProjects/tokio/tokio/src/io/driver/registration.rs:253
wasi::accept() /home/rjzak/CLionProjects/mio/src/sys/wasi/mod.rs:32
wasi::accept() listener.accept() OK /home/rjzak/CLionProjects/mio/src/sys/wasi/mod.rs:34
wasi::accept() stream.set_nonblocking(true) /home/rjzak/CLionProjects/mio/src/sys/wasi/mod.rs:36
Tokio TcpListener::accept() self.io.registration().async_io().await /home/rjzak/CLionProjects/tokio/tokio/src/net/tcp/listener.rs:167
Tokio TcpListener::accept() TcpStream::new() /home/rjzak/CLionProjects/tokio/tokio/src/net/tcp/listener.rs:169
Connection received.
Tokio TcpListener::accept() /home/rjzak/CLionProjects/tokio/tokio/src/net/tcp/listener.rs:161
Tokio Registration::async_io() /home/rjzak/CLionProjects/tokio/tokio/src/io/driver/registration.rs:250
Tokio Registration::async_io() self.readiness(READABLE) = ReadyEvent { tick: 1, ready: Ready { is_readable: true, is_writable: false, is_read_closed: false, is_write_closed: false } } returned @ /home/rjzak/CLionProjects/tokio/tokio/src/io/driver/registration.rs:253
wasi::accept() /home/rjzak/CLionProjects/mio/src/sys/wasi/mod.rs:32
```
What seems to be missing is writing the response data to the stream to send back to the client. Does work as a native Linux binary, so I'm probably just missing something...

rjzak:
Tokio side, trying to read the contents early, I'm getting "operation not supported" when trying to `peek()`, and "operation would block" when trying to `try_read()`. What's the best way to read early on the Tokio side? The socket is already set to allow blocking I/O explicitly on the Wasi side https://github.com/tokio-rs/mio/blob/master/src/sys/wasi/mod.rs#L33.

```
Listening.
Tokio TcpListener::accept() /home/rjzak/CLionProjects/tokio/tokio/src/net/tcp/listener.rs:161
Tokio Registration::async_io() /home/rjzak/CLionProjects/tokio/tokio/src/io/driver/registration.rs:250
Tokio Registration::async_io() self.readiness(READABLE) = ReadyEvent { tick: 1, ready: Ready { is_readable: true, is_writable: false, is_read_closed: true, is_write_closed: false } } returned @ /home/rjzak/CLionProjects/tokio/tokio/src/io/driver/registration.rs:253
wasi::accept() /home/rjzak/CLionProjects/mio/src/sys/wasi/mod.rs:32
wasi::accept() listener.accept() OK /home/rjzak/CLionProjects/mio/src/sys/wasi/mod.rs:34
wasi::accept() stream.set_nonblocking(true) /home/rjzak/CLionProjects/mio/src/sys/wasi/mod.rs:36
Tokio TcpListener::accept() self.io.registration().async_io().await /home/rjzak/CLionProjects/tokio/tokio/src/net/tcp/listener.rs:167
Tokio TcpListener::accept() error operation not supported on this platform trying to peek() at the buffer /home/rjzak/CLionProjects/tokio/tokio/src/net/tcp/listener.rs:175
Tokio TcpListener::accept() error operation would block trying to TcpStream::try_read() at the buffer /home/rjzak/CLionProjects/tokio/tokio/src/net/tcp/listener.rs:185
Tokio TcpListener::accept() TcpStream::new() /home/rjzak/CLionProjects/tokio/tokio/src/net/tcp/listener.rs:188
Connection received.
Tokio TcpListener::accept() /home/rjzak/CLionProjects/tokio/tokio/src/net/tcp/listener.rs:161
Tokio Registration::async_io() /home/rjzak/CLionProjects/tokio/tokio/src/io/driver/registration.rs:250
Tokio Registration::async_io() self.readiness(READABLE) = ReadyEvent { tick: 1, ready: Ready { is_readable: true, is_writable: false, is_read_closed: false, is_write_closed: false } } returned @ /home/rjzak/CLionProjects/tokio/tokio/src/io/driver/registration.rs:253
wasi::accept() /home/rjzak/CLionProjects/mio/src/sys/wasi/mod.rs:32
```

tokio/src/lib.rs
```diff
@@ -410,7 +410,9 @@ mod future;
 pub mod io;
 pub mod net;
 
+#[cfg(not(target_arch = "wasm32-wasi"))]
 mod loom;
+#[cfg(not(target_arch = "wasm32-wasi"))]
 mod park;
```
Darksonn, line 416: Surely this does not compile. The `loom` module is required by lots of stuff in `tokio::sync`.

rjzak: It does compile and works, I've been using this project to test Tokio & Axum: https://github.com/rjzak/axum-wasm-test

tokio-util/src/net/mod.rs
```diff
@@ -39,9 +39,16 @@ impl Listener for tokio::net::TcpListener {
...
+    #[cfg(not(target_os = "wasi"))]
     fn local_addr(&self) -> Result<Self::Addr> {
         self.local_addr().map(Into::into)
     }
+
+    #[cfg(target_os = "wasi")]
+    fn local_addr(&self) -> Result<Self::Addr> {
+        use std::io;
+        Err(io::ErrorKind::Unsupported.into())
+    }
```
Darksonn, line 51: Please put the cfg inside the function.
```suggestion
    fn local_addr(&self) -> Result<Self::Addr> {
        #[cfg(not(target_os = "wasi"))]
        self.local_addr().map(Into::into)

        #[cfg(target_os = "wasi")]
        Err(std::io::ErrorKind::Unsupported.into())
    }
```

tokio/src/io/driver/mod.rs
```diff
@@ -72,6 +72,7 @@ pub(super) struct Inner {
     io_dispatch: RwLock<IoDispatcher>,
 
     /// Used to wake up the reactor from a call to `turn`.
+    #[cfg(not(target_os = "wasi"))]
     waker: mio::Waker,
```
Darksonn, line 76: This needs a comment explaining why it doesn't exist on WASM.

tokio/src/io/poll_evented.rs
```diff
@@ -126,6 +126,7 @@ impl<E: Source> PollEvented<E> {
 
     /// Deregisters the inner io from the registration and returns a Result containing the inner io.
     #[cfg(any(feature = "net", feature = "process"))]
+    #[cfg(not(target_os = "wasi"))]
     pub(crate) fn into_inner(mut self) -> io::Result<E> {
```
Darksonn, line 130: Why don't these methods exist on WASM?

tokio/src/macros/cfg.rs
```diff
@@ -154,13 +155,20 @@ macro_rules! cfg_not_io_util {
 
 macro_rules! cfg_loom {
     ($($item:item)*) => {
-        $( #[cfg(loom)] $item )*
+        $(
+            #[cfg(loom)]
+            #[cfg(not(target_os = "wasi"))]
+            $item
+        )*
```
Darksonn, line 162: We can just not compile loom tests on WASM rather than disabling them like this.

tokio/src/macros/cfg.rs
```diff
@@ -61,6 +61,7 @@ macro_rules! cfg_fs {
     ($($item:item)*) => {
         $(
             #[cfg(feature = "fs")]
+            #[cfg(not(target_arch = "wasm32"))]
             #[cfg_attr(docsrs, doc(cfg(feature = "fs")))]
             $item
```
Darksonn, line 66: Here you are putting these annotations on lots of stuff in this file. Ideally, I would like to avoid that.

tokio/src/net/tcp/listener.rs
```diff
@@ -94,6 +96,7 @@ impl TcpListener {
     ///     Ok(())
     /// }
     /// ```
+    #[cfg(not(target_os = "wasi"))]
     pub async fn bind<A: ToSocketAddrs>(addr: A) -> io::Result<TcpListener> {
```
Darksonn, line 100: All of the constructors in this file appear to have been disabled. How do we create a TcpListener?

rjzak: These have to be created on behalf of the WebAssembly application by the WebAssembly runtime, which then provides the socket file descriptor to the Wasm application.

tokio/src/io/poll_evented.rs
rjzak: We can't poll events on Wasi since threading it's available.

tokio/src/macros/cfg.rs
rjzak: This disables filesystem code for Wasm, since Wasm can't access files by itself. I'm not sure how else to remove features. The cfg macros seemed like a nice global way to handle features.

tokio/src/io/driver/mod.rs
rjzak: Comment added

.github/workflows/ci.yml
```diff
@@ -159,6 +160,25 @@ jobs:
...
+            toolchain: ${{ env.rust_stable }}
+            override: true
+      - uses: Swatinem/rust-cache@v1
+      # Fetch wasm32-wasi target support
+      - name: install Wasm32 target
+        run: rustup target add wasm32-wasi
+      # Compile
+      - name: cargo build --target=wasm32-wasi
+        run: cargo build --target=wasm32-wasi
+        working-directory: tokio
```
Darksonn, line 180: Can you change this to build all of the crates? I think you can just remove the working directory option to do it.

tokio/src/net/tcp/listener.rs
npmccallum: @Darksonn To answer that question more directly: `FromRawFd`. The WebAssembly runtime pre-creates the initial file descriptors and passes them to the module. There is not currently any way to dynamically bind.

.github/workflows/ci.yml
rjzak: Done

tokio/src/io/poll_evented.rs
npmccallum: @rjzak You can poll. You just cannot wake up another thread since there are no threads (yet).

rjzak: This was fixed

tokio-macros/src/entry.rs
```diff
@@ -70,11 +79,17 @@ impl Configuration {
...
+        {
+            let runtime_str = parse_string(runtime, span, "flavor")?;
+            let runtime =
+                RuntimeFlavor::from_str(&runtime_str).map_err(|err| syn::Error::new(span, err))?;
+            self.flavor = Some(runtime);
+        }
+        #[cfg(target_os = "wasi")]
+        {
+            self.flavor = Some(RuntimeFlavor::CurrentThread);
+        }
```
Darksonn, line 92: As far as I can tell, you're just ignoring the macro options here. We shouldn't do that. If somebody specifies `#[tokio::main(flavor = "multi_thread")]` (as opposed to not specifying any flavor and using the default), then we should emit an error message saying that threading is not available on WASM. The macro already has support for detecting the difference between "use the default" and "explicitly set the option to its default value".

tokio/src/macros/cfg.rs
haraldh: the real issue is the asyncify of all fs operations with threads, since wasi does not support threads

tokio-macros/src/entry.rs
```diff
@@ -363,7 +363,13 @@ fn parse_knobs(mut input: syn::ItemFn, is_test: bool, config: FinalConfig) -> To
...
         },
         RuntimeFlavor::Threaded => quote_spanned! {last_stmt_start_span=>
-            #crate_ident::runtime::Builder::new_multi_thread()
+            #[cfg(target_os = "wasi")]
+            let a = #crate_ident::runtime::Builder::new_current_thread();
+
+            #[cfg(not(target_os = "wasi"))]
+            let a = #crate_ident::runtime::Builder::new_multi_thread();
+
+            a
```
Darksonn, line 372: It is better to implement the macro in a way such that it emits an error if you try to use the multi-threaded runtime, instead of just silently using the current-thread runtime instead. You can do this by modifying the `rt_multi_thread_available` and `default_flavor` fields here:

https://github.com/tokio-rs/tokio/blob/8aa55f1eacc2fd8930f4f0855441845feddc6482/tokio-macros/src/entry.rs#L56-L60

Then introduce a wasm-specific error here:

https://github.com/tokio-rs/tokio/blob/8aa55f1eacc2fd8930f4f0855441845feddc6482/tokio-macros/src/entry.rs#L144-L151

theduke:
Excited to see this happening, thanks for putting in the work!

rjzak:
@Thomasdezeeuw is it possible to get a Mio point release so Cargo.toml could point to a version from crates.io instead of the Git repo directly? It would resolve the last CI issue.

tokio/src/task/mod.rs
```diff
@@ -279,6 +279,7 @@ cfg_rt! {
     pub use crate::runtime::task::{JoinError, JoinHandle};
 
     mod blocking;
+    #[cfg(not(target_os = "wasi"))]
     pub use blocking::spawn_blocking;
```
Darksonn, line 283: This should probably mark `block_in_place` too.

Thomasdezeeuw:
> @Thomasdezeeuw is it possible to get a Mio point release so Cargo.toml could point to a version from crates.io instead of the Git repo directly? It would resolve the last CI issue.

I think I can do a point release this weekend.

tests-integration/tests/macros_select.rs
```diff
@@ -4,6 +4,7 @@ use futures::channel::oneshot;
 use futures::executor::block_on;
 use std::thread;
 
+#[cfg_attr(target_os = "wasi", ignore = "WASI: std::thread::spawn not supported")]
```
carllerche, line 7: I'm OK w/ this for now, but maybe we can open a tracking issue to remove this? The test itself doesn't *look* like it needs  thread spawning. I'm not exactly sure why it is the way it is.

tokio/src/net/mod.rs
```diff
@@ -33,10 +33,13 @@ cfg_net! {
 
     pub mod tcp;
     pub use tcp::listener::TcpListener;
+    #[cfg(not(target_arch = "wasm32"))]
```
carllerche, line 36: I would consider making these annotations a macro. The reason being, it is easy to typo `"wasm32"` and not notice it.

tokio/src/net/tcp/listener.rs
carllerche: Same here, I would consider making a `cfg_not_wasi` { ... }` macro

tokio/src/net/mod.rs
carllerche: Another reason for using a macro is, if we end up *removing* the macro, we can delete the macro and then find the usages using the compiler. If we want to remove `#[cfg(not(target_arch = "wasm32"))` we then have to hunt through the code base.

tokio/src/net/tcp/stream.rs
```diff
@@ -1096,10 +1104,17 @@ impl TcpStream {
...
     /// ```
+    #[cfg(not(target_os = "wasi"))]
     pub fn linger(&self) -> io::Result<Option<Duration>> {
         socket2::SockRef::from(self).linger()
     }
 
+    /// Returns `io::ErrorKind::Unsupported` as this is not supported on WebAssembly
+    #[cfg(target_os = "wasi")]
+    pub fn linger(&self) -> io::Result<Option<Duration>> {
+        Err(io::ErrorKind::Unsupported.into())
```
carllerche, line 1115: Why have this method at all if it is not supported. The method could just be omitted on platforms that do not support it?

carllerche (commented):
Looking good. I left some comments inline. Besides that, a couple of thoughts:

* Not a blocker, but consider using more cfg_macros instead of attributes. You can define new ones for wasi / not_wasi
* I would prefer removing socket methods (like `linger()`) that cannot work with wasi instead of returning an "unsupported" error. Is there a reason to keep those methods?
* Do we have tests covering the wasi functionality? I know that the FD has to be passed in instead. Maybe there can be some "integration" level tests that demonstrate this behavior and ensures we catch regressions?

tokio/src/task/mod.rs
rjzak: `block_in_place` is wrapped by `cfg_rt_multi_thread` which already excludes WebAssembly.

tokio-util/src/net/mod.rs
```diff
@@ -40,7 +40,11 @@ impl Listener for tokio::net::TcpListener {
     }
 
     fn local_addr(&self) -> Result<Self::Addr> {
-        self.local_addr().map(Into::into)
+        #[cfg(not(target_os = "wasi"))]
```
Thomasdezeeuw, line 43: I don't think is necessary, although it's not supported we can leave this up to Mio/std, which doesn't support it, but we this way we don't need a `cfg` here).

haraldh:
Suggested patch:

```diff
diff --git a/tokio/tests/io_buf_reader.rs b/tokio/tests/io_buf_reader.rs
index 0d3f6baf..7ede33ac 100644
--- a/tokio/tests/io_buf_reader.rs
+++ b/tokio/tests/io_buf_reader.rs
@@ -222,6 +222,7 @@ async fn test_short_reads() {
     assert_eq!(reader.read(&mut buf).await.unwrap(), 0);
 }
 
+#[cfg_attr(target_os = "wasi", ignore = "FIXME: empty poll in park")]
 #[tokio::test]
 async fn maybe_pending() {
     let inner: &[u8] = &[5, 6, 7, 0, 1, 2, 3, 4];
@@ -259,6 +260,7 @@ async fn maybe_pending() {
     assert_eq!(reader.read(&mut buf).await.unwrap(), 0);
 }
 
+#[cfg_attr(target_os = "wasi", ignore = "FIXME: empty poll in park")]
 #[tokio::test]
 async fn maybe_pending_buf_read() {
     let inner = MaybePending::new(&[0, 1, 2, 3, 1, 0]);
@@ -278,6 +280,7 @@ async fn maybe_pending_buf_read() {
 }
 
 // https://github.com/rust-lang/futures-rs/pull/1573#discussion_r281162309
+#[cfg_attr(target_os = "wasi", ignore = "FIXME: empty poll in park")]
 #[tokio::test]
 async fn maybe_pending_seek() {
     struct MaybePendingSeek<'a> {
```

haraldh:
Suggested patch, because of empty poll in park:
https://github.com/haraldh/tokio/commit/f4fec5e5276c528e4944652eed1cc8e535d287fd

tokio/src/util/linked_list.rs
```diff
@@ -623,68 +623,69 @@ mod tests {
...
-    proptest::proptest! {
-        #[test]
-        fn fuzz_linked_list(ops: Vec<usize>) {
-            run_fuzz(ops);
+    cfg_not_wasi! {
+        proptest::proptest! {
+            #[test]
+            fn fuzz_linked_list(ops: Vec<usize>) {
+                run_fuzz(ops);
+            }
```
haraldh, line 631: `#[cfg(not(target_arch = "wasm32"))]` is not equivalent to `cfg_not_wasi!`

tokio/src/runtime/mod.rs
```diff
@@ -173,10 +173,11 @@
 //! [`Builder::enable_all`]: crate::runtime::Builder::enable_all
 
 // At the top due to macros
-#[cfg(test)]
-#[cfg(not(target_arch = "wasm32"))]
-#[macro_use]
-mod tests;
+cfg_not_wasi! {
```
haraldh, line 176: #[cfg(not(target_arch = "wasm32"))] is not equivalent to cfg_not_wasi!

rjzak: Resolved, I put back the existing !wasm check

tokio-util/src/net/mod.rs
rjzak: Done

tokio/src/net/tcp/stream.rs
rjzak: Removed

hamza1311:
In case anyone else is wondering about lack of threading support, see [`wasi-native-threads` proposal](https://github.com/WebAssembly/wasi-native-threads). The WASI standard does not support native threads yet

tokio/tests/io_fill_buf.rs
```diff
@@ -1,5 +1,5 @@
 #![warn(rust_2018_idioms)]
-#![cfg(feature = "full")]
+#![cfg(all(feature = "full", not(target_os = "wasi")))]
 
 use tempfile::NamedTempFile;
```
Darksonn, line 4: It's unfortunate that this test is written to use a file. It means that `fill_buf` is not tested on WASM.

You don't have to fix this in your PR.

tokio/tests/join_handle_panic.rs
```diff
@@ -1,5 +1,5 @@
 #![warn(rust_2018_idioms)]
-#![cfg(feature = "full")]
+#![cfg(all(feature = "full", not(target_os = "wasi")))]
 
 struct PanicsOnDrop;
```
Darksonn, line 5: This test should run on WASM too. (Unless catching panics is somehow unsupported on WASM?)

Darksonn (commented):
My general comment is that you have disabled a lot of tests that _should_ work. This makes it seem to me that we are missing something important in the runtime.

tokio-stream/tests/watch.rs
```diff
@@ -5,6 +5,7 @@ use tokio_stream::wrappers::WatchStream;
 use tokio_stream::StreamExt;
 
 #[tokio::test]
+#[cfg_attr(target_os = "wasi", ignore = "FIXME: empty poll in park")]
 async fn message_not_twice() {
```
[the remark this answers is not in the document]
rjzak: I believe this one is due to https://github.com/tokio-rs/mio/pull/1580

tokio-util/tests/time_delay_queue.rs
```diff
@@ -778,6 +778,7 @@ async fn compact_change_deadline() {
     assert!(entry.is_none());
 }
 
+#[cfg_attr(target_os = "wasi", ignore = "FIXME: Does not seem to work with WASI")]
 #[tokio::test(start_paused = true)]
 async fn remove_after_compact() {
     let now = Instant::now();
```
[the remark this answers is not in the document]
rjzak: Because it uses `std::panic::catch_unwind()`, which isn't supported by Wasi.

tokio/src/net/tcp/listener.rs
```diff
@@ -55,68 +58,70 @@ cfg_net! {
...
-    ///
-    ///     # let _ = listener;
-    ///     Ok(())
-    /// }
-    /// ```
-    pub async fn bind<A: ToSocketAddrs>(addr: A) -> io::Result<TcpListener> {
-        let addrs = to_socket_addrs(addr).await?;
+    cfg_not_wasi! {
+        /// Creates a new TcpListener, which will be bound to the specified address.
+        ///
```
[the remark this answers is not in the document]
rjzak: I'll revert to the `#[cfg(...)]` macro here.

tokio-stream/tests/watch.rs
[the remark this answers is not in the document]
Darksonn: I understand that this is where the panic comes from, but I don't understand why this test is triggering that code. This test _should_ pass on wasi.

tokio/src/net/tcp/socket.rs
```diff
@@ -1,3 +1,4 @@
+#![cfg(not(target_arm = "wasm32"))]
 use crate::net::{TcpListener, TcpStream};
```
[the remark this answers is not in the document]
haraldh: `target_arm` is wrong anyway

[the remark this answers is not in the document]
rjzak: Fixed

tokio/src/net/tcp/stream.rs
```diff
@@ -1,16 +1,22 @@
...
+
 use crate::io::{AsyncRead, AsyncWrite, Interest, PollEvented, ReadBuf, Ready};
 use crate::net::tcp::split::{split, ReadHalf, WriteHalf};
 use crate::net::tcp::split_owned::{split_owned, OwnedReadHalf, OwnedWriteHalf};
-use crate::net::{to_socket_addrs, ToSocketAddrs};
+
+cfg_not_wasi! {
+    use crate::net::{to_socket_addrs, ToSocketAddrs};
+    use std::time::Duration;
+}
```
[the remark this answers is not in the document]
rjzak: Fixed

tokio/src/net/tcp/stream.rs
```diff
@@ -227,9 +235,12 @@ impl TcpStream {
     /// [`std::net::TcpStream`]: std::net::TcpStream
     /// [`set_nonblocking`]: fn@std::net::TcpStream::set_nonblocking
     pub fn into_std(self) -> io::Result<std::net::TcpStream> {
-        #[cfg(unix)]
+        #[cfg(any(unix, target_os = "wasi"))]
         {
+            #[cfg(unix)]
             use std::os::unix::io::{FromRawFd, IntoRawFd};
+            #[cfg(target_os = "wasi")]
+            use std::os::wasi::io::{FromRawFd, IntoRawFd};
```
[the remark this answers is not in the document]
rjzak: Fixed

tokio/src/task/blocking.rs
```diff
@@ -78,122 +80,124 @@ cfg_rt_multi_thread! {
...
-    where
-        F: FnOnce() -> R + Send + 'static,
-        R: Send + 'static,
-    {
-        crate::runtime::spawn_blocking(f)
+cfg_not_wasi! {
+    cfg_rt! {
+        /// Runs the provided closure on a thread where blocking is acceptable.
+        ///
+        /// In general, issuing a blocking call or performing a lot of compute in a
```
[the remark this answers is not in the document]
rjzak: Fixed

tokio/tests/async_send_sync.rs
```diff
@@ -127,70 +127,101 @@ macro_rules! assert_value {
...
+        )*
+    }
+}
+
+cfg_not_wasi! {
+    mod fs {
+        use crate::require_send;
+        use crate::require_sync;
+        use crate::require_unpin;
+        use crate::AmbiguousIfUnpin;
```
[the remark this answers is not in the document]
rjzak: Fixed

tokio/tests/async_send_sync.rs
```diff
@@ -127,70 +127,101 @@ macro_rules! assert_value {
...
+        async_assert_fn!(tokio::fs::File::into_std(_): Send & Sync & !Unpin);
+        async_assert_fn!(
+            tokio::fs::File::set_permissions(_, std::fs::Permissions): Send & Sync & !Unpin
+        );
+    }
+}
+
+cfg_not_wasi! {
+    assert_value!(tokio::net::TcpSocket: Send & Sync & Unpin);
+    assert_value!(tokio::net::UdpSocket: Send & Sync & Unpin);
```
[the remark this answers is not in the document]
rjzak: Fixed

tokio/src/macros/cfg.rs
```diff
@@ -160,7 +161,10 @@ macro_rules! cfg_loom {
 
 macro_rules! cfg_not_loom {
     ($($item:item)*) => {
-        $( #[cfg(not(loom))] $item )*
+        $(
+            #[cfg(any(not(loom), target_os = "wasi"))]
+            $item
+        )*
```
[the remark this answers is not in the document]
rjzak: Reverted to original state, seems to be fine on Wasi

tokio-util/tests/time_delay_queue.rs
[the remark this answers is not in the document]
rjzak: Many of the tests disabled for Wasi just have `#[cfg(not(target_os = 'wasi'))]`. Should the disabled tests have that, or have an ignore message? Just thinking about consistency.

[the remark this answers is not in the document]
Darksonn: Prefer using ignore, but that doesn't always work because it might call methods that are missing. In that case just use `#[cfg(not(...))]`. I would prefer a comment explaining why in both cases.

haraldh:
So, for a quick fix of all the polling, I might suggest:
```
diff --git a/tokio/src/io/driver/mod.rs b/tokio/src/io/driver/mod.rs
index 265396bd..c7015974 100644
--- a/tokio/src/io/driver/mod.rs
+++ b/tokio/src/io/driver/mod.rs
@@ -168,6 +168,12 @@ impl Driver {
         match self.poll.poll(&mut events, max_wait) {
             Ok(_) => {}
             Err(ref e) if e.kind() == io::ErrorKind::Interrupted => {}
+            Err(e) if e.kind() == io::ErrorKind::InvalidInput => {
+                // In case of wasm32_wasi this error happens, when trying to poll without subscriptions
+                // just return from the park, as there would be nothing, which wakes us up.
+                #[cfg(not(target_os = "wasi"))]
+                return Err(e);
+            }
             Err(e) => return Err(e),
         }
```

haraldh:
> After disabling `tokio::spawn()`, the echo server example works! I need to figure out the implications here. https://github.com/rjzak/tokio-echo-test/blob/main/src/main.rs#L65

@rjzak you should update the initial comment. Also for me it works even with  `tokio::spawn()` now.

tokio/src/net/tcp/socket.rs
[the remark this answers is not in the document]
rjzak: Fixed

tokio/tests/fs_copy.rs
```diff
@@ -1,5 +1,5 @@
 #![warn(rust_2018_idioms)]
-#![cfg(all(feature = "full", not(target_os = "wasi")))]
+#![cfg(all(feature = "full", not(target_os = "wasi")))] // Wasi does not support file operations
```
haraldh, line 2: Well, technically WASI *does* support file operations. It's just, that it needs a preopened directory as the base directory. For the tests one could define a root directory FD preopened by wasmtime.

But, the big point is, that tokio uses threads for all blocking filesystem operations.

Darksonn: If the WASI file operations are async-compatible, then we certainly could use them for `tokio::fs` in the future. However, I don't think that needs to happen in this PR.

rjzak: I'll have a follow-up to address any remaining concerns.

tokio/tests/io_buf_reader.rs
```diff
@@ -277,7 +277,6 @@ async fn maybe_pending_buf_read() {
     assert_eq!(v, []);
 }
 
-// https://github.com/rust-lang/futures-rs/pull/1573#discussion_r281162309
 #[tokio::test]
 async fn maybe_pending_seek() {
```
Darksonn, line 281: Don't remove this comment.

tokio/tests/task_local_set.rs
```diff
@@ -355,6 +369,7 @@ fn with_timeout(timeout: Duration, f: impl FnOnce() + Send + 'static) {
     thread.join().expect("test thread should not panic!")
 }
 
+#[cfg_attr(target_os = "wasi", ignore = "FIXME: Does not seem to work with WASI")]
 #[test]
 fn drop_cancels_remote_tasks() {
     // This test reproduces issue #1885.
```
Darksonn, line 375: It seems like you have been able to remove the majority of these, but this one is still here.

Darksonn (commented):
Looking through the change, it looks pretty good! I mainly found a few tests where it looks like you had just forgotten to enable the test or similar.

tokio/src/io/driver/mod.rs
```diff
@@ -164,6 +168,12 @@ impl Driver {
         match self.poll.poll(&mut events, max_wait) {
             Ok(_) => {}
             Err(ref e) if e.kind() == io::ErrorKind::Interrupted => {}
+            Err(e) if e.kind() == io::ErrorKind::InvalidInput => {
+                // In case of wasm32_wasi this error happens, when trying to poll without subscriptions
+                // just return from the park, as there would be nothing, which wakes us up.
+                #[cfg(not(target_os = "wasi"))]
+                return Err(e);
+            }
```
Darksonn, line 176: I think you can simplify this.
```suggestion
            #[cfg(target_os = "wasi")]
            Err(e) if e.kind() == io::ErrorKind::InvalidInput => {
                // In case of wasm32_wasi this error happens, when trying to poll without subscriptions
                // just return from the park, as there would be nothing, which wakes us up.
            }
```

tokio/tests/time_timeout.rs
```diff
@@ -17,6 +17,7 @@ async fn simultaneous_deadline_future_completion() {
     assert_ready_ok!(fut.poll());
 }
 
+#[cfg_attr(target_os = "wasi", ignore = "Wasi does not support threads")]
 #[tokio::test]
 async fn completed_future_past_deadline() {
     // Wrap it with a deadline
```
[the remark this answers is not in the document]
rjzak: Seems that `assert_ready_ok!()` fails, and the `tokio::spawn().poll()` function eventually tries to call a `MockTask::enter()`.

tokio/tests/task_local_set.rs
```diff
@@ -377,6 +392,7 @@ fn drop_cancels_remote_tasks() {
     });
 }
 
+#[cfg(not(target_os = "wasi"))]
 #[test]
 fn local_tasks_wake_join_all() {
     // This test reproduces issue #2460.
```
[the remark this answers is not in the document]
rjzak: `task::spawn_local().await.unwrap()` fails

tokio/tests/task_local_set.rs
rjzak: `unwrap()` in `with_timeout()` panics on Wasi

tokio/tests/io_buf_reader.rs
rjzak: Sorry... fixed.

tokio/tests/sync_mpsc.rs
```diff
@@ -2,12 +2,12 @@
 #![warn(rust_2018_idioms)]
 #![cfg(feature = "sync")]
 
-#[cfg(target_arch = "wasm32")]
+#[cfg(all(target_arch = "wasm32", not(target_os = "wasi")))]
 use wasm_bindgen_test::wasm_bindgen_test as test;
-#[cfg(target_arch = "wasm32")]
+#[cfg(all(target_arch = "wasm32", not(target_os = "wasi")))]
```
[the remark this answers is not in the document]
rjzak: Added

tokio/tests/time_pause.rs
```diff
@@ -1,5 +1,5 @@
 #![warn(rust_2018_idioms)]
-#![cfg(feature = "full")]
+#![cfg(all(feature = "full", not(target_os = "wasi")))] // Wasi doesn't support threads
 
 use rand::SeedableRng;
```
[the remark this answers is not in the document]
rjzak: Done

tokio/src/net/tcp/listener.rs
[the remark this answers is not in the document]
carllerche: I would stick w/ the macro for now. We should generally fix the macro formatting issue, but since it is widespread, I wouldn't avoid using it here.

.github/workflows/ci.yml
```diff
@@ -489,11 +489,27 @@ jobs:
...
+        run: cargo test -p tokio-stream --target wasm32-wasi --features time,net,io-util,sync
+        env:
+          CARGO_TARGET_WASM32_WASI_RUNNER: "wasmtime run --"
+          RUSTFLAGS: --cfg tokio_unstable -Dwarnings
 
       - name: test tests-integration --features wasi-rt
         # TODO: this should become: `cargo hack wasi test --each-feature`
         run: cargo wasi test --test rt_yield --features wasi-rt
         working-directory: tests-integration
+        env:
```
carllerche, line 514: We need to run this test w/o `tokio_unstable`

Darksonn (approved):
I would like to have a release before I click the merge button, but I believe this PR is ready as-is.

CryZe:
Wish I had seen this earlier. `target_family = "wasm"` is preferred over checking for `wasm32` to be forwards compatible with `wasm64`.

Darksonn:
Can you please comment that on the meta issue mentioned above?

Darksonn:
It's unstable. You need to enable the `tokio_unstable` flag to use it. You can read how to do that [here](https://docs.rs/tokio/latest/tokio/index.html#unstable-features).

rjzak:
@izissise Thanks for pointing this out, I updated the readme on that example project.

AshleySetter:
@rjzak I was trying to build your example at https://github.com/rjzak/tokio-echo-test but I'm a relative beginner at rust and I was met with an error I couldn't resolve, any help would be much appreciated. 

```
RUSTFLAGS="--cfg tokio_unstable" cargo build --target=wasm32-wasi
warning: unused manifest key: build
    Updating git repository `https://github.com/rjzak/tokio`
error: failed to get `tokio` as a dependency of package `tokio-echo-test v0.0.1 (/home/ash/Documents/Junk/tokio-echo-test)`

Caused by:
  failed to load source for dependency `tokio`

Caused by:
  Unable to update https://github.com/rjzak/tokio?branch=wasi_wip#9dfcecbd

Caused by:
  object not found - no match for id (9dfcecbdc71c8403125df1271ebdc2ea79f659e2); class=Odb (9); code=NotFound (-3)
```

rjzak:
Hello! The Git branch `wasi_wip` was deleted when this PR was merged, so change `Cargo.toml` to point to `tokio-rs/tokio` instead of `rjzak`, and set `branch = "master"`.

AshleySetter:
@rjzak Brilliant, thanks, that works. 

Is that the same for https://github.com/rjzak/axum-wasm-test as well? Which git repo and branch of axum should be used with that demo?

rjzak:
Yes, and I'm working on fixing some things up. I noticed that the latest Tokio release don't block `socket2` when compiling for Wasi, but is fixed in a later commit https://github.com/tokio-rs/tokio/commit/b2ada60e701d5c9e6644cf8fc42a100774f8e23f. So for the time being, Tokio should be pulled in with `git = "https://www.github.com/tokio-rs/tokio", rev = "b2ada60"` or `git = "https://www.github.com/tokio-rs/tokio", branch = "master"`

PR #4731, opened 2022-06-01
Check page capacity before obtaining base pointer

djkoloski:
This doesn't cause any issues in practice because this is a private API that is only used in ways that cannot trigger UB. Indexing into `slots` is not sound until after we've asserted that the page is allocated, since that aliases the first slot which may not be allocated. This PR also switches to using `as_ptr` to obtain the base pointer for clarity.

## Motivation

While reviewing an upgrade of tokio for Fuchsia, we identified a potential soundness issue in `Slots::index_for`. In the future, changes to the uses of the structure could cause UB.

## Solution

The assertion that the page is not empty has been moved earlier in the function and the method for getting a pointer to the first `Vec` element has been changed from `&slots[0] as *const _` to `slots.as_ptr()` which was stabilized in 1.37 and meets MSRV.


Darksonn (approved):
Looks good to me.

PR #4739, opened 2022-06-03
net: Add `take_error` to `TcpSocket` and `TcpStream`

mxinden:
## Motivation

When manually creating a `TcpSocket` or `TcpStream` e.g. via `tokio::net::TcpStream::try_from` one should be able to check for errors after awaiting the connection to be writable (`writable()`) before assuming the connection to be successfully established.

This is needed e.g. in rust-libp2p, see https://github.com/libp2p/rust-libp2p/pull/2458.

## Solution

The first commit reintroduces https://github.com/tokio-rs/tokio/pull/4364 which first merged but later on reverted in https://github.com/tokio-rs/tokio/pull/4392. See also https://github.com/tokio-rs/tokio/pull/4392#issuecomment-1035542164.

The second commit adds the same method to `TcpStream`.

Tagging @taiki-e as the initial author.

### DIFF
--- a/tokio/src/net/tcp/socket.rs
+++ b/tokio/src/net/tcp/socket.rs
@@ -424,6 +424,11 @@ impl TcpSocket {
         self.inner.local_addr().and_then(convert_address)
     }
 
+    /// Returns the value of the `SO_ERROR` option.
+    pub fn take_error(&self) -> io::Result<Option<io::Error>> {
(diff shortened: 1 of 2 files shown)


Darksonn (approved):
Thanks!

mxinden:
Thanks a lot, both for the quick merge and for the general maintenance! :pray:

PR #4758, opened 2022-06-08
`sync::watch`: Add `has_changed` flag and method to `Ref`

uklotzde:
Extend `sync::watch::Receiver` to cover another use case.

## Motivation

As a `sync::watch::Receiver` I would like to know, if the value obtained from `borrow_and_update()` has actually changed and marked as seen or not. Subsequent calculations could be skipped without actually inspecting the value if no changes occurred.

## Solution

Add an ~~`updated`~~ `has_changed` flag and method to `Ref`.

## Discarded solution (informational, just for the record )

Add a new function `borrow_and_update_if_changed()` that returns a tuple, both the reference and a flag indicating if the value has been detected as *changed* and marked as seen, i.e. *updated*. Using a boolean flag is consistent with the return value of `has_changed()` on success.

This function could otherwise not be implemented in terms of the public API, because invoking `has_changed()` and `borrow_and_update()` independently is not thread-safe.

~~As temporary a workaround you could invoke `has_changed()` twice, i.e. before and after invoking `borrow_and_update()`. This will cause some false positives and is less efficient. This doesn't work correctly if the `Sender` has been dropped, because then `has_changed()` always returns with an error.~~ Doesn't work as `Receiver` is already mutably borrowed after `borrow_and_update()`.


tokio/src/sync/watch.rs
```diff
@@ -365,9 +364,52 @@ impl<T> Receiver<T> {
...
+    ///                                               |  // will block
+    ///                                               |  let _ = tx.send(());
+    /// // may deadlock                               |
+    /// let (_ref2, _updated2) =                      |
+    ///         rx2.borrow_and_update_if_changed();   |
+    /// ```
+    /// </details>
+    ///
+    /// [`changed`]: Receiver::changed
+    pub fn borrow_and_update_if_changed(&mut self) -> (Ref<'_, T>, bool) {
```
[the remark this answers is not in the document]
Darksonn: That seems like a good way to go around it.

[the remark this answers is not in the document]
uklotzde: Good point! I was already experimenting with such a design in my external wrapper crate. Just wanted to avoid introducing new, named types.

I didn't consider the fact that this solutions enables us to extend the current API without any breaking changes, which make it even more appealing :)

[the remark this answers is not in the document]
uklotzde: I was blind :see_no_evil: Your proposal turned out as the perfect solution that only affects `borrow_and_update()` when focusing on the `updated`  aspect and not considering the more general `changed` aspect: https://github.com/tokio-rs/tokio/pull/4758/commits/63c9987da30bf5256eeedd62ff1d60580d06009b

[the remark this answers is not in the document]
uklotzde: I noticed that commits will be squashed before merging, so I added the changes as a new commit.

tokio/src/sync/watch.rs
```diff
@@ -112,6 +112,20 @@ pub struct Sender<T> {
...
+    /// Indicates if the shared value has been detected as _changed_
+    /// **and** marked as seen.
+    ///
+    /// The result is `true` only if this reference has been obtained
+    /// from [`Receiver::borrow_and_update()`] in case changes have been
+    /// detected and marked as seen. In all other cases the result will
+    /// be `false`.
+    pub fn updated(&self) -> bool {
+        self.updated
+    }
```
Darksonn, line 128: It seems more useful to have this function return whether the value was marked as seen before the call. Then the value is also useful for the `borrow` method.

uklotzde: This would mean to
- add an invocation of `load()` to `borrow()` and
- rename the function to `has_changed()`.

I was unsure if the extra `load()` in `borrow()` is desired. It probably doesn't have any noticeable effect and a more consistent interface might be preferable.

uklotzde: Done: https://github.com/tokio-rs/tokio/pull/4758/commits/d081f83f762c43cb826c70945a73ce3bb89f619f

I also added a comprehensive example.

tokio/src/sync/watch.rs
```diff
@@ -112,6 +112,56 @@ pub struct Sender<T> {
...
     inner: RwLockReadGuard<'a, T>,
+    has_changed: bool,
+}
+
+impl<'a, T> Ref<'a, T> {
+    /// Indicates if the borrowed value is considered as _changed_ since the last
+    /// time it has been marked as seen.
+    ///
+    /// Other than [`Receiver::has_changed()`] it does not fail if the [`Receiver`]
+    /// is orphaned after the [`Sender`] has been dropped.
```
Darksonn, line 123: What do you mean by this?

uklotzde: `Receiver::has_changed()` fails with an error if the channel is considered as "closed". This happens when the `Sender` is dropped. Which doesn't makes much sense to me, because the shared value (and therefore the "channel"'s content) is still accessible.

I am not a native speaker. If this needs to be reworded please do so.

Darksonn:
```suggestion
    /// Unlike [`Receiver::has_changed()`], this method does not fail if the channel is closed.
```

Darksonn (approved):

PR #4767, opened 2022-06-14
rt: clean up arguments passed to basic scheduler

carllerche:
Extracts the refactor from #4518.

The basic scheduler takes many configuration options as arguments to the
constructor. This cleans it up a bit by defining a `Config` struct and
using that to pass arguments to the constructor.


hawkw (approved):
already approved this previously, so feel free to ship it!

PR #4791, opened 2022-06-27
time: add track_caller to public APIs

hds:
## Motivation

`time::interval` with 0 duration) then the user currently sees the line number
of the panic call inside tokio. It would be more informative for the user to see
the place where they called the panicking function.

hidden.

This change is the 4th in a series towards closing #4413 (starting with #4772),
this change is for the time functions in the main `tokio` crate.

## Solution

Functions that may panic can be annotated with #[track_caller] so that
source.

This change adds #[track_caller] to all the non-unstable public APIs in
tokio-util where the documentation describes how the function may panic
due to incorrect context or inputs.

In cases where `#[track_caller]` does not work, it has been left out. For
example, it currently does not work on async functions, blocks, or
closures. So any call stack that passes through one of these before
reaching the actual panic is not able to show the calling site outside
of tokio as the panic location.

The public functions that could not have `#[track_caller]` added for
this reason are:
* `time::advance`

Tests are included to cover each potentially panicking function. In the
following cases, `#[track_caller]` had already been added, and only
tests have been added:
* `time::interval`
* `time::interval_at`

Refs: #4413

### DIFF
--- a/tokio/src/time/clock.rs
+++ b/tokio/src/time/clock.rs
@@ -96,6 +96,7 @@ cfg_test_util! {
     ///
     /// [`Sleep`]: crate::time::Sleep
     /// [`advance`]: crate::time::advance
+    #[track_caller]
     pub fn pause() {
         let clock = clock().expect("time cannot be frozen from outside the Tokio runtime");
         clock.pause();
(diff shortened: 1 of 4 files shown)


Darksonn (approved):
Thanks.

PR #4803, opened 2022-07-02
tests: alter integration tests for stdio to not use a handrolled cat implementation

### LINKED ISSUE #4801: process: Windows tests failing on 1.62
With the new version of rust having landed, windows tests are failing. Specifically, it seems that the `pipe_from_one_command_to_another` test is failing.

I'm currently investigating what exactly is prompting this to fail on windows. Once the root cause is identified, we can figure out how to proceed from there. It may be an std issue or it may be on us.

CI build failure: https://github.com/tokio-rs/tokio/runs/7155496538?check_suite_focus=true#step:8:228

Noah-Kennedy:
This fixes #4801, where, as a result of https://github.com/rust-lang/rust/pull/95469, our implementation of cat used for this test no longer works, as stdio functions on windows now can abort the process if the pipe is set to nonblocking mode.

Unfortunately in windows, setting one end of the pipe to be nonblocking makes the whole thing nonblocking, so when, in tokio::process we set the child pipes to nonblocking mode, it causes serious problems for any rust program at the other end.

Fixing this issue is for another day, but fixing the tests is for today.

### DIFF
--- a/tests-integration/Cargo.toml
+++ b/tests-integration/Cargo.toml
@@ -7,6 +7,7 @@ publish = false
 
 [[bin]]
 name = "test-cat"
+required-features = ["rt-process-io-util"]
 
 [[bin]]
 name = "test-mem"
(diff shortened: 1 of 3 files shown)


carllerche (approved):

ipetkov (commented):
Overall LGTM! Though I do wonder, should we keep our cat utility blocking on Unix systems?

It would be good to maintain it as a regression test for Unix behavior (even though the stdlib maintains this for us we can find out if it breaks)

Noah-Kennedy:
@ipetkov It might not be a bad idea to split the pipes test into two tests, one with sync, one with async.

PR #4825, opened 2022-07-12
chore: fix ci by hard-coding nightly version for miri

Noah-Kennedy:
It would seem that the latest nightly is having issues, so I went with the one from yesterday (2022-7-10).


ipetkov (approved):
Nice thanks for fixing this!

PR #4829, opened 2022-07-12
runtime: tiny typo in shutdown_background docs

ekzhang:
## Motivation

I noticed that the `shutdown_background` function mentions a non-existent `Duration::of_nanos` function

## Solution

Change to `Duration::from_nanos`


Darksonn (approved):
Thanks!

Issue #4836, opened 2022-07-13
Clippy lint firing with `tokio::task_local` macro


sbromling:
**Version**
 1.20.0

**Platform**
 x86_64 GNU/Linux

**Description**
After adopting the latest tokio 1.20.0 release, and using the latest stable rust toolchain (1.62.0), clippy reports the  https://rust-lang.github.io/rust-clippy/master/index.html#declare_interior_mutable_const lint against `tokio::task_local!` macro uses.

Seems like this is similar to https://github.com/rust-lang/rust-clippy/issues/8493, i.e. this may be a false positive.

rukai:
This lint is still firing on the latest rust nightly so I raised on issue on the clippy repo https://github.com/rust-lang/rust-clippy/issues/9224

PR #4837, opened 2022-07-14
task: propagate attrs on task_local items to the generated items

Arnavion:
Discovered while trying to add a `#[allow]` to try to work around #4836

That is,

```rust
tokio::task_local! {
	#[allow(clippy::declare_interior_mutable_const)]
	pub(crate) static TASK_LOCAL_LOGGER: env_logger::Logger;
}
```

... did not suppress the clippy lint, since the macro was not propagating the attr to the generated `LocalKey`

I did not find any indication in the commit and PR discussion that added `task_local!` that this was done intentionally, so it seems to have just been missed.


Darksonn (approved):
Thank you.

PR #4848, opened 2022-07-20
task: add track_caller to public APIs

hds:
## Motivation

`spawn()` from outside a tokio runtime) then the user currently
sees the line number of the panic call inside tokio. It would be more informative
for the user to see the place where they called the panicking function.

hidden.

This change is the 8th in a series towards closing #4413 (starting with #4772),
this change is for the functions in the task module in the main tokio crate.

## Solution

source.

This change adds `#[track_caller]` to all the public APIs in the task
module of the tokio crate where the documentation describes how the
function may panic due to incorrect context or inputs.

In cases where `#[track_caller]` does not work, it has been left out.
For example, it currently does not work on async functions, blocks, or
closures. So any call stack that passes through one of these before
reaching the actual panic is not able to show the calling site outside
of tokio as the panic location.

The following functions have call stacks that pass through closures:
* `task::block_in_place`
* `task::local::spawn_local`

The following functions already had `#[track_caller]` applied everywhere
it was needed and only tests have been added:
* `task::spawn`
* `task::LocalKey::sync_scope`

Refs: #4413

### DIFF
--- a/tokio/src/runtime/mod.rs
+++ b/tokio/src/runtime/mod.rs
@@ -475,6 +475,7 @@ cfg_rt! {
         /// ```
         ///
         /// [handle]: fn@Handle::block_on
+        #[track_caller]
         pub fn block_on<F: Future>(&self, future: F) -> F::Output {
             #[cfg(all(tokio_unstable, feature = "tracing"))]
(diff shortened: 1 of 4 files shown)


Darksonn (approved):
Thanks!

Noah-Kennedy (approved):

Issue #4773, opened 2022-06-17
`Builder::worker_threads` doesn't panic when used with a `current_thread` runtime


hds:
**Version**

tokio v1.19.2 (also master branch)

**Platform**

`Linux sync-z2 4.15.0-162-generic #170-Ubuntu SMP Mon Oct 18 11:38:05 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

**Description**

The documentation for the function `Builder::worker_threads` states that if it is used
with `current_thread` runtime it will panic. However it doesn't panic, it appears to
ignore the value set by the call instead.

Documentation:
https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.worker_threads

This test demonstrates the behaviour (it would live in `tokio/tests`):

```rust
#![warn(rust_2018_idioms)]
#![cfg(feature = "full")]

use tokio::runtime::Builder;

#[test]
#[should_panic]
fn test_worker_threads_panics() {
    let rt = Builder::new_current_thread()
        .worker_threads(4)
        .build()
        .unwrap();

    rt.block_on(async {
        tokio::spawn(async {}).await.unwrap();
    });
}
```
The test fails as nothing panics.

Additionally, the documentation for `worker_threads` contains 2 panic sections.
The other panic section describes a case that does panic (`val = 0`).

The 2 options I see here are:
1. Leave the functionality as it is (silently ignore `worker_threads` for a `current_thread` runtime) and fix the documentation
2. Change the functionality so that it does panic (ideally directly within `worker_threads` itself) and also fix the documentation so that all panics are described in a single heading

I would lean towards option 1 as this makes it easier to choose the runtime `Kind` in the builder at execution time. It will also not break any existing code.

PR #4849, opened 2022-07-20
doc: remove incorrect panic section for Builder::worker_threads

hds:
## Motivation

The documentation for the function `Builder::worker_threads` states that if it is used
with `current_thread` runtime it will panic. However it doesn't panic, it appears to
ignore the value set by the call instead.

Documentation:
https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.worker_threads

While this is probably not going to catch a lot of people out, it would be better if it
was correct.

Also, this function documentation had 2 `Panics` sections in it. 🤷‍♀️

## Solution

The documentation for the runtime `Builder::worker_threads` function
incorrectly stated that it would panic if used when constructing a
`current_thread` runtime. In truth, the call to the function has no
effect.

Since adding the described panic to the code could cause new panics in
existing code using tokio, the documentation has been modified to
describe the existing behavior.

Fixes: #4773

### DIFF
--- a/tokio/src/runtime/builder.rs
+++ b/tokio/src/runtime/builder.rs
@@ -289,11 +289,7 @@ impl Builder {
     ///
     /// The default value is the number of cores available to the system.
     ///
-    /// # Panics
-    ///
-    /// When using the `current_thread` runtime this method will panic, since
-    /// those variants do not allow setting worker thread counts.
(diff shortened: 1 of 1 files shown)


Darksonn (approved):
Thanks.

PR #4886, opened 2022-08-07
Add #[must_use] to lock guards

xxchan:


Noah-Kennedy:
Ah, this will bring us in line with the behavior of the std implementations more nicely.

You're gonna need to fix the CI failures though.

Darksonn (approved):
Thanks.

PR #4893, opened 2022-08-10
Only use a single `to` in `AsyncSeekExt::rewind`'s documentation

devtomio:
## Motivation

There are double `to`'s in `AsyncSeekExt::rewind`'s docs

## Solution

Remove a single `to` from the docs


Darksonn (approved):
Thanks!

PR #4897, opened 2022-08-10
Change panic on seeking busy file to error.

sebpuetz:
## Motivation

Following https://github.com/tokio-rs/tokio/issues/4875#issuecomment-1210653815 this PR changes the panic in `File`'s `AsyncSeek` implementation to an error. 

## Solution

The panic is changed to an error.

I only managed to reproduce the panic with the specific example using a `BufReader<File>` as provided in #4875, the added test also uses a BufReader and panics without the change to `File`'s seek impl.

---

I raised a question regarding inconsistent behavior between AsyncSeek impls for BufReader / BufWriter in #4875

### DIFF
--- a/tokio/src/fs/file.rs
+++ b/tokio/src/fs/file.rs
@@ -565,29 +565,30 @@ impl AsyncSeek for File {
         let me = self.get_mut();
         let inner = me.inner.get_mut();
 
-        loop {
-            match inner.state {
-                Busy(_) => panic!("must wait for poll_complete before calling start_seek"),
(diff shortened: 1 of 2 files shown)


Darksonn (approved):
Thanks. I see you removed the loop, but that also appears to be correct.

sebpuetz:
> Thanks. I see you removed the loop, but that also appears to be correct.

Yep, clippy complained about never looping. The previous `panic!()` probably threw clippy off, the loop never ran in the original version, too!

PR #4915, opened 2022-08-15
ci: upgrade to `nightly-2022-07-26` and pin `miri` nightly

jdisanti:
## Motivation

As discovered in #4914, the `miri` CI check wasn't pinned to a specific nightly, and the nightly chosen in #4903 didn't have a `miri` component.

## Solution

This PR pins the `miri` check to the `rust_nightly` env value and updates the nightly to a version that is compatible with everything else.


taiki-e (approved):
Thanks.

PR #4922, opened 2022-08-18
docs: fix indent of `tokio::sync::watch` example code

rhysd:
## Motivation

I found that [the example code of `tokio::sync::watch`](https://docs.rs/tokio/latest/tokio/sync/watch/index.html#examples) had odd indent.

image

## Solution

This PR removes the unnecessary indentation from the example code.


taiki-e (approved):
Thanks.

Issue #4921, opened 2022-08-17
Request: a way of catching any windows signal.


jkelleyrtp:
Right now it seems that tokio only exports signal handlers for ctrl_c and ctrl_break for windows, while for Unix, you can trap any signal covered by SignalKind.

It would be great if there was a method that would allow creating an RxFuture for any of the signal types in the winapi crate.

This method already exists, but it's marked private. It would be great if this were public and given a new name so I could trap the various other windows signals that exist.

https://github.com/tokio-rs/tokio/blob/4b1c4801b1383800932141d0f6508d5b3003323e/tokio/src/signal/windows/sys.rs#L20

Noah-Kennedy:
I don't have any issues with doing this, but I'd like to know why it's marked as private first.

ipetkov:
> but I'd like to know why it's marked as private first.

@Noah-Kennedy just to start out with a conservative API surface and no one has asked for it until now 😄 

@jkelleyrtp which signals are you interested in receiving?

I see win API has other events like  `CTRL_CLOSE_EVENT`/`CTRL_LOGOFF_EVENT`/`CTRL_SHUTDOWN_EVENT` we can easily add these in as a concrete public API. I'm not philosophically opposed to having more of a general signal solution like we do on Unix platforms, though I'm not very familiar with the windows signal semantics and if they are analogous to the Unix ones

schultetwin1:
The only event we need for now is `CTRL_CLOSE_EVENT`. `CTRL_LOGFF_EVENT` and `CTRL_SHUTDOWN_EVENT` can be added as well but if you want to keep with the them of a conservative API, we'd be happy with just `CTRL_CLOSE_EVENT`.

jkelleyrtp:
I'll put up a branch here in a sec.

PR #4924, opened 2022-08-18
Feature: Add more windows signal handlers

jkelleyrtp:
Add the other _event types from the winapi crate. Addresses #4921 

Includes:
- Console Close
- Shutdown
- Logoff

It might be worth tying console close into the ctrl_c handling framework, depending on how the Tokio authors consider it to fit.

Console Close I believe is analogous to unix's SIGHUP, which doesn't seem to be in the ctrl_c framework.

### DIFF
--- a/tokio/src/signal/windows.rs
+++ b/tokio/src/signal/windows.rs
@@ -260,7 +260,7 @@ pub fn ctrl_close() -> io::Result<CtrlClose> {
 #[must_use = "streams do nothing unless polled"]
 #[derive(Debug)]
 pub struct CtrlClose {
-    inner: RxFuture
+    inner: RxFuture,
 }
 
 impl CtrlClose {
@@ -357,7 +357,7 @@ pub fn ctrl_shutdown() -> io::Result<CtrlShutdown> {
 #[must_use = "streams do nothing unless polled"]
 #[derive(Debug)]
 pub struct CtrlShutdown {
-    inner: RxFuture
+    inner: RxFuture,
 }
 
 impl CtrlShutdown {
@@ -454,7 +454,7 @@ pub fn ctrl_logoff() -> io::Result<CtrlLogoff> {
 #[must_use = "streams do nothing unless polled"]
 #[derive(Debug)]
 pub struct CtrlLogoff {
-    inner: RxFuture
+    inner: RxFuture,
 }
 
 impl CtrlLogoff {
--- a/tokio/src/signal/windows.rs
+++ b/tokio/src/signal/windows.rs
@@ -221,3 +221,295 @@ pub fn ctrl_break() -> io::Result<CtrlBreak> {
         inner: self::imp::ctrl_break()?,
     })
 }
+
+/// Creates a new stream which receives "ctrl-close" notifications sent to the
+/// process.
+///
+/// # Examples
+///
+/// ```rust,no_run
+/// use tokio::signal::windows::ctrl_close;
+///
+/// #[tokio::main]
+/// async fn main() -> Result<(), Box<dyn std::error::Error>> {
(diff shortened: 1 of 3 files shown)


Darksonn (commented):
You also need to add them to `src/signal/windows.rs`

ipetkov (commented):
Approach LGTM but Alice is right that these aren't publicly exposed yet

tokio/src/signal/windows.rs
```diff
@@ -221,3 +221,295 @@ pub fn ctrl_break() -> io::Result<CtrlBreak> {
...
+    ///         self.ctrl_shutdown.poll_recv(cx)
+    ///     }
+    /// }
+    /// ```
+    pub fn poll_recv(&mut self, cx: &mut Context<'_>) -> Poll<Option<()>> {
+        self.inner.poll_recv(cx)
+    }
+}
+
+/// Creates a new stream which receives "ctrl-c" notifications sent to the
```
ipetkov, line 420: Should be "ctrl-logoff"

ipetkov (approved):

ipetkov (changes requested):
I forgot that `OsStorage` also needs to be updated to dispatch the signals appropriately. Might not be a bad idea to add tests for all the events to make sure they don't break.

We could do them in a big loop so we don't have to copy paste the same test case N times but will leave it up to you

tokio/src/signal/windows/sys.rs
```diff
@@ -25,24 +37,33 @@ fn new(signum: DWORD) -> io::Result<RxFuture> {
...
 
 impl Storage for OsStorage {
     fn event_info(&self, id: EventId) -> Option<&EventInfo> {
         match DWORD::try_from(id) {
-            Ok(CTRL_C_EVENT) => Some(&self.ctrl_c),
             Ok(CTRL_BREAK_EVENT) => Some(&self.ctrl_break),
+            Ok(CTRL_CLOSE_EVENT) => Some(&self.ctrl_close),
+            Ok(CTRL_C_EVENT) => Some(&self.ctrl_c),
+            Ok(CTRL_LOGOFF_EVENT) => Some(&self.ctrl_logoff),
+            Ok(CTRL_SHUTDOWN_EVENT) => Some(&self.ctrl_shutdown),
```
ipetkov, line 66: These need to be qualified now

```suggestion
            Ok(wincon::CTRL_BREAK_EVENT) => Some(&self.ctrl_break),
            Ok(wincon::CTRL_CLOSE_EVENT) => Some(&self.ctrl_close),
            Ok(wincon::CTRL_C_EVENT) => Some(&self.ctrl_c),
            Ok(wincon::CTRL_LOGOFF_EVENT) => Some(&self.ctrl_logoff),
            Ok(wincon::CTRL_SHUTDOWN_EVENT) => Some(&self.ctrl_shutdown),
```

jkelleyrtp: Sorry, I'm developing on a Mac and have been using the tokio CI as my cargo check :) Just got a windows machine to test out the new signals.

ipetkov:
@jkelleyrtp LGTM thanks for implementing this! Just a few more things and I'm happy to merge:

- [ ] we'll need to rebase/merge as this branch has fallen behind (there may be a merge conflict that the UI is warning about, not sure)
- [ ] update the `ctrl_logoff` method to specify "ctrl-logoff" (not "ctrl-c")
- [ ] update the tests that each of the handlers works (right now we have to manually dispatch the received signals so it would be good to get test coverage there)

Feel free to ask if you have any questions!

jkelleyrtp:
All three tasks complete - thanks for being so responsive!

ipetkov (approved):
Thanks for addressing all the feedback!

PR #4933, opened 2022-08-23
doc: correct `cargo doc` command in contrib guide

hds:
## Motivation

The command to build docs listed in the contribution guide was missing an
environment variable needed to build `tokio-stream`. Additionally it turns out
that there was outdated information about not being able to build docs from
the root of the workspace. After the resolution of rust-lang/cargo#9274, this
is not longer correct. 

## Solution

The command to build the documentation locally provided in the
Contribution Guide did not work for all crates in the project workspace.
Specifically, to build the docs for `tokio-stream` the flag `--cfg
docsrs` needs to be in the environment variable `RUSTFLAGS` in addition
to being in `RUSTDOCFLAGS`.

Additionally, there was text describing that the docs cannot be built
from the root of the workspace with a link to rust-lang/cargo#9274. That
issue has since been closed as complete and the listed commands do now
work from the root of the workspace. As such, that text has been
removed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -154,7 +154,7 @@ required for various parts of Tokio are missing. To build the documentation
 correctly, use this command:
 
 ```
-RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --all-features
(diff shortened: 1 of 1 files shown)


Darksonn (approved):
Thanks.

PR #4936, opened 2022-08-23
rt: add unstable option to disable the LIFO slot

carllerche:
Depends on: #4935

The multi-threaded scheduler includes a per-worker LIFO slot to
store the last scheduled task. This can improve certain usage patterns,
especially message passing between tasks. However, this LIFO slot is not
currently stealable.

Eventually, the LIFO slot **will** become stealable. However, as a
stop-gap, this unstable option lets users disable the LIFO task when
doing so improves their application's overall performance.

### DIFF
--- a/tokio/src/runtime/builder.rs
+++ b/tokio/src/runtime/builder.rs
@@ -87,6 +87,11 @@ pub struct Builder {
 
     #[cfg(tokio_unstable)]
     pub(super) unhandled_panic: UnhandledPanic,
+
+    /// When true, the multi-threade scheduler LIFO slot should not be used.
+    ///
+    /// This option should only be exposed as unstable.
+    pub(super) disable_lifo_slot: bool,
 }
 
 cfg_unstable! {
@@ -252,6 +257,8 @@ impl Builder {
 
             #[cfg(tokio_unstable)]
             unhandled_panic: UnhandledPanic::Ignore,
+
+            disable_lifo_slot: false,
         }
     }
 
@@ -781,6 +788,37 @@ impl Builder {
             self.unhandled_panic = behavior;
             self
         }
+
+        /// Disables the LIFO task scheduler heuristic.
+        ///
+        /// The multi-threaded scheduler includes a heuristic for optimizing
+        /// message-passing patterns. This heuristic results in the **last**
+        /// scheduled task being polled first.
+        ///
+        /// To implement this heuristic, each worker thread has a slot which
+        /// holds the task that should be polled next. However, this slot cannot
+        /// be stolen by other worker threads, which can result in lower total
+        /// throughput when tasks tend to have longer poll times.
+        ///
+        /// This configuration option will disable this heuristic resulting in
+        /// all scheduled tasks being pushed into the worker-local queue, which
+        /// is stealable.
+        ///
+        /// Eventually, the LIFO slot will become stealable and this option will
+        /// probably go away.
+        ///
+        /// ```
+        /// use tokio::runtime;
+        ///
+        /// let rt = runtime::Builder::new_multi_threaded()
+        ///     .disable_lifo_slot
+        ///     .build()
+        ///     .unwrap();
+        /// ```
+        pub fn disable_lifo_slot(&mut self) -> &mut Self {
+            self.disable_lifo_slot = true;
+            self
+        }
     }
 
     fn build_basic_runtime(&mut self) -> io::Result<Runtime> {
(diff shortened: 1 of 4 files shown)


Noah-Kennedy (approved):

tokio/src/runtime/builder.rs
```diff
@@ -87,6 +87,11 @@ pub struct Builder {
 
     #[cfg(tokio_unstable)]
     pub(super) unhandled_panic: UnhandledPanic,
+
+    /// When true, the multi-threade scheduler LIFO slot should not be used.
+    ///
+    /// This option should only be exposed as unstable.
+    pub(super) disable_lifo_slot: bool,
```
hds, line 94: Would it make sense to move this above `unhandled_panic` so that we keep the unstable member down the bottom?

Same question in `runtime::config::Config`.

tokio/src/runtime/builder.rs
```diff
@@ -781,6 +788,37 @@ impl Builder {
...
+        /// holds the task that should be polled next. However, this slot cannot
+        /// be stolen by other worker threads, which can result in lower total
+        /// throughput when tasks tend to have longer poll times.
+        ///
+        /// This configuration option will disable this heuristic resulting in
+        /// all scheduled tasks being pushed into the worker-local queue, which
+        /// is stealable.
+        ///
+        /// Eventually, the LIFO slot will become stealable and this option will
+        /// probably go away.
```
hawkw, line 808: this is maybe a bit of a nitpick, but I feel like if documentation is going to say "this will change in the future", it's good to link to an issue or something so that people can follow along?

tokio/src/runtime/builder.rs
```diff
@@ -781,6 +788,37 @@ impl Builder {
...
+        /// Disables the LIFO task scheduler heuristic.
+        ///
+        /// The multi-threaded scheduler includes a heuristic for optimizing
+        /// message-passing patterns. This heuristic results in the **last**
+        /// scheduled task being polled first.
+        ///
+        /// To implement this heuristic, each worker thread has a slot which
+        /// holds the task that should be polled next. However, this slot cannot
+        /// be stolen by other worker threads, which can result in lower total
+        /// throughput when tasks tend to have longer poll times.
```
hawkw, line 801: might be nice if these docs mentioned why this heuristic is useful, and why you would generally not want to disable it?

tokio/src/runtime/builder.rs
```diff
@@ -781,6 +788,37 @@ impl Builder {
...
+        ///
+        /// This configuration option will disable this heuristic resulting in
+        /// all scheduled tasks being pushed into the worker-local queue, which
+        /// is stealable.
+        ///
+        /// Eventually, the LIFO slot will become stealable and this option will
+        /// probably go away.
+        ///
```
hawkw, line 809: nit
```suggestion
        ///
        /// # Examples
```

hawkw, line 809: the docs for the `unhandled_panic` method have a section mentioning that the method is unstable, and what that means:
https://github.com/tokio-rs/tokio/blob/97f6693cc1c759900c03d6e5205278bb57ea2da7/tokio/src/runtime/builder.rs#L751-L756

can we add something similar here?

tokio/src/runtime/builder.rs
```diff
@@ -781,6 +788,37 @@ impl Builder {
...
+        /// scheduled task being polled first.
+        ///
+        /// To implement this heuristic, each worker thread has a slot which
+        /// holds the task that should be polled next. However, this slot cannot
+        /// be stolen by other worker threads, which can result in lower total
+        /// throughput when tasks tend to have longer poll times.
+        ///
+        /// This configuration option will disable this heuristic resulting in
+        /// all scheduled tasks being pushed into the worker-local queue, which
+        /// is stealable.
```
hawkw, line 805: might be worth stating somewhere that this option does nothing with the single-threaded scheduler?

hawkw (commented):
looks good to me overall. i had some docs suggestions.

tokio/src/runtime/builder.rs
carllerche: In theory, it should though 😆

hawkw: ah, then maybe we should change this paragraph, which only mentioned the multi-threaded scheduler? https://github.com/tokio-rs/tokio/blob/97f6693cc1c759900c03d6e5205278bb57ea2da7/tokio/src/runtime/builder.rs#L794-L797 and/or add an example disabling the lifo slot in a single threaded scheduler?

PR #4945, opened 2022-08-25
rt, chore: rename internal scheduler types

carllerche:
For historical reasons, the current-thread runtime's scheduler was named
BasicScheduler and the multi-thread runtime's scheduler were named
ThreadPool. This patch renames the schedulers to mirror the runtime
names to increase consistency. This patch also moves the scheduler
implementations into a new `runtime::scheduler` module.


Noah-Kennedy (approved):
I'm more than fine with this.

hds (approved):

hds:
This is a really nice change, it helps internal readability a lot!

Darksonn (approved):
I agree that renaming the basic scheduler is a good idea.

PR #4956, opened 2022-08-29
signal: make `SignalKind` methods const

0xdeafbeef:
`SignalKind` is just a libc int wrapper, so creation methods can be const.


ipetkov (approved):

Noah-Kennedy:
Rerunning CI due to spurious failures.

Noah-Kennedy:
@0xdeafbeef you might need to push an empty commit to restart CI. For some reason it cancelled a check and I can't restart CI again.

PR #4970, opened 2022-09-02
io: reduce syscalls in `poll_write`

Noah-Kennedy:
This applies the same optimization made in #4840 to writes.

### DIFF
--- a/tokio/src/io/poll_evented.rs
+++ b/tokio/src/io/poll_evented.rs
@@ -188,7 +188,26 @@ feature! {
             &'a E: io::Write + 'a,
         {
             use std::io::Write;
-            self.registration.poll_write_io(cx, || self.io.as_ref().unwrap().write(buf))
+
+            loop {
+                let evt = ready!(self.registration.poll_write_ready(cx))?;
+
+                match self.io.as_ref().unwrap().write(buf) {
+                    Ok(n) => {
+                        // if we write only part of our buffer, this is sufficient on unix to show
+                        // that the socket buffer is full
+                        if n > 0 && (!cfg!(windows) && n < buf.len()) {
+                            self.registration.clear_readiness(evt);
+                        }
+
+                        return Poll::Ready(Ok(n));
+                    },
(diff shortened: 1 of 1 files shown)


tokio/src/io/poll_evented.rs
```diff
@@ -188,7 +188,26 @@ feature! {
...
+            loop {
+                let evt = ready!(self.registration.poll_write_ready(cx))?;
+
+                match self.io.as_ref().unwrap().write(buf) {
+                    Ok(n) => {
+                        // if we write only part of our buffer, this is sufficient on unix to show
+                        // that the socket buffer is full
+                        if n > 0 && (!cfg!(windows) && n < buf.len()) {
+                            self.registration.clear_readiness(evt);
+                        }
```
hawkw, line 201: this is admittedly very nitpicky, but would it make sense to write this instead:
```suggestion
                        // if we write only part of our buffer, this is sufficient on unix to show
                        // that the socket buffer is full
                        #[cfg(not(windows))]
                        {
                            if n > 0 && n < buf.len() {
                                self.registration.clear_readiness(evt);
                            }
                        }

```
so that we skip the `n > 0` check on windows, where the `if` condition will always evaluate `false` anyway? or is the compiler smart enough to const-fold that too?

hawkw (approved):

tokio/src/io/poll_evented.rs
Noah-Kennedy: I don't particularly care either way, but I would be very surprised if LLVM wasn't able to optimize this away, considering that we are inserting a const boolean value here that guarantees the outcome of the whole expression.

Noah-Kennedy: I think I'm going leave this as-is, since we are already doing the same thing for reads, and I'd like to keep this change as just "copy over the behavior for reads". I'm perfectly happy with another PR that changes both in the manner you suggested though.

I might also be slightly influenced by not wanting to rerun CI here as well.

hawkw: fair enough, i'm definitely fine with this as-is!

PR #4983, opened 2022-09-06
rt: move time driver into `runtime` module

carllerche:
This patch moves the time driver into the runtime module. The time driver
is a runtime concern and is only used by the runtime. Moving the drivers
is the first step to cleaning up some Tokio internals. There will be
follow-up patches that integrate the drivers and other runtime concerns
more closely.

This is an internal refactor and should not impact any public APIs.

Similar to #4942


Noah-Kennedy:
It looks like some tests were deleted. What was the reasoning there?

carllerche:
@Noah-Kennedy they were not necessary. Both were duplicates.

Noah-Kennedy (approved):

PR #4991, opened 2022-09-07
rt: rm internal Park,Unpark traits

carllerche:
Years ago, Tokio was organized as a cluster of separate crates. This architecture required a trait to represent "park the thread and do work.". When all crates were combined into a single crate, the `Park` and `Unpark` traits remained as internal details.

This patch removes these traits as they are no longer needed. This is in service of a future refactor that will decouple the various resource drivers and store them in a single struct instead of nesting them. However, in the meantime, removing the Park/Unpark traits adds a bit of messy code to make the conditional compilation work. This code will (hopefully) be short-lived.

I am trying to submit refactors in small, approachable chunks to avoid having to review massive PRs.


tokio/src/runtime/driver.rs
```diff
@@ -1,20 +1,27 @@
...
 cfg_io_driver! {
-    type IoDriver = crate::runtime::io::Driver;
-    type IoStack = crate::park::either::Either<ProcessDriver, ParkThread>;
+    pub(crate) type IoDriver = crate::runtime::io::Driver;
     pub(crate) type IoHandle = Option<crate::runtime::io::Handle>;
 
-    fn create_io_stack(enabled: bool) -> io::Result<(IoStack, IoHandle, SignalHandle)> {
-        use crate::park::either::Either;
+    #[derive(Debug)]
+    pub(crate) enum IoStack {
```
carllerche, line 14: Some boilerplate / duplication here, but I aim to clean this up later.

tokio/src/runtime/io/mod.rs
```diff
@@ -145,7 +144,35 @@ impl Driver {
         }
     }
 
-    fn turn(&mut self, max_wait: Option<Duration>) -> io::Result<()> {
+    // TODO: remove this in a later refactor
+    cfg_not_rt! {
```
carllerche, line 148: Same, will clean up later.

.github/workflows/ci.yml
```diff
@@ -281,6 +281,8 @@ jobs:
         run: cargo install cargo-hack
       - name: check --each-feature
         run: cargo hack check --all --each-feature -Z avoid-dev-deps
+      - name: check net,time
+        run: cargl check -p tokio --no-default-features --features net,time -Z avoid-dev-deps
```
Noah-Kennedy, line 285: ph'nglui mglw'nafh Ferris `Cargl` wgah'nagl fhtagn

Noah-Kennedy (approved):
I think this is fine once you get it through CI.

hawkw (approved):

PR #4993, opened 2022-09-08
rt: remove `Handle` reference from time driver

carllerche:
A step towards the goal of consolidating thread-local variables, this patch removes the `time::Handle` reference from the time driver struct. The reference is moved a level up and passed into methods that need it.

The end goal is to have a single `Driver` struct that holds each sub-driver and a single `Handle` ref-counted struct that holds all the misc handles. Further work on the time driver is blocked by some other refactoring that will happen in follow-up patches.

### DIFF
--- a/tokio/src/runtime/driver.rs
+++ b/tokio/src/runtime/driver.rs
@@ -147,7 +147,10 @@ cfg_not_process_driver! {
 cfg_time! {
     #[derive(Debug)]
     pub(crate) enum TimeDriver {
-        Enabled(crate::runtime::time::Driver),
+        Enabled {
+            driver: crate::runtime::time::Driver,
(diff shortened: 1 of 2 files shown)


tokio/src/runtime/driver.rs
```diff
@@ -147,7 +147,10 @@ cfg_not_process_driver! {
 cfg_time! {
     #[derive(Debug)]
     pub(crate) enum TimeDriver {
-        Enabled(crate::runtime::time::Driver),
+        Enabled {
```
carllerche, line 150: The `TimeDriver` enum is temporary and just used to help make the PRs more approachable.

Noah-Kennedy (approved):

PR #4997, opened 2022-09-11
Fix LengthDelimitedCodec buffer over-reservation.

yotamofek:
The `LengthDelimitedCodec` calls `BytesMut::reserve` on two occasions:
- after the header is read, to allocate memory for the expected frame
- after a frame is read, to allocate memory for another header that may follow

In both cases, the implementation currently calls `src.reserve(n)` which reserves the necessary `n` bytes for [**additional** data](https://docs.rs/bytes/latest/bytes/struct.BytesMut.html#method.reserve), disregarding the fact that the buffer might already have additional data present that just wasn't decoded yet.

## Motivation

This causes excess memory allocation, which might be quite substantial in certain cases.

## Solution

Take into account the buffer's current length when calculating the amount of memory that should be reserved.

As a small drive-by cleanup, I also removed a check that `n != 0` before calling to `src.advance(n)`, since that is redundant according to [the docs](https://docs.rs/bytes/latest/bytes/trait.Buf.html#tymethod.advance):
> A call with cnt == 0 should never panic and be a no-op.

Not sure if/how to write tests for this, any feedback would be welcome.

### DIFF
--- a/tokio-util/src/codec/length_delimited.rs
+++ b/tokio-util/src/codec/length_delimited.rs
@@ -522,15 +522,11 @@ impl LengthDelimitedCodec {
             }
         };
 
-        let num_skip = self.builder.get_num_skip();
-
-        if num_skip > 0 {
-            src.advance(num_skip);
-        }
+        src.advance(self.builder.get_num_skip());
 
         // Ensure that the buffer has enough space to read the incoming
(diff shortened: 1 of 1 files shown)


Darksonn (approved):
Thanks.

PR #4036, opened 2021-08-12
The test-util feature depends on rt, sync, and time

asomers:
Fixes #4035

## Motivation

The build will fail for a crate that uses tokio with the test-util feature but not the rt, sync, and time features.

## Solution

Add the features dependencies to Cargo.toml

## Tests

This bug was revealed by running `cargo hack check --feature-powerset --no-dev-deps`.  However, that command takes 205 minutes with this patch, and even longer without it, so it can't be added to CI.

PR #5007, opened 2022-09-13
ci: use --feature-powerset --depth 2 in features check

taiki-e:
## Motivation

As has been pointed out a few times in the past (e.g., https://github.com/tokio-rs/tokio/pull/4036#issuecomment-897252076), each-feature is not sufficient for features check.

Ideally, we'd like to check all combinations of features, but there are too many combinations.
So limit the max number of simultaneous feature flags to 2 by [`--depth` option](https://github.com/taiki-e/cargo-hack/pull/59). I think this should be sufficient in most cases as @carllerche said in https://github.com/taiki-e/cargo-hack/issues/58.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -280,10 +280,10 @@ jobs:
       - name: Install cargo-hack
         uses: taiki-e/install-action@cargo-hack
       - name: check --feature-powerset
-        run: cargo hack check --all --feature-powerset --depth 2 -Z avoid-dev-deps
+        run: cargo hack check --all --feature-powerset --depth 2 -Z avoid-dev-deps --keep-going
       # Try with unstable feature flags
       - name: check --feature-powerset --unstable
-        run: cargo hack check --all --feature-powerset --depth 2 -Z avoid-dev-deps
+        run: cargo hack check --all --feature-powerset --depth 2 -Z avoid-dev-deps --keep-going
         env:
           RUSTFLAGS: --cfg tokio_unstable -Dwarnings
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -279,13 +279,11 @@ jobs:
       - uses: Swatinem/rust-cache@v1
       - name: Install cargo-hack
         uses: taiki-e/install-action@cargo-hack
-      - name: check --each-feature
-        run: cargo hack check --all --each-feature -Z avoid-dev-deps
-      - name: check net,time
-        run: cargo check -p tokio --no-default-features --features net,time -Z avoid-dev-deps
+      - name: check --feature-powerset
+        run: cargo hack check --all --feature-powerset --depth 2 -Z avoid-dev-deps
       # Try with unstable feature flags
-      - name: check --each-feature --unstable
-        run: cargo hack check --all --each-feature -Z avoid-dev-deps
+      - name: check --feature-powerset --unstable
+        run: cargo hack check --all --feature-powerset --depth 2 -Z avoid-dev-deps
         env:
           RUSTFLAGS: --cfg tokio_unstable -Dwarnings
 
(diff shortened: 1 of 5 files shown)


Darksonn (approved):
LGTM

taiki-e:
hmm...

```
info: running `cargo check -Z avoid-dev-deps --no-default-features --features rt,signal` on tokio (56/225)
   Compiling tokio v1.21.1 (/home/runner/work/tokio/tokio/tokio)
  --> tokio/src/runtime/driver.rs:46:58
   |
46 |                 IoStack::Enabled(v) => IoUnpark::Enabled(v.handle()),
   |                                        ----------------- ^^^^^^^^^^ expected struct `runtime::io::Handle`, found struct `signal::unix::driver::Handle`
   |                                        |
   |                                        arguments to this enum variant are incorrect
   |
note: tuple variant defined here
  --> tokio/src/runtime/driver.rs:20:9
   |
20 |         Enabled(crate::runtime::io::Handle),
   |         ^^^^^^^

error: could not compile `tokio` due to previous error
error: process didn't exit successfully: `/home/runner/.rustup/toolchains/nightly-2022-07-26-x86_64-unknown-linux-gnu/bin/cargo check -Z avoid-dev-deps --manifest-path tokio/Cargo.toml --no-default-features --features rt,signal` (exit status: 101)
```

taiki-e:
The problem seems to be that the type returned by the `handle` method of the `SignalDriver` (used as `ProcessDriver` when the `process` feature is disabled) is different from that of `ProcessDriver`.

https://github.com/tokio-rs/tokio/blob/0fddb765d4e1f14766271f38c0a9162968abce0d/tokio/src/runtime/driver.rs#L125-L143

However, `crate::signal::unix::driver::Handle` that returned by `SignalDriver::handle` doesn't have `unpark` method, so the following simple patch doesn't seem to work.

```diff
diff --git a/tokio/src/runtime/driver.rs b/tokio/src/runtime/driver.rs
index 2a0deb80..6ef98825 100644
--- a/tokio/src/runtime/driver.rs
+++ b/tokio/src/runtime/driver.rs
@@ -17,7 +17,7 @@ pub(crate) enum IoStack {
     }
 
     pub(crate) enum IoUnpark {
-        Enabled(crate::runtime::io::Handle),
+        Enabled(ProcessHandle),
         Disabled(UnparkThread),
     }
 
@@ -127,6 +127,7 @@ fn create_signal_driver(io_driver: IoDriver) -> io::Result<(SignalDriver, Signal
 
 cfg_process_driver! {
     type ProcessDriver = crate::process::unix::driver::Driver;
+    type ProcessHandle = crate::runtime::io::Handle;
 
     fn create_process_driver(signal_driver: SignalDriver) -> ProcessDriver {
         crate::process::unix::driver::Driver::new(signal_driver)
@@ -136,6 +137,7 @@ fn create_process_driver(signal_driver: SignalDriver) -> ProcessDriver {
 cfg_not_process_driver! {
     cfg_io_driver! {
         type ProcessDriver = SignalDriver;
+        type ProcessHandle = crate::signal::unix::driver::Handle;
 
         fn create_process_driver(signal_driver: SignalDriver) -> ProcessDriver {
             signal_driver
```

```conole
$ cargo check -Z avoid-dev-deps --manifest-path tokio/Cargo.toml --no-default-features --features rt,signal
error[E0599]: no method named `unpark` found for reference `&signal::unix::driver::Handle` in the current scope
  --> tokio/src/runtime/driver.rs:77:43
   |
77 |                 IoUnpark::Enabled(v) => v.unpark(),
   |                                           ^^^^^^ method not found in `&signal::unix::driver::Handle`

For more information about this error, try `rustc --explain E0599`.
```

taiki-e:
cc @ipetkov

carllerche:
Seems like we do need this!

carllerche:
@taiki-e I can work on fixing this build error since it is from my work.

carllerche:
This build error happened because I incorrectly combined "handle" and "unpark" when I got rid of the `Park` trait. I will keep them separate for now to fix the build.

carllerche:
Thanks for this!

PR #5008, opened 2022-09-13
rt: store driver handles next to scheduler handle

carllerche:
In an earlier PR (#4629), driver handles were moved into the scheduler handle (`Spawner`). This was done to let the multi-threaded scheduler have direct access to the thread pool spawner.

However, we are now working on a greater decoupling of the runtime internals. All drivers and schedulers will be peers, stored in a single thread-local variable, and the scheduler will be passed the full runtime::Handle. This will achieve the original goal of giving the scheduler access to the thread-pool while also (hopefully) simplifying other aspects of the code.

### DIFF
--- a/tokio/src/runtime/builder.rs
+++ b/tokio/src/runtime/builder.rs
@@ -833,6 +833,7 @@ impl Builder {
 
     fn build_current_thread_runtime(&mut self) -> io::Result<Runtime> {
         use crate::runtime::{Config, CurrentThread, HandleInner, Kind};
+        use std::sync::Arc;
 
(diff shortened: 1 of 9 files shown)


carllerche:
When I tried to tease out the time Arcs from the time driver, I discovered that I needed `runtime::Handle` to be defined without the `rt` feature flag enabled. However, getting this to work with the current struct layout is non trivial since `Spawner` is an `rt` concern. By moving the driver handles to the top level of `Handle`, a follow-up PR will be able to define `Handle` without `rt` (though the type will be private).

Noah-Kennedy (approved):

PR #5012, opened 2022-09-14
rt: update TimerEntry to use runtime::Handle

carllerche:
The `TimerEntry` struct is the internal integration point for public time APIs (`sleep`, `interval`, ...) with the time driver. Currently, `TimerEntry` holds a ref-counted reference to the time driver handle.

This patch replaces the reference to the time driver handle with a reference to the runtime handle. This is part of a larger effort to consolidate internal handles across the runtime.

### DIFF
--- a/tokio/src/lib.rs
+++ b/tokio/src/lib.rs
@@ -498,6 +498,12 @@ cfg_rt! {
 }
 cfg_not_rt! {
     // The `runtime` module is used when the IO or time driver is needed.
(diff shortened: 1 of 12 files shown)


carllerche:
You probably want to review this with "hide whitespace"

Noah-Kennedy (approved):

Noah-Kennedy:
CI truly is the best reviewer sometimes

PR #5022, opened 2022-09-16
rt: split internal runtime::Handle concerns

carllerche:
The `runtime::Handle` struct is part of the public API but is also used internally. This has created a bit of tension. An earlier patch made defined Handle as a private struct in some cases when `rt` is not enabled.

This patch splits out internal handle concerns into a new `scheduler::Handle` type, which will only be internal. This also defines a `Handle` type for each scheduler variant. Eventually, the per-scheduler `Handle` types will replace the per-scheduler `Spawner` types, but more work is needed before we can make that change.


carllerche:
I would suggest disabling whitespace in the diff. `runtime::Handle` is mostly an indentation change.

Noah-Kennedy:
I think it might be advantageous to not have overlapping names for these handles. Could we name the internal handle "SchedulerHandle" or "InternalHandle" or something else?

carllerche:
@Noah-Kennedy I can look into cleanup later, but IMO it is more idiomatic to use mod prefix. ie. `scheduler::Handle`. This avoids the "stuttering" that you would get with `scheduler::SchedulerHandle`. See this discussion: https://github.com/rust-lang/api-guidelines/discussions/66

Noah-Kennedy:
@carllerche Good point, that is more readable.

Noah-Kennedy (approved):

PR #5023, opened 2022-09-16
rt: update CurrentThread to use its own Handle

carllerche:
This is the next step in the internal runtime refactoring. This patch updates `CurrentThread` to use `Arc<Handle>` for all of its cross-thread needs instead of `Arc<Shared>`.

The effect of this change is that the current-thread scheduler only has a single `Arc` type, which includes the driver handles, keeping all "shared state" together.

### DIFF
--- a/tokio/src/runtime/scheduler/current_thread.rs
+++ b/tokio/src/runtime/scheduler/current_thread.rs
@@ -372,7 +372,7 @@ impl Context {
 impl Handle {
     /// Spawns a future onto the `CurrentThread` scheduler
     pub(crate) fn spawn<F>(
-        self: &Arc<Self>,
+        me: &Arc<Self>,
         future: F,
         id: crate::runtime::task::Id,
     ) -> JoinHandle<F::Output>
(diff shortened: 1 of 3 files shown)


tokio/src/runtime/scheduler/current_thread.rs
```diff
@@ -57,10 +56,7 @@ pub(crate) struct Handle {
 /// a function that will perform the scheduling work and acts as a capability token.
 struct Core {
     /// Scheduler run queue
-    tasks: VecDeque<task::Notified<Arc<Shared>>>,
-
-    /// Sendable task spawner
-    spawner: Spawner,
+    tasks: VecDeque<task::Notified<Arc<Handle>>>,
```
Noah-Kennedy, line 59: I don't think these comments are accurate anymore with this change.

tokio/src/runtime/scheduler/current_thread.rs
```diff
@@ -78,18 +74,13 @@ struct Core {
...
-#[derive(Clone)]
-pub(crate) struct Spawner {
-    shared: Arc<Shared>,
-}
-
 /// Scheduler state shared between threads.
 struct Shared {
     /// Remote run queue. None if the `Runtime` has been dropped.
-    queue: Mutex<Option<VecDeque<task::Notified<Arc<Shared>>>>>,
+    queue: Mutex<Option<VecDeque<task::Notified<Arc<Handle>>>>>,
```
Noah-Kennedy, line 80: Similar comment to above.

Noah-Kennedy (approved):

hawkw (approved):

PR #5026, opened 2022-09-17
rt: move driver unpark out of multi-thread parker

carllerche:
This patch removes the driver Unpark handle out of the multi-thread parker and passes a reference in when it is needed. This is a first step towards getting rid of the separate driver unpark handle in favor of just using the regular driver handle. Because the regular driver handle is owned at a higher level (at the top of the worker struct).


carllerche:
See #5027 for the next step in the refactoring.

Darksonn (approved):

hawkw (approved):

PR #5032, opened 2022-09-19
`is_closed` for `mpsc::{Receiver, UnboundedReceiver}`

farnz:
There's a use case for channels that wants to know up-front (before calling `recv` or `try_recv`)that there's a possibility of new messages coming in. Make it possible to check whether the channel is closed and empty, which implies that there will never be new messages on this channel.

Fixes: #4638

## Motivation

Taken from #4638 - if you're building a pipe-type abstraction over an MPSC channel carrying byte buffers, then it's useful to know that the pipe is now terminated without having to handle the possibility of data arriving when you do the check.

## Solution

Refactor so that I can see if the current block is closed without reading the value. I believe this avoids safety issues, since I never go near the actual data.

This is incomplete as-is, since it only allows you to look to see if the channel is closed, not check to see if there's data to be read, or to wait until data is available. I am working on those two items, to let you build a useful `TcpStream` equivalent from an MPSC channel.

### DIFF
--- a/tokio/src/sync/mpsc/block.rs
+++ b/tokio/src/sync/mpsc/block.rs
@@ -104,6 +104,15 @@ impl<T> Block<T> {
         other_index.wrapping_sub(self.start_index) / BLOCK_CAP
     }
 
+    /// Check to see if the given slot is a tx closed entry
+    pub(crate) fn is_closed(&self, slot_index: usize) -> bool {
+        let offset = offset(slot_index);
+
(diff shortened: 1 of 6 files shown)


Noah-Kennedy:
@faern #5016 is already in review.

farnz:
> @faern #5016 is already in review.

Thanks - for some reason, GitHub chose not to show me that pull request when I hit refresh this morning.

PR #1228, opened 2019-06-29
WIP: Rewrite examples with `async`.

rubdos:
## Motivation

> We would love to get community involvement in this work. There is a lot to do. The best way to get involved is to:

> The examples should be updated to idiomatic async / await syntax.

> The first alpha release will happen once the examples have been updated.

## Solution

I started rewriting the examples using the new `#[tokio::main]` and `async` and `await` keywords. I will have every example in it's own commit, such that you can cherry pick what you'd like to keep.

I'm on Gitter and IRC, so feel free to ping me there.

PR #5044, opened 2022-09-22
examples: `hello_world`: Use `write_all` instead of `write`

dongcarl:
The existing code for the `hello_world` example uses `tokio::io::AsyncWriteExt::write`, which works for most sanely configured TCP servers but may mislead newcomers as to the correct usage of the `tokio::io::AsyncWriteExt` API.

The more robust and correct call is `tokio::io::AsyncWriteExt::write_all`, which is equivalent to the `io::write_all` that was used in this example prior to #1228.

Usage prior to #1228:

https://github.com/tokio-rs/tokio/blob/ceed29586b446f87cce381d7d1e81bd01421f115/tokio/examples_old/hello_world.rs#L30


Darksonn (approved):
Thanks. The CI failures are unrelated to your change.

PR #5062, opened 2022-09-28
use random port in tests, not 8080

farnz:
These tests fail on my system because I have a local service already
bound to 127.0.0.1:8080 - the exact port number doesn't matter, so allow
the kernel to choose a random port instead.

## Motivation

These tests fail whenever I have my own local service running on port 8080. Similar problems would apply to any port we choose, and `tcp_stream_from_std_panic_caller` avoids the problem already by listening on a kernel-chosen port.

## Solution

Reuse the solution from `tcp_stream_from_std_panic_caller` - we can listen to any random port successfully, so let the kernel choose an unused port for our test.


farnz:
The macOS CI failure appears unrelated - it's in a test case I've not touched.

Darksonn (approved):
Thanks.

PR #5071, opened 2022-10-02
chore: use `std::AtomicPtr` and `std::AtomicU8`

xudong963:
## Motivation
Some annotations are out of date, and may be misleading.

## Solution

Remove useless annotaions.


tokio/src/loom/std/atomic_ptr.rs
```diff
@@ -1,7 +1,6 @@
 use std::fmt;
 use std::ops::{Deref, DerefMut};
 
-/// `AtomicPtr` providing an additional `load_unsync` function.
```
farnz, line 4: This no longer provides any extra functions - `loom` directly imports `std::sync::atomic::AtomicBool` because it doesn't need to add extras.

Would it make sense to remove this file completely, and change `tokio/src/loom/std/mod.rs` to use `std::sync::atomic::AtomicPtr` directly?

tokio/src/loom/std/atomic_u32.rs
```diff
@@ -2,7 +2,6 @@ use std::cell::UnsafeCell;
 use std::fmt;
 use std::ops::Deref;
 
-/// `AtomicU32` providing an additional `load_unsync` function.
```
farnz, line 5: This still provides `unsync_load` - should the comment be updated instead of removed?

tokio/src/loom/std/atomic_u8.rs
```diff
@@ -2,7 +2,6 @@ use std::cell::UnsafeCell;
 use std::fmt;
 use std::ops::Deref;
 
-/// `AtomicU8` providing an additional `load_unsync` function.
```
farnz, line 5: As `AtomicPtr` - would it be more sensible to just remove this completely?

tokio/src/loom/std/atomic_u32.rs
xudong963: Sorry, my mistake -- Will update.

tokio/src/loom/std/atomic_u8.rs
xudong963: fixed

Darksonn (approved):
LGTM.

PR #5078, opened 2022-10-06
remove unnecessary `Option` for `mio::Events` in `Driver`

cjwcommuny:
## Motivation

`tokio::runtime::io::Driver` has a field `events` with the type`Option<mio::Events>`. The `Option` wrapper is used to solve the second mutable borrow conflicts in the method `Driver::turn`. This requires to take the inner `mio::Events` every time `Driver::turn` is called, which introduce unnecessary overheads.

## Solution

1. Change `Option<mio::Events>` to `mio::Events` so that no `Option::take` is needed to call in the method `Driver::turn`.
2. Change the function signature of `Driver::dispatch` from:
    ```rust
    fn dispatch(&mut self, token: mio::Token, ready: Ready)
    ```
    to

    ```rust
    fn dispatch(resources: &mut Slab<ScheduledIo>, tick: u8, token: mio::Token, ready: Ready)
    ```

    so that the compiler won't complain.

### DIFF
--- a/tokio/src/runtime/io/mod.rs
+++ b/tokio/src/runtime/io/mod.rs
@@ -27,7 +27,7 @@ pub(crate) struct Driver {
     tick: u8,
 
     /// Reuse the `mio::Events` value across calls to poll.
-    events: Option<mio::Events>,
+    events: mio::Events,
 
     /// Primary slab handle containing the state for each resource registered
     /// with this driver.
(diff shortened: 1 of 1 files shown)


Darksonn (approved):
This seems good, thanks.

cjwcommuny:
I reformat the code.

PR #5095, opened 2022-10-11
task: wake local tasks to the local queue when woken by the same thread

hawkw:
## Motivation

Currently, when a task spawned on a `LocalSet` is woken by an I/O driver
or time driver running on the same thread as the `LocalSet`, the task is
pushed to the `LocalSet`'s locked remote run queue, rather than to its
unsynchronized local run queue. This is somewhat unfortunate, as it
negates some of the performance benefit of having an unsynchronized
local run queue. Instead, tasks are only woken to the local queue when
they are woken by other tasks also running on the local set.

This occurs because the local queue is only used when the `CONTEXT`
thread-local contains a `Context` that's the same as the task's
`Schedule` instance (an `Arc<Shared>`)'s `Context`. When the `LocalSet`
is not being polled, the thread-local is unset, and the local run queue
cannot be accessed by the `Schedule` implementation for `Arc<Shared>`.

## Solution

This branch fixes this by moving the local run queue into `Shared` along
with the remote run queue. When an `Arc<Shared>`'s `Schedule` impl wakes
a task and the `CONTEXT` thread-local is `None` (indicating we are not
currently polling the `LocalSet` on this thread), we now check if the
current thread's `ThreadId` matches that of the thread the `LocalSet`
was created on, and push the woken task to the local queue if it was.

Moving the local run queue into `Shared` is somewhat unfortunate, as it
means we now have a single field on the `Shared` type which must not be
accessed from other threads, and must add an `unsafe impl Sync for Shared`. 
owever, it's the only viable way to wake to the local queue
from the `Schedule` impl for `Arc<Shared>`, so I figured it was worth
the additional unsafe code. I added a debug assertion to check that the
local queue is only accessed from the thread that owns the `LocalSet`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,10 +80,6 @@ jobs:
       - uses: Swatinem/rust-cache@v1
       - name: Install cargo-hack
         uses: taiki-e/install-action@cargo-hack
-      # trying to figure out what the panic i cant repro was lol
-      - name: test tokio full
-        run: cargo test --features full --test task_local_set -- --nocapture --test-threads=1
-        working-directory: tokio
 
       # Run `tokio` with `full` features. This excludes testing utilities which
       # can alter the runtime behavior of Tokio.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,10 @@ jobs:
       - uses: Swatinem/rust-cache@v1
       - name: Install cargo-hack
         uses: taiki-e/install-action@cargo-hack
+      # trying to figure out what the panic i cant repro was lol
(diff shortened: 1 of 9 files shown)


hawkw:
uhhh...looks like the assertion is failing on freebsd (https://cirrus-ci.com/task/5075280174776320?logs=test#L1661)?

that's very surprising given that `LocalSet` being `!Send` means that the localset will always be dropped on the thread that created it and the assertion failure appears to be in a `Drop` (given that the process gets SIGABRT'd)...and it works fine for me (on Linux). can...can thread IDs change under freebsd? are they _supposed_ to ?

carllerche (approved):
LGTM, thanks.

PR #5100, opened 2022-10-12
macros: rename `thread_local!` to `tokio_thread_local!`

hawkw:
## Motivation

Currently, Tokio uses an internal `thread_local!` macro which compiles to either `std::thread_local!` or `loom::thread_local!`, and switches between using `const` expressions in thread locals or not based on cfg attributes. Because this macro has the same name as `std::thread_local!`, though, which is ambiently available, it's not always clear which macro is being used, and it seems like there are sometimes name resolution issues that can result in the redefined macro not being used as we'd expect.

## Solution

This branch renames the internal macro to `tokio_thread_local!` to make it explicit that we want to use it, rather than `std::thread_local!`. I've removed some `cfg` attributes in `local_set.rs` that shouldn't be necessary, since our macro *should* handle that already.

### DIFF
--- a/tokio/src/coop.rs
+++ b/tokio/src/coop.rs
@@ -31,7 +31,7 @@
 
 use std::cell::Cell;
 
-thread_local! {
+tokio_thread_local! {
     static CURRENT: Cell<Budget> = const { Cell::new(Budget::unconstrained()) };
 }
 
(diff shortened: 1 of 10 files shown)


hawkw:
This is probably gonna break something, so I'm kicking off a full ci build to figure out what...

hawkw:
okay yup cool it broke stuff

carllerche (approved):

carllerche:
Closed by #5103

PR #5103, opened 2022-10-13
tokio: rename internal thread_local macro

carllerche:
Tokio maintains an internal thread_local macro that abstracts some conditional build logic. Before this patch, the macro was named the same as the `std` macro (`thread_local`). This resulted in confusion as to whether or not the internal macro or the std macro was being called.

This patch renames the internal macro to `tokio_thread_local`, making it more obvious.

### DIFF
--- a/tokio/src/coop.rs
+++ b/tokio/src/coop.rs
@@ -31,7 +31,7 @@
 
 use std::cell::Cell;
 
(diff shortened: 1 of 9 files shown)


hds (approved):
This makes a lot of sense.

PR #5104, opened 2022-10-13
rt: remove a conditional compilation clause

carllerche:
The `LocalSet` implementation includes a conditional compilation clause that removes the `const` statement from the `thread_local` definition. However, there already is an internal macro that does this: `tokio_thread_local`.

This patch removes the conditional compilation in favor of using the `tokio_thread_local` macro. This also fixes a conditional compilation issue with an internal utility (`RcCell`).

### DIFF
--- a/tokio/src/task/local.rs
+++ b/tokio/src/task/local.rs
@@ -272,13 +272,6 @@ pin_project! {
     }
 }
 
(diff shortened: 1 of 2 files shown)


hawkw (approved):
assuming CI passes, ship it!

PR #5106, opened 2022-10-14
runtime: update the alignment of `CachePadded`

cjwcommuny:
## Motivation & Solution

Update the alignment of `CachePadded` based on [crossbeam](https://github.com/crossbeam-rs/crossbeam/blob/fa35346b7c789bba045ad789e894c68c466d1779/crossbeam-utils/src/cache_padded.rs#L62-L127)'s implementation.


carllerche:
Could you add a comment that it is copied from crossbeam/cache_padded?

Darksonn (approved):
LGTM.

PR #5121, opened 2022-10-24
rt: move signal driver to runtime module

carllerche:
The signal feature only requires a driver with unix platforms. The unix signal
driver uses the I/O driver. A future refactor would like to make the signal driver
use internal APIs of the I/O driver. To do this, the signal driver must be moved
to the runtime module.


ipetkov (approved):
LGTM

PR #5125, opened 2022-10-25
rt: signal driver now uses I/O driver directly

carllerche:
The signal driver uses a `UnixStream` to receive signal events. Previously, the
signal driver used `PollEvented` internally in order to receive events on the
`UnixStream`. However, using `PollEvented` from within a runtime driver created
a circular link between the runtime and the `PollEvented` instance.

This patch replaces `PollEvented` usage in favor of accessing the I/O driver
directly. The I/O driver now reserves a token for signal-related events and
tracks signal readiness internally. The signal driver queries the I/O driver to
check for signal-related readiness.

### DIFF
--- a/tokio/src/runtime/io/mod.rs
+++ b/tokio/src/runtime/io/mod.rs
@@ -1,4 +1,4 @@
-#![cfg_attr(not(feature = "rt"), allow(dead_code))]
+#![cfg_attr(not(all(feature = "rt", feature = "net")), allow(dead_code))]
 
 mod registration;
 pub(crate) use registration::Registration;
@@ -25,6 +25,9 @@ pub(crate) struct Driver {
     /// Tracks the number of times `turn` is called. It is safe for this to wrap
     /// as it is mostly used to determine when to call `compact()`.
     tick: u8,
+
+    /// True when an event with the signal token is received
+    signal_ready: bool,
 
     /// Reuse the `mio::Events` value across calls to poll.
     events: mio::Events,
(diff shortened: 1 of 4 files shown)


tokio/src/runtime/io/mod.rs
```diff
@@ -86,6 +89,7 @@ enum Tick {
 // TODO: Don't use a fake token. Instead, reserve a slot entry for the wakeup
 // token.
 const TOKEN_WAKEUP: mio::Token = mio::Token(1 << 31);
+const TOKEN_SIGNAL: mio::Token = mio::Token(1 + (1 << 31));
```
carllerche, line 92: This is in the reserved space starting at `1 << 31`.

I expect a later I/O driver refactor will change this.

tokio/tests/rt_metrics.rs
```diff
@@ -375,20 +375,18 @@ fn io_driver_fd_count() {
     let rt = current_thread();
     let metrics = rt.metrics();
 
-    // Since this is enabled w/ the process driver we always
-    // have 1 fd registered.
-    assert_eq!(metrics.io_driver_fd_registered_count(), 1);
+    assert_eq!(metrics.io_driver_fd_registered_count(), 0);
```
carllerche, line 378: A nice side effect: this is actually correct now.

Darksonn (approved):

ipetkov (approved):
Looks great! Way cleaner now too

PR #5127, opened 2022-10-26
rt: start decoupling I/O driver and I/O handle

carllerche:
This is the start of applying a similar treatment to the I/O driver as the time
driver. The I/O driver will no longer hold its own reference to the I/O handle.
Instead, the handle is passed in when needed.

This patch also moves the process driver to the `runtime` module.


Noah-Kennedy (approved):
Nice!

PR #5128, opened 2022-10-26
rt: switch io::handle refs with scheduler:Handle

carllerche:
The `schedule::Handle` reference is the internal runtime handle. This
patch replaces owned refs to `runtime::io::Handle` with
`scheduler::Handle`.

### DIFF
--- a/tokio/src/runtime/driver.rs
+++ b/tokio/src/runtime/driver.rs
@@ -82,10 +82,11 @@ impl Handle {
     }
 
     cfg_io_driver! {
+        #[track_caller]
         pub(crate) fn io(&self) -> &crate::runtime::io::Handle {
             self.io
                 .as_ref()
-                .expect("A Tokio 1.x context was found, but I/O is disabled. Call `enable_io` on the runtime builder to enable I/O.")
+                .expect("A Tokio 1.x context was found, but IO is disabled. Call `enable_io` on the runtime builder to enable IO.")
         }
     }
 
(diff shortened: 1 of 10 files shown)


carllerche:
Probably easier to review w/ whitespace disabled.

tokio/src/runtime/driver.rs
```diff
@@ -82,10 +82,11 @@ impl Handle {
     }
 
     cfg_io_driver! {
+        #[track_caller]
         pub(crate) fn io(&self) -> &crate::runtime::io::Handle {
             self.io
                 .as_ref()
-                .expect("A Tokio 1.x context was found, but I/O is disabled. Call `enable_io` on the runtime builder to enable I/O.")
+                .expect("A Tokio 1.x context was found, but IO is disabled. Call `enable_io` on the runtime builder to enable IO.")
```
carllerche, line 89: There is an existing test that expected `IO` to be spelled this way. Instead of updating the test, I changed the message back to what it was before.

tokio/src/runtime/scheduler/mod.rs
```diff
@@ -62,6 +69,11 @@ cfg_rt! {
     use crate::util::RngSeedGenerator;
 
     impl Handle {
+        #[track_caller]
+        pub(crate) fn current() -> Handle {
```
carllerche, line 73: Moving the `current()` function to `scheduler::Handle` makes it easier to get the conditional compilation tests to pass.

Eventually, `runtime::context` will be moved into here, and this will be the primary type to track the "current runtime." I am punting this work to a later PR.

tokio/src/time/sleep.rs
```diff
@@ -248,76 +248,61 @@ cfg_not_trace! {
...
-                    loc.col = location.column(),
+    #[cfg_attr(not(all(tokio_unstable, feature = "tracing")), allow(unused_variables))]
+    #[track_caller]
+    pub(crate) fn new_timeout(
+        deadline: Instant,
+        location: Option<&'static Location<'static>>,
+    ) -> Sleep {
+        use crate::runtime::scheduler;
+
+        let handle = scheduler::Handle::current();
```
carllerche, line 259: Without this change, the conditional compilation test complains that `scheduler::Handle::current()` is not used when only `time` is enabled. I opted to do the work to update this code here instead of silencing the warning temporarily.

Noah-Kennedy (approved):
Looks good!

PR #5132, opened 2022-10-27
test: Improve `tokio_test::task` docs

LucioFranco:
cc @rcoh


tokio-test/src/task.rs
```diff
@@ -11,7 +36,7 @@ use std::task::{Context, Poll, RawWaker, RawWakerVTable, Waker};
 
 use tokio_stream::Stream;
 
-/// TODO: dox
+/// Spawn a future into a [`Spawn`] which wraps the future in a mocked executor.
```
rcoh, line 39: recommend also adding an example here

rcoh (approved):
LGTM, but I'd add an example in both spots so that the function is obvious how to use

tokio-test/src/task.rs
rcoh: I would add more "so what" / why docs here. maybe something like:

As a side note, I'd probably make another function for streams and bound the generic—I get that they are functionally the same but it's not at all obvious you can pass a stream in here & calling `spawn` on a stream doesn't really make sense.

```
/// Test utility to spawn the [`Future`] or [`Stream`] into a [`Spawn`] container
///
/// The `Spawn` container enables direct polling and introspection of the state of the future which is often required
/// when testing code that implements [`Future`] or [`Stream`]. The [`Spawn`] container enables you to poll the 
/// future directly without requiring you to construct your own `Context` or `Waker`. It also handles the bookkeeping required to track wakeups & future state.

**Review by rcoh — APPROVED** [review]
LGTM, but I'd consider fleshing these docs out a bit more. This is an extremely helpful feature and the more key words you can add for SEO the better, eg "`test context`" "`test waker`", "`test custom future`", `call poll without context`

File: tokio-test/src/task.rs [hunk 2]
```diff
@@ -11,24 +36,26 @@ use std::task::{Context, Poll, RawWaker, RawWakerVTable, Waker};
...
+///
+/// This can be used to spawn a [`Future`] or a [`Stream`].
+///
+/// For more information, check the module docs.
```
rcoh, line 43: link?

tokio-test/src/task.rs
```diff
@@ -11,24 +36,26 @@ use std::task::{Context, Poll, RawWaker, RawWakerVTable, Waker};
...
 pub fn spawn<T>(task: T) -> Spawn<T> {
     Spawn {
         task: MockTask::new(),
         future: Box::pin(task),
     }
 }
 
-/// Future spawned on a mock task
+/// Future spawned on a mock task that can be used to poll the future or stream
+/// without needing pinning or context types.
```
rcoh, line 52: link to module docs

tokio-test/src/task.rs
```diff
@@ -91,15 +118,17 @@ impl<T: Unpin> ops::DerefMut for Spawn<T> {
 }
 
 impl<T: Future> Spawn<T> {
-    /// Polls a future
+    /// If `T` is a [`Future`] then poll it. This will handle pinning and the context
+    /// type for the future.
```
rcoh, line 122: a little misleading—kinda makes me think that the code would be

`if t.is_future() { poll } else { do nothing }`

LucioFranco: yeah though not sure how to describe that its a type system if statement rather than a rust language one...

carllerche (approved):

PR #5135, opened 2022-10-27
rt: use signal driver handle via scheduler::Handle

carllerche:
The signal driver still uses an `Arc` internally to track if the driver
is still running, however, using the `scheduler::Handle` to access the
signal driver handle lets us delete some code.

### DIFF
--- a/tokio/src/runtime/signal/mod.rs
+++ b/tokio/src/runtime/signal/mod.rs
@@ -24,16 +24,13 @@ pub(crate) struct Driver {
     receiver: UnixStream,
 
     /// Shared state
-    inner: Arc<Inner>,
+    inner: Arc<()>,
 }
 
-#[derive(Clone, Debug, Default)]
+#[derive(Debug, Default)]
 pub(crate) struct Handle {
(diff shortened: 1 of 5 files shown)


tokio/src/runtime/signal/mod.rs
```diff
@@ -24,17 +24,14 @@ pub(crate) struct Driver {
...
     /// Shared state
-    inner: Arc<Inner>,
+    inner: Arc<()>,
 }
 
-#[derive(Clone, Debug, Default)]
+#[derive(Debug, Default)]
 pub(crate) struct Handle {
-    inner: Weak<Inner>,
+    inner: Weak<()>,
```
Noah-Kennedy, line 32: Might be a good idea to document why the Weak is here, because this is going to confuse people.

Noah-Kennedy (changes requested):

tokio/src/runtime/signal/mod.rs
carllerche: Done.

Noah-Kennedy (approved):

PR #5141, opened 2022-10-28
rt: keep driver cfgs in driver.rs

carllerche:
This patch removes driver-related cfg_* macro calls from `scheduler` in
favor of keeping it in the driver module.


carllerche:
IMO the methods that delegate accessing the specific drivers are not worth keeping around. Reducing cfg_ macros is a better trade off.

LucioFranco (approved):

Noah-Kennedy (approved):

PR #5142, opened 2022-10-28
rt: remove rng_seed+mt test and update docs

carllerche:
While setting the random number generator seed with the multi-threaded
scheduler does result in deterministic behavior related to the random
number generator, threads still introduce non-determinism, making it
hard (impossible?) to test this. There also is little value in doing so.

This patch also updates the docs to remove mention of work-stealing.


hds (approved):
Thanks!

PR #5143, opened 2022-10-29
rt: add runtime::context to unify thread-locals

carllerche:
This patch is the first step towards unifying all the thread-local variables
spread out across Tokio. A new `Context` struct is added, which will be
used to replace the various thread-locals that exist today.

Initially, `Context` only holds the current runtime handle and the random
number generator. Future PRs will add other thread-local states.

A previous PR removed `runtime::context`. At that time,
`runtime::context` was used as an extra layer to access the various
runtime driver handles. This version of `runtime::context` serves a
different purpose (unifying all the thread-locals).

### DIFF
--- a/tokio/src/runtime/context.rs
+++ b/tokio/src/runtime/context.rs
@@ -46,7 +46,7 @@ cfg_rt! {
 
     /// Sets this [`Handle`] as the current active [`Handle`].
     ///
-    /// [`Handle`]: Handle
+    /// [`Handle`]: crate::runtime::scheduler::Handle
     pub(crate) fn try_enter(handle: &scheduler::Handle) -> Option<EnterGuard> {
         let rng_seed = handle.seed_generator().next_seed();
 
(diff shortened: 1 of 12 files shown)


tokio/src/runtime/context.rs
```diff
@@ -0,0 +1,72 @@
...
+
+struct Context {
+    /// Handle to the runtime scheduler running on the current thread.
+    #[cfg(feature = "rt")]
+    scheduler: RefCell<Option<scheduler::Handle>>,
+    rng: FastRand,
+}
+
+tokio_thread_local! {
+    static CONTEXT: Context = {
```
carllerche, line 16: Once all the thread-locals are unified here, I will consider bringing back `const` and avoiding the drop.

Darksonn (approved):

Noah-Kennedy (approved):

Noah-Kennedy:
@carllerche looks like some doc links broke.

notgull (approved):

PR #5150, opened 2022-10-31
tokio: fix function name in UdpSocket recv doc

schultetwin1:
In the "cancellation safety" section of the UdpSocket recv function,
"recv_from" is referenced when it should be "recv"


Darksonn (approved):
Thanks.

PR #5155, opened 2022-11-01
rt: rm coop::budget from `LocalSet::run_until`

carllerche:
The `LocalSet::run_until` future is just a "plain" future that should run on a runtime that already has a coop budget. In other words, the `run_until` future should not get its own budget but should inherit the calling task's budget. Getting this behavior is done by removing the call to `budget` in `run_until`.


Noah-Kennedy (approved):

hds (approved):

hawkw (approved):
this seems right to me

PR #5197, opened 2022-11-16
Remove libc types from external types allow list

jdisanti:
## Motivation

Exposure of the libc ID types was removed in #5191. This change removes them from the allow list
in the `cargo-check-external-types` config.


Darksonn (approved):
Thanks for reminding me of this.

PR #5231, opened 2022-11-24
rt: fix `*_closed` false positives

satakuma:
## Motivation

Readiness futures inconsistently return current readiness of a I/O resource, if it is immediately available, or all readiness relevant for the given `Interest`, if a future needs to wait. In particular, it always returns `read_closed` for `Interest::READABLE` and `write_closed` for `Interest::WRITABLE`, what often is not true. Tokio should not tolerate false positives for `*_closed` events because they are considered final states and are not cleared internally.

In case of a `io_resource.ready(Interest::READABLE | Interest::WRITABLE)` call, this behavior may also lead to false positives of other events.

## Solution

If a `Readiness` future is woken up by a new event, store the relevant readiness and return it later.
It pumps the `Waiter` and enclosing structs a bit (8 bytes on x86_64). If it's an issue, readiness can be packed into something smaller.

Closes: #5098

### DIFF
--- a/tokio/src/runtime/io/scheduled_io.rs
+++ b/tokio/src/runtime/io/scheduled_io.rs
@@ -511,14 +511,22 @@ cfg_io_readiness! {
                         drop(waiters);
                     }
                     State::Done => {
-                        let tick = TICK.unpack(scheduled_io.readiness.load(Acquire)) as u8;
-
                         // Safety: State::Done means it is no longer shared
                         let w = unsafe { &mut *waiter.get() };
 
+                        // Note: the returned tick might be newer then the event
+                        // which notified our waker. This is ok because the future
+                        // still didn't return `Poll::Ready`.
+                        let curr = scheduled_io.readiness.load(Acquire);
+                        let tick = TICK.unpack(curr) as u8;
+
+                        // Add more readiness which might have appeared in the meantime.
+                        let curr_ready = Ready::from_usize(READINESS.unpack(curr));
+                        let ready = w.ready | (curr_ready.intersection(w.interest));
+
                         return Poll::Ready(ReadyEvent {
                             tick,
-                            ready: w.ready,
+                            ready,
                         });
                     }
                 }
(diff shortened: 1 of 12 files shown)


Darksonn (approved):
Looks reasonable to me, but I would also like a review from Carl.

tokio/src/runtime/io/scheduled_io.rs
```diff
@@ -510,14 +511,22 @@ cfg_io_readiness! {
...
 
+                        // Note: the returned tick might be newer than the event
+                        // which notified our waker. This is ok because the future
+                        // still didn't return `Poll::Ready`.
+                        let curr = scheduled_io.readiness.load(Acquire);
+                        let tick = TICK.unpack(curr) as u8;
+
+                        // Add more readiness which might have appeared in the meantime.
+                        let curr_ready = Ready::from_usize(READINESS.unpack(curr));
+                        let ready = w.ready | (curr_ready.intersection(w.interest));
```
carllerche, line 525: If a readiness bit is not included in `curr_ready` then something else already cleared it. Including `w.ready` here would result in additional false positives & unnecessary I/O ops.

Based on discord, it sounds like you want to avoid empty readiness events, but that seems better to me than false positives. The caller would loop and wait again.

carllerche (changes requested):
Thanks for doing this. You are correct the current implementation is buggy. Most of this looks good. I left a comment inline.

satakuma:
@carllerche I applied your suggestion to return the current readiness.

I also updated the docs of every `ready` method to mention false-positives and empty events. My motivation is that empty `Ready` events were not really possible before this change. As far as I can tell from the docs, there is no explicit guarantee about no empty return values, but I can still imagine a program logic which would assume that at least one of the requested state is set. For example, code which would deduce that if `!io.ready(Interest::READABLE).await?.is_readable()`, then the read must be closed. For that reason I think an explicit mention in the docs will be helpful.

PR #5246, opened 2022-11-30
chore: prepare tokio-macros v1.8.2

taiki-e:
# 1.8.2 (November 30th, 2022)

- fix a regression introduced in 1.8.1 ([#5244])

[#5244]: https://github.com/tokio-rs/tokio/pull/5244


taiki-e:
Published https://crates.io/crates/tokio-macros/1.8.2

PR #5251, opened 2022-12-01
rt: fix new yield_now behavior with block_in_place

carllerche:
PR #5223 changed the behavior of `yield_now()` to store yielded tasks and notify them *after* polling the resource drivers. This PR fixes two bugs with this new behavior when combined with `block_in_place()`.

First, we must avoid freeing the deferred task queue when exiting a runtime if it is *not* the root runtime. Because `block_in_place()` allows a user to start a new runtime from within an existing task, this check is necessary.

Second, when a worker core is stolen from a thread during a `block_in_place()` call, we need to ensure that deferred tasks are notified.

### DIFF
--- a/tokio/src/runtime/scheduler/multi_thread/worker.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/worker.rs
@@ -368,6 +368,22 @@ impl Launch {
 }
 
 fn run(worker: Arc<Worker>) {
+    struct AbortOnPanic;
+
+    impl Drop for AbortOnPanic {
+        fn drop(&mut self) {
+            if std::thread::panicking() {
+                eprintln!("worker thread panicking; aborting process");
+                std::process::abort();
+            }
+        }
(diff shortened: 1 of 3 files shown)


tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -368,6 +368,22 @@ impl Launch {
...
+                eprintln!("worker thread panicking; aborting process");
+                std::process::abort();
+            }
+        }
+    }
+
+    // Catching panics on worker threads in tests is quite tricky. Instead, when
+    // debug assertions are enabled, we just abort the process.
+    #[cfg(debug_assertions)]
+    let _abort_on_panic = AbortOnPanic;
```
carllerche, line 385: This is just for tests and hopefully catching issues sooner.

Noah-Kennedy (approved):

PR #5257, opened 2022-12-02
ci: check task::Id conditional compilation

carllerche:
This adds CI coverage for a couple of code paths that are not currently
hit in CI:

* no `const fn Mutex::new`
* no `AtomicU64`

This is done by adding some new CFG flags used only for tests in order
to force those code paths.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -268,26 +268,42 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        target:
-          - i686-unknown-linux-gnu
-          - arm-unknown-linux-gnueabihf
-          - armv7-unknown-linux-gnueabihf
-          - aarch64-unknown-linux-gnu
+        include:
+          - target: i686-unknown-linux-gnu
+          - target: arm-unknown-linux-gnueabihf
+          - target: armv7-unknown-linux-gnueabihf
+          - target: aarch64-unknown-linux-gnu
+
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -268,26 +268,39 @@ jobs:
...
-          - armv7-unknown-linux-gnueabihf
-          - aarch64-unknown-linux-gnu
+        include:
+          - target: i686-unknown-linux-gnu
+          - target: arm-unknown-linux-gnueabihf
+          - target: armv7-unknown-linux-gnueabihf
+          - target: aarch64-unknown-linux-gnu
+
+          # Run a platform without AtomicU64 and no const Mutex::new
+          - target: arm-unknown-linux-gnueabihf
```
taiki-e, line 278: FYI, arm-unknown-linux-gnueabihf is armv6 so it has AtomicU64 (i.e., the compilation will succeed even if we accidentally depends on AtomicU64). If you want to test an arm target that does not have AtomicU64, you need to use a pre-v6 target such as armv5te-unknown-linux-gnueabi. (Another option is to [define a custom target with AtomicU64 disabled](https://github.com/tokio-rs/tokio/issues/5187#issuecomment-1323280893), as in the subsequent no-atomic-u64 test.)

carllerche: Thanks. I moved away from the cross test to cover this.

taiki-e (approved):

PR #5265, opened 2022-12-05
Impl `std::io::BufRead` on `SyncIoBridge<T>` if `T: AsyncBufRead + Unpin`

NobodyXu:

## Motivation

Since `SyncIoBridge<T>` implements `std::io::Read` if `T: AsyncRead + Unpin`, it will be nature to assume that it will also implement `std::io::BufRead` if `T: AsyncBufRead + Unpin`.


Darksonn (approved):
Thanks.

PR #1538, opened 2019-09-04
timer: use our own AtomicU64 on targets with target_has_atomic less than 64

taiki-e:
## Motivation

Fixes https://github.com/tokio-rs/tokio/pull/1421#issuecomment-527911780

## Solution

Reverts part of #1421.

PR #5282, opened 2022-12-09
rt: implement task::Id using `StaticAtomicU64`

carllerche:
This patch simplifies the implementation of `task::Id` by moving conditional compilation into the `AtomicU64` definition. To handle platforms that do not include `const fn Mutex::new()`, `StaticAtomicU64` is defined, which always has a `const fn new()`. `StaticAtomicU64` is implemented with `OnceCell` when needed.

Future patches will update other code locations to use `StaticAtomicU64` (e.g. `ThreadId`).


tokio/src/loom/std/atomic_u64_as_mutex.rs
```diff
@@ -1,18 +1,24 @@
 use crate::loom::sync::Mutex;
 use std::sync::atomic::Ordering;
 
+cfg_has_const_mutex_new! {
+    #[path = "atomic_u64_static_const_new.rs"]
```
carllerche, line 5: I'm experimenting with a way to have less code **within** the `cfg` macros. This should help `rustfmt` apply to more code.

tokio/src/loom/std/atomic_u64_static_once_cell.rs
```diff
@@ -0,0 +1,36 @@
+use super::AtomicU64;
+use crate::loom::sync::{Mutex, atomic::Ordering};
+use crate::util::once_cell::OnceCell;
+
+pub(crate) struct StaticAtomicU64 {
+    init: u64,
```
carllerche, line 6: It is a bit annoying to store the initial value forever, but this is *only* for platforms that do not have `AtomicU64` **and** no `const fn Mutex::new()` (i.e. old versions of Rust), so it isn't a big deal IMO.

carllerche:
Seems like I have to figure out how to make it work with loom...

tokio/src/loom/mocked.rs
```diff
@@ -25,6 +25,13 @@ pub(crate) mod sync {
         }
     }
     pub(crate) use loom::sync::*;
+
+    pub(crate) mod atomic {
+        pub(crate) use loom::sync::atomic::*;
+
+        // TODO: implement a loom version
+        pub(crate) type StaticAtomicU64 = std::sync::atomic::AtomicU64;
```
carllerche, line 33: This is a bit (ok, a lot)  of a hack, but it is equivalent to before the patch. Unfortunately, right now, loom does not have a way to model static variables.

tokio/src/util/once_cell.rs
```diff
@@ -25,7 +25,7 @@ impl<T> OnceCell<T> {
     /// If the `init` closure panics, then the `OnceCell` is poisoned and all
     /// future calls to `get` will panic.
     #[inline]
-    pub(crate) fn get(&self, init: fn() -> T) -> &T {
+    pub(crate) fn get(&self, init: impl FnOnce() -> T) -> &T {
```
FrankReh, line 28: typo on line 22: `intiailizing' -> initializing

Darksonn (approved):

notgull (commented):
This may be a bad idea, but wouldn't it be reasonable to use something like [`atomic-polyfill`](https://crates.io/crates/atomic-polyfill) or [`portable-atomic`](https://crates.io/crates/portable-atomic) to polyfill out atomics on systems that don't have them? I know this probably doesn't apply to this PR, but I think that a sequence-lock system may be a more ergonomic way of accessing this value.

taiki-e:
@notgull
tokio does not support no-std targets and the requirement for std is to have atomic types. This is an issue regarding having 64-bit atomic, and `atomic-polyfill` does not improve anything here.

`portable-atomic` provides fallback on platforms where 64-bit atomic is not available, it can more accurately detect platforms where 64-bit atomic is available than tokio[^1], and the seqlock used by `portable-atomic` in its fallback is probably more efficient for such uses.

That said, given the differences between global locking and per-object locking and tokio's policy on dependencies, it is doubtful that it is worth adding a non-optional dependency to tokio for this purpose.

[^1]: The logic of tokio in this area is also based on [my implementation](https://github.com/tokio-rs/tokio/pull/1538), and AFAIK portable-atomic's implementation is currently a complete form for this area. (although it is complex).

notgull:
@taiki-e Thanks for explaining this to me. Would you accept a PR where the current Mutex implementation is replaced with a hand-rolled seq-lock implementation?

taiki-e:
There are not many targets where 64-bit atomic is essentially unavailable (although this fallback is used by many targets because the current detection logic is a bit rough)[^1], and I'm not sure if the complexity is worth it.[^2] Proper seqlock implementation on a 32-bit system is [somewhat complex](https://github.com/crossbeam-rs/crossbeam/pull/359) and sound seqlock implementation requires logic like [per-byte atomic memcpy](https://github.com/taiki-e/atomic-memcpy) that loom is unclear whether it understands.

What I would like to do here is to improve the detection logic.
When using `cfg(target_has_atomic = "64")` on Rust 1.60+ like `portable-atomic` does, it should be able to accurately detect targets without 64-bit atomic, greatly reducing the number of platforms that require fallback implementation. Specifically, this could lead to visible performance improvements in armv6 and armv7.

[^1]: Also, recent linux has helpers for 64-bit CAS (__kuser_cmpxchg64), so future increases in Rust's kernel requirements will further reduce the number of targets requiring fallback for 64-bit atomic.
[^2]: `portable-atomic` also supports 128-bit atomic, so the situation is somewhat different.

PR #5316, opened 2022-12-27
ci: remove uses of unmaintained actions-rs actions

taiki-e:
## Motivation

actions-rs actions are not maintained and will not be available in the future due to the use of node12.
https://github.com/actions-rs/toolchain/issues/216
https://old.reddit.com/r/rust/comments/vyx4oj/actionsrs_organization_became_unmaintained/ig54zv7/

## Solution

- Use [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) instead of actions-rs/toolchain
- Use cargo/cross directly instead of actions-rs/cargo
- Use [rustsec/audit-check](https://github.com/rustsec/audit-check) (fork of actions-rs/audit-check) instead of actions-rs/audit-check

Also use [taiki-e/install-action](https://github.com/taiki-e/install-action) for installation of cross to avoid https://github.com/cross-rs/cross/issues/1177 and reduce CI time. (we already use this action for installations such as cargo-hack, wasmtime, valgrind)

Closes #5315

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -15,15 +15,16 @@ permissions:
 jobs:
   security-audit:
     permissions:
-      checks: write  # for actions-rs/audit-check to create check
+      checks: write  # for rustsec/audit-check to create check
       contents: read  # for actions/checkout to fetch code
-      issues: write  # for actions-rs/audit-check to create issues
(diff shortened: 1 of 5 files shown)


Darksonn (approved):
Thanks!

PR #5327, opened 2022-12-30
rt: move task::Id into its own file

carllerche:
This is a minor internal cleanup.


tokio/src/runtime/task/mod.rs
```diff
@@ -191,10 +192,8 @@ mod abort;
 mod join;
 
 #[cfg(feature = "rt")]
-#[allow(unreachable_pub)] // https://github.com/rust-lang/rust/issues/57411
```
carllerche, line 194: I'm also curious if this is still needed.

Darksonn (approved):

Noah-Kennedy (approved):

PR #5335, opened 2023-01-03
net: fix named pipes server configuration builder

carllerche:
The `pipe_mode` function would erase any previously set configuration option specified using the pipe_mode fit field. This patch fixes the builder to maintain the bit field when changing the pipe mode.


carllerche:
Going to redo this.

Issue #5340, opened 2023-01-03
Adding an optional `ignore_hanging_threads` flag to `#[tokio::main]`


Finomnis:
# The problem
The current behavior of `#[tokio::main]` is to wait for spawned blocking tasks.

There are good reason for this behavior:
- It matches the default behavior of dropping a `Runtime` object
- It guarantees that after a `#[tokio::main]` decorated function returns, it has dropped all resources

There are also good reasons against this behavior, especially if the `#[tokio::main]` function is used as the actual `main()` function of the program:
- It makes it impossible to schedule a reliable shutdown due to a very unpredictable waiting time after the `main` function
  - This is not an abstract problem, it can happen for example when [`tokio::net::lookup_host()` gets cancelled during a shutdown process](https://github.com/Finomnis/tokio-graceful-shutdown/issues/50).
- It is inconsistend with `std::thread::spawn` and `tokio::spawn`, both of which get killed/cancelled at the end of `main()`

This means that in its current state, `#[tokio::main]` is not usable if a predictable shutdown has to be performed at any point of the program, which I would argue is a somewhat regular usecase.

# The solution I'd like
The solution I'd like most is if `#[tokio::main]` would perform a [`shutdown_background()`](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.shutdown_background) after the `main()` function is finished, as I'd argue that this is the most commonly desired and most intuitive behavior, although I understand that this is not possible as it would be a breaking API behavior change.

So I'd argue that a new flag for `tokio::main` should be introduced to make this an opt-in behavior.

# Alteratives I've considered
To perform a reliable shutdown, only two possible solutions came to my mind:
- Tell people that `#[tokio::main]` isn't compatible with reliable shutdowns
- Introduce a thread that kills main after a grace period, like this:
```rust
use std::{thread, time::Duration};

#[tokio::main]
async fn main() {
    tokio::task::spawn_blocking(|| {
        thread::sleep(Duration::from_secs(10));
    });

    thread::spawn(|| {
        thread::sleep(Duration::from_millis(1000));
        println!("Shutdown seems to hang. Killing program ...");
        std::process::exit(1);
    });
}
```

Both of which are either non-ideal or bad workarounds.

# Additional thoughts

[...]

PR #5360, opened 2023-01-06
rt: make drop timeout configurable

Finomnis:
## Motivation

The current behaviour of the `Runtime` object is to wait for all `spawn_blocking` tasks before `Runtime::drop()` returns.

There are legitimate situations where this behavior is undesired, such as at the end of the program where it might be an active decision of the programmer not to wait for stale threads to increase shutdown reliability, but it is hard-coded in the `Runtime`'s `drop` implementation and cannot be changed.

There **is** the possibility of calling `Runtime::shutdown_background` to avoid this behavior, but this isn't always reliable, for example when panicking:

```rust
use std::{thread, time::Duration};
use tokio::{sync::oneshot, task};

fn main() {
    let runtime = tokio::runtime::Builder::new_current_thread()
        .enable_all()
        .build()
        .unwrap();

    runtime.block_on(async move {
        task::spawn_blocking(move || {
            tx.send(()).unwrap();
            thread::sleep(Duration::from_secs(10));
        });
        rx.await.unwrap();

        panic!();
    });

    // This is never reached because `panic` will unwind
    // out of `block_on` and drop `runtime` directly
    runtime.shutdown_background();
}
```

## Solution

Introducing the builder option `Builder::drop_timeout(Duration)` which can set the timeout used in the `Runtime::drop()` function:

```rust
use std::thread;
use std::time::{Duration, Instant};
use tokio::{sync::oneshot, task};

fn main() {
    let runtime = tokio::runtime::Builder::new_current_thread()
        .enable_all()
        .drop_timeout(Duration::from_nanos(0))
        .build()
        .unwrap();

    runtime.block_on(async move {
        task::spawn_blocking(move || {
            tx.send(()).unwrap();
            thread::sleep(Duration::from_secs(10));
        });
        rx.await.unwrap();

        panic!();
    });

    let now = Instant::now();
    drop(runtime);
    assert!(now.elapsed().as_secs() < 1);
}
```

Refs: #5340

### DIFF
--- a/tokio/src/runtime/blocking/pool.rs
+++ b/tokio/src/runtime/blocking/pool.rs
@@ -17,6 +17,7 @@ use std::time::Duration;
 pub(crate) struct BlockingPool {
     spawner: Spawner,
     shutdown_rx: shutdown::Receiver,
+    drop_timeout: Option<Duration>,
 }
 
 #[derive(Clone)]
@@ -232,6 +233,7 @@ impl BlockingPool {
                 }),
             },
             shutdown_rx,
+            drop_timeout: builder.drop_timeout.clone(),
         }
     }
 
@@ -275,7 +277,7 @@ impl BlockingPool {
 
 impl Drop for BlockingPool {
     fn drop(&mut self) {
-        self.shutdown(None);
+        self.shutdown(self.drop_timeout);
     }
 }
 
--- a/tokio/src/runtime/blocking/pool.rs
+++ b/tokio/src/runtime/blocking/pool.rs
@@ -233,7 +233,7 @@ impl BlockingPool {
                 }),
             },
             shutdown_rx,
-            drop_timeout: builder.drop_timeout.clone(),
+            drop_timeout: builder.drop_timeout,
         }
     }
 
--- a/tokio/src/runtime/builder.rs
+++ b/tokio/src/runtime/builder.rs
@@ -60,6 +60,9 @@ pub struct Builder {
     /// Cap on thread usage.
     max_blocking_threads: usize,
 
+    /// Timeout for joining hanging threads on drop
+    pub(super) drop_timeout: Option<Duration>,
+
     /// Name fn used for threads spawned by the runtime.
     pub(super) thread_name: ThreadNameFn,
 
@@ -243,6 +246,8 @@ impl Builder {
             worker_threads: None,
 
             max_blocking_threads: 512,
+
+            drop_timeout: None,
 
             // Default thread name
             thread_name: std::sync::Arc::new(|| "tokio-runtime-worker".into()),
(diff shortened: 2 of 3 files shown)


carllerche:
Thanks for the PR. I will comment here w/ my thoughts from Discord.

Before adding a configuration change, it is worth evaluating if the current behavior should be changed. We may want to *not* wait for the runtime to shut down during a panic. Wait on drop is intended to ensure a clean shutdown, but a panic means all bets are off.

If we add a configuration change, how do we explain to users when to use it and how to decide what to set it to? Also, what should the default be?

Finomnis:
@carllerche 
> Before adding a configuration change, it is worth evaluating if the current behavior should be changed. We may want to _not_ wait for the runtime to shut down during a panic. Wait on drop is intended to ensure a clean shutdown, but a panic means all bets are off.

While I do talk about panic here specifically, this is also about the case where the user orchestrates a shutdown via `CancellationToken` and wants to guarantee that after a timeout the program actually stops. Yes, `Runtime::shutdown_background()` exists, but is as reliable as a manual Mutex release, meaning, while it works most of the time, you don't get the hard guarantee as you would from a guard object (if the analogy makes sense). 

> If we add a configuration change, how do we explain to users when to use it and how to decide what to set it to? Also, what should the default be?

That one is kind of hard. Both sides have legitimate reasons for being the default. I'll only look at the "one global runtime which wraps the entire program" case because that's what most people use Tokio for.

Waiting has the guarantee that atomic operations like file writes finish in a proper state and don't get killed half way in a system call, which could cause file corruption. The only thing that gets canceled that way is futures that are in their waiting state, which is a lot safer. So I'd argue that this is a good default.

On the other hand, exiting immediately and killing background threads in the process also has its points. It mimics the behavior of std threads. Further, if something blocks forever, the user has to manually Ctrl-C the process to get out, and if Tokio catches those, it's even required to sigkill the entire process, which is probably not a sane default. Further, given that this behavior differs from std threads, people might think they made shutdown deterministic, just to be confused by a wait time they don't understand and need quite sophisticated Tokio knowledge for to debug.

Finomnis:
@carllerche Just for reference: the original issue that brought me to discover this problem was that `lookup_host()` has the potential to block the shutdown in the case of networking problems, which is unexpected for most users. Just to demonstrate that there are more use cases than just correct panicking behavior. Although I suspect that this should rather be a bug report for `lookup_host()` instead.

Finomnis:
@carllerche In case the previous answers were too long to read:

- Default: `None` (as otherwise it would be a breaking change)
- Explanation for users: see documentation in the code of this PR

carllerche:
Thanks for the additional context. I am still leaning against a configuration option at this point.

You argue that configurable behavior is desirable to have "guard type" safety even without panic. In that case, a configuration setting is not ideal, as there is no type-level guarantee that the configuration setting has been set correctly. Instead, I propose you define your type that wraps `Runtime` and implement drop to call `shutdown_background`.

There still is the separate question of whether or not the **current** behavior is correct in the face of panic. That discussion probably merits its issue.

Finomnis:
It was decided against this change.

PR #5394, opened 2023-01-23
clarify what happens to spawned work during runtime shutdown

e00E:
The current documentation around this is confusing because it does not
clearly differentiate the cases in which runtime shutdown blocks and
background work continues to run after runtime shutdown.

- Clear this up.
- Use more consistent wording.
- Remove a section refering to runtime internals because it is not
relevant for users, uses terminology that is not defined anywhere, and is
out of date.

closes https://github.com/tokio-rs/tokio/issues/4231


Darksonn (approved):
Thanks.

PR #5413, opened 2023-01-30
Clarify `process::Command` docs (fixing #5406)

mTsBucy1:
## Motivation

As mentioned in #5406, the documentation of `process::Command` does not respect the borrow checker.

## Solution

I've changed this, analogously to the standard library, to directly call `output().await.unwarp()`. This should now lead to usable examples.

The alternative would have been to create a `let mut command` binding and then calling `command.somefunction()`, which I found to be less readable.

Closes: #5406


Darksonn (approved):
Thanks.

PR #5434, opened 2023-02-06
rt: remove Arc from Clock

carllerche:
This patch removes `Arc` from Tokio's internal clock source. Instead of cloning `Clock` when needed, a reference is passed into functions that need to get the current instant.


Darksonn (approved):
LGTM.

hawkw (approved):
looks good when clippy passes!

carllerche:
@Darksonn @hawkw I had to change how this was implemented to get tests to pass. We need to maintain `#[track_caller]`, which means we can't panic in closures. I had the inner functions return `Result` and panic outside of the closure.

Darksonn (approved):

PR #5480, opened 2023-02-19
sync: fix docs for Send/Sync bounds in broadcast

satakuma:
`broadcast` channel does not require sent values to be `Sync`.


Darksonn (approved):
Thanks.

PR #5486, opened 2023-02-20
Use poll_fn in copy_bidirectional

xfix:
## Motivation

I didn't particularly like putting an artificial structure just to convert `Poll` into a `Future`. I think it's clearer with `poll_fn`.

## Solution

Use `poll_fn` rather than a manual `Future` implementation.


Darksonn (approved):
Thanks.

Issue #5490, opened 2023-02-21
Document the timing at which messages sent to a channel are Dropped


fulmicoton:
When sending message to a mpsc channel, messages seem dropped when the `Receiver` is dropped.

This seems like the most natural behavior, but this is not "goes without saying" situation.
Some alternative library like flume for instance, require all Senders to also be dropped for messages to be dropped.

It would be nice, if this behavior is confirmed, to add that information in the documentation.

hds:
@fulmicoton I've made some changes to the documentation in the attached PR. Do you think this would have been sufficient if you'd found it during your initial search?

fulmicoton:
@hds yes thank you!

PR #5497, opened 2023-02-22
sync: document drop behavior for channels

hds:
## Motivation

Some users mentioned that the behavior of a channel when the receivers
and/or senders are dropped isn't explicitly documented.

## Solution

This change adds wording to the documentation for each channel in the
sync module, explaining under which conditions messages in a channel are
dropped with respect to dropping the senders and the receivers.

Refs: #5490

### DIFF
--- a/tokio/src/sync/mpsc/mod.rs
+++ b/tokio/src/sync/mpsc/mod.rs
@@ -33,7 +33,8 @@
 //!
 //! If the [`Receiver`] handle is dropped, then messages can no longer
 //! be read out of the channel. In this case, all further attempts to send will
-//! result in an error.
+//! result in an error. Additionally, all unread messages will drained from the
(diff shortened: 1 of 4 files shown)


tokio/src/sync/mpsc/mod.rs
```diff
@@ -33,7 +33,8 @@
 //!
 //! If the [`Receiver`] handle is dropped, then messages can no longer
 //! be read out of the channel. In this case, all further attempts to send will
-//! result in an error.
+//! result in an error. Additionally, all unread messages will drained from the
```
satakuma, line 36:
```suggestion
//! result in an error. Additionally, all unread messages will be drained from the
```

satakuma (approved):
Looks good!

PR #5504, opened 2023-02-25
fix empty `join!` and `try_join!`

### LINKED ISSUE #5502: join! and try_join! deadlock with empty invocation
**Version**
Starting with 1.19.0 (i. e. everything including 1872a425e2b56530b46d44baf7bb45c1ee503807 / #4624).

**Platform**
any

**Description**
`join!()` and `try_join!()` emit an empty `loop {}`, see [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f93c69a9abfa85ff7b98e61c417d6254). I would expect this to either work (preferred) or be rejected at compile time (acceptable).

adrianheine:
This fixes #5502, i. e. `join!()` and `try_join!()` emitting `loop {}`. It does so in two ways:
1. The current macro implemention entry points are changed to not accept empty invocations
2. New entry points for empty invocations are added

### DIFF
--- a/tokio/src/macros/join.rs
+++ b/tokio/src/macros/join.rs
@@ -158,7 +158,9 @@ macro_rules! join {
 
     // ===== Entry point =====
 
-    ( $($e:expr),* $(,)?) => {
+    ( $($e:expr),+ $(,)?) => {
         $crate::join!(@{ () (0) } $($e,)*)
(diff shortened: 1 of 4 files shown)


Darksonn (approved):
Thanks.

PR #5529, opened 2023-03-04
Fix typo in doc: "with regards"

AaronChen0:
## Motivation

The noun `regard` means "respect" or "relation", so saying `with regard to` is like saying "with respect to" or "with relation to". The plural form of this noun, `regards`, means "good wishes" or "affection", which is significantly different from the meaning of the singular form.

## Solution
`with regards` => `with regard to`.


satakuma:
Thanks!

satakuma (approved):

PR #5557, opened 2023-03-18
Accept path as crate rename as well

daxpedda:
## Motivation

Currently Tokio supports renaming the crate inside the macro with the `crate` attribute. I was playing around with a personal library that simplifies some stuff around multiple macros, including the Tokio macro.
To implement this with macro hygiene in mind I was re-exporting Tokio, but turns out Tokio only accepts idents, not full paths.

## Solution

The solution was quite simple, just let Tokio accept paths, by using `syn::Path`, instead of only accepting an ident. There should be no functional change, as idents can also be parsed as paths.

### DIFF
--- a/tokio-macros/src/entry.rs
+++ b/tokio-macros/src/entry.rs
@@ -1,7 +1,7 @@
 use proc_macro::TokenStream;
-use proc_macro2::{Ident, Span};
+use proc_macro2::Span;
 use quote::{quote, quote_spanned, ToTokens};
-use syn::parse::Parser;
+use syn::{parse::Parser, Ident, Path};
 
 // syn::AttributeArgs does not implement syn::Parse
 type AttributeArgs = syn::punctuated::Punctuated<syn::NestedMeta, syn::Token![,]>;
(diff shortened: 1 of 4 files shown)


tokio-macros/src/entry.rs
```diff
@@ -354,16 +353,17 @@ fn parse_knobs(mut input: syn::ItemFn, is_test: bool, config: FinalConfig) -> To
         (start, end)
     };
 
-    let crate_name = config.crate_name.as_deref().unwrap_or("tokio");
-
-    let crate_ident = Ident::new(crate_name, last_stmt_start_span);
+    let crate_path = config
+        .crate_name
+        .map(ToTokens::into_token_stream)
+        .unwrap_or_else(|| Ident::new("tokio", last_stmt_start_span).into_token_stream());
```
daxpedda, line 359: It took me some time to figure out that if you just use `quote` here without a proper span you will trigger all sorts of lints. Not using a span was bad to being with anyway. Happy to build a custom `Path` here instead if desired. But I could also just use `quote_spanned`.

Darksonn (approved):
LGTM. Thanks.

PR #5577, opened 2023-03-23
stream: add StreamExt::timeout_repeating

jrray:
## Motivation

The behavior of `StreamExt::timeout` is it only fires the timeout once (per value produced from the wrapped stream) and while this behavior is documented it is easy to overlook and assume that timeouts are generated repeatedly. I had made use of `timeout` in a project assuming that timeouts did repeat and it led to a subtle bug. It is useful for my use case for timeouts to fire repeatedly, so this change adds that in `timeout_repeating` as an alternative to `timeout`. For example, one might want to log a message at a fixed interval while no messages are being produced from a stream.

## Solution

The non-repeating timeout behavior of `timeout` was not tested or demonstrated by the doc test, so first this was added.

A new `enum TimeoutMode` has been added to describe the two possible behaviors. `stream_ext::Timeout` was modified with a new `timeout_mode: TimeoutMode` field to share an implementation between the two behaviors.

### DIFF
--- a/tokio-stream/src/stream_ext.rs
+++ b/tokio-stream/src/stream_ext.rs
@@ -60,7 +60,7 @@ cfg_time! {
     pub(crate) mod timeout_repeating;
     use timeout::Timeout;
     use timeout_repeating::TimeoutRepeating;
-    use tokio::time::Duration;
+    use tokio::time::{Duration, Interval};
     mod throttle;
     use throttle::{throttle, Throttle};
     mod chunks_timeout;
@@ -997,15 +997,16 @@ pub trait StreamExt: Stream {
 
     /// Applies a per-item timeout to the passed stream.
     ///
-    /// `timeout_repeating()` takes a `Duration` that represents the maximum amount of
-    /// time each element of the stream has to complete before timing out.
+    /// `timeout_repeating()` takes an [`Interval`](tokio::time::Interval) that
+    /// controls the time each element of the stream has to complete before
+    /// timing out.
     ///
     /// If the wrapped stream yields a value before the deadline is reached, the
     /// value is returned. Otherwise, an error is returned. The caller may decide
     /// to continue consuming the stream and will eventually get the next source
     /// stream value once it becomes available. Unlike `timeout()`, if no value
     /// becomes available before the deadline is reached, additional errors are
-    /// returned at the specified interval.  See [`timeout`](StreamExt::timeout)
+    /// returned at the specified interval. See [`timeout`](StreamExt::timeout)
     /// for an alternative where the timeouts do not repeat.
     ///
     /// # Notes
@@ -1027,7 +1028,7 @@ pub trait StreamExt: Stream {
     /// use std::time::Duration;
     /// # let int_stream = stream::iter(1..=3);
     ///
-    /// let int_stream = int_stream.timeout_repeating(Duration::from_secs(1));
+    /// let int_stream = int_stream.timeout_repeating(tokio::time::interval(Duration::from_secs(1)));
     /// tokio::pin!(int_stream);
     ///
     /// // When no items time out, we get the 3 elements in succession:
@@ -1055,7 +1056,7 @@ pub trait StreamExt: Stream {
     /// // Timeout errors will be continuously produced at the specified
     /// // interval until the wrapped stream yields a value.
     /// let interval_stream = IntervalStream::new(tokio::time::interval(Duration::from_millis(23)));
-    /// let timeout_stream = interval_stream.timeout_repeating(Duration::from_millis(9));
+    /// let timeout_stream = interval_stream.timeout_repeating(tokio::time::interval(Duration::from_millis(9)));
     /// tokio::pin!(timeout_stream);
     ///
     /// // Multiple timeouts will be received between values in the source stream.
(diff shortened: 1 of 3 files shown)


tokio-stream/src/stream_ext.rs
```diff
@@ -977,7 +990,86 @@ pub trait StreamExt: Stream {
...
+    /// // interval until the wrapped stream yields a value.
+    /// let interval_stream = IntervalStream::new(tokio::time::interval(Duration::from_millis(23)));
+    /// let timeout_stream = interval_stream.timeout_repeating(Duration::from_millis(9));
+    /// tokio::pin!(timeout_stream);
+    ///
+    /// // Multiple timeouts will be received between values in the source stream.
+    /// assert!(timeout_stream.try_next().await.is_ok());
+    /// assert!(timeout_stream.try_next().await.is_err());
+    /// assert!(timeout_stream.try_next().await.is_err());
+    /// assert!(timeout_stream.try_next().await.is_ok());
```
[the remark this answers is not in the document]
Darksonn: You may be able to use Tokio's ability to mock the timer to make the test consistent.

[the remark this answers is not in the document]
jrray: Thank you, I've updated the tests to use that approach and removed the loop.

tokio-stream/src/stream_ext.rs
```diff
@@ -977,7 +990,87 @@ pub trait StreamExt: Stream {
...
+    /// # }
+    /// ```
+    #[cfg(all(feature = "time"))]
+    #[cfg_attr(docsrs, doc(cfg(feature = "time")))]
+    fn timeout_repeating(self, duration: Duration) -> Timeout<Self>
+    where
+        Self: Sized,
+    {
+        Timeout::new(self, duration, TimeoutMode::Repeating)
     }
```
Darksonn, line 1074: Please make two separate structs instead of adding a mode.

Darksonn (commented):
This PR fell between the cracks. Sorry about the delay in review.

tokio-stream/src/stream_ext.rs
jrray: Thanks for having a look! I removed the mode and timeout.rs is now mostly unchanged. I took the liberty of reusing the `Elapsed` struct for timeouts.

tokio-stream/src/stream_ext/timeout_repeating.rs
```diff
@@ -0,0 +1,66 @@
...
+                }
+                return Poll::Ready(v.map(Ok));
+            }
+            Poll::Pending => {}
+        };
+
+        ready!(me.deadline.as_mut().poll(cx));
+        let next = Instant::now() + *me.duration;
+        me.deadline.reset(next);
+        Poll::Ready(Some(Err(Elapsed::new())))
```
Darksonn, line 57: This will not yield the errors at a regular rate. The duration between two errors will be increased by the time between the timeout triggering and `poll_next` being called. Perhaps we should use an `Interval` instead?

jrray: I changed it to use an `Interval`. The infrequent polling problem has another edge case: if set to duration `x`, and the underlying stream doesn't produce a value until `x + t1`, but the stream is not polled until `x + t2`, then when polled would it be expected to produce a timeout or the value from the underlying stream?

Darksonn: It probably makes sense to add a [`MissedTickBehavior`][1] argument.

> The infrequent polling problem has another edge case: if set to duration `x`, and the underlying stream doesn't produce a value until `x + t1`, but the stream is not polled until `x + t2`, then when polled would it be expected to produce a timeout or the value from the underlying stream?

I'm not sure. What do you think? Maybe it should depend on the `MissedTickBehavior`.

[1]: https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html

jrray: For my use case, it isn't sensitive to these kinds of timing or ordering issues, but of course I would like this to be as generally useful as possible.

Returning a timeout when there is an actual value waiting to be returned feels a little pedantic but would arguably be the most correct behavior if not the most useful. I lean towards favoring the actual value, or making this a configurable behavior.

tokio-stream/src/stream_ext.rs
```diff
@@ -969,6 +973,17 @@ pub trait StreamExt: Stream {
...
+    /// // Once a timeout error is received, no further events will be received
+    /// // unless the wrapped stream yields a value (timeouts do not repeat).
+    /// let interval_stream = IntervalStream::new(tokio::time::interval(Duration::from_millis(100)));
+    /// let timeout_stream = interval_stream.timeout(Duration::from_millis(10));
+    /// tokio::pin!(timeout_stream);
+    ///
+    /// // Only one timeout will be received between values in the source stream.
+    /// assert!(timeout_stream.try_next().await.is_ok());
+    /// assert!(timeout_stream.try_next().await.is_err(), "expected one timeout");
+    /// assert!(timeout_stream.try_next().await.is_ok(), "expected no more timeouts");
```
Darksonn, line 986: I would make this a separate example.

Darksonn (commented):
You will need to merge in master to fix the CI errors.

tokio-stream/src/stream_ext.rs
jrray: I put the new code into a new section, and now the original code in `timeout` is unmodified. I think this is what you were going for, let me know if I misunderstood. Thanks!

Darksonn (approved):
Thanks, looks good to me.

PR #5589, opened 2023-03-30
Fix typo in private poll_proceed() doc comment

teor2345:
## Motivation

It looks like there's a spurious `CURRENT.with` autocompletion in the middle of the poll_proceed() doc comment.

## Solution

Remove the typo.


Darksonn (approved):
Thanks.

PR #5590, opened 2023-03-31
unsafe impl<'a> BufMut for ReadBuf<'a>

amab8901:
Solves https://github.com/tokio-rs/tokio/issues/5588

## Motivation

Explained in https://github.com/tokio-rs/tokio/issues/5588

## Solution

`unsafe impl<'a> BufMut for ReadBuf<'a>`


tokio/src/io/read_buf.rs
```diff
@@ -270,6 +271,20 @@ impl<'a> ReadBuf<'a> {
...
+    fn remaining_mut(&self) -> usize {
+        self.remaining()
+    }
+
+    unsafe fn advance_mut(&mut self, cnt: usize) {
+        self.advance(cnt);
+    }
+
+    fn chunk_mut(&mut self) -> &mut UninitSlice {
+        unsafe { UninitSlice::from_raw_parts_mut(10 as *mut u8, 10) }
```
paolobarbolini, line 287: It looks like you're missing a piece here. `10 as *mut u8` :eyes:

tokio/src/io/read_buf.rs
```diff
@@ -284,7 +284,10 @@ unsafe impl<'a> BufMut for ReadBuf<'a> {
     }
 
     fn chunk_mut(&mut self) -> &mut UninitSlice {
-        unsafe { UninitSlice::from_raw_parts_mut(10 as *mut u8, 10) }
+        let ptr = 10 as *mut u8; // any ideas?
```
paolobarbolini, line 287: I guess [`unfilled_mut()`](https://docs.rs/tokio/latest/tokio/io/struct.ReadBuf.html#method.unfilled_mut)?
Also because of something I read somewhere while spying on PRs like this I think it might be necessary to read the `len` before getting the pointer.

amab8901: image

tokio/src/io/read_buf.rs
```diff
@@ -270,6 +273,23 @@ impl<'a> ReadBuf<'a> {
...
+    fn remaining_mut(&self) -> usize {
+        self.remaining()
+    }
+
+    unsafe fn advance_mut(&mut self, cnt: usize) {
+        self.advance(cnt);
+    }
+
+    fn chunk_mut(&mut self) -> &mut UninitSlice {
+        let len = self.filled().len();
```
paolobarbolini, line 287: This should too be `unfilled_mut`

tokio/src/io/read_buf.rs
```diff
@@ -270,6 +273,25 @@ impl<'a> ReadBuf<'a> {
     }
 }
 
+#[cfg(feature = "bytes")]
```
Darksonn, line 276:
```suggestion
#[cfg(feature = "io-util")]
#[cfg_attr(docsrs, doc(cfg(feature = "io-util")))]
```

tokio/src/io/read_buf.rs
```diff
@@ -270,6 +273,25 @@ impl<'a> ReadBuf<'a> {
...
+unsafe impl<'a> BufMut for ReadBuf<'a> {
```
Darksonn, line 277: Instead of the cfg import at the top, just do like this.
```suggestion
unsafe impl<'a> bytes::BufMut for ReadBuf<'a> {
```

Darksonn (commented):
Please add a test for this. (Or use it in something that has a test.)

tokio/tests/io_read_buf.rs
```diff
@@ -34,3 +34,42 @@ async fn read_buf() {
     assert_eq!(n, 11);
     assert_eq!(buf[..], b"hello world"[..]);
 }
+
+#[tokio::test]
+#[cfg(feature = "io-util")]
+#[cfg_attr(docsrs, doc(cfg(feature = "io-util")))]
```
Darksonn, line 40: This annotation is used only for documentation, so it isn't used in tests.
```suggestion
```

tokio/src/io/read_buf.rs
```diff
@@ -270,6 +270,33 @@ impl<'a> ReadBuf<'a> {
...
+unsafe impl<'a> bytes::BufMut for ReadBuf<'a> {
+    fn remaining_mut(&self) -> usize {
+        self.remaining()
+    }
+
+    unsafe fn advance_mut(&mut self, cnt: usize) {
+        if self.initialized > 0 {
+            self.advance(cnt);
+        }
+    }
```
Darksonn, line 284:
```suggestion
    unsafe fn advance_mut(&mut self, cnt: usize) {
        // SAFETY: The caller of `advance_mut` guarantees that at least `cnt` bytes have been initialized.
        self.assume_init(cnt);
        self.advance(cnt);
    }
```

PR #5603, opened 2023-04-06
chore: bump `socket2` and `tempfile`

attila-lin:
make them use `windows-sys`


attila-lin:
Oops, It breaks the mini rustc version

PR #5617, opened 2023-04-12
Hide constant in select! macro for docs

mTsBucy1:
Using a nighly rustdoc the internal `BRANCHES` constant may leak into private documentation.
For background see https://github.com/serde-rs/serde/pull/2426.

Using the example from `select!` ([src](https://github.com/tokio-rs/tokio/blob/3b45e8614d45a44ebe6edf7b7939805f692f90ec/tokio/src/macros/select.rs#L151-L169)) and using `RUSTDOCFLAGS="--enable-index-page -Zunstable-options" cargo +nightly doc --no-deps --document-private-items` show the problem:
- Before the change:
image
- After the change:
image

Of course there may be other such instances, I haven't exhaustively searched for them.

### DIFF
--- a/tokio/src/macros/select.rs
+++ b/tokio/src/macros/select.rs
@@ -449,6 +449,7 @@ macro_rules! select {
         use $crate::macros::support::Pin;
         use $crate::macros::support::Poll::{Ready, Pending};
 
+        #[doc(hidden)]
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks!

PR #5621, opened 2023-04-13
Make entrypoints more efficient part 2 (closes #4513)

udoprog:
This is a different stab at #4513.

We use syn to parse the signature into our own `ItemFn`, maintaining most of the existing API but avoiding parsing (and reparsing) all statements except the parts we are interested in.

For signature parsing, we rely almost entirely on `syn`. The only notable trouble is the parsing of inner attributes, for which I believe I've matched `syn`'s behavior: The inner attributes are expanded as outer attributes instead. In theory we could expand them as inner attributes to the newly generated function but that doesn't seem to be necessary and I did it like this to match the previous expansion more precisely.

As a basic test, [I modified the tinyhttp example](https://github.com/udoprog/tokio/commit/8f621c862e991f9abcb367a287d1bd065c7aa638) and put everything in the main function, these were the timing differences I got for the macro:

Before:

```
[tokio-macros/src/entry.rs:466] std::time::Instant::now().duration_since(start) = 13.318901ms
```

After:

```
[tokio-macros/src/entry.rs:447] std::time::Instant::now().duration_since(start) = 1.075437ms
```

Note: We're still not down to microseconds like the other parsing method, but now at least we don't have to worry about signatures and [how to remove `syn` as a direct dependency](https://github.com/tokio-rs/tokio/pull/4513#issuecomment-1090872775) without msrv stuff breaking :smile:

As before, this is most notable for me when I was working on large entrypoints in rust-analyzer, for which large kinds of input latency caused diagnostics to be noticeably delayed. I've since refactored all of my entrpoints to quickly call another function to avoid this, but this should still be valuable for the future.

### DIFF
--- a/tokio-macros/src/entry.rs
+++ b/tokio-macros/src/entry.rs
@@ -414,7 +414,7 @@ fn parse_knobs(mut input: ItemFn, is_test: bool, config: FinalConfig) -> TokenSt
 }
 
 fn token_stream_with_error(mut tokens: TokenStream, error: syn::Error) -> TokenStream {
-    tokens.extend(TokenStream::from(error.into_compile_error()));
+    tokens.extend(error.into_compile_error());
     tokens
 }
 
@@ -521,7 +521,7 @@ impl ItemFn {
             block_expr.to_tokens(tokens);
         });
 
-        tokens.into()
+        tokens
     }
 }
 
--- a/tokio-macros/src/entry.rs
+++ b/tokio-macros/src/entry.rs
@@ -1,7 +1,7 @@
-use proc_macro::TokenStream;
-use proc_macro2::Span;
+use proc_macro2::{Span, TokenStream, TokenTree};
 use quote::{quote, quote_spanned, ToTokens};
-use syn::{parse::Parser, Ident, Path};
+use syn::parse::{Parse, ParseStream, Parser};
+use syn::{braced, Attribute, Ident, Path, Signature, Visibility};
 
 // syn::AttributeArgs does not implement syn::Parse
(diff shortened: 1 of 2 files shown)


tokio-macros/src/entry.rs
```diff
@@ -466,25 +447,145 @@ pub(crate) fn test(args: TokenStream, item: TokenStream, rt_multi_thread: bool)
...
+            stmts: &self.stmts,
+        }
+    }
+
+    /// Convert our local function item into a token stream.
+    fn into_tokens(
+        self,
+        header: proc_macro2::TokenStream,
+        body: proc_macro2::TokenStream,
+        block_expr: proc_macro2::TokenStream,
```
taiki-e, line 498: I think a name like `last_block` or `last_stmt` is easier to understand than block_expr.

taiki-e (commented):
LGTM. This is more robust and much easy to maintain (and understand) than #4513.

tokio-macros/src/entry.rs
udoprog: It's borrowed from what was already three. If we rename one we should [probably rename the other to maintain parity?](https://github.com/tokio-rs/tokio/pull/5621/files/61faf682ab51e68f1b8ed0731d11ecde89ce4eec#diff-3778ff38bebb2786ca2b45f4704af832e54cd75369b7b9215eae010cddc58a09L379)

taiki-e: I would prefer to replace all occurrences of "block_expr".

taiki-e (approved):
Thanks!

Darksonn (approved):

PR #5622, opened 2023-04-13
ci: Gate costly checks behind basic ones

udoprog:
A common source for iterating over a PR is to fail a basic test, this causes a lot of builds to fire off needlessly.

This defines a `basics` job which depends on jobs we want to pass before firing off more expensive ones, it has been added to other jobs as a dependency. In principle I think we could also add a fairly simple `cargo check --all-targets --workspace` for better feedback here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,19 @@ jobs:
     steps:
       - run: exit 0
 
+  # Basic actions that must pass before we kick of more expensive tests.
+  basics:
+    name: basic checks
+    runs-on: ubuntu-latest
+    needs:
+      - clippy
+      - fmt
+      - docs
+    steps:
+      - run: exit 0
+
   test:
+    needs: basics
     name: test tokio full
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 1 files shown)


udoprog:
[See it in action here](https://github.com/tokio-rs/tokio/actions/runs/4686299508)

image

.github/workflows/ci.yml
```diff
@@ -62,7 +62,19 @@ jobs:
     steps:
       - run: exit 0
 
+  # Basic actions that must pass before we kick of more expensive tests.
```
Darksonn, line 65:
```suggestion
  # Basic actions that must pass before we kick off more expensive tests.
```

.github/workflows/ci.yml
```diff
@@ -358,6 +380,7 @@ jobs:
 
   minrust:
     name: minrust
+    needs: basics
     runs-on: ubuntu-latest
```
Darksonn, line 384: I think it would make sense to include this one in the basic checks.

udoprog: For sure, it's a 30s action. Will add it.

Darksonn (approved):
Seems good to me.

Darksonn:
Thanks!

PR #5628, opened 2023-04-16
runtime: Add a metric that counts currently active tasks

matildasmeds:
Adds a new metric for the number of active tasks  for the Runtime. We count the elements on the OwnedTasks LinkedList, and access that through RuntimeMetrics.

I have added simple tests for the new metric, mimicking the style of existing tests. LinkedList does not have tests currently, so I did not add tests for LinkedList::count() either. Happy to add those tests if requested.

Runtime metrics come in two flavors, either prefixed with `num_` or postfixed with `_count`. The latter seems to be more common, so I have opted for that, but `num_active_tasks()` and even just `active_tasks()` as suggested in the original issue are certainly valid options as well.

Looking forward to any feedback! Thank you for taking a look. 😄 

Closes: #5400


matildasmeds:
Force pushed after rebasing with master, to resolve the FreeBSD checks and one merge conflict. Given no comments were yet given, it felt like a reasonable thing to do. 🙏

tokio/src/util/linked_list.rs
```diff
@@ -174,6 +174,21 @@ impl<L: Link> LinkedList<L, L::Target> {
...
+
+        while let Some(node) = current {
+            unsafe {
+                count += 1;
+                current = L::pointers(node).as_ref().get_next()
+            }
+        }
+
+        count
+    }
```
Darksonn, line 190: This makes the `active_task_count` call rather expensive compared to other metrics. It might be better to add a counter.

matildasmeds: It's true, also we are locking the list. I can give it a shot.

carllerche: Yes, I would add an additional task count atomic and inc / dec it as needed.

tokio/src/util/linked_list.rs
```diff
@@ -25,6 +26,9 @@ pub(crate) struct LinkedList<L, T> {
 
     /// Node type marker.
     _marker: PhantomData<*const L>,
+
+    /// Tracks the length of the linked list
+    _count: AtomicUsize,
```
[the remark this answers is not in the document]
matildasmeds: Would you rather see a counted variant of the LinkedList, and that as the type of owned  tasks? We could also do a benchmark to check how much overhead the counting adds if that is of interest.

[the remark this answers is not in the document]
Darksonn: A linked list wrapper with a counter would be fine. :+1:

tokio/src/util/linked_list.rs
```diff
@@ -243,6 +229,53 @@ impl<L: Link> fmt::Debug for LinkedList<L, L::Target> {
     }
 }
 
+// ===== impl CountedLinkedList ====
+
+// Delegates operations to the base LinkedList implementation, and adds a counter to the elements
+// in the list.
+pub(crate) struct CountedLinkedList<L: Link, T> {
+    _list: LinkedList<L, T>,
+    _count: AtomicUsize,
```
Darksonn, line 238: These fields are not unused, so their names should not start with `_`.
```suggestion
    list: LinkedList<L, T>,
    count: AtomicUsize,
```

Darksonn, line 238: You have a mutable reference any time you modify it, so you don't need an atomic.
```suggestion
    _list: LinkedList<L, T>,
    _count: usize,
```
If you want an atomic, then you should place it outside of the mutex. In that case, the logic would need to go in `OwnedTasks` rather than in the linked list.

tokio/src/runtime/task/list.rs
```diff
@@ -54,9 +54,13 @@ cfg_not_has_atomic_u64! {
 }
 
 pub(crate) struct OwnedTasks<S: 'static> {
-    inner: Mutex<OwnedTasksInner<S>>,
+    inner: Mutex<CountedOwnedTasksInner<S>>,
     id: u64,
 }
+struct CountedOwnedTasksInner<S: 'static> {
+    list: CountedLinkedList<Task<S>, <Task<S> as Link>::Target>,
```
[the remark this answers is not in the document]
matildasmeds: Although it is not necessary for the code to function correctly, I believe using different types for OwnedTasks and LocalOwnedTasks helps convey our intention more clearly. If we use CountedLinkedList for LocalOwnedTasks, the reader may think we could be counting those too. 

While it does add a few lines of code, I believe it is worth it to ensure clarity and readability. Looking forward to hearing your thoughts on this one!

tokio/src/util/linked_list.rs
Darksonn: You seem to have marked this as resolved, but it does not appear to have been addressed.

The point of using an atomic is to avoid locking the mutex when accessing the counter. However, because the atomic is inside the mutex, you are still locking the mutex when accessing the counter.

matildasmeds: My apologies marking this resolved without actually implementing the changes you had requested. It was not intentional, and I regret any inconvenience this may have caused. I will make sure to be more careful in the future to avoid similar mistakes. Thank you for your time and understanding.

Darksonn: Thanks, it looks better now.

There's no need to apologize for something like this. I was not offended or unhappy about it — I just wanted to clarify what I meant in case my review comment was unclear.

Darksonn:
You can fix the CI failures by merging master into your branch.

Darksonn (approved):
This looks good to me. Thank you for the PR. I can merge it once it passes the CI checks.

tokio/src/util/linked_list.rs
matildasmeds: Makes sense, all good! 😄

PR #5633, opened 2023-04-20
Expose futures for mpsc::Sender::reserve[_owned]()

coolreader18:
## Motivation
It's useful in a lot of cases to be able to store future types inside a manually implemented future, especially if the sub-futures are fairly primitive operations (like `reserve()`). e.g. with this change, it's possible to implement `tokio_util::sync::PollSender` fully statically, without any trait objects. (see [this branch](https://github.com/coolreader18/tokio/tree/static-pollsender) that I plan to open a PR from if this one merges.)

## Solution
Manually implement the futures.

### DIFF
--- a/tokio/src/sync/mpsc/bounded.rs
+++ b/tokio/src/sync/mpsc/bounded.rs
@@ -1,5 +1,5 @@
 use crate::loom::sync::Arc;
-use crate::sync::batch_semaphore::{self as semaphore, TryAcquireError};
+use crate::sync::batch_semaphore::{self as semaphore, Acquire, TryAcquireError};
 use crate::sync::mpsc::chan;
 use crate::sync::mpsc::error::{SendError, TryRecvError, TrySendError};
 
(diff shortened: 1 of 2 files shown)


tokio/src/sync/mpsc/bounded.rs
```diff
@@ -1103,6 +1130,99 @@ impl<T> fmt::Debug for Sender<T> {
...
+    /// Gets a reference to the [`Sender`] for the channel this future is reserving a permit in.
+    pub fn sender(&self) -> &Sender<T> {
+        self.tx
+            .as_ref()
+            .expect("ReserveOwned::sender called after completion")
+    }
+}
+
+impl<T> Future for ReserveOwned<T> {
+    type Output = Result<OwnedPermit<T>, SendError<()>>;
```
coolreader18, line 1190: Not that this can be changed without it being breaking, but I noticed this should probably return a `SendError<Sender<T>>`, like `try_reserve_owned()` does.

carllerche:
Thanks for the suggestion. 

I generally lean against naming futures like this specifically because it complicates our code. It also prevents us from using async await for that function in the future, which would be unfortunate given that Rust at some point will support naming anon types. 

You can either use boxed future trait objects or a separate crate that includes named futures.

PR #5636, opened 2023-04-20
docs: Fix typo in #[tokio::test] docs

adamchalmers:
Both current- and multi-thread runtime claimed to be the default, but only current thread actually is.


Darksonn (approved):
Thanks.

PR #5647, opened 2023-04-24
sync: useful Debug impl for RwLock 

FHTMitchell:
## Motivation

Currently the `std::fmt::Debug` impl for `tokio::sync::RwLock` is just what `#[derive(Debug)]` provides and looks like

```
RwLock { mr: 536870911, s: Semaphore { permits: 536870911 }, c: UnsafeCell { .. } }
```

which isn't very useful. Instead we can show the data *if* no thread currently holds the write lock. This matches what is done in `std::sync::RwLock` and in `tokio::sync::Mutex`.

## Solution

The updated code is basically a 1 to 1 mapping from the tokio [Mutex implementation](https://docs.rs/tokio/latest/src/tokio/sync/mutex.rs.html#787-799).


Darksonn (approved):
Looks good to me.

PR #5653, opened 2023-04-25
fs: reduce blocking ops in `fs::read_dir`

icedrocket:
This change prevents spawning blocking threads once the iteration is complete.


icedrocket:
The commit that was force pushed has some minor style changes, but the actual code remains unchanged.

PR #5682, opened 2023-05-10
Add ci step to check tokio-stream/fuzz compiles

hootio:
## Motivation

Solves the `/tokio-stream/` piece of #5601 

## Solution

Similar to the `/tokio/` check, adds step in CI to run `cargo fuzz check` on `/tokio-stream/` dir

Closes:  #5601


Darksonn (approved):
Thanks.

PR #5685, opened 2023-05-12
rt: instrument task poll times with a histogram

carllerche:
Adds support for instrumenting the poll times of all spawned tasks. Data is tracked in a histogram. The user must specify the histogram scale and bucket ranges. Implementation-wise, the same strategy is used in the runtime where we are just using atomic counters. Because instrumenting each poll duration will result in frequent calls to `Instant::now()`, I think it should be an opt-in metric.

There are still some open questions related to the builder API, but those can be punted as it is an unstable API. I will track the questions in the metrics stabilization tracking issue.


tokio/src/runtime/metrics/worker.rs
```diff
@@ -38,10 +40,19 @@ pub(crate) struct WorkerMetrics {
     /// Number of tasks currently in the local queue. Used only by the
     /// current-thread scheduler.
     pub(crate) queue_depth: AtomicUsize,
+
+    /// If `Some`, tracsk the poll times in `ns`
```
tobz, line 44:
```suggestion
    /// If `Some`, tracks the poll times in nanoseconds.

File: tokio/src/runtime/metrics/batch.rs [hunk 2]
```diff
@@ -32,11 +32,29 @@ pub(crate) struct MetricsBatch {
...
+
+    /// Instant at which work last resumed (continued after park).
     last_resume_time: Instant,
+
+    /// If `Some`, tracks poll times in `ns`
+    poll_timer: Option<PollTimer>,
+}
+
+struct PollTimer {
+    /// Histogram of mean time spent polling within each band.
```
tobz, line 44:
```suggestion
    /// Histogram of time spent polling within each band.
```

There's no mean here, since we're measuring the actual poll time.

tokio/src/runtime/metrics/batch.rs
```diff
@@ -32,11 +32,29 @@ pub(crate) struct MetricsBatch {
 
     /// The total busy duration in nanoseconds.
     busy_duration_total: u64,
+
+    /// Instant at which work last resumed (continued after park).
     last_resume_time: Instant,
+
+    /// If `Some`, tracks poll times in `ns`
```
tobz, line 39:
```suggestion
    /// If `Some`, tracks poll times in nanoseconds.
```

tokio/src/runtime/metrics/batch.rs
```diff
@@ -81,8 +113,23 @@ impl MetricsBatch {
...
+
+        if let Some(poll_timer) = &mut self.poll_timer {
+            poll_timer.poll_started_at = Instant::now();
+        }
+    }
+
+    pub(crate) fn end_poll(&mut self) {
+        if let Some(poll_timer) = &mut self.poll_timer {
+            let elapsed = duration_as_u64(poll_timer.poll_started_at.elapsed());
+            poll_timer.poll_times.measure(elapsed, elapsed);
```
tobz, line 130: This doesn't make sense.

You're adding `elapsed` samples to the bucket that `elapsed` falls within, when really you should just be tracking a single sample.

Continuing from that premise, there's seemingly also no reason to even have a `count` parameter on `measure` given that you're only ever tracking single occurrences.

tobz (dismissed):
Some doc comment nits, but a few major questions.

tokio/src/runtime/metrics/batch.rs
tobz: Sorry, to say it another way: a user _could_ calculate the mean poll times from this data, but the data as it exists is itself not the mean.

tokio/src/runtime/builder.rs
```diff
@@ -877,6 +881,29 @@ impl Builder {
         }
     }
 
+    cfg_metrics! {
+        pub fn metrics_poll_count_histogram(&mut self, histogram_scale: crate::runtime::HistogramScale) -> &mut Self {
```
hds, line 885: My preference here would be for a single function that takes either all necessary parameters or a builder or config struct.

I've had poor experiences in the past with public APIs where setting 1 of N values implicitly enables some functionality. If the poll count histogram is going to be opt-in, then I believe that a more self explanatory API would be a single function like `metrics_poll_count_histogram(&mtu self, config: PollCountHistogramConfig)`. You can then implement `Default` for the config struct to allow the user to modify only some values if that is desireable (which I think it is).

tokio/src/runtime/metrics/runtime.rs
```diff
@@ -578,6 +579,50 @@ impl RuntimeMetrics {
         self.handle.inner.worker_local_queue_depth(worker)
     }
 
+    pub fn poll_count_histogram_enabled(&self) -> bool {
```
hds, line 582: Following from my other comment, if we expose a (non-exhaustive) `PollCountHistogramConfig` struct, then these 4 methods could be replaced by one that returns `Option<PollCountHistogramConfig>`.

hds (commented):
Couple of comments on the public API. They reflect my personal preferences and experience, so take them as you wish. (-:

tokio/src/runtime/builder.rs
carllerche: Multiple arguments to a builder fn is not a forwards-compatible option, unfortunately. The same goes with a `Config` struct. If we make the struct `#[non_exaustive],` users cannot initialize it. That leaves nesting builders, which gets messy.

What issues have you had in the past?

hds: User confusion issues in both directions.
* How do I enable X function?
* I didn't realise that setting this value would enable X function.
* I didn't realise that *not* setting this value would disable X function.

Also, there's currently no way to "enable poll count histogram with default values", you always have to override one value to enable, even if you want to stick to the defaults.

My personal feeling is to go with nesting builders, because in the default case the user would pass `PollCountHistogramConfig::default()`, which I think isn't so bad.

hds: Of course, as you said on Discord, it's an unstable API, so trying it out as is isn't such an issue.

carllerche: Good points. To address those issues, I added an `enable_metrics_poll_count_histogram` builder fn. This is the only way to enable that metric. The other builder fns will not implicitly enable or disable the feature.

Noah-Kennedy (approved):

tokio/src/runtime/builder.rs
```diff
@@ -877,6 +887,127 @@ impl Builder {
         }
     }
 
+    cfg_metrics! {
+        /// Enables tracking the distribution of task poll times.
+        ///
+        /// Task poll times are not instrumented by default as doing so requires
+        /// calling [`Instant::now()]` twice per task poll, which could add
```
hds, line 894: Small typo:

```suggestion
        /// calling [`Instant::now()`] twice per task poll, which could add
```

tokio/src/runtime/builder.rs
```diff
@@ -877,6 +887,127 @@ impl Builder {
...
+            // Sanity check the argument and also make the cast below safe.
+            assert!(resolution <= Duration::from_secs(1));
+
+            let resolution = resolution.as_nanos() as u64;
+            self.metrics_poll_count_histogram.resolution = resolution;
+            self
+        }
+
+        /// Sets the number of buckets for the histogram tracking the
+        /// distribution of task poll times.
```
hds, line 990: I think it would be good to explain here what happens to poll times that fall outside resolution * buckets (or the log equivalent). Are those times included in the largest bucket? Stored separately? Dropped?

tokio/src/runtime/metrics/runtime.rs
```diff
@@ -578,6 +579,189 @@ impl RuntimeMetrics {
...
+    ///                     let count = metrics.poll_count_histogram_bucket_count(worker, i);
+    ///                     println!("Poll count {}", count);
+    ///                 }
+    ///             }
+    ///         });
+    /// }
+    /// ```
+    ///
+    /// [`poll_count_histogram_bucket_range()`]: crate::runtime::RuntimeMetrics::poll_count_histogram_bucket_range
+    pub fn poll_count_histogram_bucket_count(&self, worker: usize, bucket: usize) -> u64 {
```
hds, line 755: nit: Perhaps this could be annotated with `#[track_caller]`, even though we haven't annotated all the unstable functions as such, it would be nice to add it with new ones that can panic.

tokio/src/runtime/metrics/runtime.rs
```diff
@@ -578,6 +579,189 @@ impl RuntimeMetrics {
...
+    ///                 let range = metrics.poll_count_histogram_bucket_range(i);
+    ///                 println!("Histogram bucket {} range: {:?}", i, range);
+    ///             }
+    ///         });
+    /// }
+    /// ```
+    ///
+    /// [`metrics_poll_count_histogram_resolution()`]:
+    ///     crate::runtime::Builder::metrics_poll_count_histogram_resolution
+    pub fn poll_count_histogram_bucket_range(&self, bucket: usize) -> Range<Duration> {
```
hds, line 685: Will this method panic if `bucket` represents an invalid bucket?

tokio/src/runtime/builder.rs
carllerche: Applied.

tokio/src/runtime/builder.rs
carllerche: Done.

tokio/src/runtime/metrics/runtime.rs
carllerche: Done

tokio/src/runtime/metrics/runtime.rs
carllerche: Added a panics section to the docs.

hds (approved):
Looks good to me!

PR #5694, opened 2023-05-15
Link to latest version of tokio-util docs

a-robinson:
## Motivation

I was writing some code that used `mpsc::Sender`, and clicked on the link in the docs to `PollSender`, then wrote a bunch of code based on the docs, and only afterwards realized that I had been linked to the 0.6 tokio-util docs and its API is completely different in 0.7. Hopefully this change will save others from the same fate.

## Solution

I updated all remaining links to tokio-util docs that didn't already point to latest to now do so. I also updated a couple of links to tokio docs themselves that were in the same file, since it's hard to imagine we'd want to be linking to an old version of our own docs.

I'd guess the same type of change might be eventually useful for all the links to tokio-stream's 0.1 docs as well, but for now 0.1 is still the latest version of tokio-stream and I don't know enough about it to know if there's a reason not to change those links.

### DIFF
--- a/tokio-stream/src/lib.rs
+++ b/tokio-stream/src/lib.rs
@@ -63,12 +63,12 @@
 //! [`tokio-util`] provides the [`StreamReader`] and [`ReaderStream`]
 //! types when the io feature is enabled.
 //!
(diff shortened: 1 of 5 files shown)


Darksonn (approved):
Thanks.

PR #5700, opened 2023-05-18
rt: add internal counters to threaded runtime.

carllerche:
These counters are enabled using the `tokio_internal_mt_counters` and are intended to help with debugging performance issues.

Whenever I work on the threaded runtime, I often find myself adding these counters, then removing them before submitting a PR. I think keeping them in will save time in the long run and shouldn't impact dev much.

### DIFF
--- a/tokio/src/runtime/scheduler/multi_thread/worker.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/worker.rs
@@ -59,7 +59,7 @@
 use crate::loom::sync::{Arc, Mutex};
 use crate::runtime;
 use crate::runtime::context;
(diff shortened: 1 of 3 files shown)


tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -830,7 +839,16 @@ impl Handle {
         }
     }
 
-    fn notify_parked(&self) {
+    fn notify_parked_local(&self) {
```
carllerche, line 842: Another reason is split with `local` is the notification logic will change in the future between notifying locally and notifying from outside of the runtime.

Darksonn (approved):
Seems fine to me.

PR #5701, opened 2023-05-18
rt: small current_thread scheduler cleanup

carllerche:
There should be no functional changes.

### DIFF
--- a/tokio/src/runtime/scheduler/current_thread.rs
+++ b/tokio/src/runtime/scheduler/current_thread.rs
@@ -46,7 +46,7 @@ pub(crate) struct Handle {
 /// a function that will perform the scheduling work and acts as a capability token.
 struct Core {
     /// Scheduler run queue
-    tasks: VecDeque<task::Notified<Arc<Handle>>>,
+    tasks: VecDeque<Notified>,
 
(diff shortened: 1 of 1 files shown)


tokio/src/runtime/scheduler/current_thread.rs
```diff
@@ -250,7 +252,20 @@ impl fmt::Debug for CurrentThread {
 // ===== impl Core =====
 
 impl Core {
-    fn pop_task(&mut self, handle: &Handle) -> Option<task::Notified<Arc<Handle>>> {
```
carllerche, line 253: Renamed `pop_task` -> `next_task_*` to match names in the multi-threaded scheduler.

tokio/src/runtime/scheduler/current_thread.rs
```diff
@@ -95,6 +95,8 @@ struct Context {
     core: RefCell<Option<Box<Core>>>,
 }
 
+type Notified = task::Notified<Arc<Handle>>;
```
carllerche, line 98: Use an alias like the MT scheduler

Darksonn (approved):

PR #5702, opened 2023-05-18
rt: use task::Inject with current_thread scheduler

carllerche:
Depends on #5701

Previously, the current_thread scheduler used its own injection queue instead of sharing the same one as the multi-threaded scheduler. This patch updates the current_thread scheduler to use the same injection queue as the multi-threaded one (`task::Inject`).

`task::Inject` includes an optimization where it does not need to acquire the mutex if the queue is empty.

The primary objective for this PR is to unify the injection queue code paths for future tweaks. The benchmark results between the current `master` and this change are within the margin of error.

**Before:**

```
test spawn_many_local       ... bench:     222,721 ns/iter (+/- 5,288)
test spawn_many_remote_busy ... bench:   5,576,420 ns/iter (+/- 175,418)
test spawn_many_remote_idle ... bench:     228,768 ns/iter (+/- 1,763)
```

**After:**

```
test spawn_many_local       ... bench:     216,111 ns/iter (+/- 4,268)
test spawn_many_remote_busy ... bench:   5,329,780 ns/iter (+/- 143,615)
test spawn_many_remote_idle ... bench:     212,560 ns/iter (+/- 2,962)
````

### DIFF
--- a/tokio/src/runtime/scheduler/current_thread.rs
+++ b/tokio/src/runtime/scheduler/current_thread.rs
@@ -217,6 +217,9 @@ impl CurrentThread {
                 drop(task);
             }
 
+            // Close the injection queue
+            handle.shared.inject.close();
+
             // Drain remote queue
             while let Some(task) = handle.shared.inject.pop() {
                 drop(task);
--- a/tokio/src/runtime/scheduler/current_thread.rs
+++ b/tokio/src/runtime/scheduler/current_thread.rs
@@ -66,7 +66,7 @@ struct Core {
 
 /// Scheduler state shared between threads.
 struct Shared {
-    /// Remote run queue. None if the `Runtime` has been dropped.
+    /// Remote run queue
     inject: Inject<Arc<Handle>>,
 
     /// Collection of all active tasks spawned onto this executor.
--- a/tokio/src/runtime/scheduler/current_thread.rs
+++ b/tokio/src/runtime/scheduler/current_thread.rs
@@ -1,8 +1,8 @@
 use crate::future::poll_fn;
 use crate::loom::sync::atomic::AtomicBool;
-use crate::loom::sync::{Arc, Mutex};
+use crate::loom::sync::Arc;
 use crate::runtime::driver::{self, Driver};
-use crate::runtime::task::{self, JoinHandle, OwnedTasks, Schedule, Task};
+use crate::runtime::task::{self, Inject, JoinHandle, OwnedTasks, Schedule, Task};
 use crate::runtime::{blocking, context, scheduler, Config};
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Review by Darksonn — APPROVED** [review]

**Comment by hawkw** [comment]
It looks like the `close`, `push_batch`, and `push_batch_inner` methods on the injector queue need to be behind `cfg_rt_multi_thread!`, as they're unused when building without `features = "rt-multi-thread"`: https://github.com/tokio-rs/tokio/actions/runs/5026769568/jobs/9015455907?pr=5702#step:7:175

File: tokio/src/runtime/scheduler/current_thread.rs [hunk 1]
```diff
@@ -67,7 +67,7 @@ struct Core {
 /// Scheduler state shared between threads.
 struct Shared {
     /// Remote run queue. None if the `Runtime` has been dropped.
```
hawkw, line 69: I think the comment that this is `None` if the runtime has been dropped needs to be removed, as this is no longer an `Option`.

tokio/src/runtime/scheduler/current_thread.rs
```diff
@@ -549,14 +536,9 @@ impl Schedule for Arc<Handle> {
                 // Track that a task was scheduled from **outside** of the runtime.
                 self.shared.scheduler_metrics.inc_remote_schedule_count();
 
-                // If the queue is None, then the runtime has shut down. We
-                // don't need to do anything with the notification in that case.
```
hawkw, line 553: do we still need to do something to handle the case where the runtime is shutting down?

hawkw (approved):
Looks good to me overall! I noticed a comment that probably needs to be updated, and it looks like some of the methods on `Inject` need to be cfg-flagged, but other than that, this looks good to me!

tokio/src/runtime/scheduler/current_thread.rs
carllerche: Yes, not sure why CI still passes. I tried adding a new test, but it is a hard case to trigger I guess.

tokio/src/runtime/scheduler/current_thread.rs
carllerche: Done

hawkw:
Looks like [the WASM tests are failing](https://github.com/tokio-rs/tokio/actions/runs/5027141640/jobs/9016327685?pr=5702#step:9:681) because the new [`shutdown_concurrent_spawn`](https://github.com/tokio-rs/tokio/pull/5702/files#diff-863a5a8b3d134daa64a3f48e856431501050d5dbe3810cd09e8ca10ea96ac73dR1322) test is not ignored for WASM targets, and spawns a thread, which isn't supported in WASM...

hawkw:
Now Miri is sad...

PR #5715, opened 2023-05-23
rt(threaded): basic self-tuning of injection queue

carllerche:
Proof of concept showing how we could implement some basic self-tuning to check the interval queue (and other maintenance tasks) at variable interval rates.

This is a proof of concept. This shouldn't be the final implementation.

For the benchmark in the description of #5712, this change reduces the run time from 2s to 115ms (on my laptop).


carllerche:
Superseded by #5720

PR #5720, opened 2023-05-24
rt(threaded): basic self-tuning of injection queue

carllerche:
Each multi-threaded runtime worker prioritizes pulling tasks off of its local queue. Every so often, it checks the injection (global) queue for work submitted there. Previously, "every so often," was a constant "number of tasks polled" value. Tokio sets a default of 61, but allows users to configure this value.

If workers are under load with tasks that are slow to poll, the injection queue can be starved. To prevent starvation in this case, this commit implements some basic self-tuning. The multi-threaded scheduler tracks the mean task poll time using an exponentially-weighted moving average. It then uses this value to pick an interval at which to check the injection queue.

This commit is a first pass at adding self-tuning to the scheduler. There are other values in the scheduler that could benefit from self-tuning (e.g. the maintenance interval). Additionally, the current-thread scheduler could also benfit from self-tuning. However, we have reached the point where we should start investigating ways to unify logic in both schedulers. Adding self-tuning to the current-thread scheduler will be punted until after this unification.

With this change, I can now add the benchmark mentioned in #5712 and this brings it down from 2s -> 100ms about.


tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -506,29 +522,25 @@ impl Context {
                     }
                 };
 
-                // If task poll times is enabled, measure the poll time. Note
-                // that, if the `core` is stolen, this means `block_in_place`
-                // was called, turning the poll into a "blocking op". In this
-                // case, we don't want to measure the poll time as it doesn't
-                // really count as an async poll anymore.
-                core.metrics.end_poll();
```
carllerche, line 514: These changes might be a bit controversial, but when we poll the LIFO slot, we aren't "really" polling a new task, but we are batching polling the lifo task under the initially polled task.

carllerche:
I will work on fixing CI.

I will also investigate tests, but I think that might be hard because they are timing based.

Noah-Kennedy:
We should probably test this change on a few different workloads to see how it performs.

carllerche:
@Noah-Kennedy Go for it. Tokio's benchmarks (except the one this targets) remain unchanged (margin of error).

benches/rt_multi_threaded.rs
```diff
@@ -57,19 +59,60 @@ fn spawn_many_remote_idle(b: &mut Bencher) {
     });
 }
 
-fn spawn_many_remote_busy(b: &mut Bencher) {
+fn spawn_many_remote_busy1(b: &mut Bencher) {
```
hawkw, line 62: maybe worth a comment here describing the different cases these benchmarks are exercising?

tokio/src/runtime/builder.rs
```diff
@@ -82,7 +82,7 @@ pub struct Builder {
     pub(super) keep_alive: Option<Duration>,
 
     /// How many ticks before pulling a task from the global/remote queue?
```
hawkw, line 84: can this comment explain what it means for this value to be `None`, now that it's an `Option`?

tokio/src/runtime/scheduler/multi_thread/worker.rs
hawkw: i think this feels reasonable. can we add a comment explaining that rationale?

tokio/src/runtime/scheduler/multi_thread/stats.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    /// Exponentially-weighted moving average of time spent polling scheduled a
+    /// task.
+    ///
+    /// Tracked in nanoseconds, stored as a f64 since that is what we use with
+    /// the EWMA calculations
+    task_poll_time_ewma: f64,
+}
+
+// How to weigh each individual poll time, value is plucked from thin air.
```
hawkw, line 30:
```suggestion
/// How to weigh each individual poll time, value is plucked from thin air.
```

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -841,6 +858,19 @@ impl Core {
...
     }
+
+    fn tune_global_queue_interval(&mut self, worker: &Worker) {
+        let next = self
+            .stats
+            .tuned_global_queue_interval(&worker.handle.shared.config);
+
+        // Smooth out jitter
+        //
+        // `u32::abs_diff` is not available on Tokio's MSRV.
```
hawkw, line 869: nit, not a big deal: i might have put this comment on the abs_diff function definition, instead?

tokio/tests/rt_threaded.rs
```diff
@@ -588,6 +589,99 @@ async fn test_block_in_place4() {
     tokio::task::block_in_place(|| {});
 }
 
+#[test]
+#[cfg(not(tokio_cross))]
+fn test_tuning() {
```
hawkw, line 594: maybe i overlooked it, but...does this test ever actually assert anything? how do we know if the tuning is working correctly?

hawkw (approved):
the implementation looks really solid! i had a few minor nitpicks but nothing blocking.

tokio/tests/rt_threaded.rs
carllerche: If it doesn't work, it runs forever.

Darksonn (approved):

PR #5723, opened 2023-05-25
ci: speed up multi-threaded runtime loom tests.

carllerche:
Increase max preemption back to 2 while running the tests in under 90 minutes.


carllerche:
Well, loom passes in 5hrs :thinking: I will see if I can find other ways to get it to run faster.

Darksonn (approved):

PR #5727, opened 2023-05-26
rt: move scheduler ctxs to runtime::context

carllerche:
This commit eliminates the current_thread::CURRENT and multi_thread::CURRENT thread-local variable in favor of using `runtime::context`. This is another step towards reducing the total number of thread-local variables used by Tokio.


tokio/src/macros/mod.rs
```diff
@@ -23,9 +23,11 @@ cfg_trace! {
     mod trace;
 }
 
-#[macro_use]
-#[cfg(feature = "rt")]
-pub(crate) mod scoped_tls;
+cfg_rt_multi_thread! {
+    #[macro_use]
+    #[cfg(feature = "rt")]
```
hawkw, line 28: does this still need the feature flag on it now that it's behind `cfg_rt_multi_thread!`?
```suggestion
```

tokio/src/runtime/context/scoped.rs
```diff
@@ -0,0 +1,61 @@
+use std::cell::Cell;
+use std::marker::PhantomData;
+use std::ptr;
+
+/// Scoped thread-local storage
+pub(super) struct Scoped<T> {
+    pub(super) inner: Cell<*const ()>,
+    pub(crate) _p: PhantomData<T>,
```
hawkw, line 8: wait, why do we have a `Cell<*const ()>` _and_ a `PhantomData<T>`, and cast the pointer to `*const T` when it's actually dereferenced? can't this just be `Cell<*const T>`? Is this some kind of variance thing? If so, it would be nice to have a comment explaining what's up with this...

carllerche: Not sure; this was lifted from `scoped_tls`. I could see what happens.

hawkw: seems like it would be nicer that way, but 🤷‍♀️

carllerche: I switched it.

tokio/src/macros/mod.rs
carllerche: That block was removed entirely

hawkw (approved):

PR #5748, opened 2023-06-01
rt: move Inject to `runtime::scheduler`

carllerche:
Previously, `Inject` was defined in `runtime::task`. This was because it used some internal fns as part of the intrusive linked-list implementation.

In the future, we want to remove the mutex from Inject and move it to the scheduler proper (to reduce mutex ops). To set this up, this commit moves `Inject` to `runtime::scheduler`. To make this work, we have to `pub(crate)` `task::RawTask` and use that as the interface to access the next / previous pointers.


Noah-Kennedy (approved):

Darksonn (approved):

carllerche:
CI is spuriously failing. #5752 attempts to resolve the issue.

PR #5752, opened 2023-06-01
ci: fix spurious CI failure

carllerche:
PR #5720 introduced runtime self-tuning. It included a test that attempts to verify self-tuning logic. The test is heavily reliant on timing details. This patch tries to make the test more reliable by not assuming tuning will converge within a set amount of time.

### DIFF
--- a/tokio/tests/rt_threaded.rs
+++ b/tokio/tests/rt_threaded.rs
@@ -627,7 +627,6 @@ fn test_tuning() {
     }
 
     // Now, hammer the injection queue until the interval drops.
-    let mut i = 0;
     let mut n = 0;
     loop {
         let curr = interval.load(Relaxed);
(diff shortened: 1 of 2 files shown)


tokio/tests/rt_threaded.rs
```diff
@@ -645,18 +644,16 @@ fn test_tuning() {
             break;
         }
 
-        let counter = counter.clone();
-        let interval = interval.clone();
+        if Arc::strong_count(&interval) < 5_000 {
+            let counter = counter.clone();
+            let interval = interval.clone();
 
-        if i <= 5_000 {
```
carllerche, line 651: Because the test expects 3 "valid" measurements, with unreliable enough thread scheduling (CI), it is possible t not to observe these measurements in 5k spawned tasks.

Noah-Kennedy (approved):

PR #5753, opened 2023-06-02
rt(threaded): adjust transition_from_parked behavior after introducing disable_lifo_slot feature

### LINKED ISSUE #5751: Adjust Core::transition_from_parked behavior after introducing disable_lifo_slot feature
### Discussed in https://github.com/tokio-rs/tokio/discussions/5749

<div type='discussions-op-text'>

<sup>Originally posted by **hidva** June  1, 2023</sup>
After introducing the `disable_lifo_slot` feature in Tokio, I think that the behavior of `Core::transition_from_parked` also needs to be adjusted. For example, when `core.run_queue.is_stealable()` is true, park should also be terminated. Additionally, to avoid task contention, `Context::park_timeout` should only wake up others when `core.run_queue.len() >= 2`.

Please let me know your thoughts on this proposed adjustment.

Best regards,
盏一</div>

hidva:
## Motivation

Fix #5751

## Solution

As described in  #5751

### DIFF
--- a/tokio/src/runtime/scheduler/multi_thread/queue.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/queue.rs
@@ -105,9 +105,12 @@ pub(crate) fn local<T: 'static>() -> (Steal<T>, Local<T>) {
 }
 
 impl<T> Local<T> {
-    /// Returns true if the queue has entries that can be stolen.
-    pub(crate) fn is_stealable(&self) -> bool {
-        !self.inner.is_empty()
+    /// Returns the number of entries in the queue
+    pub(crate) fn tasks_num(&self) -> (usize /* stealable */, usize /* total */) {
+        let n = self.inner.len() as usize;
+        // I'm not sure if it's really necessary to return a tuple, but based on has_tasks's comments,
+        // it seems that there may be tasks in queue that cannot be stolen?
+        (n, n)
     }
 
     /// How many tasks can be pushed into the queue
--- a/tokio/src/runtime/scheduler/multi_thread/queue.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/queue.rs
@@ -106,11 +106,8 @@ pub(crate) fn local<T: 'static>() -> (Steal<T>, Local<T>) {
 
 impl<T> Local<T> {
     /// Returns the number of entries in the queue
-    pub(crate) fn tasks_num(&self) -> (usize /* stealable */, usize /* total */) {
-        let n = self.inner.len() as usize;
-        // I'm not sure if it's really necessary to return a tuple, but based on has_tasks's comments,
-        // it seems that there may be tasks in queue that cannot be stolen?
-        (n, n)
+    pub(crate) fn len(&self) -> usize {
+        self.inner.len() as usize
     }
 
     /// How many tasks can be pushed into the queue
--- a/tokio/src/runtime/scheduler/multi_thread/worker.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/worker.rs
@@ -854,11 +854,7 @@ impl Core {
         if self.is_searching {
             return false;
         }
-        let (stealable_tasks, total_tasks) = self.run_queue.tasks_num();
-        if stealable_tasks == 0 {
-            return false;
-        }
-        self.lifo_slot.is_some() as usize + (total_tasks - stealable_tasks) > 0
+        self.lifo_slot.is_some() as usize + self.run_queue.len() > 1
     }
 
     /// Prepares the worker state for parking.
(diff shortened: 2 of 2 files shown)


hidva:
I'm sorry, I tried but couldn't come up with this test case. As shown in the code below, my intention was to prevent other workers from leaving the park state through on_thread_unpark, but it seems that run_task -> transition_from_searching -> transition_worker_from_searching always wakes up a parked worker.

```rust
#[test]
fn issue_5751() {
    use std::thread::ThreadId;
    use std::collections::HashMap;
    struct ParkState {
        park_cnt: HashMap<ThreadId, (usize /* park */, usize /* unpark */)>,
        no_unpark: bool,
    }
    impl ParkState {
        fn new() -> Self {
            Self {
                park_cnt: HashMap::new(),
                no_unpark: false,
            }
        }
    }

    fn wait_park(park_state: &Mutex<ParkState>, exclude: Option<ThreadId>, thread_num: usize) {
        'outer: loop {
            std::thread::sleep(std::time::Duration::from_millis(10));
            let park_state = park_state.lock().unwrap();
            if park_state.park_cnt.len() < thread_num {
                continue;
            }
            for (&thdid, &cnts) in &park_state.park_cnt {
                if let Some(exclude) = exclude {
                    if exclude == thdid {
                        continue;
                    }
                }
                if cnts.0 <= cnts.1 {
                    continue 'outer;
                }
            }
            break;
        }
    }

    let thread_num = 4;
    let park_state = Arc::new(Mutex::new(ParkState::new()));
        .disable_lifo_slot()
        .worker_threads(thread_num)
        .on_thread_park({
            let park_state = Arc::clone(&park_state);
            move || {
                let current_thd = std::thread::current().id();
                let mut park_state = park_state.lock().unwrap();
                if let Some(v) = park_state.park_cnt.get_mut(&current_thd) {
                    v.0 += 1;
                } else {
                    park_state.park_cnt.insert(current_thd, (1, 0));
                }
            }
        })
        .on_thread_unpark({
            let park_state = Arc::clone(&park_state);
            move || {
                let current_thd = std::thread::current().id();
                let mut park_state = park_state.lock().unwrap();
                if park_state.no_unpark {
                    panic!("unexpected unpark");
                }
                park_state.park_cnt.get_mut(&current_thd).unwrap().1 += 1;
            }
        })
        .build()
        .unwrap();

    // wait for all workers to enter the park state
    wait_park(&park_state, None, thread_num);

    rt.block_on(async {
        tokio::spawn(async move {
            std::thread::sleep(std::time::Duration::from_secs(7));
            let current_thd = std::thread::current().id();
            wait_park(&park_state, Some(current_thd), thread_num);
            {
                let mut park_state = park_state.lock().unwrap();
                park_state.no_unpark = true;
            }
            assert_eq!(current_thd, std::thread::current().id());
            std::thread::sleep(std::time::Duration::from_secs(7));
            {
                let mut park_state = park_state.lock().unwrap();
                park_state.no_unpark = false;
            }
        })
        .await
        .unwrap();
    })
}
```

carllerche:
I dug into writing a test for this. It turns out that some *other* issue (relatively minor) counters this one and prevents any bugs :shrug:

The other issue is `notify_parked_local` sometimes will notify the current worker when coming from a parked state.

I wouldn't worry too much about this, I am in the process of rewriting all of that logic, so I will fix it then.

tokio/src/runtime/scheduler/multi_thread/queue.rs
```diff
@@ -106,9 +106,12 @@ pub(crate) fn local<T: 'static>() -> (Steal<T>, Local<T>) {
 }
 
 impl<T> Local<T> {
-    /// Returns true if the queue has entries that can be stolen.
-    pub(crate) fn is_stealable(&self) -> bool {
-        !self.inner.is_empty()
+    /// Returns the number of entries in the queue
+    pub(crate) fn tasks_num(&self) -> (usize /* stealable */, usize /* total */) {
```
carllerche, line 110: It should be fine to return a single `usize` here. I am not sure which comment you are referencing.

carllerche (changes requested):
Looks good, one smaller change is requested.

tokio/src/runtime/scheduler/multi_thread/queue.rs
hidva: > "protect" some tasks from stealing
https://github.com/tokio-rs/tokio/blob/fb4d43017d6cb7b7cbdfbe5fc189bca0291ff215/tokio/src/runtime/scheduler/multi_thread/queue.rs#L122-L128

hidva:
@carllerche @Darksonn Is there anything else I need to do here?

carllerche (approved):

PR #5804, opened 2023-06-19
fs: always try to get file type in `fs::read_dir`

icedrocket:
The current implementation of `fs::read_dir` pre-stores the file type, except for a few operating systems that are known to require an extra call to get the file type. The cfg attributes we use to check such operating systems comes from [the std implementation](
https://github.com/rust-lang/rust/blob/3c9e0705ba0c1c845fe7cdbd0bdf4a914f49cc8e/library/std/src/sys/unix/fs.rs#L862-L869), but this was recently updated and no longer matches.

I think I'd rather just delete those cfg attributes, because when we read a directory, we also read the file type, except in very few cases.

### DIFF
--- a/tokio/src/fs/read_dir.rs
+++ b/tokio/src/fs/read_dir.rs
@@ -135,12 +135,6 @@ impl ReadDir {
             let success = ret.is_ok();
 
             buf.push_back(ret.map(|std| DirEntry {
-                #[cfg(not(any(
-                    target_os = "solaris",
-                    target_os = "illumos",
-                    target_os = "haiku",
-                    target_os = "vxworks"
-                )))]
                 file_type: std.file_type().ok(),
                 std: Arc::new(std),
             }));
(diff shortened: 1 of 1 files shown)


icedrocket:
> > but this was recently updated and no longer matches.
> 
> We must still support older compilers, so if these methods are not available there, then we cannot remove the cfgs yet.

The std implementations I linked to in my description are just the parts that decide whether to make an additional call. Targets that support std should be fine, as they also support `DirEntry::file_type`.

icedrocket:
> > but this was recently updated and no longer matches.
> 
> We must still support older compilers, so if these methods are not available there, then we cannot remove the cfgs yet.

The std implementation I linked to in my description is just the part that decides whether to make an additional call. Targets that support std should be fine, as they support `DirEntry::file_type`.

icedrocket:
Those cfg attributes were added by me, but I didn't see much benefit in them, so I made this change.

icedrocket:
Superseded by #5806.

PR #5809, opened 2023-06-21
rt: pad the task struct to avoid false sharing

carllerche:
This change pads the task struct to avoid false sharing. These structs can overlap cache lines without this alignment.

I avoided adding a `CachePadded` wrapper struct as I am not sure how the padded wrapper struct would impact the pointer stuff we do in the task harness (casting `*Header` to `*Cell` as well as the other pointer math). 

I'm also not 100% sure what happens when you combine `repr(C)` with `repr(align(n))` but I would guess (hope) it does the right thing.

Thoughts @Darksonn ?

As part of a larger scheduler refactor that I am working on, I have observed false sharing happened and confirmed the false sharing using `perf c2c`. Adding the padding here resolves the issue.

### DIFF
--- a/tokio/src/runtime/task/core.rs
+++ b/tokio/src/runtime/task/core.rs
@@ -28,6 +28,94 @@ use std::task::{Context, Poll, Waker};
 ///
 /// Any changes to the layout of this struct _must_ also be reflected in the
 /// const fns in raw.rs.
+///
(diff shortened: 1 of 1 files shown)


Noah-Kennedy (approved):
I'll stamp but I agree that we should have @Darksonn weigh in on the `repr` stuff.

Darksonn (approved):

PR #5814, opened 2023-06-23
test: fetch actions from mock handle before write

### LINKED ISSUE #5813: tokio_test::io::Mock does not read actions from handle when used as writer
**Version**

```
├── tokio v1.28.2
│   └── tokio-macros v2.1.0 (proc-macro)
└── tokio-test v0.4.2
    ├── tokio v1.28.2 (*)
    └── tokio-stream v0.1.14
        └── tokio v1.28.2 (*)
```

**Platform**
Linux lenovo-arch 6.3.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Wed, 14 Jun 2023 20:10:31 +0000 x86_64 GNU/Linux
```toml
[dependencies]
tokio = {version = "1.28.2", features = ["macros", "rt-multi-thread", "io-util", "rt"]}
tokio-test = "0.4.2"
```
**Description**
```rust
use tokio_test::io::Builder;

use tokio::io::{AsyncReadExt, AsyncWriteExt};
#[tokio::main]
async fn main() {
    let mut wr = Builder::new().write(b"hello").build();
    println!("{:?}", AsyncWriteExt::write(&mut wr, b"hello").await);

    let (mut wr, mut wr_h) = Builder::new().build_with_handle();
    wr_h.write(b"hello");
    println!("{:?}", AsyncWriteExt::write(&mut wr, b"hello").await);

    let (mut wr, mut wr_h) = Builder::new().build_with_handle();
    wr_h.read(b"world");
    let mut buf = vec![0; 8];
    println!("{:?}", AsyncReadExt::read(&mut wr, &mut buf).await);
}
```

Expected output:
```
Ok(5)
Ok(5)
Ok(5)
```

Actual output:
```
Ok(5)
Err(Kind(BrokenPipe))
Ok(5)
```

**Additional notes**
https://github.com/tokio-rs/tokio/blob/2e62374e4a643015ea9ecf5a1d012c3429eed42f/tokio-test/src/io.rs#L422-L425

I suspect this is not correct
EDIT: I suspect instead that this is fine, but `self.inner.poll_action` should be called before `self.inner.write`. Working on PR.

pbrenna:
## Motivation

Fixes #5813
When the Mock object is created without actions (such as for use with its Handle), it can happen that actions from the handle are never fetched, causing an unexpected broken pipe error.

## Solution
Before executing the write, if there are no actions in the Mock, try receiving from the handle if an action is ready.


Darksonn (approved):
Thanks, LGTM.

PR #5818, opened 2023-06-26
std::error::Error import style fix

wathenjiang:
`std::error::Error` import style fix, the same way of use case in https://doc.rust-lang.org/std/error/trait.Error.html


spongecaptain:
The tool `cargo clippy` should warn, but it does not work when I run it in the `tokio` directory.

Darksonn (approved):
Thanks.

PR #5820, opened 2023-06-26
docs: `broadcast::Sender::send`: refer to lagging

marcospb19:
It's worth mentioning that receiver's lag is a possible cause of missing messages.


Darksonn (approved):
Thanks.

PR #5821, opened 2023-06-26
sync::broadcast: use usize instead of u64

marcospb19:
Doing back-and-forth casting between usize and u64 is unnecessary, if the broadcast buffer is bound to `usize` anyways, we can use that directly.

### DIFF
--- a/tokio/src/sync/broadcast.rs
+++ b/tokio/src/sync/broadcast.rs
@@ -207,7 +207,7 @@ pub struct Receiver<T> {
     shared: Arc<Shared<T>>,
 
     /// Next position to read from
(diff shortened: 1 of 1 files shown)


Noah-Kennedy:
This would be a breaking change.

marcospb19:
oh wow, this is a breaking change in `tokio` and `tokio-stream` (as CI just revealed to me), the `u64` value was exposed even though it was impossible to have anything other than what's bounded by `usize`, so maybe it's better to close this and avoid the breakage?

Noah-Kennedy:
Probably. The u64 is exposed via the error types.

PR #5833, opened 2023-06-28
io: remove slab in favor of Arc and allocations

carllerche:
This patch removes the custom slab in favor of regular allocations an `Arc`. Originally, the slab was used to be able to pass indexes as tokens to the I/O driver when registering I/O resources. However, this has the downside of having a more expensive token lookup path. It also pins a `ScheduledIo` to a specific I/O driver. Additionally, the slab is approaching custom allocator territory.

We plan to explore migrating I/O resources between I/O drivers. As a step towards that, we need to decouple `ScheduledIo` from the I/O driver. To do this, the patch uses plain-old allocation to allocate the `ScheduledIo` and we use the pointer as the token. To use the token, we need to be very careful about releasing the `ScheduledIo`. We need to make sure that the associated I/O handle is deregistered from the I/O driver **and** there are no polls. The strategy in this PR is to let the I/O driver do the final release between polls, but I expect this strategy to evolve over time.


tokio/src/runtime/io/driver.rs
```diff
@@ -196,35 +174,23 @@ impl Driver {
...
-            }
-        }
-
-        handle.metrics.incr_ready_count_by(ready_count);
-    }
-
-    fn dispatch(resources: &mut Slab<ScheduledIo>, tick: u8, token: mio::Token, ready: Ready) {
-        let addr = slab::Address::from_usize(ADDRESS.unpack(token.0));
+                let ready = Ready::from_mio(event);
+                let ptr: *const ScheduledIo = token.0 as *const _;
```
hawkw, line 178: Miri probably hates this...it could, at least, be worth adding a comment about using [`ptr::from_exposed_addr`](https://doc.rust-lang.org/nightly/std/ptr/fn.from_exposed_addr.html) if and when it becomes stable?

tokio/src/runtime/io/registration_set.rs
```diff
@@ -0,0 +1,134 @@
...
+        if synced.is_shutdown {
+            return Err(io::Error::new(
+                io::ErrorKind::Other,
+                crate::util::error::RUNTIME_SHUTTING_DOWN_ERROR,
+            ));
+        }
+
+        let ret = Arc::new(ScheduledIo::default());
+
+        // Push a ref into the
```
hawkw, line 63: "push a ref into the" ... ?

tokio/src/runtime/io/scheduled_io.rs
```diff
@@ -81,48 +82,31 @@ enum State {
...
             waiters: Mutex::new(Default::default()),
         }
     }
 }
 
 impl ScheduledIo {
-    pub(crate) fn generation(&self) -> usize {
-        GENERATION.unpack(self.readiness.load(Acquire))
+    pub(crate) fn token(&self) -> mio::Token {
+        mio::Token(self as *const _ as usize)
```
hawkw, line 109: similarly to my above comment, this is where we would want to use `pointer::expose_addr` eventually, if it becomes stable..

tokio/src/runtime/io/registration_set.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    // List of all registrations tracked by the set
+    registrations: LinkedList<Arc<ScheduledIo>, ScheduledIo>,
+
+    // Registrations that are pending drop. When a `Registration` is dropped, it
+    // stores its `ScheduledIo` in this list. The I/O driver is responsible for
+    // dropping it. This ensures the `ScheduledIo` is not freed while it can
+    // still be included in an I/O event.
+    pending_release: Vec<Arc<ScheduledIo>>,
+}
```
hawkw, line 27: what's the motivation for not just having this be 
```suggestion
pub(super) struct RegistrationSet {
    num_pending_release: AtomicUsize,
    synced: Mutex<Synced>,
}

struct Synced {
    // True when the I/O driver shutdown. At this point, no more registrations
    // should be added to the set.
    is_shutdown: bool,

    // List of all registrations tracked by the set
    registrations: LinkedList<Arc<ScheduledIo>, ScheduledIo>,

    // Registrations that are pending drop. When a `Registration` is dropped, it
    // stores its `ScheduledIo` in this list. The I/O driver is responsible for
    // dropping it. This ensures the `ScheduledIo` is not freed while it can
    // still be included in an I/O event.
    pending_release: Vec<Arc<ScheduledIo>>,
}
``` 
and changing all the methods that take a `&self, &mut Synced` to just lock `self.synced` internally? as far as I could tell, there's nowhere where we lock `synced` for multiple operations, unless i can't read and missed something?

hawkw (approved):
overall i'm +1 on this change, although it's a bummer that there's currently no way to do it that's provenance-preserving under Miri. i left some comments but there aren't any hard blockers from me.

tokio/src/runtime/io/registration_set.rs
carllerche: I plan on unifying mutexes in a later PR.

carllerche:
I'm pretty sure the hung CI is spurious and from the tuning test. I will try to fix in a separate PR.

PR #5856, opened 2023-07-09
example: use copy_bidirectional in proxy.rs

kiron1:
## Motivation

The `examples/proxy.rs` was created before `copy_bidirectional` https://github.com/tokio-rs/tokio/pull/3572 was added to tokio. In https://github.com/tokio-rs/tokio/pull/2590 the proxy example was updated to close pending connections. By reusing the `copy_bidirectional` the example can be simplified.

## Solution

Reuse [`copy_bidirectional`](https://docs.rs/tokio/latest/tokio/io/fn.copy_bidirectional.html) in example.

### DIFF
--- a/examples/proxy.rs
+++ b/examples/proxy.rs
@@ -22,8 +22,7 @@
 
 #![warn(rust_2018_idioms)]
 
-use tokio::io;
-use tokio::io::AsyncWriteExt;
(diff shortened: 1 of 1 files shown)


Darksonn (approved):
Thanks.

PR #5868, opened 2023-07-14
Add cooperative UCred impl for espidf and other process-less OSes

jasta:
## Motivation

espidf fails to compile due to missing get_peer_cred APIs.  Since there are no processes (it runs FreeRTOS), we can just
pretend everything runs as root.

## Solution

Dummy implementation keeps existing code happy and isn't technically wrong since it would show all UNIX domain sockets as being owned by the same user.


Darksonn (approved):
Thanks.

PR #5870, opened 2023-07-16
docs: BytesMut::with_capacity does not guarantee exact capacity

mahkoh:
## Motivation

The documentation implies that BytesMut::with_capacity can be used in protocols that need to read exact numbers of bytes (and where read_exact is too inefficient.) This is incorrect as described in the commit message.

## Solution

Adjust the documentation.


carllerche (approved):
Nice, thanks.

PR #5881, opened 2023-07-18
poll: Do not clear readiness on short read/writes.

jasta:
## Motivation

The new mio_unsupported_force_poll_poll behaviour works the same as Windows (using level-triggered APIs to mimic edge-triggered ones) and it depends on intercepting an EAGAIN result to start polling the fd again.  Without this change, short reads don't reregister with the poll Selector and it causes the socket to stall in a read async call forever.

## Solution

Per #5866, going with the simplest solution of just adding another `cfg!` check alongside the Windows one.

### DIFF
--- a/tokio/src/io/poll_evented.rs
+++ b/tokio/src/io/poll_evented.rs
@@ -165,7 +165,9 @@ feature! {
                 match self.io.as_ref().unwrap().read(b) {
                     Ok(n) => {
                         // if we read a partially full buffer, this is sufficient on unix to show
-                        // that the socket buffer has been drained
+                        // that the socket buffer has been drained.  Unfortunately this assumption
+                        // fails for level-triggered selectors (like on Windows or poll even for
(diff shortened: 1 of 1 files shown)


tokio/src/io/poll_evented.rs
```diff
@@ -166,7 +166,7 @@ feature! {
                     Ok(n) => {
                         // if we read a partially full buffer, this is sufficient on unix to show
                         // that the socket buffer has been drained
-                        if n > 0 && (!cfg!(windows) && n < len) {
+                        if n > 0 && (!cfg!(windows) && !cfg!(mio_unsupported_force_poll_poll) && n < len) {
```
Noah-Kennedy, line 169: You should add some inline comments to these files explaining why that cfg is there.

Noah-Kennedy (changes requested):
Add some comments inline so folks don't get confused in the future, then I think we're good.

Noah-Kennedy (approved):
Looks good.

PR #5891, opened 2023-07-28
Fix clippy warnings

NobodyXu:
## Motivation

There are a few clippy warnings to fix.

## Solution

Fix them and it also simplifies the code a bit.


Darksonn (approved):
Thanks.

PR #5893, opened 2023-07-28
Update CI config

taiki-e:
- Remove duplicate target from cross-test
- Remove needless --test-threads=1 as RUST_TEST_THREADS=1 already specifies it
- Remove --cfg tokio_no_ipv6 as setup-cross-toolchain-action supports it
- Automatically cancel all outdated workflows on PR (currently this is only set for ci.yml)
- Remove --cfg tokio_no_atomic_u64 as tests/builds with --target target-specs/i686-unknown-linux-gnu.json handles these cases

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -538,7 +538,7 @@ jobs:
           cargo test --doc -Zbuild-std --target target-specs/i686-unknown-linux-gnu.json -p tokio --all-features
         env:
           RUST_TEST_THREADS: 1
-          RUSTFLAGS: --cfg tokio_unstable --cfg tokio_taskdump -Dwarnings --cfg tokio_no_atomic_u64 --cfg tokio_no_tuning_tests
+          RUSTFLAGS: --cfg tokio_unstable --cfg tokio_taskdump -Dwarnings --cfg tokio_no_tuning_tests
 
   no-atomic-u64-check:
     name: Check tokio --feature-powerset --depth 2 on i686-unknown-linux-gnu without AtomicU64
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -464,10 +461,10 @@ jobs:
         run: |
           set -euxo pipefail
           cargo nextest run -p tokio --all-features --target ${{ matrix.target }}
-          cargo test --doc -p tokio --all-features --target ${{ matrix.target }} -- --test-threads 1
+          cargo test --doc -p tokio --all-features --target ${{ matrix.target }}
         env:
           RUST_TEST_THREADS: 1
-          RUSTFLAGS: --cfg tokio_unstable -Dwarnings --cfg tokio_no_ipv6 --cfg tokio_no_tuning_tests ${{ matrix.rustflags }}
+          RUSTFLAGS: --cfg tokio_unstable -Dwarnings --cfg tokio_no_tuning_tests ${{ matrix.rustflags }}
```
NobodyXu, line 467: I just realized these `RUST*` flags should be set before `Swatinem/rust-cache@v2`:

```
 - name: Set RUST env
   run: |
     echo RUST_TEST_THREADS=1 >> "$GITHUB_ENV"
     echo 'RUSTFLAGS=--cfg tokio_unstable -Dwarnings --cfg tokio_no_tuning_tests ${{ matrix.rustflags }}' >> "$GITHUB_ENV"
```

Darksonn (approved):

.github/workflows/ci.yml
taiki-e: Considering how Swatinem/rust-cache works, it may be best for some workflows that meet certain conditions, but since we don't change rustflags often and sometimes set different rustflags for each step, I think it is also fine as-is.

PR #5915, opened 2023-08-05
Create StreamMock for testing Streams

shade:
Inroduce a new mock type to tests streams and eventually sinks. Only includes next() and wait() for now.

## Motivation

Basically making a generic Stream mock type https://github.com/tokio-rs/tokio/issues/4106 

Next PR I'd like to add a Sink<T> implementation. Since Sink requires an error type, I think we can do something like this.

```
impl Sink<T> for StreamMock<T> {
// Consume implementation
}

impl <T: Unpin, E: Error> Sink<T> for StreamMock<Result<T,E>> {
   type Error = E;
}


```

Main motivation is so we can support splitting the stream mock into a AsyncRead/AsyncWrite.

### DIFF
--- a/tokio-test/src/stream_mock.rs
+++ b/tokio-test/src/stream_mock.rs
@@ -1,0 +1,163 @@
+#![cfg(not(loom))]
+
+//! A mock stream implementing [`Stream`].
+//!
+//! # Overview
+//! This crate provides a `StreamMock` that can be used to test code that interacts with streams.
+//! It allows you to mock the behavior of a stream and control the items it yields and the waiting
+//! intervals between items.
+//!
+//! # Usage
+//! To use the `StreamMock`, you need to create a builder using[`StreamMockBuilder`]. The builder
+//! allows you to enqueue actions such as returning items or waiting for a certain duration.
+//!
+//! # Example
+//! ```rust
+//!
+//! use futures_util::StreamExt;
+//! use std::time::Duration;
+//! use tokio_test::stream_mock::StreamMockBuilder;
+//!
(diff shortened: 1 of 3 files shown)


tokio-test/src/stream_mock.rs
```diff
@@ -0,0 +1,163 @@
...
+                    Poll::Pending
+                }
+            },
+            None => Poll::Ready(None),
+        }
+    }
+}
+
+impl<T: Unpin> Drop for StreamMock<T> {
+    fn drop(&mut self) {
```
Noah-Kennedy, line 147: You should first verify that we aren't panicking already to avoid a double panic, like we do here: https://github.com/tokio-rs/tokio/blob/master/tokio-test/src/io.rs#L473

Noah-Kennedy (changes requested):

tokio-test/src/stream_mock.rs
```diff
@@ -0,0 +1,163 @@
...
+//!     assert_eq!(stream_mock.next().await, None);
+//! }
+//! ```
+
+use std::collections::VecDeque;
+use std::pin::Pin;
+use std::task::Poll;
+use std::time::Duration;
+
+use futures_core::Future;
```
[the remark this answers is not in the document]
shade: Was the auto-import from VSCode, will fix.

tokio-test/src/stream_mock.rs
shade: Good call out, especially for tests that makes a lot of sense. Will update.

Noah-Kennedy (approved):

PR #5932, opened 2023-08-14
io: AsyncFd minor tweaks

folkertdev:
## Motivation

Fixing or documenting some things I found confusing

## Solution

- simplify the logic of `try_io` by only checking once whether the error is `WouldBlock`
- document why `AsyncFd` uses an `Option` around its `inner` field.
- remove some `.into()` calls (clearer this way, I think, but also reduced code bloat)


Darksonn (approved):
Thanks.

PR #5935, opened 2023-08-15
rt(alt): tweak some constants to improve scalability

carllerche:
This patch aims to reduce the number of threads that get no-op wakeups.

### DIFF
--- a/tokio/src/runtime/io/scheduled_io.rs
+++ b/tokio/src/runtime/io/scheduled_io.rs
@@ -17,6 +17,93 @@ use std::task::{Context, Poll, Waker};
 
 /// Stored in the I/O driver resource slab.
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


tokio/src/runtime/io/scheduled_io.rs
```diff
@@ -17,6 +17,93 @@ use std::task::{Context, Poll, Waker};
 
 /// Stored in the I/O driver resource slab.
 #[derive(Debug)]
+// # This struct should be cache padded to avoid false sharing. The cache padding rules are copied
```
carllerche, line 20: Yes, I know it is time to factor this out into a CachePadded abstraction. I pinky promise I'll do it in a follow-up, but I don't want to mix up concerns in this PR.

Darksonn (approved):

PR #5947, opened 2023-08-25
sync: improve cancel-safety documentation for mpsc::Sender::send

sunshowers:
## Motivation

This specific issue (data loss because a send got cancelled) has bitten our team a couple of times over the last few months. We've switched to recommending this kind of reserve pattern instead.

## Solution

Update the documentation to make it clear that data loss can occur if a send is cancelled.

### DIFF
--- a/tokio/src/sync/mpsc/bounded.rs
+++ b/tokio/src/sync/mpsc/bounded.rs
@@ -439,7 +439,7 @@ impl<T> Sender<T> {
     ///
     /// If `send` is used as the event in a [`tokio::select!`](crate::select)
     /// statement and some other branch completes first, then it is guaranteed
-    /// that the message was not sent. **However, in that case the message
+    /// that the message was not sent. **However, in that case, the message
     /// is dropped and will be lost.**
     ///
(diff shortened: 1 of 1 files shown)


tokio/src/sync/mpsc/bounded.rs
```diff
@@ -439,7 +439,11 @@ impl<T> Sender<T> {
     ///
     /// If `send` is used as the event in a [`tokio::select!`](crate::select)
     /// statement and some other branch completes first, then it is guaranteed
-    /// that the message was not sent.
+    /// that the message was not sent. **However, in that case the message
```
hawkw, line 442: tiny nit here:
```suggestion
    /// that the message was not sent. **However, in that case, the message
```

hawkw (approved):

sunshowers:
The issue appears to be due to Rust 1.72, unrelated to this PR.

Darksonn (approved):

hawkw:
#5951 should fix the CI issues here, so I'm updating with that change and rerunning CI.

PR #5956, opened 2023-08-27
Semaphore example using an Arc<Semaphore> for semaphore passing

alexanderkirilin:
A simple example of limiting request processing using a task-safe semaphore and demonstrating `acquire_owned` for: https://github.com/tokio-rs/tokio/issues/5933

### DIFF
--- a/tokio/src/sync/semaphore.rs
+++ b/tokio/src/sync/semaphore.rs
@@ -47,7 +47,23 @@ use std::sync::Arc;
 /// }
 /// ```
 ///
-/// Use [`Semaphore::acquire_owned`] to move permits across tasks:
+/// Limit number of incoming requests being handled at the same time, using a task-safe semaphore
+/// via [`Semaphore::acquire_owned`].
+///
+/// Similar to limiting the numer of simultaneous opened files, network handles are a limited resource.
+/// Allowing an unbounded amount of requests to be processed could result in a denial-of-service,
+/// and many other issues.
+///
+/// However, in contrast to the file example, this example uses a non-static, `Arc`-wrapped `Semaphore`
+/// for more fine-grained lifetime and ownership control. The `Arc` allows multiple threads and tasks
+/// to share ownership of the semaphore. Here we create a shallow-copy of the `Arc<Semaphore>` reference
+/// using `clone`. Then we move the `Arc` into our task, where we will wait to acquire a new permit
+/// via [`Semaphore::acquire_owned`] so we can start processing. Once our processing is done, we
+/// will drop the permit to allow other tasks to acquire it.
+///
+/// If we leave the scope where the `Arc<Semaphore>` was defined, and references still exist in any
+/// of our tasks to it, the `Semaphore` will continue to exist in a static-like context, until all
+/// `Arc`s have been dropped.
 ///
 /// ```
 /// use std::sync::Arc;
@@ -59,17 +75,22 @@ use std::sync::Arc;
 ///     let mut join_handles = Vec::new();
 ///
 ///     for _ in 0..5 {
-///         let permit = semaphore.clone().acquire_owned().await.unwrap();
+///         let semaphore_clone = semaphore.clone();
 ///         join_handles.push(tokio::spawn(async move {
-///             // perform task...
-///             // explicitly own `permit` in the task
-///             drop(permit);
+///             perform_task(semaphore_clone).await;
 ///         }));
 ///     }
 ///
 ///     for handle in join_handles {
 ///         handle.await.unwrap();
 ///     }
+/// }
+///
+/// async fn perform_task(semaphore: Arc<Semaphore>) {
+///     let permit = semaphore.acquire_owned().await.unwrap();
+///     // perform task...
+///     // explicitly own `permit` in the task
+///     drop(permit);
 /// }
 /// ```
 ///
--- a/tokio/src/sync/semaphore.rs
+++ b/tokio/src/sync/semaphore.rs
@@ -64,11 +64,10 @@ use std::sync::Arc;
 /// of our tasks to it, the `Semaphore` will continue to exist in a static-like context, until all
 /// `Arc`s have been dropped.
 ///
-/// ```no_run
+/// ```compile_fail
 /// use std::sync::Arc;
 /// use tokio::sync::Semaphore;
-/// use tokio::net::{TcpListener, TcpStream};
-/// use tokio::io::{AsyncReadExt, AsyncWriteExt};
+/// use tokio::net::{TcpListener};
 ///
 /// #[tokio::main]
 /// async fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -78,39 +77,15 @@ use std::sync::Arc;
 ///     loop {
 ///         // Acquire permit before accepting, else the connection will wait without purpose
 ///         let permit = semaphore.clone().acquire_owned().await.unwrap();
-///         let (socket, _) = listener.accept().await?;
+///         let (mut socket, _) = listener.accept().await?;
 ///
 ///         tokio::spawn(async move {
-///             handle_connection(socket).await;
+///             // Do work, and drop socket after
+///             handle_connection(&mut socket).await;
 ///             // Explicitly drop the permit, so more tasks can be created
 ///             drop(permit);
 ///         });
 ///     }
-/// }
-///
-/// async fn handle_connection(mut socket: TcpStream) {
-///     let mut buf = [0; 1024];
-///
-///     // In a loop, read data from the socket and write the data back.
-///     loop {
-///         let n = match socket.read(&mut buf).await {
-///             // socket closed
-///             Ok(n) if n == 0 => break,
-///             Ok(n) => n,
-///             Err(e) => {
-///                 eprintln!("failed to read from socket; err = {:?}", e);
-///                 break;
-///             }
-///         };
-///
-///         // Write the data back
-///         if let Err(e) = socket.write_all(&buf[0..n]).await {
-///             eprintln!("failed to write to socket; err = {:?}", e);
-///             break;
-///         }
-///     }
-///
-///     // Implicitly drop the socket here
 /// }
 /// ```
 ///
--- a/tokio/src/sync/semaphore.rs
+++ b/tokio/src/sync/semaphore.rs
@@ -76,33 +76,34 @@ use std::sync::Arc;
 ///     let listener = TcpListener::bind("127.0.0.1:8080").await?;
 ///
 ///     loop {
+///         // Acquire permit before accepting, else the connection will wait without purpose
 ///         let permit = semaphore.clone().acquire_owned().await.unwrap();
 ///         let (mut socket, _) = listener.accept().await?;
 ///
 ///         tokio::spawn(async move {
-///                 let mut buf = [0; 1024];
+///              let mut buf = [0; 1024];
 ///
-///                 // In a loop, read data from the socket and write the data back.
-///                 loop {
-///                     let n = match socket.read(&mut buf).await {
-///                         // socket closed
-///                         Ok(n) if n == 0 => break,
-///                         Ok(n) => n,
-///                         Err(e) => {
-///                             eprintln!("failed to read from socket; err = {:?}", e);
-///                             break;
-///                         }
-///                     };
+///              // In a loop, read data from the socket and write the data back.
+///              loop {
+///                  let n = match socket.read(&mut buf).await {
+///                      // socket closed
+///                      Ok(n) if n == 0 => break,
+///                      Ok(n) => n,
+///                      Err(e) => {
+///                          eprintln!("failed to read from socket; err = {:?}", e);
+///                          break;
+///                      }
+///                  };
 ///
-///                     // Write the data back
-///                     if let Err(e) = socket.write_all(&buf[0..n]).await {
-///                         eprintln!("failed to write to socket; err = {:?}", e);
-///                         break;
-///                     }
-///                 }
+///                  // Write the data back
+///                  if let Err(e) = socket.write_all(&buf[0..n]).await {
+///                      eprintln!("failed to write to socket; err = {:?}", e);
+///                      break;
+///                  }
+///              }
 ///
-///                 // Drop the permit, so more tasks can be created
-///                 drop(permit);
+///              // Drop the permit, so more tasks can be created
+///              drop(permit);
 ///         });
 ///     }
 /// }
--- a/tokio/src/sync/semaphore.rs
+++ b/tokio/src/sync/semaphore.rs
@@ -101,7 +101,7 @@ use std::sync::Arc;
 ///     let listener = TcpListener::bind("127.0.0.1:8080").await?;
 ///
 ///     loop {
-///         // Acquire permit before accepting, to avoid needlessly taking up a (finite) file descriptor
+///         // Acquire permit before accepting the next socket.
 ///         let permit = semaphore.clone().acquire_owned().await.unwrap();
 ///         let (mut socket, _) = listener.accept().await?;
 ///
--- a/tokio/src/sync/semaphore.rs
+++ b/tokio/src/sync/semaphore.rs
@@ -93,7 +93,7 @@ use std::sync::Arc;
 /// ```no_run
 /// use std::sync::Arc;
 /// use tokio::sync::Semaphore;
-/// use tokio::net::{TcpListener, TcpStream};
+/// use tokio::net::TcpListener;
 ///
 /// #[tokio::main]
 /// async fn main() -> std::io::Result<()> {
--- a/tokio/src/sync/semaphore.rs
+++ b/tokio/src/sync/semaphore.rs
@@ -96,7 +96,7 @@ use std::sync::Arc;
 /// use tokio::net::{TcpListener, TcpStream};
 ///
 /// #[tokio::main]
-/// async fn main() -> Result<(), Box<dyn std::error::Error>> {
+/// async fn main() -> std::io::Result<()> {
 ///     let semaphore = Arc::new(Semaphore::new(3));
 ///     let listener = TcpListener::bind("127.0.0.1:8080").await?;
 ///
--- a/tokio/src/sync/semaphore.rs
+++ b/tokio/src/sync/semaphore.rs
@@ -66,23 +66,31 @@ use std::sync::Arc;
 /// `Arc`s have been dropped.
 ///
 /// ```
+/// use std::sync::Arc;
+/// use tokio::sync::Semaphore;
+///
 /// #[tokio::main]
 /// async fn main() {
 ///     let semaphore = Arc::new(Semaphore::new(3));
 ///     let mut join_handles = Vec::new();
 ///
 ///     for _ in 0..5 {
-///         let permit = semaphore.clone().acquire_owned().await.unwrap();
+///         let semaphore_clone = semaphore.clone();
 ///         join_handles.push(tokio::spawn(async move {
-///             // perform task...
-///             // explicitly own `permit` in the task
-///             drop(permit);
+///             perform_task(semaphore_clone).await;
 ///         }));
 ///     }
 ///
 ///     for handle in join_handles {
 ///         handle.await.unwrap();
 ///     }
+/// }
+///
+/// async fn perform_task(semaphore: Arc<Semaphore>) {
+///     let permit = semaphore.acquire_owned().await.unwrap();
+///     // perform task...
+///     // explicitly own `permit` in the task
+///     drop(permit);
 /// }
 /// ```
 ///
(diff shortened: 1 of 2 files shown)


Darksonn:
Your branch is behind #5939, which has already been merged. Would you be able to update your branch to latest master? You should be able to do this by merging master into your branch and fixing the resulting conflicts.

tokio/src/sync/semaphore.rs
```diff
@@ -59,18 +75,23 @@ use std::sync::Arc;
...
 ///
 ///     for _ in 0..5 {
-///         let permit = semaphore.clone().acquire_owned().await.unwrap();
+///         let semaphore_clone = semaphore.clone();
 ///         join_handles.push(tokio::spawn(async move {
-///             // perform task...
-///             // explicitly own `permit` in the task
-///             drop(permit);
+///             perform_task(semaphore_clone).await;
 ///         }));
```
Darksonn, line 81: I think we probably still want to acquire the permit before spawning. That way, we don't spawn more tasks than we have permits.

Also, with the way you're using it now, you don't need `acquire_owned`. Using `acquire` would work fine, since the permit itself is not moved across tasks.

alexanderkirilin: Fair enough. I was assuming you would queue up the tasks, and they would run non-stop as permits were free'd -- but this makes more sense resource-wise. I'll change it back to the original, and rework the doc string.

Can you tell me the significance about moving permits (or anything for that matter) across tasks? Is this a matter of scoping and lifetimes?

alexanderkirilin: Ahh, I understand. Thank you.

How's this?

tokio/src/sync/semaphore.rs
```diff
@@ -47,7 +47,23 @@ use std::sync::Arc;
 /// }
 /// ```
 ///
-/// Use [`Semaphore::acquire_owned`] to move permits across tasks:
+/// Limit number of incoming requests being handled at the same time, using a task-safe semaphore
+/// via [`Semaphore::acquire_owned`].
```
Darksonn, line 51: Let's drop the `acquire_owned` part from the first line, and instead explain its use in the text.
```suggestion
/// Limit number of incoming requests being handled at the same time.
```

tokio/src/sync/semaphore.rs
```diff
@@ -47,7 +47,23 @@ use std::sync::Arc;
...
+///
+/// Similar to limiting the numer of simultaneous opened files, network handles are a limited resource.
+/// Allowing an unbounded amount of requests to be processed could result in a denial-of-service,
+/// and many other issues.
```
Darksonn, line 55: Network handles are a good example. I suggest that we update the example to accept on a `TcpListener` in a very similar way to the example in the README of this repository. Except this time, we acquire a permit before accepting a connection, and we drop the permit after handling an incoming connection, inside `tokio::spawn`.

I suggest this because right now, we're not really handling "incoming requests". It's just a simple loop.

alexanderkirilin: Got it. I've modified the request handler in the README, and repurposed it here.

Do you think that would be prudent -- or would simply sticking to the basic format that was there before be better? For example, removing the inner loop and just have the old "// do work here".

alexanderkirilin:
@Darksonn I think the event handler loop in the updated example is causing issues with CI.

The test suite has been running for 30 minutes so far, and I can't find a way to cancel the workflows.

Would you be able to stop it?

tokio/src/sync/semaphore.rs
```diff
@@ -47,28 +47,63 @@ use std::sync::Arc;
...
+/// to share ownership of the semaphore. Here we create a shallow-copy of the `Arc<Semaphore>` reference
+/// using `clone`. Then we acquire a permit through the `Arc` from the `Semaphore` via [`Semaphore::acquire_owned`],
+/// and move it inside the task. This ensures no non-`'static` variables are referenced from within
+/// said task.
+///
+/// If we leave the scope where the `Arc<Semaphore>` was defined, and references still exist in any
+/// of our tasks to it, the `Semaphore` will continue to exist in a static-like context, until all
+/// `Arc`s have been dropped.
+///
+/// ```no_run
```
alexanderkirilin, line 67: I think there's an issue with running the non-terminating outer loop, that will cause the test suite to spin forever. 

I've opted to have it compile, but not run. If it's necessary for all docstrings to run, then I think it could be replaced with a `for` loop, with some comments explaining that normally it would be a regular, bare loop.

tokio/src/sync/semaphore.rs
```diff
@@ -47,28 +47,63 @@ use std::sync::Arc;
...
+///                     if let Err(e) = socket.write_all(&buf[0..n]).await {
+///                         eprintln!("failed to write to socket; err = {:?}", e);
+///                         break;
+///                     }
+///                 }
 ///
-///     for handle in join_handles {
-///         handle.await.unwrap();
+///                 // Drop the permit, so more tasks can be created
+///                 drop(permit);
```
Darksonn, line 105: I would drop the socket first. Otherwise, there's technically a short period where the socket is still open, even though we have released the permit.
```suggestion
///                 drop(socket);
///                 // Drop the permit, so more tasks can be created
///                 drop(permit);
```

tokio/src/sync/semaphore.rs
```diff
@@ -47,28 +47,63 @@ use std::sync::Arc;
...
-///         join_handles.push(tokio::spawn(async move {
-///             // perform task...
-///             // explicitly own `permit` in the task
-///             drop(permit);
-///         }));
-///     }
+///         let (mut socket, _) = listener.accept().await?;
+///
+///         tokio::spawn(async move {
+///                 let mut buf = [0; 1024];
```
Darksonn, line 83: There's something wrong with the indentation here.

Darksonn (commented):
I think this is looking pretty good. The example could use a few more comments on the semaphore-related stuff.

tokio/src/sync/semaphore.rs
Darksonn: A `no_run` marker is fine.

tokio/src/sync/semaphore.rs
alexanderkirilin: I think the third option seems best. It gives you the basic structure in the main body, while still giving a useful "do work" example -- without cluttering everything. 

Plus we don't have to remember to explicitly drop the socket first this way.

tokio/src/sync/semaphore.rs
```diff
@@ -47,28 +47,63 @@ use std::sync::Arc;
...
 /// #[tokio::main]
-/// async fn main() {
+/// async fn main() -> Result<(), Box<dyn std::error::Error>> {
 ///     let semaphore = Arc::new(Semaphore::new(3));
-///     let mut join_handles = Vec::new();
+///     let listener = TcpListener::bind("127.0.0.1:8080").await?;
 ///
-///     for _ in 0..5 {
+///     loop {
 ///         let permit = semaphore.clone().acquire_owned().await.unwrap();
```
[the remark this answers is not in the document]
alexanderkirilin: Added.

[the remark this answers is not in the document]
Darksonn: The connection will wait either way, whether it's in the OS queue or in our program. The difference is when the socket starts taking up a file descriptor, and those are a limited resource.

(And I guess there's also the difference that if it's closed while in the OS queue, `accept` wont return it.)

tokio/src/sync/semaphore.rs
Darksonn: Makes sense. At that point, we might even entirely remove the `handle_connection` method from the example.

> Plus we don't have to remember to explicitly drop the socket first this way.

Hmm. It could make sense to do `handle_connection(&mut socket).await` to show explicitly dropping it in the example.

tokio/src/sync/semaphore.rs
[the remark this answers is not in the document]
alexanderkirilin: Oh! My mistake. I wasn't assuming 1 extra socket would make a difference -- but I remember now it does in memory-constrained environments. 

I'll fix that.

tokio/src/sync/semaphore.rs
alexanderkirilin: Yeah, at this point the details of handling the connection are irrelevant.

tokio/src/sync/semaphore.rs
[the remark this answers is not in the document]
Darksonn: It's not really about the memory, but the number of file descriptors.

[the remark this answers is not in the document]
alexanderkirilin: I've update the comment.

tokio/src/sync/semaphore.rs
alexanderkirilin: Removed the implementation of `handle_connection`; and so on.

tokio/src/sync/semaphore.rs
```diff
@@ -47,28 +47,44 @@ use std::sync::Arc;
...
+/// However, in contrast to the file example, this example uses a non-static, `Arc`-wrapped `Semaphore`
+/// for more fine-grained lifetime and ownership control. The `Arc` allows multiple threads and tasks
+/// to share ownership of the semaphore. Here we create a shallow-copy of the `Arc<Semaphore>` reference
+/// using `clone`. Then we acquire a permit through the `Arc` from the `Semaphore` via [`Semaphore::acquire_owned`],
+/// and move it inside the task. This ensures no non-`'static` variables are referenced from within
+/// said task.
+///
+/// If we leave the scope where the `Arc<Semaphore>` was defined, and references still exist in any
+/// of our tasks to it, the `Semaphore` will continue to exist in a static-like context, until all
+/// `Arc`s have been dropped.
```
Darksonn, line 65: I would probably explain this along these lines:

 1. This example uses an `Arc<Semaphore>` rather than a global. (You don't have to justify why.)
 2. We need to acquire permits before spawning because reasons.
 3. We need to drop the permit inside the task because reasons.
 4. To move the permit across the task boundary, we must use `acquire_owned`.
 5. If the semaphore was in a global, then we could use `acquire` instead, because that lets us get a `Permit<'static>`, which can also be moved across task boundaries, since the semaphore lives forever.

tokio/src/sync/semaphore.rs
```diff
@@ -47,28 +47,44 @@ use std::sync::Arc;
...
-///     }
+///         let (mut socket, _) = listener.accept().await?;
 ///
-///     for handle in join_handles {
-///         handle.await.unwrap();
+///         tokio::spawn(async move {
+///             // Do work, and drop socket after
+///             handle_connection(&mut socket).await;
+///             // Explicitly drop the permit, so more tasks can be created
+///             drop(permit);
```
Darksonn, line 86:
```suggestion
///             // Do work using the socket.
///             handle_connection(&mut socket).await;
///             // Drop socket while the permit is still live.
///             drop(socket);
///             // Drop the permit, so more tasks can be created.
///             drop(permit);
```

alexanderkirilin: I'll add it, but I was going to ask about this: if we move the reference in, wouldn't that not automatically drop the socket? 

For example, if we were going for an implicit `drop(socket)` at the end of our `handle_connection` wouldn't the right way to go about it be just `mut socket`?

tokio/src/sync/semaphore.rs
```diff
@@ -47,28 +47,44 @@ use std::sync::Arc;
 /// }
 /// ```
 ///
-/// Use [`Semaphore::acquire_owned`] to move permits across tasks:
+/// Limit number of incoming requests being handled at the same time.
 ///
-/// ```
+/// Similar to limiting the numer of simultaneous opened files, network handles are a limited resource.
+/// Allowing an unbounded amount of requests to be processed could result in a denial-of-service,
+/// and many other issues.
```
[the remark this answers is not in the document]
alexanderkirilin: Fixed the typo and moved this example underneath the file one.

tokio/src/sync/semaphore.rs
```diff
@@ -102,6 +76,54 @@ use std::sync::Arc;
...
+///             drop(socket);
+///             // Drop the permit, so more tasks can be created.
+///             drop(permit);
+///         });
+///     }
+/// }
+///
+/// # async fn handle_connection(socket: &mut TcpStream) {
+/// #   // Do work
+/// # }
```
Darksonn, line 124: The empty line here shows up in the generated documentation.
```suggestion
/// # async fn handle_connection(socket: &mut TcpStream) {
/// #   // Do work
/// # }
```

tokio/src/sync/semaphore.rs
```diff
@@ -102,6 +76,54 @@ use std::sync::Arc;
...
+/// use tokio::sync::Semaphore;
+/// use tokio::net::{TcpListener, TcpStream};
+///
+/// #[tokio::main]
+/// async fn main() -> Result<(), Box<dyn std::error::Error>> {
+///     let semaphore = Arc::new(Semaphore::new(3));
+///     let listener = TcpListener::bind("127.0.0.1:8080").await?;
+///
+///     loop {
+///         // Acquire permit before accepting, to avoid needlessly taking up a (finite) file descriptor
```
Darksonn, line 107: I don't think using finite like this is grammatically correct.
```suggestion
///         // Acquire permit before accepting the next socket.
```

tokio/src/sync/semaphore.rs
```diff
@@ -102,6 +76,54 @@ use std::sync::Arc;
 /// }
 /// ```
 ///
+/// Limit the number of incoming requests being handled at the same time.
+///
+/// Similar to limiting the number of simultaneously opened files, network handles
+/// are a limited resource. Allowing an unbounded amount of requests to be processed
+/// could result in a denial-of-service, among many other issues.
+///
+/// However, in contrast to the file example, this example uses an `Arc<Semaphore>`.
```
Darksonn, line 85:
```suggestion
/// This example uses an `Arc<Semaphore>` instead of a global variable.
```

tokio/src/sync/semaphore.rs
```diff
@@ -102,6 +76,54 @@ use std::sync::Arc;
...
+/// Before we spawn a new task to process a request, we must acquire a permit from
+/// this semaphore, in order to limit usage of system resources (like file descriptors,
+/// memory usage, CPU time, etc.). Once acquired, a new task is spawned; and once
```
Darksonn, line 88: This could be simplified to:

> To limit the number of requests that can be processed at the time, we acquire a permit for each task before spawning it.

tokio/src/sync/semaphore.rs
```diff
@@ -102,6 +76,54 @@ use std::sync::Arc;
...
+/// finished, the permit is dropped inside of the task to allow others to spawn.
+/// Permits must be acquired via [`Semaphore::acquire_owned`], in order to be
+/// movable across the task boundary. Contrastingly, if the semaphore were a global,
```
Darksonn, line 91: You can almost always replace "in order to" with "to".
```suggestion
/// Permits must be acquired via [`Semaphore::acquire_owned`] to be
/// movable across the task boundary. Contrastingly, if the semaphore were a global,
```

tokio/src/sync/semaphore.rs
```diff
@@ -102,6 +76,54 @@ use std::sync::Arc;
...
+/// as in the file-limiting example, only ['Semaphore::acquire'] would be necessary.
+/// This is because a `Permit<'static>` would be returned, which is movable across
+/// task boundaries, since the semaphore lives forever.
```
Darksonn, line 94: I know I suggested putting the part about `Permit<'static>`, but I changed my mind.
```suggestion
/// movable across the task boundary. (Since our semaphore is not a global variable — if it was, then `acquire` would be enough.)
```

Darksonn (commented):
It's looking good. I have a few comments to simplify the language a bit.

tokio/src/sync/semaphore.rs
Darksonn: You're not wrong. Taking ownership of `socket` would probably be nicer here, if this was real code. My motivation is that I think the example will be more helpful this way.

When people write this kind of code themselves, they will either:

 * Move the socket into another function.
 * Not move the socket into another function.

If they're basing their code on this example and moving the socket, then I'm sure that they will figure out that they don't need the `drop(socket)`. However, if they're not moving it into another function (e.g., maybe they process it directly in `tokio::spawn` like in the readme), then there's a good chance that they forget the `drop(socket)`, and I want the example to remind them.

tokio/src/sync/semaphore.rs
Darksonn: Also, the single quote in your `Semaphore::acquire` link is incorrect. Links should use the backtick character instead.

tokio/src/sync/semaphore.rs
alexanderkirilin: As far as I can tell, using Oxford Languages as a casual source, it is grammatically correct -- though it would be less flourishy if the parens were removed. 

See: 
finite-ex

Are we also removing the explanation/justification for acquiring a permit before accepting a socket? The document is always changing, so I understand if this fits better within the whole.

tokio/src/sync/semaphore.rs
alexanderkirilin: Yes, I need to improve on my technical writing.

alexanderkirilin:
@Darksonn Thank you for the suggestions. This looks much more concise and focused than what I had begun with; and I've learned a few things about technical writing by reflecting on them.

tokio/src/sync/semaphore.rs
Darksonn: Sorry, let me correct myself. I agree that it is grammatically correct, but I don't think using finite in this way has the meaning you intended.

I think you wanted to say that file descriptors are a finite resource, correct? However, you're saying that the file descriptor itself is finite, which is a different statement.

alexanderkirilin: You're right. It was a sloppy error on my part.

tokio/src/sync/semaphore.rs
```diff
@@ -102,6 +76,50 @@ use std::sync::Arc;
...
+/// [`Semaphore::acquire_owned`] to be movable across the task boundary.
+/// (Since our semaphore is not a global variable — if it was, then `acquire` would be enough.)
+///
+/// ```no_run
+/// use std::sync::Arc;
+/// use tokio::sync::Semaphore;
+/// use tokio::net::{TcpListener, TcpStream};
+///
+/// #[tokio::main]
+/// async fn main() -> Result<(), Box<dyn std::error::Error>> {
```
Darksonn, line 99: I don't like using `Box<dyn Error>` in examples. The type is not `Send`, so it can't really be used outside of the main function, since `tokio::spawn` requires the task to be `Send`.
```suggestion
/// async fn main() -> std::io::Result<()> {
```

tokio/src/sync/semaphore.rs
```diff
@@ -102,6 +76,50 @@ use std::sync::Arc;
...
+/// we acquire a permit for each task before spawning it. Once acquired,
+/// a new task is spawned; and once finished, the permit is dropped inside
+/// of the task to allow others to spawn. Permits must be acquired via
+/// [`Semaphore::acquire_owned`] to be movable across the task boundary.
+/// (Since our semaphore is not a global variable — if it was, then `acquire` would be enough.)
+///
+/// ```no_run
+/// use std::sync::Arc;
+/// use tokio::sync::Semaphore;
+/// use tokio::net::{TcpListener, TcpStream};
```
Darksonn, line 96: The `TcpStream` type is not visible in the example. I would use the full path in the declaration of `handle_connection` instead.
```suggestion
/// use tokio::net::TcpListener;
```

tokio/src/sync/semaphore.rs
```diff
@@ -102,6 +76,50 @@ use std::sync::Arc;
...
+///
+/// #[tokio::main]
+/// async fn main() -> Result<(), Box<dyn std::error::Error>> {
+///     let semaphore = Arc::new(Semaphore::new(3));
+///     let listener = TcpListener::bind("127.0.0.1:8080").await?;
+///
+///     loop {
+///         // Acquire permit before accepting the next socket.
```
Darksonn, line 104:
```suggestion
///         // Acquire permit before accepting the next socket.
///         //
///         // We use `acquire_owned` so that we can move `permit` into
///         // other tasks.
```

Darksonn (commented):
I think these suggestions are the last ones. Then I'm happy to merge this.

Darksonn (approved):
Thanks! Looks good to me.

alexanderkirilin:
@Darksonn Merged latest commits from upstream. Would this be able to be merged?

Darksonn:
Sorry, seems like I forgot to press the button.

PR #5961, opened 2023-08-29
[feature] Add check for  LinesCodec when input contains '\n'

arthur-zhang:
## Motivation

we should check weather input contains '\n'. For example, if we encode "abc\n", we get "abc\n\n".
But when we decode, we get "abc"


wathenjiang:
Unfortunately the docs of `LinesCodec` do not specify which line delimiter it use. It currently use `\n` as line delimiter.
Different OSs may use differenct line delimiter :).
We may specify it clearly, and support more line delimiter, such as `\r`.

PR #5965, opened 2023-08-30
Cargo: Fix minimum Rust version

progval:
## Motivation

Building on v1.63 fails with:

```
error[E0658]: use of unstable library feature 'process_set_process_group'
   --> ~/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.29.0/src/process/mod.rs:777:18
    |
777 |         self.std.process_group(pgroup);
    |                  ^^^^^^^^^^^^^
    |
    = note: see issue #93857 <https://github.com/rust-lang/rust/issues/93857> for more information
```

## Solution

Raised the minimum rust version to v1.64, which compiles successfully

### DIFF
--- a/tokio/Cargo.toml
+++ b/tokio/Cargo.toml
@@ -8,7 +8,7 @@ name = "tokio"
 # - Create "v1.x.y" git tag.
 version = "1.32.0"
 edition = "2021"
-rust-version = "1.63"
+rust-version = "1.64"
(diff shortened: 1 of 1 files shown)


taiki-e:
That function is under tokio_unstable, and MSRV does not apply to unstable APIs.

https://github.com/tokio-rs/tokio/blob/37bb47c4a2aff8913e536767645772f15650e6cd/tokio/src/process/mod.rs#L766-L772

PR #5971, opened 2023-09-01
SyncIOBridge: Add `into_inner`

cgwalters:
This keeps us in sync with other I/O wrapper types (e.g. `BufWriter`).

I was working on this change
https://github.com/ostreedev/ostree-rs-ext/pull/527/commits/13455cce4923ac0c3237d3983c2d4044842fb139 and it would have been helpful to be able to call this API so I can return the input stream back to the caller in order to extend its lifetime.


Darksonn (approved):
Thanks.

PR #5972, opened 2023-09-01
tokio: render taskdump documentation on docs.rs

jswrenn:
Taskdump documentation is currently not rendered by docs.rs. This PR modifies `package.metadata.docs.r` so that `--cfg tokio_taskdump` is used by docs.rs when building documentation.

Ref. #5638.


Darksonn (approved):
Thanks.

PR #5980, opened 2023-09-04
chore: list 1.32.x as LTS release

hds:
List that 1.32.x will be supported until September 2024. The 1.20.x LTS
which is supported until September 2023 has been left, to be removed in
October.


Darksonn (approved):
Thanks

Issue #5983, opened 2023-09-05
DuplexStream does not support vectored writes


qrnch-jan:
When `write_vectored` is used on a `DuplexStream` it only writes the first buffer in the scatter/gather list.

It would be nice if it wrote as much as it can.

The following code illustrates the problem:

```rust
use std::io::IoSlice;
use bytes::Bytes;
use tokio::io::AsyncWriteExt;

#[tokio::main]
async fn main() {
    let bufs = vec![Bytes::from("hello,"), Bytes::from(" world")];
    let (mut ep1, _ep2) = tokio::io::duplex(64*1024);
    let iov: Vec<IoSlice> = bufs.iter().map(|b| IoSlice::new(b)).collect();
    println!("iov.len={}", iov.len());
    let total_size = iov.iter().fold(0, |acc, e| acc + e.len());
    println!("total size: {}", total_size);
    let written = ep1.write_vectored(&iov[..]).await.unwrap();
    println!("wrote {} bytes", written)
}
```

The program outputs that it has written 6 bytes, rather than the expected 12.

PR #5985, opened 2023-09-05
implement write_vectored for DuplexStream

maminrayej:
The `poll_write_vectored` function for `DuplexStream` currently uses the default implementation in `AsyncWrite`, which only writes the first non-empty buffer in `bufs`. This PR adds proper vectored write support for `DuplexStream`.

## Solution

Instead of only copying the first non-empty buffer, this implementation tries to fill all available space in the internal buffer.

Resolves Issue #5983.


Darksonn (approved):
Thanks. Looks good to me.

PR #5997, opened 2023-09-10
Add `?Sized` bound for `MutexGuard::map` and `OwnedMutexGuard::map`.

ennis:
## Motivation

I can't return a reference to an unsized type in `MutexGuard::map` and `OwnedMutexGuard::map` because of an implicit `Sized` bound in their signature.

## Solution

Add a `?Sized` bound to the signatures of the two methods. It seems safe at a glance (the mapped guard types already have `?Sized`), but there might be some subtlety somewhere that I overlooked.


Darksonn (approved):
This looks good to me.

PR #5999, opened 2023-09-11
fix ci

maminrayej:
See #5960 for more info.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,7 +220,7 @@ jobs:
         with:
           toolchain: ${{ env.rust_nightly }}
           override: true
-      - uses: Swatinem/rust-cache@v1
+      # - uses: Swatinem/rust-cache@v1
       - name: asan
         run: cargo test --workspace --all-features --target x86_64-unknown-linux-gnu --tests -- --test-threads 1
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,9 +71,7 @@ jobs:
         run: rustup update stable
       - uses: Swatinem/rust-cache@v1
       - name: Install cargo-hack
-        uses: taiki-e/install-action@v2 
-        with: 
-            tool: cargo-hack@0.3.0
+        run: cargo install cargo-hack --version 0.3.0
 
       # Run `tokio` with `full` features. This excludes testing utilities which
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -71,7 +71,7 @@ jobs:
         run: rustup update stable
       - uses: Swatinem/rust-cache@v1
       - name: Install cargo-hack
-        run: cargo install cargo-hack --version 0.5.26
+        run: cargo install cargo-hack --version 0.3.0
```
taiki-e, line 74: I would recommend using install-action like master does.

https://github.com/tokio-rs/tokio/blob/65e7715909694d8d7d15b081cc3790b4ea1ad300/.github/workflows/ci.yml#L165-L168

maminrayej: I replaced:
```
run: cargo install cargo-hack --version 0.3.0
```
with
```
uses: taiki-e/install-action@v2 
with: 
  tool: cargo-hack@0.3.0 
```
some jobs failed. I believe the reason was this warning:
```
Warning: cargo-hack@0.3.0 for 'linux' is not supported; fallback to cargo-binstall
info: install-action does not support cargo-hack@0.3.0; fallback to cargo-binstall
```
So I decided to revert back to installing it using `cargo install`.

taiki-e: > tool: cargo-hack@0.3.0 

Removing '@0.3.0' should fix the error.

maminrayej: But I need the 0.3.0 version. It's the latest version that works.

taiki-e: Why 0.3.0 is needed?

taiki-e: To be clear: cargo-hack is also compatible with much older versions of cargo than required for installation.

https://github.com/taiki-e/cargo-hack?tab=readme-ov-file#from-source

> cargo-hack is usually runnable with Cargo versions older than the Rust version required for installation (e.g., `cargo +1.31 hack check`). Currently, to run cargo-hack requires Cargo 1.26+.

maminrayej: You're right. I mistakenly thought the action compiles `cargo-hack`. I've made the necessary changes.

taiki-e (approved):
Thanks.

PR #6000, opened 2023-09-11
util: replace `sync::reusable_box::Pending` with `std::future::Pending`

victor-timofei:
Now that MSRV is 1.63 we can remove our `Pending` imlementation.


taiki-e (approved):
Thanks!

Issue #5979, opened 2023-09-04
Proposal: use criterion crate instead of bencher crate in benchmark


wathenjiang:
The current benchmark crate is [bencher](https://crates.io/crates/bencher), but the [criterion](https://crates.io/crates/criterion) is more modern and being maintained, which can be found [here](https://github.com/tokio-rs/tokio/blob/95fb599664eb0dc56251b59a2d746a3ec25de58e/benches/Cargo.toml#L12).

I would like to do more benchmarks, but I found the API of [bencher](https://crates.io/crates/bencher) is kind of poor, such as the lack of [iter_custom](https://docs.rs/criterion/latest/criterion/struct.Bencher.html#method.iter_custom) in criterion.

If this proposal can be passed, I would like to help compelete it.

PR #6001, opened 2023-09-12
reduce the lock contention in task spawn.

wathenjiang:
We have to use a OwnedTasks to track the pointers of all Tasks, this is a multi-threaded scenario. When the Multi-threaded concurrent `tokio::spawn` is called, the lock contention would be very busy. 

I propose a solution without losing any single-threaded spawn task performance, to improve the multi-threaded spawn task performance aboout `15% ~ 20%`.

This solution is that: let tasks with different task ID to be put into the different LinkedList. The different LinkedList use own Mutex to make sure multi-threaded concurrency safety instead of a runtime unique Mutex.

The origin version:
```
spawn_tasks_current_thread
                        time:   [654.87 ns 664.26 ns 672.13 ns]
                        change: [-4.5814% -1.2511% +2.1604%] (p = 0.48 > 0.05)

spawn_tasks_current_thread_parallel
                        time:   [537.55 ns 540.00 ns 542.20 ns]
                        change: [-1.0603% -0.2591% +0.6283%] (p = 0.58 > 0.05)
  1 (1.00%) high severe

spawn_tasks             time:   [1.0447 µs 1.0533 µs 1.0610 µs]
                        change: [+0.9379% +2.2768% +3.7191%] (p = 0.00 < 0.05)
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low severe
  2 (2.00%) low mild

spawn_tasks_parallel    time:   [992.38 ns 1.0002 µs 1.0073 µs]
                        change: [+0.1973% +1.4194% +2.5819%] (p = 0.02 < 0.05)
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low severe
  2 (2.00%) low mild
  1 (1.00%) high mild
  2 (2.00%) high severe
```

This version:

```
spawn_tasks_current_thread
                        time:   [705.92 ns 724.31 ns 739.78 ns]

spawn_tasks_current_thread_parallel
                        time:   [529.33 ns 531.00 ns 532.61 ns]
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

spawn_tasks             time:   [881.56 ns 892.21 ns 902.10 ns]
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) low mild
  1 (1.00%) high mild

spawn_tasks_parallel    time:   [815.00 ns 819.87 ns 824.60 ns]
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe
```


carllerche:
Thanks for doing this. I tried something similar but did not measure an improvement. What CPU are you running this on?

wathenjiang:
> Thanks for doing this. I tried something similar but did not measure an improvement. What CPU are you running this on?

The above benchmark test runs on the Linux, the CPU is `Intel(R) Xeon(R) Gold 6133 CPU @ 2.50GHz`. 

I also run the same benchmark test on other cpu architecture like ARM (mac m1), I found the silimar improvement too.

The lock contention problem would be heavier, when we use more threads to spwan tasks concurrently. Do you make sure the `task::spawn` is doing on diffrenet worker threads concurrently?

The reason why I propose to use criterion crate in https://github.com/tokio-rs/tokio/issues/5979 is that we could do conrrent jobs more conveniently.


```rust
async fn job(iters: usize, procs: usize) {
    for _ in 0..procs {
        let mut threads_handles = Vec::with_capacity(procs);
        threads_handles.push(tokio::spawn(async move {
            let mut thread_handles = Vec::with_capacity(iters / procs);
            for _ in 0..iters / procs {
                thread_handles.push(tokio::spawn(async {
                    let val = 1 + 1;
                    black_box(val)
                }));
            }
            for handle in thread_handles {
                handle.await.unwrap();
            }
        }));
        for handle in threads_handles {
            handle.await.unwrap();
        }
    }
}
```

We can get `iters` now, which is not reachable in bencher crate. This means we can make sure the number of `tokio::spawn` is same, when the way `tokio::spawn` changes(single thread or multi-thread concurrency).

The falmegraph told the same improvement

The origin version:
image

This version:

image

tokio/src/runtime/task/core.rs
```diff
@@ -168,11 +168,14 @@ pub(crate) struct Header {
...
     /// removed from the list.
-    pub(super) owner_id: UnsafeCell<Option<NonZeroU64>>,
+    pub(super) owner_id: UnsafeCell<Option<NonZeroU32>>,
 
     /// The tracing ID for this instrumented task.
     #[cfg(all(tokio_unstable, feature = "tracing"))]
     pub(super) tracing_id: Option<tracing::Id>,
+
+    /// The task's ID, is used for deciding which list to put it in.
+    pub(super) task_id: Id,
```
Darksonn, line 178: We already store the task id. It's in the core struct.

tokio/src/runtime/task/list.rs
```diff
@@ -111,25 +111,27 @@ impl<S: 'static> OwnedTasks<S> {
...
+        self.push_inner(id, task);
+        Some(notified)
+    }
+
+    #[inline]
+    fn push_inner(&self, task_id: super::Id, task: Task<S>) {
+        let mut lock = self.lists[task_id.0 as usize & (self.grain - 1)].lock();
+        lock.push_front(task);
+        self.count.fetch_add(1, Ordering::Relaxed);
     }
```
Darksonn, line 128: This doesn't look right to me. The `OwnedTasks` could be closed between the call to `self.closed.load(Acquire)` and the call to `self.lists[...].lock()`. If that happens, then we push something to the list, but it's never removed because the close operation has already completed.

Darksonn: You must check the `closed` flag _after_ calling `lock()` to fix this.

tokio/src/runtime/task/core.rs
Darksonn: Does the `header_lte_cache_line` test still pass when you put it here?

wathenjiang: > header_lte_cache_line

Yes. I reduce the size of `owner_id` to u32, I believe this size for owner_id is big enough. And I noticed that the `#[repr(C)]` is used, current struct fields 


I noticed that the tag a is used

tokio/src/runtime/task/list.rs
Darksonn: No, I don't agree that this is correct. If I spawn something from outside the runtime during shutdown, this problem can result in the task being pushed to the `OwnedTasks`, but not removed by the `close_and_shutdown_all` call.

wathenjiang: Does the `Runtime.enter()` will cause such concurrency problem?

Darksonn: You can spawn tasks during shutdown with `tokio::runtime::Handle`.

However, it doesn't really matter. Even if the API prevents you from triggering this, we should still fix it. Having correctness rely on such things makes it very easy to introduce bugs when changing things.

wathenjiang: In this cases, If a thread entered the runtime, then it become a legal worker thread.

This thread shoud be responsible for cleaning tasks, when the`EnterGuard` is dropped, let it check if the runtime is closed, if it is true, then cleaning tasks.

Does it make sense?

wathenjiang: > You can spawn tasks during shutdown with `tokio::runtime::Handle`.
> 
> However, it doesn't really matter. Even if the API prevents you from triggering this, we should still fix it. Having correctness rely on such things makes it very easy to introduce bugs when changing things.

I just saw this. What I said above may be a solution, but it may be more appropriate to directly use locks to solve the problem. I think your solution is more reasonable, let it easy.

Darksonn: I see you moved the check into the lock, so I'll resolve this.

tokio/src/runtime/task/list.rs
```diff
@@ -73,13 +58,23 @@ struct OwnedTasksInner<S: 'static> {
...
 
 impl<S: 'static> OwnedTasks<S> {
-    pub(crate) fn new() -> Self {
+    /// grain must be an integer power of 2
+    pub(crate) fn new(grain: u32) -> Self {
+        assert_eq!(
+            grain & (grain - 1),
+            0,
+            "the grain of OwnedTasks must be an integer power of 2"
+        );
```
hawkw, line 67: it looks like `OwnedTasks::new` is only ever called with the values 1 and 16. What do you think about making this constructor private, and having the `pub(crate)` constructors be `new_current_thread` and `new_multi_thread` or something? that way, the value (which feels like an internal implementation detail) isn't leaked to the places where the `OwnedTasks` is constructed, and we can change the value in one place, rather than having to update _every_ call to `OwnedTasks::new`?

if we did that, we could probably also remove the assertion, or change it to a `debug_assert!`, since the caller can't pass in a value that isn't a power of two.
```suggestion
    pub(crate) fn new_current_thread() -> Self {
        Self::new(1)
    }
    
    pub(crate) fn new_multi_thread() -> Self {
        Self::new(16)
    }

    /// grain must be an integer power of 2
    fn new(grain: u32) -> Self {
        debug_assert_eq!(
            grain & (grain - 1),
            0,
            "the grain of OwnedTasks must be an integer power of 2"
        );
```

tokio/src/runtime/task/list.rs
```diff
@@ -111,25 +111,36 @@ impl<S: 'static> OwnedTasks<S> {
...
+    pub(crate) fn push_inner(
+        &self,
+        task_id: super::Id,
+        task: Task<S>,
+        notified: Notified<S>,
+    ) -> Option<Notified<S>>
+    where
+        S: Schedule,
+    {
+        let mut lock = self.lists[task_id.0 as usize & (self.grain - 1) as usize].lock();
```
hawkw, line 127: this code for indexing the set of task lists by a task ID is somewhat complex, and it's repeated in a few places. what do you think about potentially adding a function that takes a task ID and returns a `MutexGuard` so that we don't have to repeat this?

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -287,7 +287,7 @@ pub(super) fn create(
             remotes: remotes.into_boxed_slice(),
             inject,
             idle,
-            owned: OwnedTasks::new(),
+            owned: OwnedTasks::new(16),
```
hawkw, line 290: how was 16 chosen as the value for this? i see that you ran the benchmarks on a Xeon Gold 6133, which has 16 CPU cores...would it be more correct to determine this value based on `num_cpus` to ensure that it's always the number of CPU cores, regardless of which CPU is in use?

hawkw (commented):
thanks for working on this, and for sharing the benchmark results! the performance improvement looks promising. i had a few questions, and some minor suggestions.

tokio/src/runtime/scheduler/multi_thread/worker.rs
wathenjiang: I run on a `Intel(R) Xeon(R) Gold 6133 CPU @ 2.50GHz.` , which has 8 CPU cores. The reason I choose 16 is for performance testing, and this size can be discussed further.

Actually I decided to make this size twice the number of CPU cores, because the number of worker threads in tokio is the number of CPU core. Meanwhile, I want it to be a integer power of 2.

So we could just set it to a big enough vlaue for most platform, like 64, and let people to configure this. But it might cause CPU cache problem.

Or, set it to an integer power of two that is at least twice the number of CPU cores.

tokio/src/runtime/task/list.rs
wathenjiang: Yes, this is duplicate code, I have move the way of get list by task id into the method `segment_inner`.

tokio/src/runtime/task/list.rs
wathenjiang: I would make it configurable in `Builder` of the Runtime.

tokio/src/runtime/task/list.rs
```diff
@@ -24,47 +26,30 @@ use std::num::NonZeroU64;
...
-    inner: Mutex<CountedOwnedTasksInner<S>>,
-    pub(crate) id: NonZeroU64,
-}
-struct CountedOwnedTasksInner<S: 'static> {
-    list: CountedLinkedList<Task<S>, <Task<S> as Link>::Target>,
-    closed: bool,
+    lists: Box<[Mutex<ListInner<S>>]>,
+    pub(crate) id: NonZeroU32,
+    closed: AtomicBool,
+    pub(crate) grain: u32,
```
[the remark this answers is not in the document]
wathenjiang: Yes, I've renamed it `segment_size` which descibes the segment lock design better.

tokio/src/runtime/task/core.rs
wathenjiang: We could consider removing the task_id from the Core struct and, when a task is selected into Harness for execution, pass the task_id in the header to the Core method by parameter.

I'm not sure if this is the best approach.

tokio/src/runtime/task/list.rs
```diff
@@ -179,22 +184,46 @@ impl<S: 'static> OwnedTasks<S> {
...
+    unsafe fn remove_inner(&self, task: &Task<S>) -> Option<Task<S>> {
+        let mut lock = self.segment_inner(task.task_id() as usize);
+        match lock.remove(task.header_ptr()) {
+            Some(task) => {
+                drop(lock);
+                self.count.fetch_sub(1, Ordering::Relaxed);
+                Some(task)
+            }
+            None => None,
+        }
```
hawkw, line 200: style nit, take it or leave it: i think we can simplify this code to just
```suggestion
        let task = self.segment_inner(task.task_id() as usize).remove(task.header_ptr())?;
        self.count.fetch_sub(1, Ordering::Relaxed);
        Some(task)
```

tokio/src/runtime/task/list.rs
```diff
@@ -140,34 +146,33 @@ impl<S: 'static> OwnedTasks<S> {
...
+                let task = match lock.pop_back() {
+                    Some(task) => {
+                        drop(lock);
+                        self.count.fetch_sub(1, Ordering::Relaxed);
+                        task
+                    }
+                    None => break,
+                };
+                task.shutdown();
+            }
```
hawkw, line 170: style nit: i think we can simplify this to
```suggestion
            while let Some(task) = self.segment_inner(i).pop_back() {
                self.count.fetch_sub(1, Ordering::Relaxed);
                task.shutdown();
            }
```
since the `lock` is dropped at the end of the expression `self.segment_inner(i).pop_back()`. of course, if you think it improves readability to make the duration of the mutex lock more explicit, we could also write
```suggestion
            loop {
                let mut lock = self.segment_inner(i);
                match lock.pop_back() {
                    Some(task) => {
                        drop(lock);
                        self.count.fetch_sub(1, Ordering::Relaxed);
                        task.shutdown();
                    }
                    None => break,
                };
            }
```
but IMO, the `while` loop is a bit nicer. it's up to you, though.

hawkw (commented):
it looks like the `loom` tests are failing on this branch because the use of task IDs for indexing the task lists has introduced nondeterminism into the tests: https://github.com/tokio-rs/tokio/actions/runs/6276681507/job/17046907750?pr=6001#step:5:294

this is because task IDs are generated by a non-`loom` atomic (and it can't be a loom atomic because it lives in a `static`). previously, task IDs didn't control actual branching behavior that occurred within the loom tests, so generating them non-deterministically was fine., however, the task ID may differ based on the order in which the `loom` tests run. now that we actually use them to determine which list to access, they now result in non-deterministic branching inside of the loom tests.

i think we'll need to change how task IDs are generated when `#[cfg(loom)]` is set to ensure these tests are still deterministic.

tokio/src/runtime/task/list.rs
wathenjiang: Yes! We use `?` here to resolve `None` better, but I would like to drop mutex manully here.

tokio/src/runtime/task/list.rs
wathenjiang: When using `while let Some(task) = self.segment_inner(i).pop_back() {`, we encounter the undesirable effect of holding the mutex for a longer time. The MutexGuard won't be released until the end of an iteration of the while loop.

The `task.shutdown` operation should be performed when we do not hold the lock.

wathenjiang: I would like to take the second suggestion, but I made the mistake of taking the first suggestion so I chose to submit the code myself instead of adding suggestion to batch.

wathenjiang:
Hey, do you have some suggetions about resolving this loom test problem?

The simplest way I can think of is that task_id is a constant or we always limit the number of OwnedTasks in lists to 1, or any other way to make all tasks always put into single one list, when #[cfg(loom)] is set. But I am not sure that is loom test want.

tokio/src/runtime/task/list.rs
hawkw: > When using while let Some(task) = self.segment_inner(i).pop_back() {, we encounter the undesirable effect of holding the mutex for a longer time. The MutexGuard won't be released until the end of an iteration of the while loop.

hmm, i thought that the MutexGuard would be dropped at the end of the `self.segment_inner(i).pop_back()` expression, so the mutex is locked for the same period of time as in the current code. but, if that's not actually the case, i agree that the current code is better, since the critical section is only as long as it takes to pop the task.

would introducing an explicit scope in the `while` loop also have that effect? like this:
```rust
while let Some(task) = { self.segment_inner(i).pop_back() } {
    // ...
}
```
all locals to the scope (like the guard) should be dropped as soon as the scope is evaluated, right?

hawkw: in any case, i don't have a problem with the current code as-is, i'm just curious about whether the while loop would have similar behavior or not.

wathenjiang: The lifetime of `MutexGuard` is as follows:
```rust
while let Some(task) = self.segment_inner(i).pop_back() { // get MutexGuard here
      self.count.fetch_sub(1, Ordering::Relaxed);
      let _lock = self.segment_inner(i); // dead lock here
      task.shutdown();
}// drop MutexGuard here
```
The dead lock line reflects the lifetime of `MutexGuard`.

I belive the auto drop feature of Rust is very useful for avoid unexpected resources not being released, but in order to achieve the precise lock control, the better way is that we drop lock manully like other languages which has no auto drop.

wathenjiang: The following does not change the lifetime of `MutexGuard` at all.
```
while let Some(task) = { self.segment_inner(i).pop_back() } {
    // ...
}
```
- we create `MutexGuard` in ` { self.segment_inner(i).pop_back() } ` code block
- and then move it from the block
- the `MutexGuard` still will not be dropped until th end of an iteration of this while loop.

This behavior of `MutexGuard in while let` also can be found in https://stackoverflow.com/questions/58968488/why-is-this-mutexguard-not-dropped

wathenjiang:
I want to run the loom tests local by `cargo test --lib --release --features full -- --nocapture loom_current_thread`, and the env variables:
```
export RUSTFLAGS="-Dwarnings --cfg loom --cfg tokio_unstable -C debug_assertions"
export LOOM_MAX_PREEMPTIONS=2
export LOOM_MAX_BRANCHES=10000
export RUST_BACKTRACE=1
export rust_stable=stable
export CARGO_INCREMENTAL=0
export ARGO_TERM_COLOR=always
export CACHE_ON_FAILURE=false
```
But got the following errors:
```
   Compiling tokio v1.32.0 (/root/rustproject/fork-tokio/tokio/tokio)
   Compiling tokio-util v0.7.9 (/root/rustproject/fork-tokio/tokio/tokio-util)
error[E0432]: unresolved import `tokio::net`
 --> tokio-util/src/udp/frame.rs:4:26
  |
4 | use tokio::{io::ReadBuf, net::UdpSocket};
  |                          ^^^ could not find `net` in `tokio`

error[E0433]: failed to resolve: could not find `net` in `tokio`
 --> tokio-util/src/net/unix/mod.rs:7:26
  |
7 | impl Listener for tokio::net::UnixListener {
  |                          ^^^ could not find `net` in `tokio`

error[E0433]: failed to resolve: could not find `net` in `tokio`
 --> tokio-util/src/net/unix/mod.rs:8:22
  |
8 |     type Io = tokio::net::UnixStream;
  |                      ^^^ could not find `net` in `tokio`

error[E0433]: failed to resolve: could not find `net` in `tokio`
 --> tokio-util/src/net/unix/mod.rs:9:24
  |
9 |     type Addr = tokio::net::unix::SocketAddr;
  |                        ^^^ could not find `net` in `tokio`

error[E0433]: failed to resolve: could not find `net` in `tokio`
  --> tokio-util/src/net/mod.rs:34:26
   |
34 | impl Listener for tokio::net::TcpListener {
   |                          ^^^ could not find `net` in `tokio`

error[E0433]: failed to resolve: could not find `net` in `tokio`
  --> tokio-util/src/net/mod.rs:35:22
   |
35 |     type Io = tokio::net::TcpStream;
   |                      ^^^ could not find `net` in `tokio`

Some errors have detailed explanations: E0432, E0433.
For more information about an error, try `rustc --explain E0432`.
error: could not compile `tokio-util` (lib) due to 6 previous errors
```

tokio/src/runtime/task/core.rs
wathenjiang: Why don't we expand using the existing `owner_id` field in `Header` struct instead of moving the task_id from `Core` to `Header`?

What we should do is that owner_id is still 64 bits(in current version I change it to u32), but now its lower 32 bits are the index of lists, and the upper 32 bits are the unique ID of ownedTasks. Of cause, we can use another u32 field to store the index, and also use u32 to store owner_id instead of u64.

When we need to remove the task from OwnedTasks, we use the lower 32 bits of the `owner_id` to find the list store it, we don't do caculating for index every time, it already caculated.

It can decouple index implementation details(which depends on task_id) from `Header` struct. Header owns the index of list have the tasks, instead of having the data to calculate index.

@Darksonn Do you think my above thought is more reasonable than the current PR design?

hawkw:
> Hey, do you have some suggetions about resolving this loom test problem?
> 
> The simplest way I can think of is that task_id is a constant or we always limit the number of OwnedTasks in lists to 1, or any other way to make all tasks always put into single one list, when #[cfg(loom)] is set. But I am not sure that is loom test want.
> 
> Or use [Lazy::new](https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html#example) to alllow AtomicU64 in loom can be used in generating task_id?

Hi, sorry, I meant to reply to this earlier but I was quite busy.

I think one option would be to just continue using the stdlib `AtomicU64` for generating task IDs in `loom`, but store the task ID counter in a `thread_local!` rather than in a `static`, _only while `cfg(loom)` is set_. This way, each loom test (running in its own thread) gets a totally fresh task ID counter, making them deterministic.

> I want to run the loom tests locally by `cargo test --lib --release --features full -- --nocapture loom_current_thread`, and the env variables:

Hmm, I think the problem is that you're running `cargo test` in the repo root directory, which also tries to compile `tokio-util` and run its tests. But, `tokio-util` depends on `tokio` APIs that aren't available with `cfg(loom)`. I think you should be able to `cd` into the `tokio` crate directory and then it should work?

wathenjiang:
Thanks.

After `cd tokio`, the loom tests can run on my local machine.

And in loom tests, making `task_id_counter` in thread local can make them deterministic.

I found `crate::loom::sync::atomic::AtomicBool` is just the alias type of `std::sync::atomic::AtomicBool`, but I do use `AtomicBool` to make a happens-before relationship with `count: AtomicUsize` in OwnedTasks. So I make the type of `count` is `std::sync::atomic::AtomicUsize` for avoid loom tests of this happens-before relationship undeterministic :

```rust
debug_assert!(self.shared.owned.is_shutdown());
debug_assert!(self.shared.owned.is_empty());
```

And finally all the tests passed. This is partly because I am not familiar with the tokio testing system :)

And as I said above, I make the segment index to be stored in `owner_id`, it will be calculated depending on task_id when the task created. After that, we will not calculate segment index again, we just use the segment index directly which is read from `owner_id`.

tokio/src/runtime/task/core.rs
```diff
@@ -157,8 +157,8 @@ pub(crate) struct Header {
     /// Table of function pointers for executing actions on the task.
     pub(super) vtable: &'static Vtable,
 
-    /// This integer contains the id of the OwnedTasks or LocalOwnedTasks that
-    /// this task is stored in. If the task is not in any list, should be the
+    /// The lower 32 bits of this value is the id of OwnedTasks, and higher 32 bits of this value is
+    /// the segment id of inner segment list of OwnedTasks. If the task is not in any list, should be the
```
Darksonn, line 161: Why not just recompute this every time you access the list? I think that would simplify the code.

Is the problem with being able to access the task id inside `list.rs`?

tokio/src/runtime/task/id.rs
```diff
@@ -74,11 +74,17 @@ impl fmt::Display for Id {
 
 impl Id {
     pub(crate) fn next() -> Self {
-        use crate::loom::sync::atomic::{Ordering::Relaxed, StaticAtomicU64};
+        #[cfg(loom)]
+        return Self(context::next_task_id().unwrap());
```
Darksonn, line 78: I think what you can do is use loom's lazy_static wrapper, which you can read about at https://docs.rs/loom/latest/loom/macro.lazy_static.html

This avoids a problem where tasks created in different threads get the same task id, since thread locals have a separate counter per thread.

By using the lazy_static wrapper, you should be able to put a `loom::sync::atomic::AtomicU64` in a global.

Darksonn (commented):
> I found `crate::loom::sync::atomic::AtomicBool` is just the alias type of `std::sync::atomic::AtomicBool`

Not when we run loom tests. In that case, it's an alias of `loom::sync::atomic::AtomicBool` instead.

tokio/src/runtime/task/core.rs
wathenjiang: It is because if we use task_id to calculate the index every time, we have to store index in Header struct. But Core struct also relies on task_id to generate panic and some logic in TaskIdGuard, which means we either store a duplicate copy of task_id in Core(just like header), or pass in task_id as a parameter when calling Core-related methods.

The reason for this is not to avoid calculating the index every time, but so that thead Header struct does not need to know the task_id. 

Header knows the index and OwnedTasks, Core konws the scheduler and task_id. What do you think of it?

tokio/src/runtime/task/id.rs
wathenjiang: I have no tests on it yet, but as far as I know, the lazy_static has a little performance loss, do you think it is suitable for generating global ID?

tokio/src/runtime/task/core.rs
wathenjiang: Depend on the current design, when we got a task in `list.rs`, we can access its Header, and have no ideas about Core, the Core is used by the scheduler.

tokio/src/runtime/task/id.rs
Darksonn: You should only use lazy_static during loom tests, so this does not impact performance when not using loom.

wathenjiang: I get you pointer, I will try this way.

tokio/src/runtime/task/core.rs
wathenjiang: It definitely works! So I will use this method to caculate the index.

tokio/src/runtime/task/id.rs
wathenjiang: But when I try this way,
```
  #[cfg(loom)]
        {
            crate::loom::lazy_static! {
                static ref NEXT_ID: StaticAtomicU64 = StaticAtomicU64::new(1);
            }            
            Self(NEXT_ID.fetch_add(1, Relaxed)) 
        }
```

 this following error occurs:

```
error[E0433]: failed to resolve: could not find `lazy_static` in `loom`
  --> tokio/src/runtime/task/id.rs:84:26
   |
84 |             crate::loom::lazy_static! {
   |                          ^^^^^^^^^^^ could not find `lazy_static` in `loom`
```

wathenjiang: Oh may be we should use `[cfg(not(all(test, loom)))]` and `#[cfg(all(test, loom))]` instead.
`

wathenjiang: I  have use `lazy_static` in loom instead.

tokio/src/runtime/task/core.rs
wathenjiang: I have use `Header::get_id` method to calculate the index, and it works well.

tokio/src/runtime/builder.rs
```diff
@@ -401,6 +409,51 @@ impl Builder {
...
+    ///     .unwrap();
+    ///
+    /// rt.spawn(async move {});
+    /// ```
+    ///
+    /// # Panics
+    ///
+    /// This will panic if `val` is not larger than `0`.
+    #[track_caller]
+    pub fn spawn_concurrency_level(&mut self, mut val: usize) -> &mut Self {
```
Darksonn, line 448: It's not clear to me that we want a configuration option for this. At least, it should be unstable.

tokio/src/runtime/task/list.rs
```diff
@@ -140,34 +170,39 @@ impl<S: 'static> OwnedTasks<S> {
...
+                        drop(lock);
+                        self.count.fetch_sub(1, Ordering::Relaxed);
+                        task.shutdown();
+                    }
+                    None => break,
+                };
+            }
         }
+        // we have shut down all tasks
+        self.shutdown.store(true, Ordering::Release)
```
Darksonn, line 196: What is the purpose of this flag?

Darksonn (commented):
Okay, it seems like we have worked out the issues with the task id. I'll need to find the time to properly go through the entire change to give a full review, but here are a few comments I noticed when looking today.

tokio/src/runtime/builder.rs
wathenjiang: This is as I said above:
> The new added method `spawn_concurrency_level()` in Builder to set the number of segment lock in OwnedTasks struct is to allow us to customize this value. Its default value is 4 times the number of workers thread. An alternative approach is to set it as a constant, like 64, regardless of the number of CPU cores on different platforms. However, considering the usage of Tokio on micro-embedded devices, making the default value dependent on the number of worker threads can effectively minimize the time required for runtime creation on micro-embedded devices.

I thinke make it unstable is a good decision.

tokio/src/runtime/task/list.rs
wathenjiang: I use `count: AtomicUsize` to represent the number of tasks, and wo do operation by `Ordering::Relaxed` everywhere for the best performance, but it has no `happens-before` relationship.

The shutdown flag is write by `Ordering::Release` and read by `Ordering::Acquire`, it can ensure a `happens-before` relationship.

I use this flag here to make sure the following assert can always pass:

```
debug_assert!(self.shared.owned.is_shutdown()); // do has a `happens-before` relationship.
debug_assert!(self.shared.owned.is_empty()); // has no `happens-before` relationship.
```

wathenjiang: Because in `multi_thread` mod, the lock is always obtained first, and then the closed flag of OwnedTasks is read. So 
 I would like to delete the shutdown flag.

wathenjiang:
The failure of https://github.com/tokio-rs/tokio/actions/runs/6388358967/job/17338072509?pr=6001 puzzles me.

Assuming there are two worker threads, thread A closes the last tasks and then sets the `count` to 0. Meanwhile, thread B performs the assert `debug_assert!(self.shared.owned.is_empty());`.

Under this assumption, thread A must first acquire the lock `self.shared.shutdown_cores` and then release it. Only after thread A releases the lock can thread B obtain it and perform the assert.

Thus, due to the `happens before` relationship provided by the lock `self.shared.shutdown_cores`, the assert should always pass. 

Am I doing something wrong here? Or does the lock used in this test provide no guarantee of happens before?

Darksonn:
Imagine this ordering of events, where we assume that both threads are on the last iteration of the outer for loop.
```
thread 1              thread 2
lock.pop_back()=Some
drop(lock)
                      let mut lock = self.segment_inner(i);
                      lock.pop_back()=None
		      break
		      return
		      ...
		      debug_assert!(self.shared.owned.is_empty())
self.count.fetch_sub(1, Ordering::Relaxed);
task.shutdown();
```
Here, the counter is not yet zero because another thread still has a task that it needs to drop. However, thread 2 was able to return from `close_and_shutdown_all` because the task in question has already been removed from `OwnedTasks`.

Darksonn:
Yes. The call to `task.shutdown` can still happen after other threads return from the call, but it seems like that's already the case.

wathenjiang:
But we do the assertion `debug_assert!(self.shared.owned.is_empty());` in the `shutdown_core` method, it is executed by the last worker threads. At the moment when the assertion `debug_assert!(self.shared.owned.is_empty());` is executed, all the worker threads must have returned from `close_and_shutdown_all`, so the execution order would be as follows :

```
the `close_and_shutdown_all` method:

thread 1              thread 2
lock.pop_back()=Some
drop(lock)
                      let mut lock = self.segment_inner(i);
                      lock.pop_back()=None
		      break
		      return
		      ...		      
self.count.fetch_sub(1, Ordering::Relaxed);
task.shutdown();
return

the `shutdown_core` method:

thread 1              thread 2
                      shutdown_cores.lock();
                      return
shutdown_cores.lock();
debug_assert!(self.shared.owned.is_empty())
```

wathenjiang:
If setting the count to 0 occurs in thread 1, then thread 1's assertion must succeed, so we only need to discuss the scenario when thread 2 is responsible for the assertion.

Thread 2 making an assertion implies that thread 1 first acquires the lock, then releases it, and after that, thread 2 acquires the lock and performs the assertion.

```
thread 1              thread 2
self.count.fetch_sub(1, Ordering::Relaxed);
//... whatever happens here
shutdown_cores.lock();
cores.push(core);
return; <-------------
                     |
                     |
                     |
              happens before
                     |
                     |
                     |
                      shutdown_cores.lock();
                      debug_assert!(self.shared.owned.is_empty())
```

tokio/src/runtime/task/list.rs
```diff
@@ -179,26 +210,53 @@ impl<S: 'static> OwnedTasks<S> {
...
+    }
+
+    #[inline]
+    unsafe fn remove_inner(&self, task: &Task<S>) -> Option<Task<S>> {
+        // Safety: it is safe, because every task has one task_id
+        let task_id = unsafe { Header::get_id(task.header_ptr()) };
+        let mut lock = self.segment_inner(task_id.0 as usize);
+        let task = lock.remove(task.header_ptr())?;
+        drop(lock);
+        self.count.fetch_sub(1, Ordering::Relaxed);
```
Darksonn, line 223: You should only subtract here if `task.is_some()`.

wathenjiang: I have used the `?` operator to quickly return. So if the task is `None`, we will not subtract here.

Darksonn:
Since the error happened in `coop_disabled_in_block_in_place`, it probably has something to do with the usage of `block_in_place`. After all, when `block_in_place` is used, the thread stops being a worker thread, but the current future is still executed on that thread. So this means that the future is being executed outside the runtime worker threads.

wathenjiang:
Good insight! Thanks.

The key here is that non-worker threads will participate in the shutdown of the task, I believe there would be the following order of execution:


```
non-worker thread                           worker thread
segmented_link.lock();
segmented_link.unlock(); <--------------------
self.count.fetch_sub(1, Ordering::Relaxed);  |
                                             |
                                             |
                                             |
                                      happens before
                                             |
                                             |
                                             |
                                             segmented_link.lock();
                                             segmented_link.unlock();
                                             debug_assert!(self.shared.owned.is_empty())
```
We **CANNOT** ensure `self.count.fetch_sub(1, Ordering::Relaxed);` happens before `debug_assert!(self.shared.owned.is_empty())`.

So let us move `self.count.fetch_sub(1, Ordering::Relaxed);` into the duration when we having the lock.

wathenjiang:
I'm wondering if we could add Tasks to OwnedTasks only when necessary. The OwnedTasks is used for track all the tasks now, but for those tasks in lifo_slot, local queue, or inject, the tokio scheduler does hold the references to them. Because of that, we can release them when runtime is being dropped. In addition of these tasks, other tasks must be tracked by OwnedTaks.

I have tried not to add the Task to OwnedTasks when calling `tokio::spawn` function,and whether the task is added to the OwnedTasks depends on the scheduling status of the subsequent Task. I have not verified the feasibility of this solution. But, so far, I found that this way effectively reduces the lock performance overhead of `tokio::spawn` when creating tasks concurrently.

What's your take on this? I'd be very grateful for any advice.

Darksonn:
I think that would hurt performance overall. You would have to remove and add them to the `OwnedTasks` every time you notify or poll them. 

Like, sure, maybe this makes `tokio::spawn` faster, but we shouldn't do so at the expense of notifying or polling tasks.

wathenjiang:
My opinion is that OwnedTasks now works like a memo, which will record some tasks that **may** not be held by the scheduler, note that this is only a possibility. This means that tasks being stored in life_slot, local queue, or inject may be stored repeatedly by OwnedTasks at shutdown time.

And we need to make sure to avoid double free problems between lifo_slot, local queue, or inject and OwnedTasks during shutdown.

wathenjiang:
Therefore, when a Task is scheduled, if it is not completed or will not be scheduled again immediately (for example, waiting to be awakened by a reactor), it will try to be added to OwnedTasks. If it has already been added, nothing will be done.

When the task is been notified or polled, we do not remove it from OwnedTasks.

Darksonn:
I see the idea, and I agree that it could help if there are many tasks that never go to sleep.

However, I want to avoid making the scheduler more complicated. The current PR isn't so bad, since the complexity is localized to a single file, but this idea isn't localized in the same way.

wathenjiang:
Yes, I would separate this idea from the current pr.

Anyway, If you have any thoughts on the current pr, please point it out, I appreciate every comment.

Darksonn:
The above conversation gave me the following idea: What if we isolated the sharding into a separate file? This could seriously simplify the code, because we wouldn't have to worry about sharding in `list.rs`.

So, the idea is as follows:

 * Create a new file `src/util/sharded_list.rs` that defines a sharded linked list.
 * The struct looks something like this:

```
struct ShardedList<T: SharedListItem> {
    lists: Box<[Mutex<ListSement<S>>]>,
    count: AtomicUsize,
}

type ListSement<S> = LinkedList<Task<S>, <Task<S> as Link>::Target>;

trait ShardedListItem: Link {
    // Used to pick which list this item should go into.
    fn get_shard_index(&self) -> usize;
}
```

 * Then, we modify `src/runtime/task/list.rs` to just use the above utility.

What do you think about this? The goal would be to simplify the code in `list.rs`.

Darksonn:
To be clear, I don't think you _have_ to use this approach to get it merged. If you don't want to, we can also continue with the current implementation.

wathenjiang:
I think your idea is very great, I would like to move it to a separate file.

I also notice actually we do not need tail in OwnedTasks, because the `pop_last` seems not useful for the OwnedTasks, `pop_front` is enough. If make `sharded_list` do not depend on `LinkedList`, we might get a little bit performance win. But since modern CPUs always have very good branch prediction capabilities, this should have little effect.

What's your take on this?

Darksonn:
Okay, sounds good.

> I also notice actually we do not need tail in OwnedTasks, because the `pop_last` seems not useful for the OwnedTasks, `pop_front` is enough. If make `sharded_list` do not depend on `LinkedList`, we might get a little bit performance win. But since modern CPUs always have very good branch prediction capabilities, this should have little effect.

We can do that in a follow-up PR.

tokio/src/util/shared_list.rs
Darksonn, line 1: The filename is incorrect. It should be `sharded_list.rs`.

tokio/src/runtime/task/list.rs
```diff
@@ -111,25 +110,30 @@ impl<S: 'static> OwnedTasks<S> {
...
+
+        // double check close flag for ensuring all tasks will shutdown after OwnedTasks has been closed,
+        // it should be completed quickly.
+        if self.closed.load(Ordering::Acquire) {
+            for i in 0..self.get_shard_size() {
+                if let Some(task) = self.list.pop_back(i) {
+                    task.shutdown();
+                }
+            }
+        }
```
Darksonn, line 128: You can avoid this by defining a method like this:
```
impl<L: ShardedListItem> ShardedList<L, L::Target> {
    pub fn lock_shard(&self, val: &L::Handle) -> ShardGuard<'_, L, L::Target> {
        let id = unsafe { L::get_shared_id(L::as_raw(&val)) };
        ShardGuard {
            lock: self.shard_inner(id),
            count: &self.count,
            id,
        }
    }
}

struct ShardGuard<'a, L, T> {
    lock: MutexGuard<'a, LinkedList<L, T>>
    count: &'a AtomicUsize,
    id: usize,
}

impl<L: ShardedListItem> ShardGuard<L, L::Target> {
    /// Push a value to this shard.
    pub fn push_shard(self, val: L::Handle) {
        let id = unsafe { L::get_shared_id(L::as_raw(&val)) };
        assert_eq!(id, self.id);
        self.lock.push_front(val);
        self.count.fetch_add(1, Ordering::Relaxed);
    }
}
```
This will let you check `self.closed` while the mutex is locked.

tokio/src/runtime/task/list.rs
```diff
@@ -140,34 +138,36 @@ impl<S: 'static> OwnedTasks<S> {
 
     /// Shuts down all tasks in the collection. This call also closes the
     /// collection, preventing new items from being added.
-    pub(crate) fn close_and_shutdown_all(&self)
+    ///
+    /// The parameter start should be random among different worker threads
+    /// to reduce lock conflicts during shutdown.
+    /// Initiate shutting down the shard indexed by the start, and reset to 0
+    /// once the shar_size is reached, continuing until start - 1, it works like a ring.
```
Darksonn, line 145: This sentence is inaccurate if `start == 0`. 
```suggestion
    /// once the `shard_size` is reached, continuing until `start - 1`, it works like a ring.
```

tokio/src/runtime/task/list.rs
```diff
@@ -179,22 +179,21 @@ impl<S: 'static> OwnedTasks<S> {
...
     impl<S: 'static> OwnedTasks<S> {
-        /// Locks the tasks, and calls `f` on an iterator over them.
-        pub(crate) fn for_each<F>(&self, f: F)
+        pub(crate) fn for_each<F>(&self, mut f: F)
         where
-            F: FnMut(&Task<S>)
+            F: FnMut(&Task<S>),
         {
-            self.inner.lock().list.for_each(f)
+                self.list.for_each(&mut f);
```
Darksonn, line 196:
```suggestion
            self.list.for_each(&mut f);
```

tokio/src/runtime/task/mod.rs
```diff
@@ -503,3 +504,14 @@ unsafe impl<S> linked_list::Link for Task<S> {
...
 }
+
+/// # Safety
+///
+/// Tasks are pinned.
+unsafe impl<S> sharded_list::ShardedListItem for Task<S> {
+    unsafe fn get_shared_id(target: NonNull<Self::Target>) -> usize {
+        let task_id = unsafe { Header::get_id(target) };
+        // safety: in 32-bits machine, the lower 32 bits will be reserved
+        task_id.0 as usize
```
Darksonn, line 515: I don't understand this safety comment.

tokio/src/util/linked_list.rs
```diff
@@ -342,22 +295,11 @@ cfg_io_driver_impl! {
...
-        where
-            F: FnMut(&T::Handle),
-        {
-            self.list.for_each(f)
-        }
-    }
-
     impl<T: Link> LinkedList<T, T::Target> {
-        pub(crate) fn for_each<F>(&mut self, mut f: F)
+        pub(crate) fn for_each<F>(&mut self, f: &mut F)
```
Darksonn, line 299: Why this change?

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
+#![cfg_attr(not(feature = "full"), allow(dead_code))]
+
+use std::{ptr::NonNull, sync::atomic::Ordering};
```
Darksonn, line 3: We only use `{}` for things in the same module.
```suggestion
use std::ptr::NonNull;
use std::sync::atomic::Ordering;
```

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
+#![cfg_attr(not(feature = "full"), allow(dead_code))]
+
```
Darksonn, line 2: Please remove this one.

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
...
+use std::{ptr::NonNull, sync::atomic::Ordering};
+
+use crate::loom::sync::{Mutex, MutexGuard};
+use std::sync::atomic::AtomicUsize;
+
+use super::linked_list::{Link, LinkedList};
+
+/// An intrusive  linked list supporting high concurrent updates.
```
Darksonn, line 10:
```suggestion
/// An intrusive linked list supporting highly concurrent updates.
```

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
...
+///
+/// It currently relies on `LinkedList`, so it is the caller's
+/// responsibility to ensure the list is empty before dropping it.
+///
+/// Note: Due to its inner sharded design, the order of node cannot be guaranteed.
```
Darksonn, line 15:
```suggestion
/// Note: Due to its inner sharded design, the order of nodes cannot be guaranteed.
```

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
...
+pub(crate) struct ShardedList<L, T> {
+    lists: Box<[Mutex<LinkedList<L, T>>]>,
+    count: AtomicUsize,
+    shard_mask: usize,
+}
+
+/// Defines the id of a node in a linked list, different ids cause inner list nodes
+/// to be scattered in different mutexs.
```
Darksonn, line 23:
```suggestion
/// Determines which linked list an item should be stored in.
```

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
...
+///
+/// # Safety
+///
+/// Implementations must guarantee that `Target` types are pinned in memory. In
+/// other words, when a node is inserted, the value will not be moved as long as
+/// it is stored in the list.
```
Darksonn, line 29: You don't need this, because anything that implements this trait also implements `Link`, so they promised to follow these rules when implementing `Link`. It's not necessary to require the same thing twice.

However, you do require one other thing:
```suggestion
/// Implementations must guarantee that the id of an item does not change from
/// call to call.
```

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
...
+pub(crate) unsafe trait ShardedListItem: Link {
+    // The returned id is used to pick which list this item should go into.
+    unsafe fn get_shared_id(target: NonNull<Self::Target>) -> usize;
```
Darksonn, line 32:
```suggestion
    /// # Safety
    /// The provided pointer must point at a valid list item.
    unsafe fn get_shared_id(target: NonNull<Self::Target>) -> usize;
```

Darksonn, line 32:
```suggestion
    unsafe fn get_shard_id(target: NonNull<Self::Target>) -> usize;
```

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+impl<L, T> ShardedList<L, T> {
+    /// Creates a new and empty sharded linked list under specified size
```
Darksonn, line 36:
```suggestion
    /// Creates a new and empty sharded linked list with the specified size.
```

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub(crate) fn new(sharded_size: usize) -> Self {
+        // Find power-of-two sizes best matching arguments
+        let mut size = 1;
+        while size < sharded_size {
+            size <<= 1;
+        }
```
Darksonn, line 42: I'm not a big fan of this. It means that the size is not always accurate, and the loop will run infinitely if `sharded_size > 2^63`.
```suggestion
        assert!(sharded_size.is_power_of_two());
```

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+impl<L: ShardedListItem> ShardedList<L, L::Target> {
+    #[allow(dead_code)]
+    /// Push a value to this shard.
+    pub(crate) fn push(&self, val: L::Handle) {
+        let shard = self.lock_shard(&val);
+        shard.push(val);
+        self.count.fetch_add(1, Ordering::Relaxed);
+    }
```
Darksonn, line 71: If this function isn't used, then just delete it.

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
...
+            id,
+        }
+    }
+
+    /// Gets the count of elements in this list.
+    pub(crate) fn len(&self) -> usize {
+        self.count.load(Ordering::Relaxed)
+    }
+
+    /// Returns whether the linked list does not contain any node
```
Darksonn, line 117: You are missing these in several places. I won't point out every single one, so please go through and fix them.
```suggestion
    /// Returns whether the linked list does not contain any node.
```

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub(crate) fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Gets the shard size of this SharedList.
+    /// Used to help us to decide the parameter `shard_id`` of the `pop_back` method.
```
Darksonn, line 123: Generally, you have an empty after the first line of documentation.
```suggestion
    /// Gets the shard size of this SharedList.
    ///
    /// Used to help us to decide the parameter `shard_id`` of the `pop_back` method.
```

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub(crate) fn shard_size(&self) -> usize {
+        self.shard_mask + 1
+    }
+
+    #[inline]
+    fn shard_inner(&self, id: usize) -> MutexGuard<'_, LinkedList<L, <L as Link>::Target>> {
+        // Safety: this modulo operation ensures it is safe here.
```
Darksonn, line 130:
```suggestion
        // Safety: This modulo operation ensures that the index is not out of bounds.
```

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
...
+        self.lock.push_front(val);
+        self.count.fetch_add(1, Ordering::Relaxed);
+    }
+}
+
+cfg_taskdump! {
+    impl<L: ShardedListItem> ShardedList<L, L::Target> {
+        pub(crate) fn for_each<F>(&self, f: &mut F)
+        where
+        F: FnMut(&L::Handle),
```
Darksonn, line 149:
```suggestion
            F: FnMut(&L::Handle),
```

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
...
+        {
+            let mut guards = vec![];
```
Darksonn, line 151:
```suggestion
            let mut guards = Vec::with_capacity(self.lists.len());
```

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
...
+            for list in self.lists.iter(){
```
Darksonn, line 152:
```suggestion
            for list in self.lists.iter() {
```

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,160 @@
...
+                guards.push(list.lock());
+            }
+            for g in &mut guards{
```
Darksonn, line 155:
```suggestion
            for g in &mut guards {
```

Darksonn (commented):
Overall, I think this change looks good! I went through and found a bunch of minor issues, but they should all be relatively easy to fix.

Sorry about the delay in reviewing this.

tokio/src/runtime/task/mod.rs
wathenjiang: When I was writing this code, it occurred to me that `task_id` is actually different from `shard_id`. The size of `task_id` is 64 bits in all platform, but the `shard_id` depends on the specific platform: 
- 32 bits on 32-bit machine.
- 64 bits on 32-bit machine.

Casting 64 bits to 32 bits may be not safe, but we use this id for shard, so it is safe here.

tokio/src/util/linked_list.rs
wathenjiang: Yes, it should accept ownership of this `F` closure  instead!

tokio/src/util/sharded_list.rs
wathenjiang: Ok.

tokio/src/runtime/task/mod.rs
```diff
@@ -503,3 +504,14 @@ unsafe impl<S> linked_list::Link for Task<S> {
...
 }
+
+/// # Safety
+///
+/// Tasks are pinned.
+unsafe impl<S> sharded_list::ShardedListItem for Task<S> {
+    unsafe fn get_shard_id(target: NonNull<Self::Target>) -> usize {
+        let task_id = unsafe { Header::get_id(target) };
+        // safety: in 32-bits machine, the lower 32 bits will be reserved
+        task_id.0 as usize
```
Darksonn, line 515:
```suggestion
/// # Safety
///
/// The id of a task is never changed after creation of the task, so the return value of
/// `get_shard_id` will not change. (The cast may throw away the upper 32 bits of the task id, but
/// the shard still won't change from call to call.)
unsafe impl<S> sharded_list::ShardedListItem for Task<S> {
    unsafe fn get_shard_id(target: NonNull<Self::Target>) -> usize {
        // SAFETY: The caller guarantees that `target` points at a valid task.
        let task_id = unsafe { Header::get_id(target) };
        task_id.0 as usize
```

tokio/src/runtime/task/mod.rs
Darksonn: See my other comment for the right way to add these safety comments.

tokio/src/util/sharded_list.rs
wathenjiang: I also think it's better to leave the size of `sharded_size` to  the caller.

But considering that the number of worker threads is not necessarily exactly an integer power of 2, this loop must still be executed.

However, spawn_concurrency_level is 4 times the number of worker threads. I considered setting its maximum value to 1024, which I believe is a large enough number.

tokio/src/runtime/task/mod.rs
wathenjiang: make it.

tokio/src/runtime/task/mod.rs
wathenjiang: make it.

tokio/src/util/sharded_list.rs
wathenjiang: I moved this logic to 
```
impl Builder {
    fn spawn_concurrency_level(&self) -> usize {
        const MAX_SPAWN_CONCURRENCY_LEVEL: usize = 1 << 16;

        match self.spawn_concurrency_level {
            Some(i) => i,
            None => {
                use crate::loom::sys::num_cpus;
                let core_threads = self.worker_threads.unwrap_or_else(num_cpus);

                let mut size = 1;
                while size / 4 < core_threads && size < MAX_SPAWN_CONCURRENCY_LEVEL {
                    size <<= 1;
                }
                size.min(MAX_SPAWN_CONCURRENCY_LEVEL)
            }
        }
    }
}
```

tokio/src/util/sharded_list.rs
wathenjiang: Do you means `#![cfg_attr(not(feature = "full"), allow(dead_code))]`, if do that, we will got dede_code error in some feature.

Darksonn: Please adjust the `cfg` in `src/util/mod.rs` instead to fix that dead code warning.

wathenjiang: I use `cfg_rt!` in `src/util/mod.rs`, and move `#![cfg_attr(not(feature = "full"), allow(dead_code))]`.

wathenjiang:
I haven't paid attention to this PR because I have been too busy recently, sorry for that. I'd like to know how this PR is progressing so far. Any feedback on the design would be greatly appreciated.

As far as the current code review is concerned, we may have inconsistent views in the following areas:
- Whether the number of mutexes in sharedList used by ownedTasks should be fixed
- Whether `spawn_concurrency_level` should be supported for configuration]

Whether it should be such a complex design depends on tokio's use case, I provide some factual reference here.

I use the following code for benchmark testing:

```rust
use std::time::Instant;

use criterion::{measurement::WallTime, *};

fn bench_parallel_spawn_multi_thread(c: &mut Criterion) {
    let mut group = c.benchmark_group("spawn_parallel_multi_thread2");
    let ws: [usize; 8] = [1, 2, 4, 8, 16, 32, 64, 128];
    let ss: [usize; 12] = [1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048];

    for w in ws {
        for s in ss {
            spawn_tasks_parallel_multi_thread(&mut group, w, s);
        }
    }
}

fn spawn_tasks_parallel_multi_thread(g: &mut BenchmarkGroup<WallTime>, w: usize, s: usize) {
        .worker_threads(w)
        .spawn_concurrency_level(black_box(s))
        .build()
        .unwrap();

    g.bench_function(format!("{:03}-{:04}", w, s), |b| {
        b.iter_custom(|iters| {
            let start = Instant::now();
            runtime.block_on(async {
                black_box(spawn_job(iters as usize, w).await);
            });
            start.elapsed()
        })
    });
}

async fn spawn_job(iters: usize, procs: usize) {
    for _ in 0..procs {
        let mut threads_handles = Vec::with_capacity(procs);
        threads_handles.push(tokio::spawn(async move {
            let mut thread_handles = Vec::with_capacity(iters / procs);
            for _ in 0..iters / procs {
                thread_handles.push(tokio::spawn(async {
                    let val = 1 + 1;
                    black_box(val)
                }));
            }
            for handle in thread_handles {
                handle.await.unwrap();
            }
        }));
        for handle in threads_handles {
            handle.await.unwrap();
        }
    }
}

criterion_group!(benches, bench_parallel_spawn_multi_thread,);
criterion_main!(benches);
```
The following test results were obtained on a Linux machine with a 16 cores x86 CPU:

mutexes/workers | 1 | 2 | 4 | 8 | 16 | 32 | 64 | 128
-- | -- | -- | -- | -- | -- | -- | -- | --
1 | 803.57 | 687.35 | 986.31 | 1015.4 | 1030.8 | 1070.4 | 1005 | 1127.9
2 | 793.59 | 668.52 | 914.75 | 883.09 | 872.51 | 874.69 | 950.95 | 971.81
4 | 817.35 | 650.24 | 832.3 | 832.16 | 885.87 | 858.87 | 850.21 | 860.88
8 | 827.45 | 631.54 | 770.99 | 852.24 | 811.57 | 804.15 | 876.1 | 892.08
16 | 815.97 | 594.82 | 838.56 | 846.6 | 825.85 | 820.52 | 803.44 | 885.4
32 | 820.33 | 577.8 | 776.64 | 805.6 | 796.62 | 805.46 | 781.81 | 874.36
64 | 769.73 | 594.62 | 750.4 | 801.3 | 734.45 | 798.41 | 804.04 | 816.49
128 | 763.23 | 599.43 | 742.1 | 739.86 | 784 | 777.41 | 731.66 | 814.59
256 | 761.52 | 608.53 | 745.58 | 745.21 | 713.8 | 748.61 | 802.93 | 805.13
512 | 807.22 | 596.78 | 694.39 | 727.19 | 785.3 | 783.67 | 740.8 | 733.06
1024 | 773.73 | 621.63 | 749.25 | 781.1 | 723.71 | 775.19 | 797.04 | 811.37
2048 | 808.78 | 578.05 | 662.01 | 775.17 | 786.92 | 774.91 | 794.79 | 800.56

If the above test results are plotted into a graph, then it is as shown in the figure below:
image

We can see that:
- No matter how many Mutex configurations there are, the optimal thing is to use 2 worker threads to perform `tokio::spawn` concurrently.
- For all numbers of worker threads, there is a situation where the time taken by `tokio::spawn` first decreases as the number of Mutex increases, but eventually increases.
- The blue line is the baseline, that is, the trend chart of the performance change of `tokio::spawn` caused by a single worker thread as the number of Mutex.
- When the number of Mutex is 1, regardless of the number of worker threads, increasing the number of Mutex can always reduce the time consumption of `tokio::sapwn`.

Finally, it is difficult to estimate how many mutexes to provide in different worker threads is optimal for the time consumption of `tokio::spawn`. Perhaps a fixed-size mutex number is provided by default like 64 (currently at least 4 times the current worker threads, and is a minimum integer power of 2, which is relatively more complicated), and providing a configurable mutex number option for those extreme `tokio::spawn` highly concurrency is a good solution.

Darksonn:
Sorry for the delay in review. I've been visiting conferences last week and next week, so I'm not as actively watching the repository right now as I usually am.

> I'd like to know how this PR is progressing so far. Any feedback on the design would be greatly appreciated.

Overall, I'm pretty happy with the approach. As you mention, whether we should have a config option is my only real question. I would lean towards no because it's simpler, but if it's unstable, I do not mind having it.

Other than that, I think there were only various minor things to fix last I looked.

wathenjiang:
I have considered for some time and believe that your idea is better choice. We will as well not allow users to configure the `spawn_concurrency_level` option directly, even in unstable mode.

tokio/src/runtime/builder.rs
```diff
@@ -1314,6 +1329,15 @@ cfg_rt_multi_thread! {
...
         }
+
+        fn get_spawn_concurrency_level(core_threads : usize) -> usize {
+            const MAX_SPAWN_CONCURRENCY_LEVEL: usize = 1 << 16;
+                    let mut size = 1;
+                    while size / 4 < core_threads && size < MAX_SPAWN_CONCURRENCY_LEVEL {
+                        size <<= 1;
+                    }
+                    size.min(MAX_SPAWN_CONCURRENCY_LEVEL)
+            }
```
Darksonn, line 1340: If we are removing it as an option from the builder, then all changes related to `spawn_concurrency_level` in this file and other files should be removed. Instead, I would just move this into `runtime/task/list.rs` as a const.

tokio/src/runtime/task/list.rs
```diff
@@ -112,24 +111,23 @@ impl<S: 'static> OwnedTasks<S> {
             task.header().set_owner_id(self.id);
         }
 
-        let mut lock = self.inner.lock();
-        if lock.closed {
-            drop(lock);
-            drop(notified);
+        let shard = self.list.lock_shard(&task);
+        // check the closed flag in the lock for ensuring all tasks
+        // will shutdown after ownedTasks has been closed.
```
Darksonn, line 116:
```suggestion
        // Check the closed flag in the lock for ensuring all that tasks
        // will shut down after the OwnedTasks has been closed.
```

tokio/src/runtime/task/list.rs
```diff
@@ -179,22 +179,21 @@ impl<S: 'static> OwnedTasks<S> {
...
     pub(crate) fn is_empty(&self) -> bool {
-        self.inner.lock().list.is_empty()
+        self.list.is_empty()
     }
 }
 
 cfg_taskdump! {
     impl<S: 'static> OwnedTasks<S> {
-        /// Locks the tasks, and calls `f` on an iterator over them.
         pub(crate) fn for_each<F>(&self, f: F)
```
Darksonn, line 192: Please don't delete the documentation.

tokio/src/runtime/task/list.rs
```diff
@@ -140,34 +138,36 @@ impl<S: 'static> OwnedTasks<S> {
 
     /// Shuts down all tasks in the collection. This call also closes the
     /// collection, preventing new items from being added.
-    pub(crate) fn close_and_shutdown_all(&self)
+    ///
+    /// The parameter start should be random among different worker threads
+    /// to reduce lock conflicts during shutdown.
+    /// Initiate shutting down the shard indexed by the start, and reset to 0
+    /// once the `shard_size` is reached, continuing until `start - 1`, it works like a ring.
```
Darksonn, line 145:
```suggestion
    /// The parameter start determines which shard this method will start at.
    /// Using different values for each worker thread reduces contention.
```

tokio/src/runtime/task/mod.rs
```diff
@@ -503,3 +504,16 @@ unsafe impl<S> linked_list::Link for Task<S> {
         self::core::Trailer::addr_of_owned(Header::get_trailer(target))
     }
 }
+
+/// # Safety
+///
+/// The id of a task is never changed after creation of the task, so the return value of
+/// `get_shard_id` will not change. (The cast may throw away the upper 32 bits of the task id, but
+/// the shard still won't change from call to call.)
```
Darksonn, line 512:
```suggestion
/// the shard id still won't change from call to call.)
```

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// # Safety
+    ///
+    /// The caller **must** ensure that exactly one of the following is true:
+    /// - `node` is currently contained by `self`,
+    /// - `node` is not contained by any list,
+    /// - `node` is currently contained by some other `GuardedLinkedList`.
+    pub(crate) unsafe fn remove(&self, node: NonNull<L::Target>) -> Option<L::Handle> {
+        let id = L::get_shard_id(node);
+        let mut lock = self.shard_inner(id);
+        let node = lock.remove(node);
```
Darksonn, line 81:
```suggestion
        // SAFETY: Since the shard id cannot change, it's not possible for this node
        // to be in any other list of the same sharded list.
        let node = unsafe { lock.remove(node) };
```

tokio/src/util/linked_list.rs
Darksonn: Please undo this change and continue to use `mut f: F`. The code in `sharded_list.rs` will still work because mutable references are also `FnMut`.

Darksonn (commented):
With the changes below, I'll be happy to merge this.

tokio/src/runtime/task/list.rs
```diff
@@ -179,11 +179,30 @@ impl<S: 'static> OwnedTasks<S> {
...
+    /// Due to the above reasons, we set a maximum value for the shared list size,
+    /// denoted as `MAX_SHARED_LIST_SIZE`.
+    fn gen_shared_list_size(num_cores: usize) -> usize {
+        const MAX_SHARED_LIST_SIZE: usize = 1 << 16;
+        let mut size = 1;
+        while size / 4 < num_cores && size < MAX_SHARED_LIST_SIZE {
+            size <<= 1;
+        }
+        size.min(MAX_SHARED_LIST_SIZE)
     }
```
Darksonn, line 206: It seems like this has several typos of "shared" instead of "sharded". Or is there some reason you use the word "shared"?

tokio/src/runtime/task/list.rs
```diff
@@ -179,11 +179,30 @@ impl<S: 'static> OwnedTasks<S> {
...
         // linked list.
-        unsafe { self.inner.lock().list.remove(task.header_ptr()) }
+        unsafe { self.list.remove(task.header_ptr()) }
     }
 
     pub(crate) fn is_empty(&self) -> bool {
-        self.inner.lock().list.is_empty()
+        self.list.is_empty()
+    }
+    /// Generates the size of the shared list based on the number of worker threads.
```
Darksonn, line 188:
```suggestion
    }

    /// Generates the size of the shared list based on the number of worker threads.
```

tokio/src/runtime/task/list.rs
```diff
@@ -179,11 +179,30 @@ impl<S: 'static> OwnedTasks<S> {
...
+    ///
+    /// Due to the above reasons, we set a maximum value for the shared list size,
+    /// denoted as `MAX_SHARED_LIST_SIZE`.
+    fn gen_shared_list_size(num_cores: usize) -> usize {
+        const MAX_SHARED_LIST_SIZE: usize = 1 << 16;
+        let mut size = 1;
+        while size / 4 < num_cores && size < MAX_SHARED_LIST_SIZE {
+            size <<= 1;
+        }
+        size.min(MAX_SHARED_LIST_SIZE)
```
Darksonn, line 205:
```suggestion
        usize::min(MAX_SHARED_LIST_SIZE, num_cores.next_power_of_two()*4)
```

tokio/src/runtime/task/list.rs
```diff
@@ -179,11 +179,30 @@ impl<S: 'static> OwnedTasks<S> {
...
+    /// lock contention performance problems caused by high concurrency.
+    ///
+    /// However, as the number of shards increases, the memory continuity between
+    /// nodes in the intrusive linked list will diminish. Furthermore,
+    /// the construction time of the sharded list will also increase with a higher number of shards.
+    ///
+    /// Due to the above reasons, we set a maximum value for the shared list size,
+    /// denoted as `MAX_SHARED_LIST_SIZE`.
+    fn gen_shared_list_size(num_cores: usize) -> usize {
+        const MAX_SHARED_LIST_SIZE: usize = 1 << 16;
```
Darksonn, line 200: This maximum seems really large to me. Thoughts?

wathenjiang: I just used the settings of the previous `spawn_concurrency_level` function. The previous function could support user configuration, so this value was set to a large value.

If you want a smaller value, I'm ok with that.

As far as I know, some current server CPUs can reach close to 200 CPU cores (or hyperthreadings), such as the [AMD EPYC 9654](https://www.amd.com/en/products/cpu/amd-epyc-9654) which has 192 threads. Maybe we can set the value to 256?

wathenjiang: I'd like to retract my previous point. I believe this should be a memory-bound scenario, so we only need to ensure a certain number of mutexes. In fact, the number of mutexes should not depend on the number of worker threads, but actually depends on the possible level of concurrency.

Below, I describe the concurrency level that may actually occur, rather than what occurs in the benchmark.

In real applications that use Tokio, multiple threads may perform `tokio::spawn` concurrently. However, having too many threads concurrently performing `tokio::spawn` at the same time can be considered a poor design choice in the higher-level application architecture. It is more likely that multiple threads will perform removing tasks concurrently at the same time. Nonetheless, removing a task from the list only takes up a minimal amount of runtime during the entire task life cycle, so the concurrency level of removing tasks is usually not a significant concern.

Therefore, I agree with your idea, and it is reasonable to set the upper limit to a small value. Setting this to 64 or 128 might make sense.

Darksonn: Alright. What do you think about multiplying by 4?

wathenjiang: The total number of random memory operations is consistent, providing multiple locks. Each thread obtains one of the locks through round robin, and can only perform random memory access after obtaining the lock. I got the following test results:

image

We can find that 4 times is reasonable when the number of threads is low.

wathenjiang: The total number of random memory operations is consistent, providing multiple locks. Each thread obtains one of the locks through round robin, and can only perform random memory access after obtaining the lock. I got the following test results:
mutexes/threads | 1 | 2 | 4 | 8 | 12 | 16 | 24 | 32 | 48 | 64 | 128
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
0 | 5.980172444 | 2.899437975 | 1.447906311 | 0.828731566 | 0.689618066 | 0.612355429 | 0.589401394 | 0.587380871 | 0.525477567 | 0.578456362 | 0.552132325
1 | 7.970250774 | 17.29034894 | 22.60164692 | 25.97284605 | 28.12352579 | 33.31359697 | 31.18786342 | 31.61139126 | 29.23225856 | 30.94094675 | 31.59191497
2 | 7.883931727 | 15.97845738 | 16.11107368 | 18.73377898 | 20.34614133 | 23.02624802 | 22.69439808 | 23.15802647 | 21.80570219 | 22.48815498 | 22.98585238
4 | 7.975676415 | 10.25364766 | 11.88074538 | 15.40198137 | 15.51024255 | 16.35328034 | 15.46874828 | 15.7982897 | 15.48703267 | 15.67227903 | 15.35829948
8 | 8.058803258 | 8.138193999 | 7.619081588 | 7.936418179 | 7.654288652 | 7.901945312 | 7.642439744 | 7.861542054 | 7.730389506 | 7.821229611 | 7.748344488
16 | 9.797308994 | 6.213334839 | 4.455407945 | 4.496371955 | 4.291254249 | 4.130849346 | 4.347601475 | 4.294096757 | 3.990391527 | 4.028562691 | 4.059085994
32 | 8.742854719 | 4.847656612 | 3.301780829 | 2.578327826 | 2.480488617 | 2.331294827 | 2.388718271 | 2.306257478 | 2.421350161 | 2.278177495 | 2.26569423
64 | 8.042672888 | 4.963568223 | 3.012473492 | 2.08243512 | 1.828237002 | 1.653421053 | 1.550811454 | 1.536452054 | 1.519761769 | 1.618966043 | 1.48010674
128 | 8.62801309 | 4.978525185 | 2.637936755 | 1.777546296 | 1.549096849 | 1.359814529 | 1.43875245 | 1.385468038 | 1.238832309 | 1.249940559 | 1.248131329
256 | 8.584906215 | 4.591742459 | 2.441556366 | 1.504790937 | 1.335449235 | 1.169191715 | 1.115906268 | 1.230570609 | 1.075581823 | 1.048285585 | 1.02977064
512 | 8.171549127 | 4.182283461 | 2.37535305 | 1.54202412 | 1.1690348 | 1.054650104 | 1.015366906 | 1.153238581 | 0.993319168 | 0.998864737 | 0.981392837
1024 | 8.533398132 | 4.175120792 | 2.209645233 | 1.412410651 | 1.055442085 | 0.938202817 | 1.122801927 | 0.940661156 | 0.888767412 | 0.914867532 | 0.92237305

</byte-sheet-html-origin>

This is mainly because when the number of worker threads is relatively small, setting the number of locks to 4 times has a significant performance improvement compared to 2 times.

image

wathenjiang: For complete testing, please refer to https://gist.github.com/wathenjiang/30b689a7ef20b4ea667a2e8f358c321d

I think this performance test can provide a reference for how many mutexes are needed for OwnedTasks.

tokio/src/util/sharded_list.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /// Returns whether the linked list does not contain any node.
+    pub(crate) fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Gets the shard size of this SharedList.
+    ///
+    /// Used to help us to decide the parameter `shard_id`` of the `pop_back` method.
```
Darksonn, line 112:
```suggestion
    /// Used to help us to decide the parameter `shard_id` of the `pop_back` method.
```

Darksonn (approved):
Thank you.

PR #6010, opened 2023-09-15
Add recv_many for mpsc channels

aschweig:
Introduces recv_many that allows mpsc receivers to get all available messages on a queue in bulk, as opposed to one-at-a-time.  In the use case where message arrivals are bursty and only the last sent message matters, e.g., when updating a dashboard, this approach provides an easy approach to avoiding unnecessary work resulting in substantial speedup.  The approach can also provide a modest raw speedup over `recv` as shown in the 'uncontented_bounded_updater_*' and `uncontented_unbounded_updater_*` benchmarks included in this PR.

Sample usage from proposed test in `sync_mpsc.rs`:
```
async fn send_recv_many_unbounded() {
    let (tx, mut rx) = mpsc::unbounded_channel::<i32>();

    // Using `try_send`
    assert_ok!(tx.send(7));
    assert_ok!(tx.send(13));
    assert_ok!(tx.send(100));
    assert_ok!(tx.send(1002));

    let mut buffer = vec![0; 0];
    assert_eq!(rx.recv_many(&mut buffer).await, 4);
    assert_eq!(vec![7,13,100,1002], buffer);
    assert!(buffer.capacity() >= 4);

    drop(tx);

    assert!(rx.recv().await.is_none());
}
```

Calls to `recv_many` will clear the passed-in buffer and, if needed, reserve additional capacity.

Pull request for feature enhancement discussed here:

https://github.com/tokio-rs/tokio/issues/5844

edit:clarified origin of code sample.


Darksonn:
>  In the use case where message arrivals are bursty and only the last sent message matters, e.g., when updating a dashboard, this approach provides an easy approach to avoiding unnecessary work resulting in substantial speedup.

Why not use the watch channel for this?

aschweig:
Agree that the watch channel is appropriate if one target is to be updated.

`recv_many` enables use of a single channel for an arbitrary number of topics while retaining FIFO mailbox order; the latter property is important for the calculation of some rolling statistics like sliding window averages (e.g., average of last 5 observations).  It also allows the receiver to measure channel volume. In the actor use-case, I find the ergonomics of a single channel/ActorMessage preferable to using `tokio::select!` over distinct control and data channels; `recv_many` allows the single channel to achieve some of the message eliding gains of `watch`.

aschweig:
I'd like to solicit feedback on adding another parameter to `recv_many` to set the maximum number of retrieved elements in the buffer, where a non-positive value would be unbounded. For example, the following would populate no more than 1000 elements:

`rx.recv_many(&mut buffer, 1000)` // at most 1000 elements populated; helpful?

Are  there meaningful cases when memory is at a premium but an unbounded queue is required?  Or, thinking about an actor, for liveness reasons, an actor might want to limit the number of messages received in one chunk.  While actors could  `yield_now().await;` while processing very large message buffers -- the goal is to reduce complexity and this helps along these lines.

aschweig:
> Usually this kind of API never increases the capacity of the vector unless it's equal to the length. This defines a limit.

I think I am a bit confused on your meaning -- can you point me in the direction of some examples?  Couldn't this again result in unbounded capacity growth if the caller fails to clear the buffer between calls?

In the proposed chan.rs `fn recv_many` -- the capacity is currently increased if the number of acquired permits suggests more messages than the vector buffer capacity, with the goal of anticipating the total required capacity up front:

```
                            let capacity = self.inner.semaphore.num_acquired();
                            if buffer.capacity() < capacity {
                                buffer.reserve(capacity);
                            }
```

Maybe the cleanest way to bound the buffer is to allow `recv_many` to increase the capacity to some small constant, e.g., 2 if the capacity is zero. I choose the magic value 2 here as the smallest number consistent with receiving more than one.  Users who want a specific additional capacity would need to reserve that capacity up-front before the call.

Darksonn:
I mean that if `buffer.capacity() < capacity`, then it doesn't receive all of the messages.

But an argument with a limit is also fine.

tokio/src/sync/mpsc/chan.rs
```diff
@@ -293,6 +297,79 @@ impl<T, S: Semaphore> Rx<T, S> {
...
+                                        buffer.push(value);
+                                        next = rx_fields.list.peek();
+                                        true
+                                    }
+                                    _ => false,
+                                })
+                            {}
+                            self.inner.semaphore.add_permits(buffer.len());
+                            coop.made_progress();
+                            return Ready(buffer.len());
```
Darksonn, line 337: It is not obvious to me that this code is correct. Would it be possible to simplify it?

tokio/src/sync/mpsc/list.rs
```diff
@@ -252,6 +252,20 @@ impl<T> Rx<T> {
         }
     }
 
+    /// Peeks at the next value on the queue, if any.
+    pub(crate) fn peek(&mut self) -> Option<block::Read<T>> {
```
Darksonn, line 256: What does this function do? It's name is `peek`, which makes it sounds like it does _not_ consume the next value, but the implementation looks like it consumes the next value.

tokio/src/sync/mpsc/chan.rs
```diff
@@ -41,9 +41,13 @@ pub(crate) trait Semaphore {
 
     fn add_permit(&self);
 
+    fn add_permits(&self, n: usize);
+
     fn close(&self);
 
     fn is_closed(&self) -> bool;
+
+    fn num_acquired(&self) -> usize;
```
Darksonn, line 50: What does this function do? The name is unclear to me.

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -229,6 +229,14 @@ impl<T> Receiver<T> {
         use crate::future::poll_fn;
         poll_fn(|cx| self.chan.recv(cx)).await
     }
+    /// Receives the all available values for this receiver
```
Darksonn, line 232: Please put an empty line here.

tokio/src/sync/mpsc/chan.rs
aschweig: Colloquially, it is a count of yet-to-be-received successfully sent messages.  The Semaphore has the property that `is_idle` is true if this function returns 0.  I see what you mean.  Do you find `num_used` or `used_permits`[ as proposed here](https://stackoverflow.com/a/29453095) more clear?

tokio/src/sync/mpsc/list.rs
aschweig: The fn `peek` is modeled after `pop` above it.  The input is mutable due to the signature `try_advancing_head(&mut self)`.  The list maintains an index into the current message `self.index` whose contents are stored in finite-sized blocks organized as a linked list, with head `self.head`.  If the head block cannot contain the message for `self.index` then `self.head` should be advanced.  Finally, because `peek` doesn't mutate `self.index` -- the next message remains the same before and after the call.  If/when `pop` is subsequently called; the `self.head` pointer has already been updated and at that time `reclaim_blocks` can be called because the caller now has ownership of the data.

aschweig:
> Usually this kind of API never increases the capacity of the vector unless it's equal to the length. This defines a limit.

I think I am a bit confused on your meaning -- can you point me in the direction of some examples?  Couldn't this again result in unbounded capacity growth if the caller fails to clear the buffer between calls?

In the proposed chan.rs `fn recv_many` -- the capacity is currently increased if the number of acquired permits suggests more messages than the vector buffer capacity, with the goal of anticipating the total required capacity up front:

```
                            let capacity = self.inner.semaphore.num_acquired();
                            if buffer.capacity() < capacity {
                                buffer.reserve(capacity);
                            }
```

Maybe the cleanest way to bound the buffer is to allow `recv_many` to increase the capacity to some small constant, e.g., 2 if the capacity is zero. I choose the magic value 2 here as the smallest number consistent with receiving more than one.  Users who want a specific additional capacity would need to reserve that capacity up-front before the call.


> I mean that if `buffer.capacity() < capacity`, then it doesn't receive all of the messages.
> 
> But an argument with a limit is also fine.

I am going to rework my implementation to accept a capacity argument.  That will essentially eliminate the interaction with the Semaphore.  I will also simplify the logic in chan.rs `recv_many`.

tokio/src/sync/mpsc/chan.rs
Darksonn: How about `num_permits_in_use`?

aschweig:
So, I decided to let the input buffer set the capacity. In the event of a zero-capacity vector is passed, it reserves `BLOCK_CAP` capacity to avert a run-time error.  I removed `peek()` and `num_acquired()` as they are not required for the `recv_many` use-case.

tokio/src/sync/mpsc/chan.rs
```diff
@@ -293,6 +295,79 @@ impl<T, S: Semaphore> Rx<T, S> {
...
 
+    /// Receives values into `buffer` up to its capacity
+    ///
+    /// If the buffer initially has 0 capacity, reserves `super::BLOCK_CAP` elements
+    pub(crate) fn recv_many(&mut self, cx: &mut Context<'_>, buffer: &mut Vec<T>) -> Poll<usize> {
+        use super::block::Read;
+        if buffer.capacity() == 0 {
+            buffer.reserve(super::BLOCK_CAP);
+        }
+        buffer.clear();
```
Darksonn, line 306: I don't think we should clear the buffer. That's up to the user. I think it's better to support appending to the vector if the user wants that.

tokio/src/sync/mpsc/chan.rs
```diff
@@ -293,6 +295,79 @@ impl<T, S: Semaphore> Rx<T, S> {
...
+                                // dropped. Dropping a tx handle releases memory,
+                                // which ensures that if dropping the tx handle is
+                                // visible, then all messages sent are also visible.
+                                assert!(self.inner.semaphore.is_idle());
+                                coop.made_progress();
+                                return Ready(buffer.len());
+                            }
+
+                            None => {
+                                working = false; // fall through
```
Darksonn, line 340: It's okay to use `break` here.

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -230,6 +230,38 @@ impl<T> Receiver<T> {
...
+    ///     tx.send("hello").await.unwrap();
+    ///     tx.send("world").await.unwrap();
+    ///
+    ///     // At most 5 at a time
+    ///     let mut buffer : Vec<&str> = Vec::with_capacity(5);
+    ///     assert_eq!(2, rx.recv_many(&mut buffer).await);
+    ///     assert_eq!(vec!["hello","world"], buffer);
+    /// }
+    /// ```
+    pub async fn recv_many(&mut self, buffer: &mut Vec<T>) -> usize {
```
Darksonn, line 260: I would document that this method only returns zero if the channel is closed.

tokio/src/sync/mpsc/chan.rs
aschweig: There is a clear performance case to be made; particularly if this avoids copying elements in the caller.

Can you suggest the behavior if the buffer has already hit capacity at the time of the `recv_many` call?  `assert!(buffer.capacity() > buffer.len());` -- or `return Pending;` -- if there is a message that cannot be retrieved?

Along these lines, it may also make sense to eliminate `buffer.reserve(super::BLOCK_CAP);` -- I had included that behavior to make `recv_many` unfailing, but with this change the method can fail when the buffer's _full_ (len = capacity).  A zero-length buffer is -- in this sense -- always _full_.

Darksonn: There are two options that make sense to me:

 * Add a third argument that specifies the maximum number of elements to add to the vector. In this case, the capacity is irrelevant, and we will resize the buffer if necessary.
 * Or use the capacity like this:

```rs
if buffer.len() == buffer.capacity() {
    buffer.reserve(super::BLOCK_CAP);
}
let max_number_added = buffer.capacity() - buffer.len();
```

In either case, I prefer that we do not clear the buffer.

tokio/src/sync/mpsc/bounded.rs
aschweig: fixed

tokio/src/sync/mpsc/chan.rs
aschweig: simplified when reworked.

tokio/src/sync/mpsc/list.rs
aschweig: peel and try_advancing_head have been removed.

tokio/src/sync/mpsc/chan.rs
aschweig: agreed.

tokio/src/sync/mpsc/bounded.rs
aschweig: documentation was updated to reflect this.

tokio/src/sync/mpsc/chan.rs
aschweig: I adopted the second approach in the latest push.

tokio/src/sync/mpsc/unbounded.rs
```diff
@@ -172,6 +172,41 @@ impl<T> UnboundedReceiver<T> {
...
+    ///
+    /// This method returns the number of values populated in `buffer`;
+    /// it returns 0 if the channel has been closed and there
+    /// are no remaining messages in the channel.
+    /// If there are no messages in the channel's buffer, but the channel has
+    /// not yet been closed, this method will sleep until a message is sent or
+    /// the channel is closed.
+    ///
+    /// If at the time of the call the buffer has no unused capacity,
+    /// `BLOCK_CAP` additional elements are reserved.
```
Darksonn, line 185: The user doesn't know what `BLOCK_CAP` is because it's not a public constant. I wouldn't explain how much space we reserve. That will let us change it in the future.

tokio/src/sync/mpsc/unbounded.rs
```diff
@@ -172,6 +172,41 @@ impl<T> UnboundedReceiver<T> {
...
     }
 
+    /// Receives the next values for this receiver and extends `buffer`.
+    ///
+    /// This method returns the number of values populated in `buffer`;
+    /// it returns 0 if the channel has been closed and there
+    /// are no remaining messages in the channel.
+    /// If there are no messages in the channel's buffer, but the channel has
+    /// not yet been closed, this method will sleep until a message is sent or
+    /// the channel is closed.
```
Darksonn, line 182: I would split this into several paragraphs, like how it is done in the documentation for `recv`. In fact, it would be fine if you copy the documentation for `recv` and adjust it.

tokio/src/sync/mpsc/unbounded.rs
```diff
@@ -172,6 +172,41 @@ impl<T> UnboundedReceiver<T> {
...
+    ///
+    /// If at the time of the call the buffer has no unused capacity,
+    /// `BLOCK_CAP` additional elements are reserved.
+    ///
+    /// # Example:
```
Darksonn, line 187: We generally always set up the heading like this, even if there's only one example.
```suggestion
    /// # Examples
```

Darksonn (commented):
At this point, I think the main thing remaining is adjusting documentation.

tokio/src/sync/mpsc/unbounded.rs
aschweig: I reworked the documentation based on `recv`, but changed the phrase "channel's buffer" to "channel's queue" to avoid confusion with the parameter named `buffer`.

tokio/src/sync/mpsc/unbounded.rs
aschweig: fixed in latest push.

tokio/src/sync/mpsc/unbounded.rs
aschweig: Fixed in latest push.

tokio/src/sync/mpsc/unbounded.rs
```diff
@@ -172,6 +172,41 @@ impl<T> UnboundedReceiver<T> {
...
+    ///     tokio::spawn(async move {
+    ///         tx.send("hello").unwrap();
+    ///     });
+    ///
+    ///     let mut buffer : Vec<&str> = Vec::with_capacity(5);
+    ///     assert_eq!(1, rx.recv_many(&mut buffer).await);
+    ///     assert_eq!(vec!["hello"], buffer);
+    ///     assert_eq!(0, rx.recv_many(&mut buffer).await);
+    /// }
+    pub async fn recv_many(&mut self, buffer: &mut Vec<T>) -> usize {
```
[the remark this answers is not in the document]
aschweig: fixed in latest push.

tokio/src/sync/mpsc/chan.rs
```diff
@@ -293,6 +295,83 @@ impl<T, S: Semaphore> Rx<T, S> {
...
+                            None => {
+                                break; // fall through
+                            }
+                        }
+                    }
+                    let number_added = buffer.len() - initial_length;
+                    if number_added > 0 {
+                        if number_added > 0 {
+                            self.inner.semaphore.add_permits(number_added);
+                        }
```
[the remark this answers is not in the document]
aschweig: fixed. smh.

tokio/src/sync/mpsc/unbounded.rs
```diff
@@ -172,6 +172,56 @@ impl<T> UnboundedReceiver<T> {
         poll_fn(|cx| self.poll_recv(cx)).await
     }
 
+    /// Receives the next values for this receiver and extends `buffer`.
+    ///
+    /// This method returns the number of values populated in `buffer`.
```
Darksonn, line 177: I think the word "populated" isn't a great fit.
```suggestion
    /// This method returns the number of values added to `buffer`.
```

tokio/src/sync/mpsc/unbounded.rs
```diff
@@ -172,6 +172,56 @@ impl<T> UnboundedReceiver<T> {
...
+    /// further values can ever be received from this `Receiver`. The channel is
+    /// closed when all senders have been dropped, or when [`close`] is called.
+    ///
+    /// If there are no messages in the channel's queue, but the channel has
+    /// not yet been closed, this method will sleep until a message is sent or
+    /// the channel is closed.
+    ///
+    /// If at the time of the call `buffer` has no unused capacity,
+    /// additional elements are reserved. Otherwise `recv_many` populates
+    /// `buffer` with no more elements than its unused capacity.
```
Darksonn, line 190: Can we reverse this sentence? I think the case where `len < capacity` is more common, so I'd like to explain it first.

tokio/src/sync/mpsc/unbounded.rs
```diff
@@ -172,6 +172,56 @@ impl<T> UnboundedReceiver<T> {
...
+    ///
+    /// # Cancel safety
+    ///
+    /// This method is cancel safe. If `recv` is used as the event in a
```
Darksonn, line 194:
```suggestion
    /// This method is cancel safe. If `recv_many` is used as the event in a
```

tokio/src/sync/mpsc/unbounded.rs
```diff
@@ -172,6 +172,56 @@ impl<T> UnboundedReceiver<T> {
...
+    /// This method returns the number of values populated in `buffer`.
+    ///
+    /// This method returns 0 if the channel has been closed and there are
+    /// no remaining messages in the channel's queue. This indicates that no
+    /// further values can ever be received from this `Receiver`. The channel is
+    /// closed when all senders have been dropped, or when [`close`] is called.
+    ///
+    /// If there are no messages in the channel's queue, but the channel has
+    /// not yet been closed, this method will sleep until a message is sent or
+    /// the channel is closed.
```
Darksonn, line 186: I'd like to add some more emphasis on the fact that if it returns 0, then the channel is closed. Perhaps we can swap these paragraphs, and add something to the paragraph about 0 along the lines of "this method will never return 0 if the channel is not closed" or "if the channel is not closed, then this method will never return 0".

tokio/src/sync/mpsc/unbounded.rs
```diff
@@ -172,6 +172,56 @@ impl<T> UnboundedReceiver<T> {
...
+    ///     tokio::spawn(async move {
+    ///         tx.send("hello").unwrap();
+    ///     });
+    ///
+    ///     let mut buffer : Vec<&str> = Vec::with_capacity(5);
+    ///     assert_eq!(1, rx.recv_many(&mut buffer).await);
+    ///     assert_eq!(vec!["hello"], buffer);
+    ///     assert_eq!(0, rx.recv_many(&mut buffer).await);
+    /// }
+    /// ```
```
Darksonn, line 219: The examples are very different. Perhaps we could use the same example? I imagine a good example could work like this:

 * First send two messages, then receive both in one go.
 * Then spawn a task that sends a message, and receive it with `recv_many`. Include a comment saying that this call is guaranteed to not return 0 since the channel is not closed.
 * Drop the last sender.
 * Then call `recv_many` again and get 0 because the channel is closed.

tokio/src/sync/mpsc/unbounded.rs
Darksonn: Sounds good!

tokio/src/sync/mpsc/unbounded.rs
aschweig: Yes.  I like:

_If at the time of the call `buffer` has unused capacity, `recv_many` extends the buffer with no more elements than
its unused capacity. Should `buffer` have no initial unused capacity, additional elements are first reserved._

tokio/src/sync/mpsc/unbounded.rs
aschweig: The zero case is analogous to the `None` case with `recv`.  Does it make sense to change `recv`'s documentation as well?

I think the simplest accommodation is to add _only_, e.g.,:
_This method returns 0 only if..._
_This method returns `None` only if..._

tokio/src/sync/mpsc/unbounded.rs
aschweig: I've set it up so that all senders are dropped to close the channel.
Two paths not covered:

- Demonstrate that the buffer's unused capacity is a bound on the number of elements returned.  
- Close the channel but have outstanding permits/messages; similar to the first doc example for `recv`

tokio/src/sync/mpsc/unbounded.rs
Darksonn: I don't think it's quite as important for `recv` because people are less likely to be confused about `None`. I've seen confusion about 0 = closed a lot of times with the `read` method from the IO traits.

aschweig: I've again strengthened the language in today's push,

> This method will never return 0 unless the channel has been closed and
> there are no remaining messages in the channel's queue.

Does this work?

tokio/src/sync/mpsc/unbounded.rs
aschweig: I've added two new tests analogous to the doctests to sync_mpsc.rs to cover these behaviors explicitly.

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -230,6 +230,74 @@ impl<T> Receiver<T> {
...
 
+    /// Receives the next values for this receiver and extends `buffer`.
+    ///
+    /// This method returns the number of values added to `buffer`.
+    ///
+    /// If there are no messages in the channel's queue, but the channel has
+    /// not yet been closed, this method will sleep until a message is sent or
+    /// the channel is closed.  Note that if [`close`] is called, but there are
+    /// still outstanding [`Permits`] from before it was closed, the channel is
+    /// not considered closed by `recv` until the permits are released.
```
Darksonn, line 241:
```suggestion
    /// not considered closed by `recv_many` until the permits are released.
```

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -230,6 +230,74 @@ impl<T> Receiver<T> {
...
+    ///
+    /// # Cancel safety
+    ///
+    /// This method is cancel safe. If `recv_many` is used as the event in a
+    /// [`tokio::select!`](crate::select) statement and some other branch
+    /// completes first, it is guaranteed that no messages were received on this
+    /// channel.
+    ///
+    /// [`close`]: Self::close
+    /// [`Permits`]: struct@crate::sync::mpsc::Permit///
```
Darksonn, line 262:
```suggestion
    /// [`Permits`]: struct@crate::sync::mpsc::Permit
```

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -230,6 +230,74 @@ impl<T> Receiver<T> {
...
+    /// This method will never return 0 unless the channel has been closed and
+    /// there are no remaining messages in the channel's queue. This indicates
+    /// that no further values can ever be received from this `Receiver`. The
+    /// channel is closed when all senders have been dropped, or when [`close`]
+    /// is called.
+    ///
+    /// If at the time of the call `buffer` has unused capacity,
+    /// `recv_many` extends the buffer with no more elements than
+    /// its unused capacity. Should `buffer` have no initial unused
+    /// capacity, additional elements are first reserved.
```
Darksonn, line 252: I was looking at this paragraph, and I think it could be improved like this:
```suggestion
    /// If `buffer` has unused capacity, then this call will not reserve
    /// additional space in `buffer`. This means that the maximum number of
    /// received messages is `buffer.capacity() - buffer.len()`. However, if
    /// the capacity is equal to the length, then this call will increase the
    /// capacity to make space for additional elements.
```
This actually raises a question: Perhaps it makes sense to only reserve space when we return a message? This way, we don't consume extra memory until a message arrives, and if the channel gets closed, we don't reserve any space.

What do you think?

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -230,6 +230,74 @@ impl<T> Receiver<T> {
...
+    ///         tokio::spawn(async move {
+    ///             tx.send("third").await.unwrap();
+    ///         });
+    ///         // The 'tx' is dropped, but `recv_many`
+    ///         // is guaranteed not to return 0 as the channel
+    ///         // is not yet closed.  If the buffer is full, the next
+    ///         // call to `recv_many` reserves additional capacity.
+    ///         assert_eq!(1, rx.recv_many(&mut buffer).await);
+    ///         assert_eq!(vec!["first", "second", "third"], buffer);
+    ///     }
```
Darksonn, line 290: Instead of a scope, please use `drop(tx)`. I think that is more clear.

Also, please add an empty line to visually separate each part of the test.

Darksonn (commented):
I have a few more documentation nits. I know we've gone through many review iterations for this. Please let me know if it's been too many. I would be happy to merge it as it is now if so.

tokio/src/sync/mpsc/unbounded.rs
Darksonn: Yes, it looks great now.

tokio/src/sync/mpsc/bounded.rs
aschweig: I like your language here better.

I also like the idea of not allocating memory unless needed.  If the channels are very lightweight perhaps there are often created and a finite number of messages sent and then closed.  Why force the caller to always allocate one more message than is actually sent?

In chan.rs, `recv_many`:
```
let mut insufficient_capacity = buffer.capacity() == buffer.len();
...
                () => {
                    while (buffer.len() < buffer.capacity() || insufficient_capacity) {
                        match rx_fields.list.pop(&self.inner.tx) {
                            Some(Read::Value(value)) => {
                                if insufficient_capacity {
                                    buffer.reserve(super::BLOCK_CAP);
                                    insufficient_capacity = false;
                                }
                                buffer.push(value);
                            }
```

aschweig: Does it make sense to urge users to manage the buffer's capacity outside `recv_many`?
e.g.,

If `buffer` has unused capacity, then this call will not reserve
additional space in `buffer`. This means that the maximum number of
received messages is `buffer.capacity() - buffer.len()`.  
Efficient client code should ensure `buffer` has unused capacity,
but if the capacity is equal to the length and there is at least one message
in the channel's queue, then this call will _conservatively_ increase the capacity
to make space for additional elements.

tokio/src/sync/mpsc/chan.rs
```diff
@@ -293,6 +295,82 @@ impl<T, S: Semaphore> Rx<T, S> {
...
+                    }
+                    let number_added = buffer.len() - initial_length;
+                    if number_added > 0 {
+                        self.inner.semaphore.add_permits(number_added);
+                        coop.made_progress();
+                        return Ready(number_added);
+                    }
+                };
+            }
```
Darksonn, line 353: This is getting complicated again. Let's try to simplify a bit. How about this?

 * At the beginning instead of `insufficient_capacity`, define a constant called `max_length` set to `buffer.capacity()`, or `buffer.len() + BLOCK_CAP` when we want to resize.
 * In the while loop, just use `buffer.len() < max_length`.
 * We don't resize it explicitly. Instead, we just call `Vec::push`, which will resize for us.

aschweig:
My concern is on the second call after the push that effectively reserves
the buffer.  As capacity grows exponentially on push, then naive use of
recv_many eventually results in taking in all elements in future calls.
(So why initially even limit max_length to buffer.len() + BLOCK_CAP?)

tokio/src/sync/mpsc/chan.rs
aschweig: Reposted my email reply in this thread:

> My concern is on the second call to after the push that effectively reserves
> the buffer.  As capacity grows exponentially on push, then naive use of
> recv_many eventually results in taking in all elements in future calls.
> (So why initially even limit max_length to buffer.len() + BLOCK_CAP?)

_I promise to not make a habit of replying from email._

I suggest the following use cases:

1. The caller clears the buffer between calls to `recv_many`; capacity sets the number of messages reasonable to process in one call so as to not affect liveness.
2.  The caller can anticipate the total number of messages they will receive and want to pre-allocate that space by setting the capacity explicitly; liveness is known to not be a concern.
3. The caller wants to indefinitely extend the buffer.  It makes sense to the buffer's capacity grows exponentially, e.g., how push does it.
  i. The caller knows that the rate of sent messages on the channel will not affect liveness no matter how many are retrieved.
  ii. The caller needs to limit how many messages are added to the buffer at a time to ensure liveness.

Use-case 3 seems to be in conflict with use-case 1.  It seems clear that one possible (maybe the only) resolution is to pass in an argument to limit the number of messages retrieved.  Furthermore, passing in a bound to the number of retrieved messages helps generalize case 2.

Thoughts?

Darksonn: It's not clear to me how the use-cases are in conflict, but I am also ok with using an argument to limit the number of messages if you prefer that.

aschweig: I assume there is a very busy sender.

If the buffer holds say, 1,000,0000 elements and has no unused capacity, upon the call to `recv_many` the capacity is doubled and `BLOCK_CAP` (few) values are retrieved. On the _next_ call, `recv_many` will  populate 1,000,0000 - `BLOCK_CAP` elements in a single call (as necessitated by how use-case #1 works -- which allows up to the capacity to be retrieved).  In the next call, `recv_many` finds the buffer full and doubles it again, and `BLOCK_CAP` elements are retrieved.  On the next call to `recv_many` -- 2,000,000 - BLOCK_CAP elements are retrieved in a single call.

To me this is a very odd behavior where between calls it oscillates between returning a small number of values (on the step where capacity is doubled) and an exponentially increasing number of values (because of significant unused capacity.)  

This fails to satisfy use-case 3.ii -- where the "caller needs to limit how many messages are added to the buffer at a time to ensure liveness."

This was why I added constant capacity on each call -- so that naive use of `recv_many` only resulted in, at most, a fixed number of values being added to the buffer with each `recv_many` call.  (And why an earlier commit simply panicked if there was no spare capacity.)

aschweig: I assume there is unlimited flood of messages being sent.  The problem arises when a `Vec::push` adds significant unused capacity to the buffer and the behavior of use-case 1 takes over, resulting in a potential huge number of messages being put into the buffer at once, violating 3.ii -- too many messages at once can make things unresponsive. The problem becomes worse as the buffer grows. 

To resolve this -- I have re-worked to pass in a limit to the number of received values.

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -230,6 +230,82 @@ impl<T> Receiver<T> {
         poll_fn(|cx| self.chan.recv(cx)).await
     }
 
+    /// Receives the next values for this receiver and extends `buffer`.
+    ///
+    /// This method extends `buffer` by no more than a fixed number
+    /// of values as specified by `limit`.  If `limit` is zero,
+    /// then a default is used. The return value is the number
+    /// of values added to `buffer`.
```
Darksonn, line 238: If the limit is zero, then I think it would be okay to just return 0 immediately.

tokio/src/sync/mpsc/chan.rs
Darksonn: I guess that makes sense.

tokio/src/sync/mpsc/chan.rs
```diff
@@ -293,6 +295,85 @@ impl<T, S: Semaphore> Rx<T, S> {
...
+    /// For `limit` = 0, receives at most `super::BLOCK_CAP` values.
+    pub(crate) fn recv_many(
+        &mut self,
+        cx: &mut Context<'_>,
+        buffer: &mut Vec<T>,
+        limit: usize,
+    ) -> Poll<usize> {
+        use super::block::Read;
+
+        let mut to_go = if limit > 0 { limit } else { super::BLOCK_CAP };
```
Darksonn, line 310: This `to_go` variable seems weird to me. Why not use a `maximum_length` instead?

aschweig: I have renamed `to_go` to `remaining` in the latest push -- `to_go` suggests work to be done.  The decrement loop is set up that way because I think someone will pass `usize::max_value()` to `limit` with `buffer.length() > 0` length and this makes the function's behavior simple and clear.

tokio/src/sync/mpsc/bounded.rs
aschweig: If the caller wants a no-op, they can instead simply not-call the method, instead of calling with a zero limit.

Furthermore, immediately returning 0 means that this behavior becomes unique to a single value of `limit`; all other values of `limit` only return 0 if the channel is closed and no values are pending.  So it breaks the API's guarantee for a case of no practical value to someone seeking to receive messages on the channel.

I considered the following possible ways to handle a 0 value for limit:

- The fail-fast approach -- `assert!(limit > 0);`.
- Have it so that `limit=0` acts the same way as `limit=usize::max_value()`.
- Have it so that `limit=0` results in some reasonable default behavior, e.g., retrieve 2 or 8 messages or BLOCK_CAP.  I liked BLOCK_CAP because it corresponded to the internal chunking used.

Darksonn: Hmm. I think immediately returning 0 is the least surprising behavior. After all, you asked to receive zero messages, and you got zero messages.

aschweig: Recall, you wrote:

>  Perhaps we can swap these paragraphs, and add something to the paragraph about 0 along the lines of "this method will never return 0 if the channel is not closed" or "if the channel is not closed, then this method will never return 0".

Nonetheless, I see that the same no-op convention is used in unistd.h's `read` function... and evidence of users running into the types of bugs I anticipate:
https://stackoverflow.com/a/3074217

If this is acceptable, I'll update the handling and the documentation.

Darksonn: If what is acceptable?

I mainly see people be confused about length zero reads when they do this:
```rs
let mut vec = Vec::with_capacity(1024);
io.read(&mut vec);
```
and are surprised because `&mut vec` becomes a slice of length zero.

But that doesn't happen in our case.

aschweig: Understood regarding `io.read(&mut vec);` -- here the issue is resolved as we take a Vec.

The behavior of `recv_many(buf, 0)` is similar to calling [`rd.take(0)`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#method.take).  I am unclear when in an async context `take(0)` is ever useful -- but it is allowed -- so at least `recv_many(buf, 0)` is no worse.

Contrived example adapted from [Example: split a file into chunks](https://geekingfrog.com/blog/post/getting-things-done-with-async)
```
#[tokio::test]
async fn take_zero() {
    use tokio::fs::File;
    use std::path::PathBuf;
    for chunk_size in [0, 1, 1024] {
        let mut input_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
        input_path.push("tests/io_take.rs");
        let input_file = File::open(input_path).await.unwrap();
        let mut output_path = PathBuf::from(env!("CARGO_TARGET_TMPDIR"));
        output_path.push("takezero");
        if output_path.exists() {
            tokio::fs::remove_file(&output_path).await.unwrap();
        }
        assert!(!output_path.exists());

        // the `.take` method is coming from the AsyncReadExt trait imported above
        let mut reader = input_file.take(chunk_size);
        let mut output = File::create(&output_path).await.unwrap();
        loop {
            let bytes_copied = tokio::io::copy(&mut reader, &mut output).await.unwrap();
            if bytes_copied == 0 {
                // EOF on the initial reader, remove the empty file we created
                break;
            }
            // our reader is now exhausted, but that doesn't mean the underlying reader
            // is. So we recover it, and we create the next chunked reader
            reader = reader.into_inner().take(chunk_size);
        }
        let file_size = tokio::fs::metadata(output_path).await.unwrap().len();
        assert!((chunk_size > 0 && file_size > 0) || chunk_size == 0 && file_size == 0);
    }
}
```

I think this is now resolved (for me).

Darksonn: I think allowing a zero length is useful for cases where the length is computed dynamically. I've seen code along these lines:
```rs
let len = io.read_u64().await? as usize;
let mut buffer = Vec::with_capacity(len);
io.take(len).read_to_end(&mut buffer)?;
```
if there are length-zero messages, then the above makes use of `take(0)` behavior sometimes.

Darksonn (commented):
Thank you for being patient with my reviews. I'm sorry it took so many rounds. I think we're reaching the end.

tokio/tests/sync_mpsc.rs
```diff
@@ -120,6 +120,35 @@ async fn async_send_recv_with_buffer() {
...
+    let mut buffer = Vec::<i32>::with_capacity(3);
+
+    // With `limit=0` does not sleep, returns immediately
+    assert_eq!(0, rx.recv_many(&mut buffer, 0).await);
+
+    tokio::spawn(async move {
+        assert_ok!(tx.send(1).await);
+        assert_ok!(tx.send(2).await);
+        assert_ok!(tx.send(7).await);
+    });
```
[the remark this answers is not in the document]
aschweig: I have adopted this for `recv_many` in the latest push.  I note that the pattern exists in other tests --e.g., in `async_send_recv_with_buffer` -- these I have left unchanged.

tokio/tests/sync_mpsc.rs
```diff
@@ -120,6 +120,35 @@ async fn async_send_recv_with_buffer() {
...
+    while recv_count < 3 {
+        let n = rx.recv_many(&mut buffer, limit).await;
+        recv_count += n;
+        assert_eq!(buffer.len(), recv_count);
+        assert!(buffer.iter().sum::<i32>() <= 10);
+    }
+    assert_eq!(buffer.iter().sum::<i32>(), 10);
+    assert_eq!(3, recv_count);
+    assert_eq!(recv_count, buffer.len());
+    assert_eq!(None, rx.recv().await);
```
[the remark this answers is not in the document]
aschweig: I have simplified and strengthened this test with the latest push.

tokio/src/sync/mpsc/chan.rs
```diff
@@ -293,6 +295,89 @@ impl<T, S: Semaphore> Rx<T, S> {
...
+        if limit == 0 {
+            return Ready(0usize);
+        }
+
+        let mut remaining = limit;
+        let initial_length = buffer.len();
+
+        ready!(crate::trace::trace_leaf(cx));
+        // Keep track of task budget
+        let coop = ready!(crate::runtime::coop::poll_proceed(cx));
```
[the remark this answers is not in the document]
aschweig: No objections.

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -230,6 +230,84 @@ impl<T> Receiver<T> {
         poll_fn(|cx| self.chan.recv(cx)).await
     }
 
+    /// Receives the next values for this receiver and extends `buffer`.
+    ///
+    /// This method extends `buffer` by no more than a fixed number
+    /// of values as specified by `limit`.  If `limit` is zero,
+    /// the function returns immediately with `0`.
```
Darksonn, line 237:
```suggestion
    /// the function immediately returns `0`.
```

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -230,6 +230,84 @@ impl<T> Receiver<T> {
...
+    /// Note that if [`close`] is called, but there are
+    /// still outstanding [`Permits`] from before it was closed, the channel is
+    /// not considered closed by `recv_many` until the permits are released.
+    ///
+    /// For non-zero values of `limit`, this method will never return `0` unless
+    /// the channel has been closed and there are no remaining messages in the
+    /// channel's queue. This indicates that no further values can ever
+    /// be received from this `Receiver`. The channel is closed when all senders
+    /// have been dropped, or when [`close`]
+    /// is called.
```
Darksonn, line 252: Please reflow the text in documentation at 80 characters.

Darksonn (approved):
Thanks! Just a single grammar fix, and I'll merge this.

aschweig:
Thanks for your help on this.  I note that there have been changes to master since approval and that github offers the "Update branch" button.  Should I rebase this PR or leave as-is?

Darksonn:
Sorry for the delay. I've merged it now.

Cheban1996:
I generally expected that this method would wait until it accumulated messages up to limit. Then we could use it together with timeout.

Issue #5020, opened 2022-09-16
LocalEnterGuard would cause current thread runtime to become stuck


futursolo:
**Version**

`tokio v1.21.1`

**Platform**

- aarch64-apple-darwin
- x86_64-apple-darwin
- aarch64-unknown-linux-gnu
- x86_64-unknown-linux-gnu

**Description**

I tried [this code](https://github.com/futursolo/local-enter-guard-test):

```rust
use std::time::Duration;
use tokio::sync::oneshot;
use tokio::task::{spawn_local, LocalSet};
use tokio::time::sleep;

fn main() {
    let rt = tokio::runtime::Builder::new_current_thread()
        .enable_all()
        .build()
        .expect("failed to create runtime");

    let local = LocalSet::new();
    // Comment this and the program would exit as usual.
    let _guard = local.enter();

    let (tx, rx) = oneshot::channel();

    local.block_on(&rt, async move {
        spawn_local(async move {
            sleep(Duration::ZERO).await;

            tx.send(()).expect("failed to send");
        });
        assert_eq!(rx.await, Ok(()));
    });
}

```

I expected to see this happen: 

This program should exit shortly.

Instead, this happened:

The program hangs forever.

Noah-Kennedy:
I stepped through this in a debugger. Looks like the future from `sleep` never completes.

PR #6016, opened 2023-09-19
tokio: distinguish LocalSet::enter() with being polled

inq:
## Motivation

This fixes: https://github.com/tokio-rs/tokio/issues/5020

## Solution

If we enter a `LocalSet`, then the `LocalEnterGuard` sets `CURRENT` with it. But the `CURRENT` is being used with both of these two situations:
1. task::spawn_local()
2. being polled
But as we know, `LocalEnterGuard` should be related only with (1). When we schedule it, we should check if it is being polled, or just being entered. So I added a boolean variable to distinguish these situations.

### DIFF
--- a/tokio/src/task/local.rs
+++ b/tokio/src/task/local.rs
@@ -313,7 +313,9 @@ struct LocalDataEnterGuard<'a> {
 impl<'a> Drop for LocalDataEnterGuard<'a> {
     fn drop(&mut self) {
         self.local_data_ref.ctx.set(self.ctx.take());
-        self.local_data_ref.wake_on_schedule.set(self.wake_on_schedule)
+        self.local_data_ref
+            .wake_on_schedule
+            .set(self.wake_on_schedule)
     }
 }
 
@@ -402,10 +404,15 @@ pub struct LocalEnterGuard {
 
 impl Drop for LocalEnterGuard {
     fn drop(&mut self) {
-        CURRENT.with(|LocalData { ctx, wake_on_schedule }| {
-            ctx.set(self.ctx.take());
-            wake_on_schedule.set(self.wake_on_schedule);
-        })
+        CURRENT.with(
+            |LocalData {
+                 ctx,
+                 wake_on_schedule,
+             }| {
+                ctx.set(self.ctx.take());
+                wake_on_schedule.set(self.wake_on_schedule);
+            },
+        )
     }
 }
 
@@ -447,11 +454,20 @@ impl LocalSet {
     ///
     /// [`spawn_local`]: fn@crate::task::spawn_local
     pub fn enter(&self) -> LocalEnterGuard {
-        CURRENT.with(|LocalData { ctx, wake_on_schedule, .. }| {
-            let ctx = ctx.replace(Some(self.context.clone()));
-            let wake_on_schedule = wake_on_schedule.replace(true);
-            LocalEnterGuard { ctx, wake_on_schedule }
-        })
+        CURRENT.with(
+            |LocalData {
+                 ctx,
+                 wake_on_schedule,
+                 ..
+             }| {
+                let ctx = ctx.replace(Some(self.context.clone()));
+                let wake_on_schedule = wake_on_schedule.replace(true);
+                LocalEnterGuard {
+                    ctx,
+                    wake_on_schedule,
+                }
+            },
+        )
     }
 
     /// Spawns a `!Send` task onto the local task set.
--- a/tokio/src/task/local.rs
+++ b/tokio/src/task/local.rs
@@ -280,10 +280,12 @@ pin_project! {
 
 tokio_thread_local!(static CURRENT: LocalData = const { LocalData {
     ctx: RcCell::new(),
+    entered: Cell::new(false),
 } });
 
 struct LocalData {
     ctx: RcCell<Context>,
+    entered: Cell<bool>,
 }
 
 cfg_rt! {
@@ -360,12 +362,16 @@ const MAX_TASKS_PER_TICK: usize = 61;
 const REMOTE_FIRST_INTERVAL: u8 = 31;
 
 /// Context guard for LocalSet
-pub struct LocalEnterGuard(Option<Rc<Context>>);
+pub struct LocalEnterGuard {
+    ctx: Option<Rc<Context>>,
+    entered: bool,
+}
 
 impl Drop for LocalEnterGuard {
     fn drop(&mut self) {
-        CURRENT.with(|LocalData { ctx, .. }| {
-            ctx.set(self.0.take());
+        CURRENT.with(|LocalData { ctx, entered }| {
+            ctx.set(self.ctx.take());
+            entered.set(self.entered);
         })
     }
 }
@@ -408,9 +414,10 @@ impl LocalSet {
     ///
     /// [`spawn_local`]: fn@crate::task::spawn_local
     pub fn enter(&self) -> LocalEnterGuard {
-        CURRENT.with(|LocalData { ctx, .. }| {
-            let old = ctx.replace(Some(self.context.clone()));
-            LocalEnterGuard(old)
+        CURRENT.with(|LocalData { ctx, entered }| {
+            let ctx = ctx.replace(Some(self.context.clone()));
+            let entered = entered.replace(true);
+            LocalEnterGuard { ctx, entered }
         })
     }
 
@@ -667,21 +674,27 @@ impl LocalSet {
     }
 
     fn with<T>(&self, f: impl FnOnce() -> T) -> T {
-        CURRENT.with(|LocalData { ctx, .. }| {
+        CURRENT.with(|LocalData { ctx, entered, .. }| {
             struct Reset<'a> {
                 ctx_ref: &'a RcCell<Context>,
-                val: Option<Rc<Context>>,
+                entered_ref: &'a Cell<bool>,
+                ctx_val: Option<Rc<Context>>,
+                entered_val: bool,
             }
             impl<'a> Drop for Reset<'a> {
                 fn drop(&mut self) {
-                    self.ctx_ref.set(self.val.take());
+                    self.ctx_ref.set(self.ctx_val.take());
+                    self.entered_ref.set(self.entered_val)
                 }
             }
-            let old = ctx.replace(Some(self.context.clone()));
+            let ctx_old = ctx.replace(Some(self.context.clone()));
+            let entered_old = entered.replace(false);
 
             let _reset = Reset {
                 ctx_ref: ctx,
-                val: old,
+                entered_ref: entered,
+                ctx_val: ctx_old,
+                entered_val: entered_old,
             };
 
             f()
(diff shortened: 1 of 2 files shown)


tokio/tests/task_local_set.rs
```diff
@@ -573,6 +573,25 @@ async fn spawn_wakes_localset() {
     }
 }
 
+#[tokio::test]
+async fn sleep_with_local_enter_guard() {
```
hawkw, line 577: This isn't a big deal, but would you mind maybe adding a comment describing what this test reproduces (and maybe referecning issue https://github.com/tokio-rs/tokio/issues/5020)?

tokio/src/task/local.rs
```diff
@@ -693,21 +706,27 @@ impl LocalSet {
...
-            let old = ctx.replace(Some(self.context.clone()));
+            let ctx_old = ctx.replace(Some(self.context.clone()));
+            let entered_old = entered.replace(false);
 
             let _reset = Reset {
                 ctx_ref: ctx,
-                val: old,
+                entered_ref: entered,
+                ctx_val: ctx_old,
+                entered_val: entered_old,
```
hawkw, line 729: take it or leave it: the `Reset` code got substantially more complex in this change, and it's used in two different places. now that it's no longer fairly trivial, what do you think about moving this out of the `with` and `with_if_possible` functions, so that we can use one implementation in both places. Something like:
```rust
impl LocalData {
    fn enter(&self, context: Rc<Context>) -> Reset<'_> {
        let ctx_val = self.ctx.replace(Some(context));
        let entered_val = self.entered.replace(false);

        Reset {
            ctx_ref: self.ctx,
            entered_ref: self.entered,
            ctx_val: ctx_old,
            entered_val: entered_old,
        }
    }
}

struct Reset<'a> {
    ctx_ref: &'a RcCell<Context>,
    entered_ref: &'a Cell<bool>,
    ctx_val: Option<Rc<Context>>,
    entered_val: bool,
}

impl<'a> Drop for Reset<'a> {
    fn drop(&mut self) {
        self.ctx_ref.set(self.ctx_val.take());
        self.entered_ref.set(self.entered_val)
    }
}
```

and then changing `with` and `with_if_possible` to just call
```rust
let _reset = local_data.enter(self.context.clone());
```

tokio/src/task/local.rs
```diff
@@ -667,21 +674,27 @@ impl LocalSet {
     }
 
     fn with<T>(&self, f: impl FnOnce() -> T) -> T {
-        CURRENT.with(|LocalData { ctx, .. }| {
+        CURRENT.with(|LocalData { ctx, entered, .. }| {
             struct Reset<'a> {
                 ctx_ref: &'a RcCell<Context>,
-                val: Option<Rc<Context>>,
+                entered_ref: &'a Cell<bool>,
```
hawkw, line 680: nit, not a big deal: both `ctx_ref` and `entered_ref` reference fields of the `LocalData`. we could make this struct a word smaller by changing these to a single `&'a LocalData` reference, and just using that to access both `ctx` and `entered`.

tokio/src/task/local.rs
```diff
@@ -360,12 +362,16 @@ const MAX_TASKS_PER_TICK: usize = 61;
 const REMOTE_FIRST_INTERVAL: u8 = 31;
 
 /// Context guard for LocalSet
-pub struct LocalEnterGuard(Option<Rc<Context>>);
+pub struct LocalEnterGuard {
+    ctx: Option<Rc<Context>>,
+    entered: bool,
```
hawkw, line 367: can we add a comment here describing how `entered` is used and what it represents? thanks!

hawkw (commented):
Overall, this change looks good to me --- thanks for adding a test reproducing the bug!

I had some minor suggestions for potentially simplifying some aspects of the implementation a bit. Let me know what you think!

tokio/tests/task_local_set.rs
inq: Okay! I've added.

tokio/src/task/local.rs
inq: It's a great idea! I've moved out, and renamed to more verbose name.

tokio/src/task/local.rs
inq: I've added! Thanks!

tokio/src/task/local.rs
inq: Thank you! I've moved out, and renamed it to more verbose name.

inq:
@hawkw Thank you for providing an awesome review!

tokio/src/task/local.rs
```diff
@@ -288,6 +288,32 @@ struct LocalData {
...
+
+        LocalDataEnterGuard {
+            local_data_ref: self,
+            ctx,
+            entered,
+        }
+    }
+}
+
+struct LocalDataEnterGuard<'a> {
```
hawkw, line 304: could we maybe add a comment on this explaining how it differs from the public `LocalEnterGuard`?  we now have two types in this module with similar names and similar behavior, one of which is private and the other of which is public. So, it would be nice to make the difference clear to readers. We may want to consider giving this type a different name, as well, if you can come up with something more descriptive.

tokio/src/task/local.rs
```diff
@@ -364,6 +390,8 @@ const REMOTE_FIRST_INTERVAL: u8 = 31;
 /// Context guard for LocalSet
 pub struct LocalEnterGuard {
     ctx: Option<Rc<Context>>,
+
+    /// Distinguish whether the context was entered or polled.
```
hawkw, line 394:
```suggestion
    /// Distinguishes whether the context was entered or polled.
```

hawkw:
> @hawkw Thank you for providing an awesome review!

@inq My pleasure, thanks for contributing the fix! :)

tokio/src/task/local.rs
```diff
@@ -706,29 +713,8 @@ impl LocalSet {
...
-
-            let _reset = Reset {
-                ctx_ref: ctx,
-                entered_ref: entered,
-                ctx_val: ctx_old,
-                entered_val: entered_old,
-            };
-
+        let res = CURRENT.try_with(|local_data| {
+            local_data.enter(self.context.clone());
```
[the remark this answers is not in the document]
inq: Huh why did I... thank you!

tokio/src/task/local.rs
```diff
@@ -288,6 +288,32 @@ struct LocalData {
...
+        let entered = self.entered.replace(true);
+
+        LocalDataEnterGuard {
+            local_data_ref: self,
+            ctx,
+            entered,
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
inq: Okay I did it!

tokio/src/task/local.rs
```diff
@@ -288,6 +288,32 @@ struct LocalData {
     entered: Cell<bool>,
 }
 
+impl LocalData {
+    fn enter(&self, ctx: Rc<Context>) -> LocalDataEnterGuard<'_> {
```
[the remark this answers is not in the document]
inq: Of course!

tokio/src/task/local.rs
inq: Thank you!

tokio/src/task/local.rs
inq: While writing the comment, I've realized that there is no need to separate `LocalEnterGuard` with `LocalDataEnterGuard`. So I've unified them!
Was it so radical?

hawkw: hmm, so one concern with combining the two guard types is that dropping a `LocalEnterGuard` will perform a second `CURRENT.with(...) call, rather than borrowing from the already-accessed thread-local as the `Reset` drop guard did. that means that if the `f` closure in `with` and `with_if_possible` panics and the thread-local gets torn down by unwinding *before* the `Drop` impl executes, we may attempt to access the thread local after its destruction, resulting in a double panic.

i'm unsure if this actually happens in practice or not, but it is a meaningful change in semantics relative to the previous implementation. if we're going to combine the two guard types, we should make sure that this case is tested and that we didn't introduce a double panic that didn't already exist.

hawkw: there may already be such a test, I'll have to double-check.

inq: Great point! But If I've understood the situation correctly, 
1. If a panic happens, local variables are dropped first, and then thread_local would be dropped later. for example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4dd7f70a8cdea6441775101582c1e6c3
2. Even though the variable CURRENT is being dropped, then aren't Cells inside it already being dropped? then double-panic problem should be reproduced in the previous code also.

inq: So I think it would not cause the problem, but if we have remaining worries, then it would be okay to separate again!

inq: I was considering about this, but combining two guards may not be related with the original problem.
So let me revert it. Thank you for pointing it! @hawkw

inq:
@hawkw Hello! I am sorry but can we proceed this?

hawkw:
> @hawkw Hello! I am sorry but can we proceed this?

Hi, sorry for not getting back to you sooner --- I'm taking another look now!

tokio/src/task/local.rs
hawkw: Sorry if this is overly nitpicky, but I think this comment could maybe be improved a bit --- it would be nice if it explained *why* we need to differentiate between `enter` and polling the `LocalSet`, and what behavior is controlled based on that. I don't think that's currelty all that clear from reading the code.

tokio/src/task/local.rs
hawkw: Okay, I think the current implementation is fine. We can revisit this in a follow-up PR, potentially.

tokio/tests/task_local_set.rs
```diff
@@ -573,6 +573,27 @@ async fn spawn_wakes_localset() {
     }
 }
 
+/// Checks that the task wakes up with `enter`.
+/// <https://github.com/tokio-rs/tokio/issues/5020>.
```
hawkw, line 577: nit, not important: I might add:
```suggestion
/// Reproduces <https://github.com/tokio-rs/tokio/issues/5020>.
```

tokio/src/task/local.rs
```diff
@@ -967,7 +975,7 @@ impl Shared {
     fn schedule(&self, task: task::Notified<Arc<Self>>) {
         CURRENT.with(|localdata| {
             match localdata.ctx.get() {
-                Some(cx) if cx.shared.ptr_eq(self) => unsafe {
+                Some(cx) if cx.shared.ptr_eq(self) && !localdata.entered.get() => unsafe {
```
hawkw, line 978: it might be worth adding a comment here explaining why we only choose this branch if `entered` is false? although, if we expand the doc comment on `entered`, that could also be fine. really, i just think there ought to be an explicit comment somewhere stating that the `LocalSet` is woken if `entered` is true.

hawkw (approved):
Overall, this looks good to me. I think it would be nice to document the behavior with the entered flag and when the `LocalSet` future is woken, but I feel good about the implementation and tests, and would be happy to merge this PR. Thank you!

tokio/src/task/local.rs
inq: Oops, sorry! I've corrected it!

inq:
@hawkw I am really sorry for delayed works. Your kind correction helped me a lot! Thank you very much!

tokio/src/task/local.rs
```diff
@@ -280,10 +280,41 @@ pin_project! {
...
+    /// Should be called except when we call `LocalSet::enter`.
+    /// Especially when we poll a LocalSet.
+    #[must_use = "dropping this guard will reset the entered state"]
+    fn enter(&self, ctx: Rc<Context>) -> LocalDataEnterGuard<'_> {
+        let ctx = self.ctx.replace(Some(ctx));
+        let entered = self.entered.replace(false);
+        LocalDataEnterGuard {
+            local_data_ref: self,
+            ctx,
+            entered,
```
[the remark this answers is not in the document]
inq: It's because of the name of the user-facing level functions. So this `LocalData::enter` should be called when we poll, and the value `entered` should be set only if user called the `LocalSer::enter`, which is public. 
I agree that it's confused, but could not find suitable naming..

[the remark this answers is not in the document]
hawkw: I agree with @Darksonn, the `entered` field could be renamed to something clearer.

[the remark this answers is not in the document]
Nerdy5k: Agreed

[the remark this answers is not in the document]
inq: Thank you very much! I've renamed it!
But... `wake_on_spawn_local` vs `wake_on_schedule`.. which would be better?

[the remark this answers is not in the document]
Darksonn: `wake_on_schedule` also makes sense to me.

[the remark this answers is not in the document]
inq: Thank you! I've renamed it to `wake_on_schdule`!

Darksonn (approved):
Thanks.

PR #6017, opened 2023-09-19
sync::watch: Fix changed handling on version overflow

uklotzde:
## Motivation

`sync::watch::Receiver`: The asymmetric implementation of `Version::decrement()` fails after an overflow of the version number. This is an edge case that might occur in long running applications.

## Solution

Consistently use *wrapping add/sub* semantics for both `AtomicUsize` and `usize` operations when incrementing/decrementing version numbers. This also allows to start the version numbering at 0 as before.


uklotzde:
I have also added and renamed some constants. They depend on each other and their definition and documentation should be colocated.

Darksonn (approved):
LGTM.

PR #6020, opened 2023-09-20
sync::watch::Receiver: Fix code comment

uklotzde:
The `CLOSED_BIT` in the shared state tracks whether the single sender has been dropped.


Darksonn (approved):
Thanks.

PR #6021, opened 2023-09-21
watch::Receiver::wait_for(): Prevent poisoning of the lock

uklotzde:
## Motivation

If the provided closure panics the lock is poisoned.

## Solution

Adopt the `catch_unwind()` strategy that is already used elsewhere.


Darksonn (approved):
Thanks.

PR #6030, opened 2023-09-23
fix incorrect line number in sync::Mutex documentation

rebekahkim:
## Motivation

The code for dereferencing (*) is on lines 13 and 20.

## Solution

Fix line number


Darksonn (approved):
Thanks.

PR #6037, opened 2023-09-28
io: mark Interest::add as must_use

nylonicious:
Like in https://github.com/tokio-rs/mio/pull/1694. 

Marking `is_*` methods makes little bit less sense to me but if that's something desirable I'll mark those as well.


Darksonn (approved):
Thanks.

PR #6058, opened 2023-10-07
Implement std::io::Seek for tokio_util::io::SyncIoBridge

nitsky:
## Motivation

I want to use a type that implements `AsyncSeek` with sync code.

## Solution

This PR implements `std::io::Seek` for `tokio_util::io::SyncIoBridge`.


Darksonn (approved):
Looks good to me. Thanks.

PR #6097, opened 2023-10-22
time: remove cached `elapsed` value from driver state

satakuma:
The time driver stores the number of elapsed units of time in a variable and keeps it in sync with the same counter inside the timer wheel. Since the variable is only used to ensure that we poll the timer wheel with monotonic instants, we may as well use the actual counter from the wheel and avoid redundancy.


Darksonn (approved):
Thanks.

PR #6158, opened 2023-11-18
task: make task span explicit root

hds:
## Motivation

In Tokio, tasks are optionally instrumented with tracing spans to allow
analysis of the runtime behavior to be performed with tools like
tokio-console.

The span that is created for each task gets currently follows the
default tracing behavior and has a contextual parent attached to it
based on the span that is actual when `tokio::spawn` or similar is
called.

However, in tracing, a span will remain "alive" until all its children
spans are closed. This doesn't match how spawned tasks work. A task may
outlive the context in which is was spawned (and frequently does). This
causes tasks which spawn other - longer living - tasks to appear in
`tokio-console` as having lost their waker when instead they should be
shown as completed (tokio-rs/console#345). It can also cause undesired
behavior for unrelated tracing spans if a subscriber is receiving both
the other spans as well as Tokio's instrumentation.

## Solution

To fix this mismatch in behavior, the task span has `parent: None` set
on it, making it an explicit root - it has no parent. The same was
already done for all spans representing resources in #6107. This change
is made within the scope of #5792.

Due to a defect in the currently available `tracing-mock` crate, it is
not possible to test this change at a tracing level
(tokio-rs/tracing#2440). Instead, a test for the `console-subscriber`
has been written which shows that this change fixes the defect as
observed in `tokio-console` (tokio-rs/console#490).

### DIFF
--- a/tokio/src/util/trace.rs
+++ b/tokio/src/util/trace.rs
@@ -16,6 +16,7 @@ cfg_trace! {
                 let location = std::panic::Location::caller();
                 tracing::trace_span!(
                     target: "tokio::task",
+                    parent: None,
                     "runtime.spawn",
                     %kind,
                     task.name = %name.unwrap_or_default(),


hawkw:
@hds is there a specific reason why this PR is marked as a draft? is it waiting on anything in particular before it's ready for review?

hds:
@hawkw Not waiting for anything any more. Please feel free to review.

(reasons below for anyone who's interested)

The associated test in console-subscriber (poining at this branch for tokio) has a failure in it (tokio-rs/console#490). Yesterday I thought that it might have been caused by this PR or the previous one where all resource spans were made explicit roots, so I moved it to draft status.

After digging into it more today, it seems that it is just another case of the flaky tests (tokio-rs/console#473) caused by that race condition in tracing (tokio-rs/tracing#2743).

hawkw (approved):

PR #6163, opened 2023-11-21
util: add `DelayQueue::deadline`

aliu:
## Motivation

When implementing a cache with item expirations ([example from mini-redis](https://github.com/tokio-rs/mini-redis/blob/9b93f751b3cd0fb84e27ad1f938da8121ce1e83b/src/db.rs#L154-L158)), it's not possible to use `DelayQueue` as a more efficient backing store for the expirations. The reason is, when inserting/removing/modifying a cache entry there needs to be a way to determine whether the ["next expiring time" has changed](https://github.com/tokio-rs/mini-redis/blob/9b93f751b3cd0fb84e27ad1f938da8121ce1e83b/src/db.rs#L175-L178) as a result of the operation. Basically, you'd need an api similar to [`DelayQueue::peek`](https://docs.rs/tokio-util/0.7.10/src/tokio_util/time/delay_queue.rs.html#906) but for the next deadline in the queue. However, currently `DelayQueue` does not expose such an api, so you are left with having to store an independent mapping of `Key`s to expiration `Instant`s and keep it in sync with all operations on the `DelayQueue`. It would be nice to not have to maintain this bookkeeping when `DelayQueue` already knows this information.

## Solution

Originally, my first thought was to simply make [`DelayQueue::next_deadline`](https://docs.rs/tokio-util/0.7.10/src/tokio_util/time/delay_queue.rs.html#913) public. However looking closer at the internals, it doesn't consider deadlines in the `expired` queue. So a user who inserts an already expired item into the queue and calls `next_deadline` might be surprised when they get `None` back. This function is used a lot internally as well, so I thought it would be best to not modify its behavior.

Instead we could expose a `deadline` function that takes a `Key` and returns its expiration time. This way, we could still get the next deadline via `delay_queue.peek().map(|key| delay_queue.deadline(&key))`.

### DIFF
--- a/tokio-util/src/time/delay_queue.rs
+++ b/tokio-util/src/time/delay_queue.rs
@@ -671,6 +671,39 @@ impl<T> DelayQueue<T> {
         }
     }
 
+    /// Returns the deadline of the item associated with `key`.
+    ///
+    /// Since the queue operates at millisecond granularity, the returned
+    /// deadline may not exactly match the value that was given when initially
+    /// inserting the item into the queue.
+    ///
+    /// # Panics
+    ///
+    /// This function panics if `key` is not contained by the queue.
+    ///
+    /// # Examples
+    ///
+    /// Basic usage
+    ///
+    /// ```rust
+    /// use tokio_util::time::DelayQueue;
+    /// use std::time::Duration;
+    ///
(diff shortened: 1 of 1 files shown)


Darksonn (approved):
Thanks.

PR #6169, opened 2023-11-26
io: drop the `Sized` requirements from `AsyncReadExt.read_buf`

tijsvd:
## Motivation

`AsyncReadExt.read_buf` takes both `Self` and its buffer argument by mut ref, yet requires `Sized` for both of them,
disallowing `dyn Trait` for no good reason.

## Solution

Remove the explicit `Sized` requirement for `Self`, and introduce `?Sized` in relevant places.

See also #6161


Darksonn (approved):
Thank you.

PR #6180, opened 2023-11-30
chore: remove uses of unsafe from maybe_done.rs

xfix:
## Motivation

Noticed that `maybe_done.rs` file has some unnecessary `unsafe`, and decided to remove them.

## Solution

Used `pin_project!` macro to remove `unsafe` uses.


taiki-e (approved):
Thanks!

PR #4383, opened 2022-01-05
rt: reduce no-op wakeups in the multi-threaded scheduler

carllerche:
This PR reduces the number of times worker threads wake up without having work to do in the multi-threaded scheduler. Unnecessary wake-ups are expensive and slow down the scheduler. I have observed this change reduce no-op wakes by up to 50%.

The multi-threaded scheduler is work-stealing. When a worker has tasks to process, and other workers are idle (parked), these idle workers must be unparked so that they can steal work from the busy worker. However, unparking threads is expensive, so there is an optimization that avoids unparking a worker if there already exists workers in a "searching" state (the worker is unparked and looking for work). This works pretty well, but transitioning from 1 "searching" worker to 0 searching workers introduces a race condition where a thread unpark can be lost:

* thread 1: last searching worker about to exit searching state
* thread 2: needs to unpark a thread, but skip because there is a searching worker.
* thread 1: exits searching state w/o seeing thread 2's work.

Because this *should* be a rare condition, Tokio solves this by always unparking a new worker when the current worker:

* is the last searching worker
* is transitioning out of searching
* has work to process.

When the newly unparked worker wakes, if the race condition described above happened, "thread 2"'s work will be found. Otherwise, it will just go back to sleep.

Now we come to the issue at hand. A bug incorrectly set a worker to "searching" when the I/O driver unparked the thread. In a situation where the scheduler was only partially under load and is able to operate with 1 active worker, the I/O driver would unpark the thread when new I/O events are received, incorrectly transition it to "searching", find new work generated by inbound I/O events, incorrectly transition itself from the last searcher -> no searchers, and unpark a new thread. This new thread would wake, find no work and go back to sleep.

Note that, when the scheduler is **fully** saturated, this change will make no impact as most workers are always unparked and the optimization to avoid unparking threads described at the top applies.

# Benchmarks

## Hyper

This benches Hyper's "hello" server using `wrk -t1 -c400 -d10s http://127.0.0.1:3000/`

**master**

```
Requests/sec: 118495.92
Transfer/sec: 9.94MB
```

**this PR**

```
Requests/sec: 135261.99
Transfer/sec: 11.35MB
```

## mini-redis

This benches [mini-redis](github.com/tokio-rs/mini-redis) using `redis-benchmark -c 5 -t get,set`

**master**

```
====== SET ======
  100000 requests completed in 1.03 seconds
  5 parallel clients
  3 bytes payload
  keep alive: 1
  multi-thread: no

99.17% <= 0.1 milliseconds
[...]

PR #6251, opened 2023-12-27
rt: reduce the impact of CPU bound tasks on the overall runtime shceduler

wathenjiang:
We discussed in https://github.com/tokio-rs/tokio/issues/4730 that CPU-bound tasks will cause increased scheduling delays in the entire Tokio runtime. We recommend using [tokio::task::block_in_place](https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html) for these tasks to prevent latency spikes for other tasks. However, even with a multi-threaded runtime and multiple worker threads, just one single CPU-bound task can still cause significant latency.

Some async runtimes, like Go, have dedicated threads for polling drivers (Go's [sysmon thread](https://github.com/golang/go/blob/2184a394777ccc9ce9625932b2ad773e6e626be0/src/runtime/proc.go#L5939)). Tokio, however, has each worker thread responsible for polling the driver under certain conditions, allowing only one worker thread to do so at a time by using Condvar. Go's strategy may not be suitable for Tokio, as its sysmon thread also handles GC-related operations and goroutine preemption.

This PR aims to fully harness the CPU computing potential of the multi_thread tokio runtime, reduce the delay impact of CPU bound tasks on IO event processing.

**Note**: This does not solve the problem that asynchronous tasks in Rust cannot currently implement preemptive scheduling. The number of CPU bound tasks exceeding the number of worker threads will still cause the entire tokio runtime to run blocked.

Tokio's scheduling mechanism is very great. It minimizes the need for worker thread wake-ups and has significantly improved performance, which can been seen in https://github.com/tokio-rs/tokio/pull/4383 . However, the thread wake-up mechanism seems a bit too conservative. It might be worth making it slightly more aggressive to ensure a quicker I/O event response time. This PR includes the following change: 

When the worker responsible for polling the driver gets unparked, no thread will continue to poll the driver at that moment, as the current worker is going to run tasks. So, we quickly try to wake up another worker, hoping it can poll the driver.

Here is the test case for the current PR:

```rust
use std::time::Duration;
use std::time::Instant;

async fn handle_request(start_time: Instant) {
    tokio::time::sleep(Duration::from_millis(500)).await;
    println!("request took {}ms", start_time.elapsed().as_millis());
}

async fn background_job() {
    loop {
        tokio::time::sleep(Duration::from_secs(3)).await;
        // Adjust as needed
        for _ in 0..1_000 { vec![1; 1_000_000].sort() }
    }
}

#[tokio::main(flavor = "multi_thread", worker_threads = 8)]
async fn main() {
    for _ in 0..7{
        tokio::spawn(async { background_job().await });
    }

    loop {
        let start = Instant::now();
        tokio::spawn(async move {
            handle_request(start).await
        }).await.unwrap();
    }
}
```

**The test result on master:**
```rust
request took 502ms
request took 500ms
request took 501ms
request took 501ms
request took 502ms
request took 501ms
request took 501ms
request took 501ms
request took 502ms
request took 502ms
request took 501ms
request took 502ms
request took 500ms
request took 501ms
request took 899ms    <==== high-latency request
request took 502ms
request took 502ms
request took 502ms
request took 502ms
request took 501ms
request took 1072ms   <==== high-latency request
request took 1007ms   <==== high-latency request
request took 501ms
request took 502ms
request took 501ms
request took 1093ms   <==== high-latency request
request took 1067ms   <==== high-latency request
request took 856ms    <==== high-latency request
...
```

**The test result on this PR:**
```rust
request took 502ms
request took 502ms
request took 503ms
request took 502ms
request took 502ms
request took 502ms
request took 506ms
request took 502ms
request took 502ms
request took 502ms
request took 500ms
request took 502ms
request took 501ms
request took 502ms
request took 504ms
```

Referring to https://github.com/tokio-rs/tokio/pull/4383 , I did a performance test of Hyper's "hello" server:

This benches Hyper's "hello" server using `wrk -t1 -c400 -d10s http://127.0.0.1:3000/`

**Master**

```
Requests/sec: 162342.60
Transfer/sec:     13.62MB
```

**This PR**

```
Requests/sec: 161073.47
Transfer/sec:     13.52MB
```

There is almost no performance difference in this test case.

This PR will not completely solve the negative impact of CPU bound tasks (or [blocking tasks](https://ryhl.io/blog/async-what-is-blocking/)), but it can reduce the processing delay of I/O events when the task schedule is not busy.

### DIFF
--- a/tokio/src/runtime/scheduler/multi_thread/worker.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/worker.rs
@@ -917,6 +917,13 @@ impl Core {
                 .shared
                 .idle
                 .unpark_worker_by_id(&worker.handle.shared, worker.index);
+
+            // Since this worker is awakened with tasks, it cannot be responsible for
+            // polling the driver for a certain period of time.
+            // Therefore, we ensure that at least one worker is designated to poll the driver here.
+            // This is for avoid significant delays in driver event processing,
+            // as discussed in https://github.com/tokio-rs/tokio/issues/4730.
+            worker.handle.notify_parked_local();
             return true;
         }
 


carllerche:
Hey, thanks for taking a stab at this. I appreciate you jumping into complex code and trying to resolve a real issue that causes users pain.

Unfortunately, I cannot accept this PR as it is, and I will try to break down why.

First, a point that isn't critical for accepting the PR, but since you took some time to dig into this, you might want to learn more about the code. [This](https://github.com/tokio-rs/tokio/blob/master/tokio/src/runtime/scheduler/multi_thread/worker.rs#L734-L736) is where a worker notifies a peer. You would probably want to update `should_notify_others` to consolidate the wakeups; otherwise, there will be two workers woken up per tick.

That said, there is tension in the scheduler when deciding if a peer thread should be woken up. Synchronization overhead is a big issue on larger systems (over 50 cores). When I first implemented the PR to reduce no-op wakeups, it was because I was observing significant CPU overhead at scale due to thread synchronization as part of the wakeup logic. One option could be a configuration option to tune how aggressive the scheduler is at waking up peer threads.

Also, looking through worker.rs, I wonder if there are some changes that I experimented with in the [alt scheduler](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.new_multi_thread_alt) that we may want to pull in, e.g. skipping the LIFO slot and batching events from the I/O driver.

Anyway, all of this work relates to #6315, so we should try to figure out steps to move forward there.

PR #6293, opened 2024-01-17
util: simplify check for empty slice in chain

wutchzone:
Just a plain simplification of the emptiness check.


maminrayej:
Thanks.

maminrayej (approved):

PR #6300, opened 2024-01-23
task: implement `FromIterator` for `JoinSet`

erikdesjardins:
## Motivation

When using `JoinSet` I usually end up populating it from an iterator, which would be more convenient with `collect`.

## Solution

Implement `FromIterator`.

(Possibly, spawning each element of the iter is too magical / side-effecty for a `FromIterator` impl though.)

Closes: #6132

### DIFF
--- a/tokio/src/task/join_set.rs
+++ b/tokio/src/task/join_set.rs
@@ -526,6 +526,49 @@ impl<T> Default for JoinSet<T> {
     }
 }
 
+/// Collect an iterator of futures into a [`JoinSet`].
+///
+/// This is equivalent to calling [`JoinSet::spawn`] on each element of the iterator.
+///
+/// # Examples
+///
+/// The main example from [`JoinSet`]'s documentation can also be written using [`collect`]:
+///
+/// ```
+/// use tokio::task::JoinSet;
+///
+/// #[tokio::main]
+/// async fn main() {
+///     let mut set: JoinSet<_> = (0..10).map(|i| async move { i }).collect();
+///
+///     let mut seen = [false; 10];
+///     while let Some(res) = set.join_next().await {
+///         let idx = res.unwrap();
+///         seen[idx] = true;
+///     }
+///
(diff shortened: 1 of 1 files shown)


maminrayej:
Related to issue #6132. Also, I believe this has conflicts with #5924.

@Darksonn thoughts?

hawkw:
> @Darksonn does this have conflicts with #5924?
> My understanding is that this would have conflicts if it collected handles. Since this implementation spawns the futures using the `JoinSet` itself, it's fine.

If we were to make the change described in #5924, we could potentially add a `FromIterator` impl where the iterator's `Item` is `JoinHandle`. This would, potentially, be more flexible in terms of how the tasks in the iterator are spawned, because the current `FromIterator` impl spawns the tasks using `JoinSet::spawn`, but an iterator of `JoinHandle`s could also contain `JoinHandle`s that were spawned using `spawn_local` or `spawn_blocking`. However, because `JoinHandle` implements `Future`, the `FromIterator` implementation for iterators of `JoinHandle`s would conflict with the `FromIterator` implementation for iterators of `F: Future`s.

Note that you _could_ pass an iterator of `JoinHandle`s to the current `FromIterator` implementation, but it would spawn a second task that just awaits the `JoinHandle`'s completion, which introduces some unnecessary overhead (including an additional task allocation). But, it's worth mentioning that  `JoinHandle`s _can_ be used with the interface in this PR, even if they _probably shouldn't_. 

However, I think the conclusion from #5924 was that we're probably not going to add an API for adding `JoinHandle`s to a `JoinSet`, because we'd like to eventually make an internal optimization in `JoinSet` where the `JoinSet` linked list nodes are inlined into the task's allocation (see https://github.com/tokio-rs/tokio/issues/5924#issuecomment-1696316816). Therefore, I don't really think we should worry too much about the potential addition of a `FromIterator` implementation for `Iterator<Item = JoinSet>`, because we're probably not going to do that...

PR #6307, opened 2024-01-26
feat: generate function declaration in macro for better type check compile error message

ik1ne:
## Motivation
issue #6306 

`parse_knobs` method creates an async block and then calls it. While doing so, the return type of original function is erased for the async block.
This slightly changes the compiler error message from type mismatch.

## Solution
Create a temporary function with original return type preserved instead of creating an async block.

## Caveats
1. If users call __tokio_internal_async_body inside their main function it will infinitely recurse.
2. I don't think it will, but this might have some erroneous behavior on successfully compiled binary so I would appreciate if someone can find out if this implementation is sound.
3. This is a breaking change for compile error tests, since compiler error message for typecheck will change.

### DIFF
--- a/tests-build/tests/fail/macros_type_mismatch.stderr
+++ b/tests-build/tests/fail/macros_type_mismatch.stderr
@@ -57,3 +57,31 @@ error[E0308]: mismatched types
 31 |     return 1;
 32 |     ;
    |     ^ expected `()`, found integer
+
+error[E0277]: the `?` operator can only be used in an async block that returns `Result` or `Option` (or another type that implements `FromResidual`)
+  --> tests/fail/macros_type_mismatch.rs:37:11
+   |
+35 | #[tokio::main]
+   | -------------- this function should return `Result` or `Option` to accept `?`
+36 | async fn extra_semicolon_with_try_operator() -> Result<(), ()> {
+37 |     Ok(())?;
+   |           ^ cannot use the `?` operator in an async block that returns `()`
+   |
+   = help: the trait `FromResidual<Result<Infallible, _>>` is not implemented for `()`
+
+error[E0308]: mismatched types
+  --> tests/fail/macros_type_mismatch.rs:38:5
+   |
+36 | async fn extra_semicolon_with_try_operator() -> Result<(), ()> {
+   |                                                 -------------- expected `Result<(), ()>` because of return type
+37 |     Ok(())?;
+38 |     Ok(());
+   |     ^^^^^^^ expected `Result<(), ()>`, found `()`
+   |
+   = note:   expected enum `Result<(), ()>`
+           found unit type `()`
+help: try adding an expression at the end of the block
+   |
+38 ~     Ok(());;
+39 +     Ok(())
+   |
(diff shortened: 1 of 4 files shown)


tests-build/tests/fail/macros_type_mismatch.stderr
ik1ne, line 1: Except for `extra_semicolon`, the stderr has been changed by these steps:
1. Within a blank cargo crate, copy paste `macros_type_mismatch.rs`.
2. Comment `#[tokio::main]`
3. Copy the stderr to macros_type_mismatch.stderr

For `extra_semicolon`, it seems the result of syn parse removes the extra semicolon, which makes compiler error message slightly different from plain one.

tokio-macros/src/entry.rs
```diff
@@ -369,10 +370,15 @@ fn parse_knobs(mut input: ItemFn, is_test: bool, config: FinalConfig) -> TokenSt
...
+    let return_or_nothing = match &input.sig.output {
+        syn::ReturnType::Type(_, ty) if matches!(**ty, syn::Type::Never(_)) => quote! {},
+        _ => quote! { return },
+    };
+
     let last_block = quote_spanned! {last_stmt_end_span=>
         #[allow(clippy::expect_used, clippy::diverging_sub_expression)]
         {
-            return #rt
+            #return_or_nothing #rt
```
ik1ne, line 381: If we always emit `return`, the following code will result to compiler warning(unreachable statement for `return`)

macros_test.rs: 36
```rust
// https://github.com/tokio-rs/tokio/issues/4175
#[tokio::main]
pub async fn issue_4175_main_1() -> ! {
    panic!();
}```

So with this PR, the rustc properly detects that the `return` clause is meaningless for `!` type returns.

**Comment by taiki-e** [comment]
This approach was considered in the past but rejected because it would be a breaking change.
https://github.com/tokio-rs/tokio/pull/3766#issuecomment-835508651 https://github.com/tokio-rs/tokio/pull/4513#issuecomment-1043738115

While "use this approach if there are no arguments" might be fine, it's not clear that it's worth the complexity.

**Comment by ik1ne** [comment]
Thanks for the clarification, I understand why this implementation is not valid and going to close this PR.

However, the test `cargo test --all-features` has been succeeded(including `trait_method`), since `<() as A>::f` doesn't access `self` inside the function. (calling `self.f()` inside `f()` makes the test fail, while master version passes)

So, before I close this PR, would you mind if I change the `trait_method` and check if this PR's ci fails? (And then create PR for `trait_method` afterwards)

**Comment by taiki-e** [comment]
> However, the test `cargo test --all-features` has been succeeded(including `trait_method`), since `<() as A>::f` doesn't access `self` inside the function. (calling ~`self.f()`~ `self.g()` inside `f()` makes the test fail, while master version passes)
> 
> So, before I close this PR, would you mind if I change the `trait_method` and check if this PR's ci fails? (And then create PR for `trait_method` afterwards)

Yeah, it is definitely a problem that the trait_method test did not catch that this is a breaking change.
It would be great if you could fix that test so that it catches the problem.

**Comment by ik1ne** [comment]
While not every check has been finished, I checked following compiler error locally and for some of the github checks:
```
error[E0434]: can't capture dynamic environment in a fn item
  --> tokio/tests/macros_test.rs:34:13
   |
34 |             self.g()
   |             ^^^^
   |
   = help: use the `|| { ... }` closure form instead

For more information about this error, try `rustc --explain E0434`.
error: could not compile `tokio` (test "macros_test") due to previous error
```
So I'm closing this issue and going to create a PR for `trait_method`.

PR #6308, opened 2024-01-26
macros: fix trait_method breaking change detection

ik1ne:
## Motivation
Context: #6307 
The test `trait_method` was designed to prevent breaking changes for `#[tokio::main]` on trait methods.
However, previous implementation didn't access self so it was unable to catch some case of breaking changes.

## Solution
Created another method `g(self)` and called it to access `self`.


taiki-e (approved):
Thanks!

PR #6311, opened 2024-01-30
net: expose keepalive option on `TcpSocket`

### LINKED ISSUE #6214: net: expose TCP keep alive socket option
Currently, Tokio does not expose the keep-alive socket option. There have been attempts (#3082, #3146) but they have not succeeded because keep-alive behavior is not consistent between *nix and windows (possibly others) platforms.

That said, setting the keep-alive option is still a common operation. Today, work arounds exist by using libraries like `socket2`, `nix`, or `libc` directly, but we should still add an easier API.

Because there is no consistent behavior across platforms, we should consider adding this feature on a platform-by-platform basis. The biggest challenge for doing this is that I am not sure what the best API would be to add platform-specific methods that have the same name but different arguments. E.g. `set_keep_alive` on linux would probably take different arguments than on windows.

To move forward with this issue, we would need a proposal for how to add an API with a similar name, but different arguments and behavior based on the platform. This proposal should consider Tokio's existing platform-specific APIs and prior-art within the Rust ecosystem.

maminrayej:
It was decided in the Discord discussion to only support enabling the keep-alive option since it is portable across different platforms.

Resolves #6214.

### DIFF
--- a/tokio/src/net/tcp/socket.rs
+++ b/tokio/src/net/tcp/socket.rs
@@ -185,6 +185,11 @@ impl TcpSocket {
         Ok(TcpSocket { inner })
     }
 
+    /// Enable sending of keep-alive messages on connection-oriented sockets.
+    pub fn set_keepalive(&self, keepalive: bool) -> io::Result<()> {
+        self.inner.set_keepalive(keepalive)
+    }
+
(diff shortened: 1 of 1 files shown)


carllerche (approved):
Thanks!

PR #6320, opened 2024-02-02
docs: transition_to_idle doesn't return boolean

hackaugusto:
## Motivation

Small change to the docs of `transition_to_idle`. The method doesn't return `bool` but a value of the `TransitionToIdle` enum.


Darksonn (approved):
Thanks.

PR #6329, opened 2024-02-06
runtime: fix leaking registration entries when os registration fails

maminrayej:
It is possible for `add_source` to fail when registering a source with the OS and leak the allocated slot in the registration set. This PR removes the allocated slot in the case of a failure in order to prevent this issue.

I could call `registrations.deregister` and let the `io::Driver` remove the slot on its next `turn`. But since the registration failed, and there won't be any event for the provided source, I believe it is safe to remove the entry directly.

### DIFF
--- a/tokio/src/runtime/io/driver.rs
+++ b/tokio/src/runtime/io/driver.rs
@@ -220,8 +220,17 @@ impl Handle {
         let scheduled_io = self.registrations.allocate(&mut self.synced.lock())?;
         let token = scheduled_io.token();
 
-        // TODO: if this returns an err, the `ScheduledIo` leaks...
-        self.registry.register(source, token, interest.to_mio())?;
+        // we should remove the `scheduled_io` from the `registrations` set if registering
+        // the `source` with the OS fails. Otherwise it will leak the `scheduled_io`.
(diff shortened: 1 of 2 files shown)


tokio/src/runtime/io/driver.rs
```diff
@@ -220,8 +220,17 @@ impl Handle {
         let scheduled_io = self.registrations.allocate(&mut self.synced.lock())?;
         let token = scheduled_io.token();
 
-        // TODO: if this returns an err, the `ScheduledIo` leaks...
-        self.registry.register(source, token, interest.to_mio())?;
+        // we should remove the `scheduled_io` from the `registrations` set if registering
+        // the `source` with the OS fails. Otherwise it will leak the `scheduled_io`.
+        if let Err(e) = self.registry.register(source, token, interest.to_mio()) {
+            // safety: `scheduled_io` is part of the `registartions` set.
```
balliegojr, line 226:
```suggestion
            // safety: `scheduled_io` is part of the `registrations` set.
```

carllerche (approved):
Looks good to me. It would be nice to have a test for it, but that would be tough without a bigger change (adding the ability to mock Mio behavior, which could be good for testing stuff like this!)

PR #1997, opened 2019-12-20
Add unit test showing some unexpected behaviour of mpsc::channel

bikeshedder:
In ['deadpool'](https://crates.io/crates/deadpool/) I noticed some weird behaviour when calling `Receiver::try_recv` on a channel which is protected by a semaphore in such a way that the size can't exceed its limit: https://github.com/bikeshedder/deadpool/blob/6daa2e7b5bd06d0b516b83f9da1902848e7812e7/src/lib.rs#L266

I noticed more objects being returned to the pool than the `max_size` which just didn't make any sense. After some digging I found that `try_recv` sometimes does not return `Ok` even if something has already been put into the channel before via `try_send`.

This behaviour can be reproduced with the following steps:

- Create a semaphore with size 1
- Create a channel and fill it with one message
- Spawn two tasks
- One of the tasks manages to acquire the permit, receive the message from the channel and send it back. Afterwards the permit is dropped.
- Now the other task tries to do the same and fails during the `try_recv` call.

I have no idea how to fix that but please find that test case attached.

As using channel just with `try_send` and `try_lock` is a bit wasteful in my case anyways I just found [`crossbeam_queue`](https://crates.io/crates/crossbeam-queue) which supports concurrent `push/pop` without any locking.

I'm afraid I might have actually introduced that bug myself when adding `Receiver::try_recv`.

PR #6348, opened 2024-02-15
Add `is_closed`, `is_empty` and `len` to `mpsc::Receiver` and `mpsc::UnboundedReceiver`

balliegojr:
Fixes: #4638
Fixes: #6314

## Motivation

Taken from #4638

## Solution

Adds a new `is_closed` function to both `mpsc::Receivers` that check if the channel is closed and has no remaining messages in the internal buffer.

Since there are two possible ways to close a channel and they lead to different internal states, it was necessary to add two checks, one for each internal state.

### DIFF
--- a/tokio/src/sync/mpsc/bounded.rs
+++ b/tokio/src/sync/mpsc/bounded.rs
@@ -511,6 +511,25 @@ impl<T> Receiver<T> {
         self.chan.is_empty()
     }
 
+    /// Returns the number of messages in the channel.
+    ///
+    /// # Examples
+    /// ```
+    /// use tokio::sync::mpsc;
+    ///
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let (tx, rx) = mpsc::channel(10);
+    ///     assert_eq!(0, rx.len());
+    ///
+    ///     tx.send(0).await.unwrap();
+    ///     assert_eq!(1, rx.len());
+    /// }
+    /// ```
+    pub fn len(&self) -> usize {
+        self.chan.len()
+    }
+
     /// Polls to receive the next message on this channel.
     ///
     /// This method returns:
--- a/tokio/src/sync/mpsc/bounded.rs
+++ b/tokio/src/sync/mpsc/bounded.rs
@@ -463,6 +463,39 @@ impl<T> Receiver<T> {
         self.chan.close();
     }
 
+    /// Checks if a channel is closed.
+    ///
+    /// This method returns `true` if the channel has been closed and there are
+    /// no remaining messages in the channel's buffer. The channel is closed
+    /// when all senders have been dropped, or when [`close`] is called.
+    ///
+    /// # Examples
+    /// ```
+    /// use tokio::sync::mpsc;
+    ///
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let (tx, mut rx) = mpsc::channel(10);
+    ///     assert!(!rx.is_closed());
+    ///
+    ///     tx.send(100).await.unwrap();
+    ///     tx.send(200).await.unwrap();
+    ///
+    ///     rx.close();
+    ///     assert!(!rx.is_closed());
+    ///     assert_eq!(rx.recv().await, Some(100));
+    ///
+    ///     assert!(!rx.is_closed());
+    ///     assert_eq!(rx.recv().await, Some(200));
+    ///     assert!(rx.recv().await.is_none());
+    ///     
+    ///     assert!(rx.is_closed());
+    /// }
+    /// ```
+    pub fn is_closed(&mut self) -> bool {
+        self.chan.is_closed_and_empty()
+    }
+
     /// Polls to receive the next message on this channel.
     ///
     /// This method returns:
--- a/tokio/src/sync/mpsc/bounded.rs
+++ b/tokio/src/sync/mpsc/bounded.rs
@@ -465,8 +465,7 @@ impl<T> Receiver<T> {
 
     /// Checks if a channel is closed.
     ///
-    /// This method returns `true` if the channel has been closed and there are
-    /// no remaining messages in the channel's buffer. The channel is closed
+    /// This method returns `true` if the channel has been closed. The channel is closed
     /// when all [`Sender`] have been dropped, or when [`Receiver::close`] is called.
     ///
     /// [`Sender`]: crate::sync::mpsc::Sender
@@ -478,25 +477,38 @@ impl<T> Receiver<T> {
     ///
     /// #[tokio::main]
     /// async fn main() {
-    ///     let (tx, mut rx) = mpsc::channel(10);
+    ///     let (_tx, mut rx) = mpsc::channel::<()>(10);
     ///     assert!(!rx.is_closed());
-    ///
-    ///     tx.send(100).await.unwrap();
-    ///     tx.send(200).await.unwrap();
     ///
     ///     rx.close();
-    ///     assert!(!rx.is_closed());
-    ///     assert_eq!(rx.recv().await, Some(100));
-    ///
-    ///     assert!(!rx.is_closed());
-    ///     assert_eq!(rx.recv().await, Some(200));
-    ///     assert!(rx.recv().await.is_none());
     ///     
     ///     assert!(rx.is_closed());
     /// }
     /// ```
-    pub fn is_closed(&mut self) -> bool {
-        self.chan.is_closed_and_empty()
+    pub fn is_closed(&self) -> bool {
+        self.chan.is_closed()
+    }
+
+    /// Checks if a channel is empty.
+    ///
+    /// This method returns `true` if the channel has no messages.
+    ///
+    /// # Examples
+    /// ```
+    /// use tokio::sync::mpsc;
+    ///
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let (tx, mut rx) = mpsc::channel(10);
+    ///     assert!(rx.is_empty());
+    ///
+    ///     tx.send(0).await.unwrap();
+    ///     assert!(!rx.is_empty());
+    /// }
+    ///
+    /// ```
+    pub fn is_empty(&mut self) -> bool {
+        self.chan.is_empty()
     }
 
     /// Polls to receive the next message on this channel.
--- a/tokio/src/sync/mpsc/bounded.rs
+++ b/tokio/src/sync/mpsc/bounded.rs
@@ -467,7 +467,10 @@ impl<T> Receiver<T> {
     ///
     /// This method returns `true` if the channel has been closed and there are
     /// no remaining messages in the channel's buffer. The channel is closed
-    /// when all senders have been dropped, or when [`close`] is called.
+    /// when all [`Sender`] have been dropped, or when [`Receiver::close`] is called.
+    ///
+    /// [`Sender`]: crate::sync::mpsc::Sender
+    /// [`Receiver::close`]: crate::sync::mpsc::Receiver::close
     ///
     /// # Examples
     /// ```
--- a/tokio/src/sync/mpsc/bounded.rs
+++ b/tokio/src/sync/mpsc/bounded.rs
@@ -499,7 +499,7 @@ impl<T> Receiver<T> {
     ///
     /// #[tokio::main]
     /// async fn main() {
-    ///     let (tx, mut rx) = mpsc::channel(10);
+    ///     let (tx, rx) = mpsc::channel(10);
     ///     assert!(rx.is_empty());
     ///
     ///     tx.send(0).await.unwrap();
@@ -507,7 +507,7 @@ impl<T> Receiver<T> {
     /// }
     ///
     /// ```
-    pub fn is_empty(&mut self) -> bool {
+    pub fn is_empty(&self) -> bool {
         self.chan.is_empty()
     }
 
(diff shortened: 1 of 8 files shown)


tokio/src/sync/mpsc/chan.rs
```diff
@@ -241,6 +241,24 @@ impl<T, S: Semaphore> Rx<T, S> {
...
+        //  outstanding messages if the list is at the tail position.
+        //
+        //  2. When all senders are dropped.
+        //  In this case, the semaphore remains unclosed, and the `index` in the list won't
+        //  reach the tail position. It is necessary to check the list if the last block is
+        //  `closed` and has no value available.
+        self.inner.rx_fields.with_mut(|rx_fields_ptr| {
+            let rx_fields = unsafe { &mut *rx_fields_ptr };
+            (self.inner.semaphore.is_closed() && rx_fields.list.is_at_tail_position(&self.inner.tx))
+                || rx_fields.list.is_closed_and_empty(&self.inner.tx)
```
balliegojr, line 258: On alternative to this implementation would be to change the `Tx` drop implementation to also close the semaphore, this would simplify the internal state to rely only on the semaphore.

The issue with this alternative approach is that is necessary to add a constraint to the struct generic parameter.

```rust
impl<T, S: Semaphore> Drop for Tx<T, S> { // <- Tx<T, S: Semaphore>, this would propagate to a lot of places
    fn drop(&mut self) {
        if self.inner.tx_count.fetch_sub(1, AcqRel) != 1 {
            return;
        }

        // Close the list, which sends a `Close` message
        self.inner.tx.close();
        self.inner.semaphore.close();

        // Notify the receiver
        self.wake_rx();
    }
}
```

tokio/tests/sync_mpsc.rs
Darksonn, line 1: It is easier for me to understand the tests if you split them into many small tests rather than one or two large ones. For example, these small tests could be useful:

1. `is_closed` should return true after calling `close` but still has a sender
2. `is_closed` should return true after dropping all senders
3. `is_closed` should return true after dropping all senders except for a weak sender
4. `is_closed` should return false when there is a sender
4. `is_closed` should return false when there is a sender, even if enough messages have been sent to fill the channel

balliegojr: I have split the tests into smaller single purpose tests. 

Is it necessary to have the same tests for the unbounded channel as well?

tokio/tests/sync_mpsc.rs
```diff
@@ -1017,4 +1017,81 @@ async fn test_tx_capacity() {
...
+
+#[tokio::test]
+async fn test_rx_is_not_closed_when_there_are_messages_but_not_senders() {
+    // is_closed should return false when there is a permit (but no senders)
+    let (tx, mut rx) = mpsc::channel(10);
+    for i in 0..10 {
+        assert!(tx.send(i).await.is_ok());
+    }
+    drop(tx);
+    assert!(!rx.is_closed());
```
Darksonn, line 1069: The comment here looks wrong.

tokio/tests/sync_mpsc.rs
```diff
@@ -1017,4 +1017,81 @@ async fn test_tx_capacity() {
...
+
+#[tokio::test]
+async fn test_rx_is_not_closed_when_there_are_messages_and_close_is_called() {
+    // is_closed should return false when there is a permit (but no senders)
+    let (tx, mut rx) = mpsc::channel(10);
+    for i in 0..10 {
+        assert!(tx.send(i).await.is_ok());
+    }
+    rx.close();
+    assert!(!rx.is_closed());
```
Darksonn, line 1080: Here as well.

tokio/tests/sync_mpsc.rs
```diff
@@ -1017,4 +1017,81 @@ async fn test_tx_capacity() {
...
+async fn test_rx_is_closed_after_consuming_messages() {
+    // is_closed should return false when there is a permit (but no senders)
+    let (tx, mut rx) = mpsc::channel(10);
+    for i in 0..10 {
+        assert!(tx.send(i).await.is_ok());
+    }
+    drop(tx);
+
+    assert!(!rx.is_closed());
+    while (rx.recv().await).is_some() {}
```
Darksonn, line 1093: Also here. It seems like there aren't any tests for the case where there is a permit but no sender.

tokio/tests/sync_mpsc.rs
Darksonn: Thanks!

That could make sense. After all, they have different implementations of the semaphore, so we aren't *just* testing the same code twice.

tokio/src/sync/mpsc/bounded.rs
```diff
@@ -463,6 +463,39 @@ impl<T> Receiver<T> {
...
+    ///     assert_eq!(rx.recv().await, Some(100));
+    ///
+    ///     assert!(!rx.is_closed());
+    ///     assert_eq!(rx.recv().await, Some(200));
+    ///     assert!(rx.recv().await.is_none());
+    ///     
+    ///     assert!(rx.is_closed());
+    /// }
+    /// ```
+    pub fn is_closed(&mut self) -> bool {
```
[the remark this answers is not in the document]
Darksonn: So I was going through this one more time and I realized that you may be right here. Currently, it's possible to have a situation where `Sender::is_closed` returns true but `Receiver::is_closed()` returns false. That's probably not desirable. It probably makes more sense to change it to match the existing method on the sender.

Of course, another option is to rename this method to reflect its actual behavior.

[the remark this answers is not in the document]
balliegojr: > Currently, it's possible to have a situation where Sender::is_closed returns true but Receiver::is_closed() returns false. That's probably not desirable. It probably makes more sense to change it to match the existing method on the sender.

Yeah, I totally agree with this.  I think the `is_closed` should return true if `close` was called or all the senders are dropped, regardless of having any messages available in the channel.

I believe it would be a good idea to provide an additional function, maybe `is_empty` or `has_messages`, that checks if there are any messages available in the channel, this way the user would be able to check both  conditions with `rx.is_closed() && rx.is_empty()`.

What do you think?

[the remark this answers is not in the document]
Darksonn: That seems reasonable to me. Adding an `is_empty` method could make even more sense together with adding a `len` method (which would close #6314). After all, clippy emits a warning if you add `len` without `is_empty`.

But if you only want to add `is_empty`, that's also okay.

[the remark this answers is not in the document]
balliegojr: I have added the `is_empty` and `len` functions. 

The `len` implementation only checks for the `tail_position`. I am not sure if it is necessary to check the `ready` bits for valid values in each position, since that would be relatively costly because of the linked list implementation

tokio/src/sync/mpsc/list.rs
```diff
@@ -230,6 +230,28 @@ impl<T> fmt::Debug for Tx<T> {
...
+
+        unsafe {
+            let tail_block = &mut *tail;
+
+            if tail_block.is_closed() {
+                tail_position - self.index - 1
+            } else {
+                tail_position - self.index
+            }
+        }
```
Darksonn, line 252: Please reduce the scope of this unsafe block. You only need it for the `&mut *tail` operation as far as I can tell.
```Rust
let tail_block = unsafe { &mut *tail };
```

tokio/tests/sync_mpsc.rs
Darksonn, line 1: That's a lot of tests! That's awesome.

tokio/src/sync/mpsc/bounded.rs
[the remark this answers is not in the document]
Darksonn: Hmm, regarding the ready bits, I guess the main question here is concurrency-related. For example, imagine if you have the following two threads running in parallel:
```Rust
sender.send("message1").await;
assert!(receiver.len() != 0); // or similar with `is_empty`
```
```Rust
sender.send("message2").await;
```
Then, depending on how you implement `len`, it's possible that the assertion could fail. This would happen if message1 ends up being stored _after_ message2 in the channel queue, but when the assert runs, message2 is has not yet been fully sent.

We once had a similar bug with `try_recv`. See #1997. Basically, this assert could fail:
```Rust
sender.send("message1").await;
assert!(receiver.try_recv().is_some());
```
```Rust
sender.send("message2").await;
```
The assert would sometimes fail when message2 ends up before message1 in the queue. Here, `try_recv` can't return message1 because we guarantee that messages are returned in order. It also can't return message2 because the call to `send("message2")` is still running. So it returns `None` instead, even though we _know_ that a message has been successfully sent. When we fixed it in #4113, we did so by having `try_recv` sleep until `send("message2")` finished when there are fully sent messages later in the queue.

To test this, you can add these loom tests:
```Rust
// tokio/src/sync/tests/loom_mpsc.rs
#[test]
fn len_nonzero_after_send() {
    loom::model(|| {
        let (send, recv) = mpsc::unbounded_channel();
	let send2 = send.clone();

	let join = thread::spawn(move || {
	    block_on(send2.send("message2")).unwrap();
	});

        block_on(send.send("message1")).unwrap();
        assert!(recv.len() != 0);

        join.join().unwrap();
    });
}

#[test]
fn nonempty_after_send() {
    loom::model(|| {
        let (send, recv) = mpsc::unbounded_channel();
	let send2 = send.clone();

	let join = thread::spawn(move || {
	    block_on(send2.send("message2")).unwrap();
	});

        block_on(send.send("message1")).unwrap();
        assert!(!recv.is_empty());

        join.join().unwrap();
    });
}
```
I don't know whether these tests will fail, but if they do, please look at `try_recv` to see how it determines whether it should sleep or not.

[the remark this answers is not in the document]
balliegojr: Both tests failed. I am looking on how to fix it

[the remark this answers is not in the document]
balliegojr: I added both loom tests and fixed the `is_empty` implementation.

About `len_nonzero_after_send` test, I think the `len` implementation was already correct, but I ran the wrong test initially. 
At first I had written the assertion `assert!(recv.len() == 2)`, which should fail since the `join()` is after the `assert!`.

With the `assert!(recv.len() != 0);` assertion, the test passes

tokio/src/sync/mpsc/list.rs
```diff
@@ -230,6 +230,49 @@ impl<T> fmt::Debug for Tx<T> {
...
+    }
+
+    pub(crate) fn len(&self, tx: &Tx<T>) -> usize {
+        let tail_position = tx.tail_position.load(Acquire);
+        let tail = tx.block_tail.load(Acquire);
+
+        let is_closed = unsafe {
+            let tail_block = &mut *tail;
+            tail_block.is_closed()
```
Darksonn, line 266: The `is_closed` method takes `&self`. Please either change this to `&*tail` or change `is_closed` to `&mut self` depending on which one is correct.

Same in `is_empty`. In fact, it seems like you could factor out the `is_empty` logic into a separate function in `list.rs`.

tokio/src/sync/mpsc/list.rs
```diff
@@ -230,6 +230,49 @@ impl<T> fmt::Debug for Tx<T> {
...
+        };
+
+        if is_closed {
+            tail_position - self.index - 1
+        } else {
+            tail_position - self.index
+        }
```
Darksonn, line 273:
```suggestion
        tail_position - self.index - (is_closed as usize)
```

tokio/src/sync/mpsc/list.rs
```diff
@@ -230,6 +230,49 @@ impl<T> fmt::Debug for Tx<T> {
...
+        } else if tail_position - self.index == 1 {
+            let tail = tx.block_tail.load(Acquire);
+
+            unsafe {
+                let tail_block = &mut *tail;
+                tail_block.is_closed()
+            }
+        } else {
+            false
+        }
```
Darksonn, line 257: I'm thinking we could make this clearer by using a match statement:
```suggestion
        match tail_position - self.index {
            0 => true,
            1 => tx.is_closed(),
            _ => false,
        }
```

Darksonn (commented):
The code looks correct, but I think we can simplify it. Please see my suggestions below.

tokio/src/sync/mpsc/list.rs
balliegojr: Nice, thanks for the feedback! The code is much better now

tokio/src/sync/mpsc/list.rs
```diff
@@ -230,6 +239,24 @@ impl<T> fmt::Debug for Tx<T> {
...
+        // To be sure, it is necessary to check the length of the list.
+        self.len(tx) == 0
+    }
+
+    pub(crate) fn len(&self, tx: &Tx<T>) -> usize {
+        // When all the senders are dropped, there will be a last block in the tail position,
+        // but it will be closed
+        let tail_position = tx.tail_position.load(Acquire);
+        tail_position - self.index - (tx.is_closed() as usize)
+    }
```
Darksonn, line 258: Do you have a test for whether the length is correct in all of the following cases?

* There are still senders, but `Receiver::close` has been called.
* There are no more senders, but `Receiver::close` has not been called.
* There are no more senders, and `Receiver::close` has been called before the last sender dropped.
* There are no more senders, and `Receiver::close` has been called after the last sender dropped.

balliegojr: Some of the scenarios were missing, but I have added them now.

Darksonn (approved):
Thanks.

Issue #6359, opened 2024-02-19
docsrs edge case when building docs without "net" feature


qrnch-jan:
**Version**
1.36.0

**Platform**
Darwin whoop.qlan 23.2.0 Darwin Kernel Version 23.2.0: Wed Nov 15 21:54:10 PST 2023; root:xnu-10002.61.3~2/RELEASE_X86_64 x86_64

**Description**
I have a project that can use tokio to run a message passing dispatch loop, which means it only needs `tokio::task::spawn()`, and thus only needs the `rt` feature.  When I try to build the docs for my project using the feature labels feature:

```
$ RUSTFLAGS="--cfg docsrs" RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --all-features
```

I get an error while generating docs for tokio:

```
    Checking tokio v1.36.0
error[E0432]: unresolved import `self::doc::os`
   --> /Users/jan/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.36.0/src/lib.rs:635:16
    |
635 | pub(crate) use self::doc::os;
    |                ^^^^^^^^^^^^^ no `os` in `doc`
    |
note: found an item that was configured out
   --> /Users/jan/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.36.0/src/doc/mod.rs:47:9
    |
47  | pub mod os;
    |         ^^
    = note: the item is gated behind the `net` feature
help: consider importing this module instead
    |
635 | pub(crate) use std::os;
    |                ~~~~~~~

 Documenting tokio v1.36.0
For more information about this error, try `rustc --explain E0432`.
error: could not document `tokio`
```

This problem goes away if the `net` feature is added to tokio's features list in `Cargo.toml`.

The reason is pretty obvious; `lib.rs` has:

```rust
#[cfg(docsrs)]
pub mod doc;

#[cfg(docsrs)]
#[allow(unused)]
pub(crate) use self::doc::os;
```

.. while `doc/mod.rs` has:

```rust
#[cfg(feature = "net")]
pub mod os;
```

Not sure what the appropriate solution is though.  The `os` submodule looks like it's for Windows only?  Should the top level module have:

```rust
#[cfg(feature = "net")]
#[cfg(docsrs)]
#[allow(unused)]
pub(crate) use self::doc::os;
```

PR #6360, opened 2024-02-19
Put import of os submodule behind 'net' feature to fix doc generation…

qrnch-jan:
… without the 'net' feature.

Fixes https://github.com/tokio-rs/tokio/issues/6359.

## Motivation

Fixes doc generation without `net` feature.

## Solution

Put `os` submodule import into root module behind `feature="net"` gate, because of assumptions made elsewhere.

Tested on macOS.

PS. Didn't bother to update change log because it's such an edge-case.


Darksonn (approved):
Thank you.

PR #6371, opened 2024-02-29
tokio-test: mark `tokio_test::task::Spawn` as `must_use`

mox692:
A task passed to `Spawn` will not be executed unless it is polled, and it would be more helpful to add `must_use` annotation.


Darksonn (approved):
Thank you. Good call.

PR #6378, opened 2024-03-04
Update mio version

psandana:
Upgrade mio to latest version.

## Motivation

To consume latest fixes in mio.

## Solution

Upgrade mio version to `0.8.11`

### DIFF
--- a/tokio/Cargo.toml
+++ b/tokio/Cargo.toml
@@ -95,7 +95,7 @@ pin-project-lite = "0.2.11"
 
 # Everything else is optional...
 bytes = { version = "1.0.0", optional = true }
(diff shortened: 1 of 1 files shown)


psandana:
@seanmonstar CI failed one job which seems unrelated to my changes (or am I missing something?)

psandana:
@seanmonstar rebased with `master`. CI should pass now. Can you approve?

seanmonstar:
Based on my initial comment, I don't believe this is something that needs to be merged into Tokio. It doesn't reduce the amount of updates anyone else would need to do. It doesn't make it any easier to use the newest version.

So, I'm inclined to close. If any other maintainer disagrees, they can reopen.

PR #6397, opened 2024-03-13
tokio-util: remove `'static` bound on `impl Sink for PollSender`

nyibbang:
## Motivation

In PR #5665, the `'static` bound has been removed on values sent into `PollSender`. One of this bound was remaining on the `PollSender` implementation of `Sink`.

## Solution

This patch removes the bound and adds some tests on the `Sink` interface for `PollSender`.


Darksonn (approved):
Thank you.

PR #6408, opened 2024-03-16
readme: fix the guide for running loom tests

maminrayej:
I found out that the guide for running loom tests doesn't work. This PR adds:
* `-C debug_assertions` to `RUSTFLAGS` since there is a `compile_error` that prevents the loom tests in the `runtime` module to run without debug assertions.
* `LOOM_MAX_BRANCHES=10000` to make it consistent with how CI is configured.
* `#![cfg(tokio_unstable)]` to `loom_multi_thread_alt.rs` to gate this test file appropriately.

### DIFF
--- a/tokio/src/runtime/tests/loom_multi_thread_alt.rs
+++ b/tokio/src/runtime/tests/loom_multi_thread_alt.rs
@@ -1,3 +1,5 @@
+#![cfg(tokio_unstable)]
+
 mod queue;
 mod shutdown;
 mod yield_now;
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -184,9 +184,11 @@ it, `rustfmt` will update your files locally instead.
 You can run loom tests with
 ```
 cd tokio # tokio crate in workspace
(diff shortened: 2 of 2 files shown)


tokio/src/runtime/tests/loom_multi_thread_alt.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(tokio_unstable)]
```
mox692, line 1: Maybe we want to put this annotation on files under `tokio/src/runtime/tests/loom_multi_thread_alt/*` as well?

maminrayej: isn't `loom_multi_thread_alt.rs` basically the `mod.rs` file for that directory? So I believe by putting the annotation in `loom_multi_thread_alt.rs`, the whole module is affected.

mox692: Ah my bad, you're right. Annotating against `loom_multi_thread_alt.rs` is enough.

mox692 (approved):
Looks good to me.

PR #6411, opened 2024-03-18
fs: add `set_max_buf_size` to `tokio::fs::File`

mox692:
## Motivation
https://github.com/tokio-rs/tokio/pull/5397 enhanced performance by updating the `tokio::fs::File`'s MAX_BUF from 16KiB to 2MiB for operations involving `AsyncWrite` / `AsyncRead`. Related to this, it would also be useful to provide an API that allows to customize this default.

## Solution
This introduce a `set_max_buf_size` to `tokio::fs::File` to customize the maximum buffer size.
If this is not set, the current default of 2MiB will be applied. In the current implementation, this `max_buf_size` will be applicable for both read and write operations.

### DIFF
--- a/tokio/src/fs/file.rs
+++ b/tokio/src/fs/file.rs
@@ -508,6 +508,35 @@ impl File {
         let std = self.std.clone();
         asyncify(move || std.set_permissions(perm)).await
     }
+
+    /// Set the maximum buffer size for the underlying [`AsyncRead`] / [`AsyncWrite`] operation.
+    ///
+    /// Although Tokio uses a sensible default value for this buffer size, this function would be
+    /// useful for changing that default depending on the situation.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// use tokio::fs::File;
+    /// use tokio::io::AsyncWriteExt;
+    ///
+    /// # async fn dox() -> std::io::Result<()> {
+    /// let mut file = File::open("foo.txt").await?;
+    ///
+    /// // Set maximum buffer size to 8 MiB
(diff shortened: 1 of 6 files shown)


tokio/src/fs/file.rs
```diff
@@ -508,6 +508,35 @@ impl File {
...
+    ///
+    /// // Write the 1 GiB buffer in chunks up to 8 MiB each.
+    /// file.write_all(&mut buf).await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn set_max_buf_size(&mut self, max_buf_size: usize) {
+        let mut inner = self.inner.lock().await;
+        inner.state = State::Idle(Some(Buf::with_capacity_and_max_buf_size(0, max_buf_size)));
+    }
```
maminrayej, line 539: I'm a little bit confused why exclusive access is needed for this since we're locking `inner` anyway.
Why not `&self`, lock the `inner`, await any inflight operation, and then replace the buffer?

carllerche:
I think it should be doable as a sync function if you set a value somewhere (max buf size) that is checked when the buffer is written to.

mox692:
@carllerche 
If such an implementation is still acceptable, I think we can do it.
I'll give it a try tomorrow.

mox692:
Okay, I think this is ready for another look. The implementation is still simpler than I expected.

Also, I noticed that the inner `Buf` type is also used for `io::Stdin`, `io::Stdout`, but I've left these as they are, using 2MiB. (If we wanted to provide the same means of customization for `Stdin`, `Stdout`, we could easily do so.)

maminrayej (approved):

PR #6443, opened 2024-03-30
chore: update CI to clippy 1.77

Siech0:
## Motivation
This updates the CI to use clippy version 1.77 which is aligned with the rust stable version.

## Solution

**Ignored Lints**
The following functions accept the `clippy:unused_io_amount` by the virtue that they handle left over IO in that they are purposefully discarding all IO.
* tokio/src/runtime/signal/mod.rs - process
* tokio/tests/io_async_fd.rs - drain


mox692 (approved):
Thanks!

PR #6449, opened 2024-04-01
io: implement AsyncBufRead for Join

bharath-123:
## Motivation
The Join struct implements the AsyncRead and AsyncWrite interface. This PR implements the AsyncReadBuf interface.
Fixes https://github.com/tokio-rs/tokio/issues/6446

## Solution
We implement the AsyncReadBuf interface for the Join, if the Reader also implements the AsyncReadBuf interface.


mox692 (approved):
Looks good to me, thank you.

PR #6456, opened 2024-04-02
docs: add README.md for tokio-stream

aoikurokawa:
## Motivation

#6455 

## Solution

- Add README.md similar to tokio-util.


mox692 (approved):
Thanks!

PR #6465, opened 2024-04-06
Add Apple visionOS support

QuentinPerez:
## Motivation

Add the support of Apple VisionOS  [121419](https://github.com/rust-lang/rust/pull/121419)

## Solution
Add target_os = "visionos"


Darksonn (approved):
Thanks.

PR #6468, opened 2024-04-06
macros: render more comprehensible documentation for `select!`

dtolnay:
## Motivation

Screenshot from 2024-04-06 15-16-57

## Solution

Screenshot from 2024-04-06 15-17-10

### DIFF
--- a/tokio/src/macros/select.rs
+++ b/tokio/src/macros/select.rs
@@ -396,8 +396,27 @@
 ///     }
 /// }
 /// ```
+#[cfg(doc)]
 #[macro_export]
 #[cfg_attr(docsrs, doc(cfg(feature = "macros")))]
+macro_rules! select {
+    {
+        $(
+            biased;
+        )?
+        $(
+            $bind:pat = $fut:expr $(, if $cond:expr)? => $handler:expr,
+        )*
+        $(
+            else => $els:expr $(,)?
+        )?
+    } => {
+        unimplemented!()
+    };
+}
+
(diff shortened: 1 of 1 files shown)


tokio/src/macros/select.rs
```diff
@@ -396,8 +396,27 @@
...
+#[cfg(doc)]
 #[macro_export]
 #[cfg_attr(docsrs, doc(cfg(feature = "macros")))]
+macro_rules! select {
+    {
+        $(
+            biased;
+        )?
+        $(
+            $bind:pat = $fut:expr $(, if $cond:expr)? => $handler:expr /*optional comma*/,
```
Darksonn, line 408: I might even drop the comment about the comma. Doesn't seem that important.
```suggestion
            $bind:pat = $fut:expr $(, if $cond:expr)? => $handler:expr,
```

dtolnay:
> Does this hide docs from ide tooltips?

Yes, seems it does. Good call.

I found a way to fix it.

tokio/src/macros/select.rs
dtolnay: 👍

Darksonn (approved):
Thanks for the doc improvement! Looks good to me.

The extra indentation is unfortunate, but oh well.

dtolnay:
In fact the extra indentation wrecked the docs, too.

image

I pushed a different approach without indentation, that rustfmt is still okay with.

Darksonn:
Good catch.

PR #6479, opened 2024-04-12
docs: add --locked to cargo install commands

joshka:
https://github.com/tokio-rs/tokio/pull/6475\#discussion_r1561492918


taiki-e (approved):
Thanks.

PR #6496, opened 2024-04-18
docs: Runtime::metrics() documentation

kvcache:
reused the comment on Handle::metrics() to help communicate that
they are the same thing.

Resolves: https://github.com/tokio-rs/tokio/issues/6486

I tried to make a doctest but struggled to get Cargo to actually run it.
The --cfg tokio_unstable was able to get the compiler to notice syntax
errors in the cfg_metrics!{} block, but it did not run the doctest.

Rather than commit a non-running doctest, I simply followed suit with
the existing comment on runtime::Handle::metrics.


Darksonn (approved):
Thanks.

PR #6498, opened 2024-04-19
Fix obsolete reference to “stream ID” in `ReadHalf::unsplit()` documentation.

kpreid:
## Motivation

`SplitStreamId`s were added in 5bf78d77ada685826b33fd62e69df179f3de8a1c and removed the next day in c7719a2d2962f83854193b4c9131ee275a5a4475, and never made it into a release. Therefore, the documentation for `unsplit()` mentioning “comparing the stream ID” is incorrect (and has been since its introduction in tokio 0.2.10).

## Solution

Describe the actual method for checking the precondition of `ReadHalf::unsplit()`: calling `ReadHalf::is_pair_of()`.


mox692 (approved):
Thanks!

PR #6503, opened 2024-04-21
docs: fix time::{pause,resume} requirement

jainl28patel:
## Motivation
Fixes : #6495 

## Solution
feature list in rust nightly requires `#[cfg_attr(docsrs, doc(cfg(feature = "<feature-name>")))]` whose absence will result into feature requirement not being detected in doc.


Darksonn (approved):
Thanks.

PR #6507, opened 2024-04-22
Fix document on watch

takassh:
## Motivation

`tokio::sync::watch` should be a single-producer,  multi-consumer channel afaik.
Actually, it can't be cloned.
So the current documentation may lead misunderstanding.

## Solution

Fixed documentation

### DIFF
--- a/tokio/src/sync/watch.rs
+++ b/tokio/src/sync/watch.rs
@@ -1,6 +1,6 @@
 #![cfg_attr(not(feature = "sync"), allow(dead_code, unreachable_pub))]
 
-//! A multi-producer, multi-consumer channel that only retains the *last* sent
+//! A single-producer, multi-consumer channel that only retains the *last* sent
 //! value.
 //!
 //! This channel is useful for watching for changes to a value from multiple


mox692:
Thanks for the PR.

`Clone` of `watch::Sender` was introduced in [this PR](https://github.com/tokio-rs/tokio/pull/6388), and that change was shipped in the [latest v1.37.0 release](https://github.com/tokio-rs/tokio/releases/tag/tokio-1.37.0).

If you are not using v1.37.0 tokio, then you probably cannot use Sender's Clone.

takassh:
@mox692 
Ah thank you!
then this document may be wrong?
https://tokio.rs/tokio/tutorial/channels#tokios-channel-primitives

which says `watch: single-producer, multi-consumer. Many values can be sent, but no history is kept. Receivers only see the most recent value.`

Can I fix it?

mox692:
Thanks, that correction would be helpful.

takassh:
Thanks for the quick response!
let me close this.

PR #6508, opened 2024-04-23
Remove unnecessary uses of #[cfg(not(test))]

avtrujillo:
## Motivation

The main and main-rt macro functions both have the #[cfg(not(test))] with a comment indicating that its inclusion is a workaround for rust-lang/rust#62127, which has since been resolved. Because rust-analyzer enables test by default, this means that these functions are greyed out when using rust-analyzer in VSCode without configuration changes.

## Solution

Remove all uses of #[cfg(not(test))] in the tokio-macros crate that are accompanied by comments mentioning rust-lang/rust#62127 as the motivation for usage.

It's possible that the usage in https://github.com/tokio-rs/tokio/blob/8093712604e54a434658e693ca80535d906e26a7/tokio/src/lib.rs#L677 is also unnecessary, since it appears in the context of using main as an identifier. I'm less sure about that one since it doesn't have a comment mentioning rust-lang/rust#62127, so I've left it alone for now.

### DIFF
--- a/tokio-macros/src/entry.rs
+++ b/tokio-macros/src/entry.rs
@@ -418,7 +418,6 @@ fn token_stream_with_error(mut tokens: TokenStream, error: syn::Error) -> TokenS
     tokens
 }
 
-#[cfg(not(test))] // Work around for rust-lang/rust#62127
 pub(crate) fn main(args: TokenStream, item: TokenStream, rt_multi_thread: bool) -> TokenStream {
     // If any of the steps for this macro fail, we still want to expand to an item that is as close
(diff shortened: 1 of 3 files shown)


GlenDC (approved):
All good as far as I can see.
Pretty certain you can also remove the same thing in https://github.com/tokio-rs/tokio/blob/8093712604e54a434658e693ca80535d906e26a7/tokio/src/lib.rs#L677, as it is the same situation as the `cfg_rt!` case AFAIK.

A tokio maintainer can probably confirm, this and all, but to me it does look good and working as expected.

Darksonn (approved):
Thank you.

PR #6520, opened 2024-04-27
Implementation of notify_one_last method

pfreixes:
A new notify_one_last_in method is implemented for notifying the most recent waiter, LIFO, instead of the oldest one, FIFO, which is used by the current notify_one method.

## Motivation

As explained here https://github.com/tokio-rs/tokio/issues/6506 the `notify_one_last_in` method provides the user a method for waking up the most recent waiter rather than the oldest one. When using a LIFO strategy we sacrifice waiters that have been waiting longer by first dequeuing the ones that have been in the queue for shorter time, this should provide benefits in situations of contention where LIFO dequeueing should provide a better latency observation compared to the usage of the FIFO.
## Solution

Ive followed main recommendations for storing the strategy for dequeueing followed by the user as  part of the waiter notification state. No impact on the memory footprint should be observed.

### DIFF
--- a/tokio/src/sync/notify.rs
+++ b/tokio/src/sync/notify.rs
@@ -587,6 +587,8 @@ impl Notify {
     ///
     /// Check the [`notify_one()`] documentation for more info and
     /// examples.
+    ///
+    /// [`notify_one()`]: Notify::notify_one
     pub fn notify_one_last_in(&self) {
         self.notify_with_strategy(NotifyOneStrategy::Lifo);
     }
--- a/tokio/src/sync/notify.rs
+++ b/tokio/src/sync/notify.rs
@@ -224,8 +224,8 @@ struct Waiter {
     waker: UnsafeCell<Option<Waker>>,
 
     /// Notification for this waiter. Uses 2 bits to store if and how was
-    /// notified, 2 bit for storing if was woken up using FIFO or LIFO and
-    /// the rest of it are unused.
+    /// notified, 2 bits for storing if it was woken up using FIFO or LIFO, and
+    /// the rest of it is unused.
     /// * if it's `None`, then `waker` is protected by the `waiters` lock.
     /// * if it's `Some`, then `waker` is exclusively owned by the
     ///   enclosing `Waiter` and can be accessed without locking.
@@ -292,7 +292,11 @@ impl AtomicNotification {
     fn store_release(&self, notification: Notification, strategy: Option<NotifyOneStrategy>) {
         let data: usize = match strategy {
             None => notification as usize & NOTIFICATION_TYPE_MASK,
-            Some(strategy) => (((strategy as usize) << NOTIFICATION_NOTIFY_ONE_STRATEGY_SHIFT) & NOTIFICATION_NOTIFY_ONE_STRATEGY_MASK) | (notification as usize & NOTIFICATION_TYPE_MASK)
+            Some(strategy) => {
+                (((strategy as usize) << NOTIFICATION_NOTIFY_ONE_STRATEGY_SHIFT)
+                    & NOTIFICATION_NOTIFY_ONE_STRATEGY_MASK)
+                    | (notification as usize & NOTIFICATION_TYPE_MASK)
+            }
         };
         self.0.store(data, Release);
     }
@@ -305,7 +309,9 @@ impl AtomicNotification {
             NOTIFICATION_ALL => Some(Notification::All),
             _ => unreachable!(),
         };
-        let strategy = match (data & NOTIFICATION_NOTIFY_ONE_STRATEGY_MASK) >> NOTIFICATION_NOTIFY_ONE_STRATEGY_SHIFT {
+        let strategy = match (data & NOTIFICATION_NOTIFY_ONE_STRATEGY_MASK)
+            >> NOTIFICATION_NOTIFY_ONE_STRATEGY_SHIFT
+        {
             NOTIFICATION_NOTIFY_ONE_STRATEGY_NONE => None,
             NOTIFICATION_NOTIFY_ONE_STRATEGY_FIFO => Some(NotifyOneStrategy::Fifo),
             NOTIFICATION_NOTIFY_ONE_STRATEGY_LIFO => Some(NotifyOneStrategy::Lifo),
@@ -489,7 +495,7 @@ impl Notify {
     pub const fn const_new() -> Notify {
         Notify {
             state: AtomicUsize::new(0),
-            waiters: Mutex::const_new(LinkedList::new())
+            waiters: Mutex::const_new(LinkedList::new()),
         }
     }
 
@@ -597,12 +603,11 @@ impl Notify {
 
     /// Notifies the last waiting task.
     ///
-    /// This function behaves identically as `notify_one` but using a
-    /// LIFO algorithm for dequeing the waiters, if there are.
+    /// This function behaves identically to `notify_one` but using a
+    /// LIFO algorithm for dequeuing the waiters, if there are any.
     ///
     /// Check the `notify_one` documentation for more info and
     /// examples.
-    ///
     pub fn notify_one_lifo(&self) {
         self.notify_with_strategy(NotifyOneStrategy::Lifo);
     }
@@ -610,7 +615,6 @@ impl Notify {
     fn notify_with_strategy(&self, strategy: NotifyOneStrategy) {
         // Load the current state
         let mut curr = self.state.load(SeqCst);
-
 
         // If the state is `EMPTY`, transition to `NOTIFIED` and return.
         while let EMPTY | NOTIFIED = get_state(curr) {
@@ -622,9 +626,7 @@ impl Notify {
 
             match res {
                 // No waiters, no further work to do
-                Ok(_) => {
-                    return
-                }
+                Ok(_) => return,
                 Err(actual) => {
                     curr = actual;
                 }
@@ -637,6 +639,7 @@ impl Notify {
         // The state must be reloaded while the lock is held. The state may only
         // transition out of WAITING while the lock is held.
         curr = self.state.load(SeqCst);
+
         if let Some(waker) = notify_locked(&mut waiters, &self.state, curr, strategy) {
             drop(waiters);
             waker.wake();
@@ -760,7 +763,12 @@ impl Default for Notify {
 impl UnwindSafe for Notify {}
 impl RefUnwindSafe for Notify {}
 
-fn notify_locked(waiters: &mut WaitList, state: &AtomicUsize, curr: usize, strategy: NotifyOneStrategy) -> Option<Waker> {
+fn notify_locked(
+    waiters: &mut WaitList,
+    state: &AtomicUsize,
+    curr: usize,
+    strategy: NotifyOneStrategy,
+) -> Option<Waker> {
     match get_state(curr) {
         EMPTY | NOTIFIED => {
             let res = state.compare_exchange(curr, set_state(curr, NOTIFIED), SeqCst, SeqCst);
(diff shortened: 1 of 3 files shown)


tokio/src/sync/notify.rs
```diff
@@ -1138,7 +1205,9 @@ impl Drop for Notified<'_> {
             // the notification was triggered via `notify_one`, it must be sent
             // to the next waiter.
             if notification == Some(Notification::One) {
-                if let Some(waker) = notify_locked(&mut waiters, &notify.state, notify_state) {
+                if let Some(waker) =
+                    notify_locked(&mut waiters, &notify.state, notify_state, strategy.unwrap())
```
Darksonn, line 1209: I think we can drop the `Option` around the strategy. Then we don't need an `unwrap` here. We can ignore the strategy when we are waking everything.

pfreixes: Would you prefer to just pass or return the `NOTIFICATION_NOTIFY_ONE_STRATEGY_NONE`?  If so we can also avoid this two indirection using the const and the enum and having somthing like

```
 enum NotifyOneStrategy {
    None = 0
    Fifo = 1,
    Lifo = 2,
}
```

And use the `NotifyOneStrategy` as a parameter for the [store](https://github.com/tokio-rs/tokio/pull/6520/files#diff-e6a0a252981328ec25e728ff0cf7c30125c33e85a73667d31c971e8efecab68eR292) and as a return value for the [load](https://github.com/tokio-rs/tokio/pull/6520/files#diff-e6a0a252981328ec25e728ff0cf7c30125c33e85a73667d31c971e8efecab68eR304).

tokio/src/sync/notify.rs
```diff
@@ -269,30 +284,48 @@ struct AtomicNotification(AtomicUsize);
 
 impl AtomicNotification {
     fn none() -> Self {
-        AtomicNotification(AtomicUsize::new(NOTIFICATION_NONE))
+        AtomicNotification(AtomicUsize::new(0))
```
carllerche, line 287: Is there a reason this isn't a const anymore?

tokio/src/sync/notify.rs
```diff
@@ -269,30 +284,48 @@ struct AtomicNotification(AtomicUsize);
...
     }
 
     /// Store-release a notification.
     /// This method should be called exactly once.
-    fn store_release(&self, notification: Notification) {
-        self.0.store(notification as usize, Release);
+    fn store_release(&self, notification: Notification, strategy: Option<NotifyOneStrategy>) {
```
carllerche, line 292: Do we need `Option` here or could callers default to a strategy?

tokio/src/sync/notify.rs
```diff
@@ -269,30 +284,48 @@ struct AtomicNotification(AtomicUsize);
...
+                    & NOTIFICATION_NOTIFY_ONE_STRATEGY_MASK)
+                    | (notification as usize & NOTIFICATION_TYPE_MASK)
+            }
+        };
+        self.0.store(data, Release);
     }
 
-    fn load(&self, ordering: Ordering) -> Option<Notification> {
-        match self.0.load(ordering) {
+    fn load(&self, ordering: Ordering) -> (Option<Notification>, Option<NotifyOneStrategy>) {
```
carllerche, line 304: Is there ever a time when this will return `None` for `Notification` but will return `Some` for `NotifyOneStrategy`? I don't think there is. Perhaps, `Notification` the enum should take the strategy as an argument:

```rust
enum Notification {
    One(Strategy),
    All,
}
```

tokio/src/sync/notify.rs
```diff
@@ -558,6 +598,21 @@ impl Notify {
...
+    }
+
+    /// Notifies the last waiting task.
+    ///
+    /// This function behaves identically to `notify_one` but using a
+    /// LIFO algorithm to notify waiters from the queue, if there are any.
+    ///
+    /// Check the `notify_one` documentation for more info and
+    /// examples.
+    pub fn notify_one_lifo(&self) {
```
carllerche, line 611: I'm not against calling this `notify_one_lifo`. As a consideration, do you think `notify_last_in` or `notify_one_last_in` could be better options?

tokio/src/sync/notify.rs
carllerche: I commented above, but it looks like the strategy is an argument to `Notification`. Something like:

```rust
enum Notification {
    One(Strategy),
    All,
}
```

You can implement the conversion of the enum to size using a match statement, the compiler will optimize it anyway.

carllerche (changes requested):
Left some thoughts inline.

tokio/src/sync/notify.rs
pfreixes: Good point, let me change this towards your suggestion.

tokio/src/sync/notify.rs
pfreixes: `notify_one_last_in` sounds better, users will easily understand the semantics of the method. Ill change this one.

tokio/src/sync/notify.rs
pfreixes: Yeps, changing a bit this one.

tokio/src/sync/notify.rs
pfreixes: Maybe there was since bot attributes, if there were a notification and the strategy, but once Ive changed the interfaces we can go back to what was originally, so this should be "fixed" now.

tokio/src/sync/notify.rs
pfreixes: This should no longer apply after the latest refactor.

pfreixes:
@carllerche and @Darksonn PR should be ready for another look 🙇

pfreixes:
Any new feedback on this one? 🙇

tokio/src/sync/notify.rs
```diff
@@ -275,13 +287,41 @@ impl AtomicNotification {
...
+                (((NOTIFICATION_NOTIFY_ONE_STRATEGY_FIFO)
+                    << NOTIFICATION_NOTIFY_ONE_STRATEGY_SHIFT)
+                    & NOTIFICATION_NOTIFY_ONE_STRATEGY_MASK)
+                    | (NOTIFICATION_ONE & NOTIFICATION_TYPE_MASK)
+            }
+            Notification::One(NotifyOneStrategy::Lifo) => {
+                (((NOTIFICATION_NOTIFY_ONE_STRATEGY_LIFO)
+                    << NOTIFICATION_NOTIFY_ONE_STRATEGY_SHIFT)
+                    & NOTIFICATION_NOTIFY_ONE_STRATEGY_MASK)
+                    | (NOTIFICATION_ONE & NOTIFICATION_TYPE_MASK)
```
Darksonn, line 302: This is pretty hard to read. Can we just make three constants equal to `0b00`, `0b01`, `0b10` or whatever the values are?

tokio/src/sync/notify.rs
```diff
@@ -558,6 +605,21 @@ impl Notify {
     // Alias for old name in 0.x
     #[cfg_attr(docsrs, doc(alias = "notify"))]
     pub fn notify_one(&self) {
+        self.notify_with_strategy(NotifyOneStrategy::Fifo);
+    }
+
+    /// Notifies the last waiting task.
+    ///
+    /// This function behaves identically to `notify_one` but using a
+    /// LIFO algorithm to notify waiters from the queue, if there are any.
```
Darksonn, line 614:
```suggestion
    /// This function behaves similar to `notify_one`. The only difference is that it wakes
    /// the most recently added waiter instead of the oldest waiter.
```

tokio/src/sync/notify.rs
```diff
@@ -558,6 +605,21 @@ impl Notify {
...
+    ///
+    /// Check the `notify_one` documentation for more info and
```
Darksonn, line 616:
```suggestion
    /// Check the [`notify_one`] documentation for more info and
```

tokio/src/util/linked_list.rs
```diff
@@ -144,6 +144,26 @@ impl<L: Link> LinkedList<L, L::Target> {
...
 
+    /// Removes the first element from a list and returns it, or None if it is
+    /// empty.
+    pub(crate) fn pop_front(&mut self) -> Option<L::Handle> {
+        unsafe {
+            let head = self.head?;
+            self.head = L::pointers(head).as_ref().get_next();
+
+            if let Some(next) = L::pointers(head).as_ref().get_next() {
+                L::pointers(next).as_mut().set_prev(None);
```
Darksonn, line 155:
```suggestion
            if let Some(new_head) = L::pointers(head).as_ref().get_next() {
                L::pointers(new_head).as_mut().set_prev(None);
```

tokio/tests/sync_notify.rs
```diff
@@ -105,6 +137,49 @@ fn notified_multi_notify_drop_one() {
...
+    let mut notified3 = spawn(async { notify.notified().await });
+
+    // add waiters by order of poll execution
+    assert_pending!(notified1.poll());
+    assert_pending!(notified2.poll());
+    assert_pending!(notified3.poll());
+
+    notify.notify_one_last_in();
+
+    drop(notified1);
```
Darksonn, line 176: Shouldn't you drop `notified3` to trigger the forwarding logic?
```suggestion
    drop(notified3);
```

Darksonn (commented):
Hi again. Sorry for the delay. I was at RustNL this week.

tokio/tests/sync_notify.rs
pfreixes: My bad, fixed!

tokio/src/sync/notify.rs
pfreixes: Good call, Ive changed for using just constants which makes definitely more readable the code

pfreixes:
@Darksonn thanks for the review, comments should be addressed by now.

pfreixes:
@Darksonn @carllerche any feedback on this one? 🙇

Darksonn (approved):
The implementation looks good to me. But I have a question about the naming. Why `notify_one_last_in`? Last in what?

pfreixes:
> The implementation looks good to me. But I have a question about the naming. Why `notify_one_last_in`? Last in what?

last in to be added into the queue. @carllerche recommended this name over the `notify_one_lifo` since seemed more concrete, and I kinda agree.

pfreixes:
@carllerche any chance that you could take another look to the PR?

carllerche (commented):
Thanks for the changes. After a quick chat in Discord, @Darksonn convinced me that `notify_last` would be a better name.

Sorry for the churn. The code is good, could you change the name to `notify_last` and then we should be good to go.

pfreixes:
No problemo @carllerche, I already renamed it in my latest commit, PTAL when you have a moment /cc @Darksonn

tokio/src/sync/notify.rs
```diff
@@ -558,6 +577,23 @@ impl Notify {
...
+    /// Notifies the last waiting task.
+    ///
+    /// This function behaves similar to `notify_one`. The only difference is that it wakes
+    /// the most recently added waiter instead of the oldest waiter.
+    ///
+    /// Check the [`notify_one()`] documentation for more info and
+    /// examples.
+    ///
+    /// [`notify_one()`]: Notify::notify_one
+    pub fn notify_one_last(&self) {
```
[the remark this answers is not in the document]
pfreixes: 🤦 my bad.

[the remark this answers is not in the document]
pfreixes: done!

Darksonn (approved):
Thank you.

PR #6534, opened 2024-05-04
Using sharded locks instead of a global lock for Timers

wathenjiang:
## Motivation

As part of addressing https://github.com/tokio-rs/tokio/issues/6504, this PR attempts to implement a shard approach to improve `timeout` and `sleep` performance in specific scenarios.

When high concurrency, a large number of `timeout`s or `sleep`s are registered to Timer, the global lock contention is very severe. By sharding the lock, this performance issue can be significantly reduced.

## Solution

- Thread local storage random number generation is used to achieve sharding.
- Because the wheel data structure is a multi-layered linked list, the current shard size is consistent with the number of workers. I hope that the creation of a runtime will not require more time, and the serial lock acquisition in the park method will not take too much time.

### DIFF
--- a/tokio/src/runtime/scheduler/multi_thread_alt/worker.rs
+++ b/tokio/src/runtime/scheduler/multi_thread_alt/worker.rs
@@ -1312,7 +1312,7 @@ impl Context {
         &self.handle.shared
     }
 
-    #[allow(dead_code)]
+    #[cfg_attr(not(feature = "time"), allow(dead_code))]
     pub(crate) fn get_worker_index(&self) -> usize {
         self.index
     }
--- a/tokio/src/runtime/scheduler/multi_thread_alt/worker.rs
+++ b/tokio/src/runtime/scheduler/multi_thread_alt/worker.rs
@@ -1311,7 +1311,8 @@ impl Context {
     fn shared(&self) -> &Shared {
         &self.handle.shared
     }
-
+    
+    #[allow(dead_code)]
     pub(crate) fn get_worker_index(&self) -> usize {
         self.index
     }
--- a/tokio/src/runtime/scheduler/multi_thread_alt/worker.rs
+++ b/tokio/src/runtime/scheduler/multi_thread_alt/worker.rs
@@ -1311,6 +1311,10 @@ impl Context {
     fn shared(&self) -> &Shared {
         &self.handle.shared
     }
+
+    pub(crate) fn get_worker_index(&self) -> usize {
+        self.index
+    }
 }
 
 impl Core {
--- a/tokio/src/runtime/scheduler/multi_thread_alt/worker.rs
+++ b/tokio/src/runtime/scheduler/multi_thread_alt/worker.rs
@@ -1311,7 +1311,7 @@ impl Context {
     fn shared(&self) -> &Shared {
         &self.handle.shared
     }
-    
+
     #[allow(dead_code)]
     pub(crate) fn get_worker_index(&self) -> usize {
         self.index
--- a/tokio/src/runtime/time/entry.rs
+++ b/tokio/src/runtime/time/entry.rs
@@ -58,6 +58,7 @@ use crate::loom::cell::UnsafeCell;
 use crate::loom::sync::atomic::AtomicU64;
 use crate::loom::sync::atomic::Ordering;
 
+use crate::runtime::context;
 use crate::runtime::scheduler;
 use crate::sync::AtomicWaker;
 use crate::time::Instant;
@@ -498,13 +499,23 @@ impl TimerEntry {
     fn inner(&self) -> &TimerShared {
         let inner = unsafe { &*self.inner.get() };
         if inner.is_none() {
-            let shard_id =
-                super::rand::thread_rng_n(self.driver.driver().time().inner.get_shard_size());
+            let shard_id = self.get_shard_id();
             unsafe {
                 *self.inner.get() = Some(TimerShared::new(shard_id));
             }
         }
         return inner.as_ref().unwrap();
+    }
+
+    fn get_shard_id(&self) -> u32 {
+        let shard_size = self.driver.driver().time().inner.get_shard_size();
+        use scheduler::Context::MultiThread;
+        let id = context::with_scheduler(|ctx| match ctx {
+            Some(MultiThread(ctx)) => ctx.get_worker_index() as u32,
+            Some(_) => 1,
+            _ => super::rand::thread_rng_n(shard_size),
+        });
+        id % shard_size
     }
 
     pub(crate) fn deadline(&self) -> Instant {
--- a/tokio/src/runtime/time/entry.rs
+++ b/tokio/src/runtime/time/entry.rs
@@ -527,7 +527,7 @@ impl TimerEntry {
     cfg_not_rt! {
         fn get_shard_id(&self) -> u32 {
             let shard_size = self.driver.driver().time().inner.get_shard_size();
-            let id = crate::runtime::context::thread_rng_n(shard_size);
+            let id = context::thread_rng_n(shard_size);
             id % shard_size
         }
     }
--- a/tokio/src/runtime/time/entry.rs
+++ b/tokio/src/runtime/time/entry.rs
@@ -665,7 +665,7 @@ cfg_rt! {
             Some(scheduler::Context::MultiThread(ctx)) => ctx.get_worker_index() as u32,
             #[cfg(all(tokio_unstable, feature = "rt-multi-thread"))]
             Some(scheduler::Context::MultiThreadAlt(ctx)) => ctx.get_worker_index() as u32,
-            _ => context::thread_rng_n(shard_size),
+            None => context::thread_rng_n(shard_size),
         });
         id % shard_size
     }
--- a/tokio/src/runtime/time/entry.rs
+++ b/tokio/src/runtime/time/entry.rs
@@ -542,9 +542,8 @@ impl TimerEntry {
 
     /// Cancels and deregisters the timer. This operation is irreversible.
     pub(crate) fn cancel(self: Pin<&mut Self>) {
-        // Avoid calling the `clear_entry` method, because it has not been
-        // initialized or registered.
-        if !self.is_inner_init() || !self.inner().might_be_registered() {
+        // Avoid calling the `clear_entry` method, because it has not been initialized yet.
+        if !self.is_inner_init() {
             return;
         }
         // We need to perform an acq/rel fence with the driver thread, and the
--- a/tokio/src/runtime/time/entry.rs
+++ b/tokio/src/runtime/time/entry.rs
@@ -494,37 +494,13 @@ impl TimerEntry {
     fn inner(&self) -> &TimerShared {
         let inner = unsafe { &*self.inner.get() };
         if inner.is_none() {
-            let shard_id = self.get_shard_id();
+            let shard_size = self.driver.driver().time().inner.get_shard_size();
+            let shard_id = generate_shard_id(shard_size);
             unsafe {
                 *self.inner.get() = Some(TimerShared::new(shard_id));
             }
         }
         return inner.as_ref().unwrap();
-    }
-
-    // Gets the shard id. If current thread is a worker thread, we use its worker index as a shard id.
-    // Otherwise, we use a random number generator to obtain the shard id.
-    cfg_rt! {
-        fn get_shard_id(&self) -> u32 {
-            let shard_size = self.driver.driver().time().inner.get_shard_size();
-            let id = context::with_scheduler(|ctx| match ctx {
-                Some(scheduler::Context::CurrentThread(_ctx)) => 0,
-                #[cfg(feature = "rt-multi-thread")]
-                Some(scheduler::Context::MultiThread(ctx)) => ctx.get_worker_index() as u32,
-                #[cfg(all(tokio_unstable, feature = "rt-multi-thread"))]
-                Some(scheduler::Context::MultiThreadAlt(ctx)) => ctx.get_worker_index() as u32,
-                _ => context::thread_rng_n(shard_size),
-            });
-            id % shard_size
-        }
-    }
-
-    cfg_not_rt! {
-        fn get_shard_id(&self) -> u32 {
-            let shard_size = self.driver.driver().time().inner.get_shard_size();
-            let id = context::thread_rng_n(shard_size);
-            id % shard_size
-        }
     }
 
     pub(crate) fn deadline(&self) -> Instant {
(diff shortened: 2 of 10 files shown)


wathenjiang:
The following code is used for the benchmark:
```rust
use std::{
    future::pending,
    time::Duration,
};

fn main() {
    let hander = pprof_init();
        .enable_all()
        .worker_threads(8)
        .build()
        .unwrap();

    let _r = runtime.block_on(async {
        let mut handles = Vec::with_capacity(1024);
        for _ in 0..1024 {
                loop {
                    let h = timeout(Duration::from_millis(10), never_ready_job(1));
                    let _r = h.await;
                }
            }));
        }
        for handle in handles {
            handle.await.unwrap();
        }
    });
    hander.join().unwrap();
}

// a job never ready
async fn never_ready_job(n: u64) -> u64 {
    pending::<()>().await;
    n * 2
}

```

The flamegraph of the master branch:

image


The flamegraph of this PR:

image

The above shows a significant reduction in lock contention:

- `parking_lot::raw_mutex::RawMutex::lock_slow` has reduced from `6.03%` to `0.79%` in `poll`
- `parking_lot::raw_mutex::RawMutex::lock_slow` has reduced from `6.32%` to `1.24%` in `clear_entry`

tokio/src/runtime/time/mod.rs
```diff
@@ -107,34 +112,27 @@ struct Inner {
...
     ///
     /// Specifying the source of time is useful when testing.
-    pub(crate) fn new(park: IoStack, clock: &Clock) -> (Driver, Handle) {
+    pub(crate) fn new(park: IoStack, clock: &Clock, shards: u32) -> (Driver, Handle) {
         let time_source = TimeSource::new(clock);
 
+        let mut wheels = vec![];
+        for _ in 0..shards {
+            wheels.push(Mutex::new(wheel::Wheel::new()));
+        }
```
FrankReh, line 128: You may want to consider asserting that `shards` is nonzero and this construction
```Rust
        let wheels: Vec<_> = (0..shards)
            .map(|_| Mutex::new(wheel::Wheel::new()))
            .collect();
```

tokio/src/runtime/time/mod.rs
```diff
@@ -171,15 +169,24 @@ impl Driver {
...
-
         assert!(!handle.is_shutdown());
 
-        let next_wake = lock.wheel.next_expiration_time();
-        lock.next_wake =
-            next_wake.map(|t| NonZeroU64::new(t).unwrap_or_else(|| NonZeroU64::new(1).unwrap()));
+        // Finds out the min expiration time to park.
+        let mut next_wake: Option<u64> = None;
+        for id in 0..rt_handle.time().inner.get_shard_size() {
+            let lock = unsafe { rt_handle.time().inner.lock_sharded_wheel(id) };
```
FrankReh, line 177: Sorry, not sure why `unsafe` was used here. But consider this.
```rust
        let next_wake = (0..handle.inner.get_shard_size())
            .filter_map(|id| handle.inner.lock_sharded_wheel(id).next_expiration_time())
            .fold(None::<u64>, |acc, expiration_time| {
                Some(match acc {
                    Some(t) => t.min(expiration_time),
                    None => expiration_time,
                })
            });
```

tokio/src/runtime/time/mod.rs
wathenjiang: Great points.

tokio/src/runtime/time/mod.rs
wathenjiang: Thanks. The `unsafe` here is a mistake.

tokio/src/runtime/time/entry.rs
```diff
@@ -490,8 +498,10 @@ impl TimerEntry {
     fn inner(&self) -> &TimerShared {
         let inner = unsafe { &*self.inner.get() };
         if inner.is_none() {
+            let shard_id =
+                super::rand::thread_rng_n(self.driver.driver().time().inner.get_shard_size());
```
Darksonn, line 502: I'm wondering whether we could use the current worker id, when this is called from a runtime thread. What do you think?

tokio/src/runtime/time/mod.rs
```diff
@@ -406,5 +442,24 @@ impl fmt::Debug for Inner {
...
 
+mod rand {
+    use crate::util::rand::FastRand;
+    use std::cell::Cell;
+
+    // Used by `TimerEntry`.
+    pub(crate) fn thread_rng_n(n: u32) -> u32 {
+        thread_local! {
+            static THREAD_RNG: Cell<FastRand> = Cell::new(FastRand::new());
+        }
```
Darksonn, line 453: Please do not create new thread locals. Some platforms have small limits on the number of thread locals that each process can have.

There should already be a random number generator in the runtime context somewhere.

tokio/src/runtime/time/mod.rs
```diff
@@ -171,15 +169,24 @@ impl Driver {
...
-        let mut lock = handle.inner.state.lock();
-
         assert!(!handle.is_shutdown());
 
-        let next_wake = lock.wheel.next_expiration_time();
-        lock.next_wake =
-            next_wake.map(|t| NonZeroU64::new(t).unwrap_or_else(|| NonZeroU64::new(1).unwrap()));
+        // Finds out the min expiration time to park.
+        let mut next_wake: Option<u64> = None;
+        for id in 0..rt_handle.time().inner.get_shard_size() {
```
FrankReh, line 176: Just a question. What does it mean to the system if the park_internal came up with a new expiration time that wasn't the min? I don't know if other threads could be running when this function is being called, but if one can, couldn't it set a new smaller timer on a shard that the look had already just checked? But I would have a question about the earlier version too so this is not an important question. Tx.

tokio/src/runtime/time/entry.rs
wathenjiang: It's worth a try, this is the difference between `sleep`s and `task`s. The timer wheel is more like an mpsc, while the OwnedTasks I optimized before are mpmc. That is to say, when it comes to lock contention in sleep, the most severe one is insert, while clear is much smaller in lock contention. Therefore, making the create operation thread-local is worth trying for `sleep`s.

However, I still need to point out that it would be useful if our scenario involves creating sleep and timeouts in a worker thread environment. But when it needs to be created in a thread outside of the worker thread, the effect is minimal. If there are enough shards, then the performance consumed by locking is already small enough, such as the `0.79%` in the above benchmark.

tokio/src/runtime/time/mod.rs
wathenjiang: It's a good question.

When the park_internal method is executed, other worker threads may be running or sleeping on condvar. Our concern is that if a smaller expiration time timer is registered, the entire runtime can process it in a timely manner.

Therefore, when `Timer`s are registered, if necessary, we execute `unpark.unpark()` to wake up the worker thread. Please See https://github.com/tokio-rs/tokio/blob/cdf9d997dc79ac51fbe376edd04136251e73c403/tokio/src/runtime/time/mod.rs#L360


Considering this is a multi-threaded issue, both the current PR version and the master branch version have unnecessary wake-up issues. That is to say, even if the current minimum expiration time has changed, the worker thread still parks on the driver with the older and larger expiration time which is generated by the last check. In this case, the worker thread will be quickly unparked by `unpark.unpark()`.

This is  similar to [spurious wakeup](https://www.wikiwand.com/en/Spurious_wakeup), but the cost is not high.

wathenjiang: If there are some other running worker threads.
- If necessary, it will attempt to execute `unpark.unpark()`
- Alternatively, the worker thread that was originally responsible for the poll driver is not executed in the timeout way. If there is no event in the driver, it will immediately return from park_internal

tokio/src/runtime/time/entry.rs
wathenjiang: I added this logic, as expected, creating a sleep in the worker thread is almost lock free contention.

tokio/src/runtime/time/mod.rs
wathenjiang: I used the `context::thread_rng_n` instead. But this introduces unexpected additional `time` features. 

If we are not allowed to access 'context:: thread_rng'n' in the `time` feature, then our other option may be to use `std::thread::current().id().as_u64()`. However, the latter is unstable. Maybe the hash value of thread id can address this problem.

tokio/src/runtime/time/mod.rs
```diff
@@ -281,19 +295,17 @@ impl Handle {
...
-        lock.next_wake = lock
-            .wheel
+        let next_wake_up = lock
             .poll_at()
             .map(|t| NonZeroU64::new(t).unwrap_or_else(|| NonZeroU64::new(1).unwrap()));
-
         drop(lock);
 
         waker_list.wake_all();
+        next_wake_up
```
FrankReh, line 308: Now I'm confused that waker_list.wake_all is called potentially multiple times before self.inner.set_next_wake is called and not afterwards, but I will admit I haven't quite caught up my understanding of the implications in the multi threaded case.

tokio/src/runtime/time/entry.rs
FrankReh: > I added the thread-local logic, as expected, creating a sleep in the worker thread is almost lock free contention.

Assume the same would be said for creating a *timeout* in the worker thread? That it would be an almost lock free operation?

tokio/src/runtime/time/mod.rs
FrankReh: Thanks for pointing me to `reregister`. So any time a task creates a timer entry that happens to create a new earliest time for the timer wheel to fire, the task is rescheduled to be woken up immediately. This PR doesn't change that fact.

And this design favors helping systems that are busy. The busier the system, the more unlikely a new timer entry will be the earliest, as there would likely be other tasks already waiting for similar timeout events that had been created earlier. Only when the system isn't busy is it likely a new entry will often be the wheel's next entry and that's just when a little inefficiency doesn't make any difference.

tokio/src/runtime/time/entry.rs
wathenjiang: Yes, it is almost lock-free.

tokio/src/runtime/time/entry.rs
```diff
@@ -485,13 +494,39 @@ impl TimerEntry {
...
+                Some(scheduler::Context::CurrentThread(_ctx)) => 0,
+                #[cfg(feature = "rt-multi-thread")]
+                Some(scheduler::Context::MultiThread(ctx)) => ctx.get_worker_index() as u32,
+                #[cfg(all(tokio_unstable, feature = "rt-multi-thread"))]
+                Some(scheduler::Context::MultiThreadAlt(ctx)) => ctx.get_worker_index() as u32,
+                _ => context::thread_rng_n(shard_size),
+            });
+            id % shard_size
+        }
+    }
```
Darksonn, line 520: This sounds like a getter for a value that is already created, but actually it is the logic for figuring out which shard to use for a new timer. Can we rename this, and perhaps move it to a stand-alone function instead of a method on `TimerEntry`. Otherwise I am worried that someone will call it and expect it to return the same value as what the `TimerShared` is using.

Darksonn (commented):
Overall LGTM.

tokio/src/runtime/time/entry.rs
```diff
@@ -643,3 +654,25 @@ impl Drop for TimerEntry {
...
+// Otherwise, we use a random number generator to obtain the shard id.
+cfg_rt! {
+    fn generate_shard_id(shard_size: u32) -> u32 {
+        let id = context::with_scheduler(|ctx| match ctx {
+            Some(scheduler::Context::CurrentThread(_ctx)) => 0,
+            #[cfg(feature = "rt-multi-thread")]
+            Some(scheduler::Context::MultiThread(ctx)) => ctx.get_worker_index() as u32,
+            #[cfg(all(tokio_unstable, feature = "rt-multi-thread"))]
+            Some(scheduler::Context::MultiThreadAlt(ctx)) => ctx.get_worker_index() as u32,
+            _ => context::thread_rng_n(shard_size),
```
Darksonn, line 668:
```suggestion
            None => context::thread_rng_n(shard_size),
```

tokio/src/runtime/time/mod.rs
```diff
@@ -91,8 +93,11 @@ pub(crate) struct Driver {
 
 /// Timer state shared between `Driver`, `Handle`, and `Registration`.
 struct Inner {
-    // The state is split like this so `Handle` can access `is_shutdown` without locking the mutex
-    pub(super) state: Mutex<InnerState>,
+    /// The earliest time at which we promise to wake up without unparking.
+    next_wake: AtomicU64,
```
Darksonn, line 97: Perhaps it would result in clearer code to wrap this in an `AtomicOptionNonZeroU64` utility type? Or at least we should simplify `NonZeroU64::new(t).unwrap_or_else(|| NonZeroU64::new(1).unwrap())`.

wathenjiang: Yes, it shoud be simplified. I have added a new helper type exactly called `AtomicOptionNonZeroU64`.

Compared to `unwrap_or_else` and `unwarp`, I prefer `match` here.

tokio/src/runtime/time/mod.rs
```diff
@@ -25,9 +26,35 @@ use crate::time::error::Error;
...
+struct AtomicOptionNonZeroU64(AtomicU64);
+
+// A helper type to store the `next_wake`.
+impl AtomicOptionNonZeroU64 {
+    fn new(val: Option<u64>) -> Self {
+        Self(AtomicU64::new(Self::turn(val)))
+    }
+
+    fn store(&self, val: Option<u64>) {
+        self.0.store(Self::turn(val), Ordering::Relaxed);
```
Darksonn, line 42: This stores `Some(1)` if passed `Some(0)`. Can you add a comment that explains why?

Actually, it might be more clear to change this to `Option<NonZeroU64>` and move the call to `turn` to where you call this method.

wathenjiang: I add the helper function `next_wake_time` instead and some explains about it.

tokio/src/runtime/scheduler/multi_thread_alt/worker.rs
```diff
@@ -1311,6 +1311,11 @@ impl Context {
     fn shared(&self) -> &Shared {
         &self.handle.shared
     }
+
+    #[allow(dead_code)]
```
Darksonn, line 1315:
```suggestion
    #[cfg_attr(not(feature = "time"), allow(dead_code))]
```

Darksonn (approved):
Looks good to me. Thanks.

PR #6561, opened 2024-05-15
util: Use ptr::addr_of over pointer arithmetics in linked_list

udoprog:
Minor nit, since we no longer should need to use pointer arithmetics since the stabilization of `ptr::addr_of_mut!`.

The SAFETY documentation is a bit odd, since it previously didn't motivate why accessing the field through `UnsafeCell` was OK. However, ~~I'm not sure `UnsafeCell<T>` is even necessary any longer with `addr_of*` since we avoid intermediate references directly over `self`~~. Nevermind! Just needed a refresher.


Darksonn (approved):
Thanks. This type is used in intrusive linked lists, which is why we don't want references inside it.

PR #6563, opened 2024-05-15
time: remove the true_when field in TimerShared.

wathenjiang:
Since the `true_when` field in the `TimerShared` struct actually has no effect,  we should remove the `true_when` field.

It will be a win in memory type size, which reduces the size of `TimerShared` from 72 bytes to 64 bytes. It should be beneficial for memory cache.


Darksonn (approved):
Thanks.

PR #6564, opened 2024-05-15
docs: gate taskdump compiler_error behind not(doc)

rcoh:
## Motivation

Docs don't build on macOS because of the `compile_error!` warning intended to prevent customers from using task dumps on unsupported platforms.

## Solution

Add `not(doc)` so that rustdoc doesn't hit the compile failure


Darksonn (approved):
Thanks.

PR #6569, opened 2024-05-17
Improved the code with clippy 1.80

41Leahcim:
I improved the code with clippy 1.80 (nightly  2024-05-16). The code compiles, doesn't cause any warnings with clippy 1.77 and 1.80, and all tests succeed. I know you don't like it when people use a newer version of clippy, let alone the current nightly version, but I wanted to contribute and I don't know enough to contribute more.

## Motivation
I'm trying to improve the code of libraries I use and I use Tokio a lot.

## Solution
Replaced cmp::max(2, cmp::min(...)) with clamp, removed usize imports, replaced std::usize:: and std::u32:: with usize:: and u32:: respectively, moved parentheses to put the future into the blackbox instead of the unit value.

### DIFF
--- a/benches/time_timeout.rs
+++ b/benches/time_timeout.rs
@@ -40,7 +40,7 @@ fn do_timeout_test(c: &mut Criterion, workers: usize, name: &str) {
         b.iter_custom(|iters| {
             let start = Instant::now();
             runtime.block_on(async {
-                black_box(spawn_timeout_job(iters as usize, workers).await);
+                black_box(spawn_timeout_job(iters as usize, workers)).await;
             });
             start.elapsed()
         })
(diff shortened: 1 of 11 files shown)


Darksonn (commented):
Thanks. Can you please update our ci configuration?

https://github.com/tokio-rs/tokio/blob/35c6fd99243616db0a677c6686e37082c7c68a6f/.github/workflows/ci.yml#L20

41Leahcim:
Do you want me to update it to the latest nightly, beta or stable version?

Darksonn (approved):
Oh, I did not realize that 1.80 was not a stable release. In that case, let's keep it as-is.

PR #6574, opened 2024-05-20
metrics: add MetricAtomicU64 and use in metrics

rcoh:
## Motivation

As we move to stabilize metrics, we need a graceful way to support U64 on platforms without 64-bit atomics. 

**Note:** this is a breaking change on platforms without 32-bit atomics. Previously, they would have used a lock fallback provided by loom. Now, they simply do not have read access to these metrics.

## Solution

This commit adds MetricAtomicU64 which is empty on platforms that do not support 64-bit atomics. Write operations are no-ops and read operations are not defined. This also allows us to remove the loom gating on `cfg_metrics`. This will allow #6556 to avoid the need for a separate `cfg_metrics` gate.


Darksonn:
Is this intended to be merged before or after #6556?

rcoh:
Before — can't build the other cr without it

tokio/src/util/metric_atomics.rs
```diff
@@ -0,0 +1,47 @@
...
+
+        pub(crate) fn fetch_add(&self, value: u64, ordering: Ordering) -> u64 {
+            self.value.fetch_add(value, ordering)
+        }
+    }
+
+    cfg_no_64bit_metrics! {
+        pub(crate) fn store(&self, _val: u64, _ordering: Ordering) { }
+        // on platforms without 64-bit atomics, fetch-add returns unit
+        pub(crate) fn fetch_add(&self, _value: u64, _ordering: Ordering) {  }
```
Darksonn, line 44: Then I would probably just return unit from both. (And perhaps rename to `add`.)

Darksonn (commented):
Overall lgtm.

tokio/src/macros/cfg.rs
```diff
@@ -220,25 +220,45 @@ macro_rules! cfg_metrics {
         $(
             // For now, metrics is only disabled in loom tests.
             // When stabilized, it might have a dedicated feature flag.
-            #[cfg(all(tokio_unstable, not(loom)))]
+            #[cfg(tokio_unstable)]
```
mox692, line 223: Perhaps this comment is also out of date?

Darksonn (approved):
Looks good to me.

mox692 (approved):

PR #6581, opened 2024-05-22
docs: `stdin` is always `inherit` by default

### LINKED ISSUE #6577: Doc bug in `Command::stdin`?
**Version**
1.37.0

**Platform**
N/A / the documentation

**Description**

[The documentation for `Command::stdin`](https://docs.rs/tokio/latest/tokio/process/struct.Command.html#method.stdin) reads,

> Sets configuration for the child process’s standard input (stdin) handle.
>
> Defaults to [`inherit`](https://doc.rust-lang.org/nightly/std/process/struct.Stdio.html#method.inherit) when used with `spawn` or `status`, and defaults to [`piped`](https://doc.rust-lang.org/nightly/std/process/struct.Stdio.html#method.piped) when used with `output`.

N.b., that this documentation is essentially copy/pasted between `stdin`, `stdout`, and `stderr`, with references to `stdin` above replaced appropriately; I think this is a bug in that `stdin`'s docs were merely copy/pasted from `stdout`/`stderr` without enough thought; in particular, let's look at the second paragraph.

In the second paragraph: assume we call `output`, which captures the _output_ of the process: why should such a call presume `piped` as a default for `stdin`? (It would make sense for `stderr` and `stdout`. But `stdin`?)

If we look at [the source for `output`](https://github.com/tokio-rs/tokio/blob/tokio-1.37.0/tokio/src/process/mod.rs#L954-L963):

```rust
    pub fn output(&mut self) -> impl Future<Output = io::Result<Output>> {
        self.std.stdout(Stdio::piped());
        self.std.stderr(Stdio::piped());

        let child = self.spawn();

        async { child?.wait_with_output().await }
    }
}
```
… this seems to suggest that `stdin` is not piped merely by calling `Command::output`.

I think the behavior is fine, but the docs seem to not relay the correct behavior. This paragraph should probably just be truncated to "Defaults to `inherit`." for `stdin` only.

roy-work:
The code for `output` indicates that only sets `stdout` and `stderr` (as one would expect, IMO), yet the docs for `stdin` indicated that it too would be set. This seems like it was just a simple copy/paste typo, so correct `stdin` to note that it just defaults to `inherit`.

Fixes #6577.

## Motivation

See #6577; the docs for `stdin` seem to disagree with the code for `output`.

## Solution

Adjust the documentation to match the code.

### DIFF
--- a/tokio/src/process/mod.rs
+++ b/tokio/src/process/mod.rs
@@ -544,11 +544,9 @@ impl Command {
 
     /// Sets configuration for the child process's standard input (stdin) handle.
     ///
-    /// Defaults to [`inherit`] when used with `spawn` or `status`, and
-    /// defaults to [`piped`] when used with `output`.
+    /// Defaults to [`inherit`].
     ///
     /// [`inherit`]: std::process::Stdio::inherit
-    /// [`piped`]: std::process::Stdio::piped
     ///
     /// # Examples
     ///


mox692 (approved):
Thank you!

PR #6589, opened 2024-05-26
Add `SimplexStream` structure

wutchzone:
I think that it is unfortunate that the underlying structure backing the `DuplexStream` is not exposed. There are cases where a single direction will suffice, for example, when I have only a single writer and single reader, thus not requiring bidirectionality.

In this patch, I renamed the `Pipe` to the `SimplexStream` for coherence with the already existing `DuplexStream` and made it public

### DIFF
--- a/tokio/src/io/util/mem.rs
+++ b/tokio/src/io/util/mem.rs
@@ -1,6 +1,6 @@
 //! In-process memory IO types.
 
-use crate::io::{AsyncRead, AsyncWrite, ReadBuf};
+use crate::io::{split, AsyncRead, AsyncWrite, ReadBuf, ReadHalf, WriteHalf};
 use crate::loom::sync::Mutex;
 
 use bytes::{Buf, BytesMut};
@@ -47,15 +47,34 @@ use std::{
 #[derive(Debug)]
 #[cfg_attr(docsrs, doc(cfg(feature = "io-util")))]
 pub struct DuplexStream {
-    read: Arc<Mutex<Pipe>>,
-    write: Arc<Mutex<Pipe>>,
+    read: Arc<Mutex<SimplexStream>>,
+    write: Arc<Mutex<SimplexStream>>,
 }
 
-/// A unidirectional IO over a piece of memory.
+/// A unidirectional pipe to read and write bytes in memory.
 ///
-/// Data can be written to the pipe, and reading will return that data.
+/// It can be constructed by [`simplex`] function which will create a pair of
+/// reader and writer or by calling [`SimplexStream::new_unsplit`] that will
+/// create a handle for both reading and writing.
+///
+/// # Example
+///
+/// ```
+/// # async fn ex() -> std::io::Result<()> {
+/// # use tokio::io::{AsyncReadExt, AsyncWriteExt};
+/// let (mut receiver, mut sender) = tokio::io::simplex(64);
+///
+/// sender.write_all(b"ping").await?;
+///
+/// let mut buf = [0u8; 4];
+/// receiver.read_exact(&mut buf).await?;
+/// assert_eq!(&buf, b"ping");
+/// # Ok(())
+/// # }
+/// ```
 #[derive(Debug)]
-struct Pipe {
+#[cfg_attr(docsrs, doc(cfg(feature = "io-util")))]
+pub struct SimplexStream {
     /// The buffer storing the bytes written, also read from.
     ///
     /// Using a `BytesMut` because it has efficient `Buf` and `BufMut`
@@ -83,8 +102,8 @@ struct Pipe {
 /// written to a side before the write returns `Poll::Pending`.
 #[cfg_attr(docsrs, doc(cfg(feature = "io-util")))]
 pub fn duplex(max_buf_size: usize) -> (DuplexStream, DuplexStream) {
-    let one = Arc::new(Mutex::new(Pipe::new(max_buf_size)));
-    let two = Arc::new(Mutex::new(Pipe::new(max_buf_size)));
+    let one = Arc::new(Mutex::new(SimplexStream::new_unsplit(max_buf_size)));
+    let two = Arc::new(Mutex::new(SimplexStream::new_unsplit(max_buf_size)));
 
     (
         DuplexStream {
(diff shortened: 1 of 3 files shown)


Darksonn:
Adding a stream with one direction of communication seems reasonable, but I don't love the name `SimplexStream`. Maybe it's the best option, but can we come up with anything else? I guess there's the option of just keeping `Pipe` ...

wutchzone:
My reasoning for not using `Pipe` is that it is already [an overloaded term used for OS primitive](https://docs.rs/tokio/latest/tokio/net/unix/pipe/index.html), and someone could get confused by reusing it again. And since we have the `DuplexStream` I think™ that a better name is `SimplexStream` because these two structures are somewhat related.

But if you think it is negligible I will change it back to the `Pipe`.

robgjansen:
A `SimplexStream` would be very useful and something I'm missing! You can `split` a `TcpStream` into read and write halves, so for testing I'd like to replace those with `SimplexStream`s. I also wonder if we might want to be able to `split` or `into_split` a `DuplexStream` into two `SimplexStreams`s?

Darksonn:
Sorry, this PR has gotten lost. I'm happy to take it up again if you're still interested.

wutchzone:
I would be glad to continue working on this feature (and possibly extending it into other structures, as @ robgjansen mentioned) . However, I am still uncertain whether we agree to name it `SimplexStream` or keep the original `Pipe`.

Darksonn:
We can go with SimplexStream. Sorry for not replying earlier.

tokio/src/io/util/mem.rs
```diff
@@ -161,19 +181,24 @@ impl Drop for DuplexStream {
...
+#[cfg_attr(docsrs, doc(cfg(feature = "io-util")))]
+pub fn simplex(max_buf_size: usize) -> SimplexStream {
+    SimplexStream {
+        buffer: BytesMut::new(),
+        is_closed: false,
+        max_buf_size,
+        read_waker: None,
+        write_waker: None,
     }
+}
```
Darksonn, line 199: This provides a single object that is both the reader and writer, but in practice I think people will want those to be two different objects.

robgjansen: I agree with the previous comment. In my case I would like to use these stream objects as drop-in replacements for `TcpStream` and related types.

A `TcpStream` is specific to one "node", i.e., the client has a `TcpStream`, and the server has a `TcpStream`. For testing code that uses a `TcpStream`, we can replace with `DuplexStream`s

```rust
let (client_rw, server_rw) = tokio::io::duplex(65_536);
```

Here we have both the client-side and server-side objects, and both are Read+Write. Thus, code written to generically handle IO on a `TcpStream` will function identically when replaced with the `DuplexStream`.

Now in my use case, each `TcpStream` is split into a reader (`OwnedReadHalf`) and a writer (`OwnedWriteHalf`), 

```rust
let (read_half, write_half) = tcp_stream.into_split();
```

and my code operates generically on the Read and on the Write objects. To replicate that, I would need the `SimplexStream` to work similarly:

```rust
let (client_read_half, server_write_half) = tokio::io::simplex(65_536);
```

And then you could stack two of these together to get functionality similar to the split of a `TcpStream`

```rust
let (client_read_half, server_write_half) = tokio::io::simplex(65_536);
let (server_read_half, client_write_half) = tokio::io::simplex(65_536);
```

Then I could pass `client_read_half` / `client_write_half` to my client-side code, and `stream_read_half` / `stream_write_half ` to my server-side code.

There is a bit of a question about the API. Do we want `tokio::io::simplex` to return a `SimplexStream` that is Read+Write, and then support `SimplexStream::into_split` that returns a `SimplexStreamReadHalf` and `SimplexStreamWriteHalf`? Or should `tokio::io::simplex` directly return the tuple `(SimplexStreamReadHalf, SimplexStreamWriteHalf)` as in my examples above?

(Also, I'm not sure if it makes sense to mix DuplexStreams and SimplexStreams or not, it sorta depends on how the simplex API looks.)

wutchzone: IMO, we should not reinvent the wheel and try to reuse already existing APIs and facilities. It is true that for most of the use cases, we should rather return a pair of reader and writer instead of a single structure, that is a great point. Also, most of the channel-like structures already do this.
Furthermore, since `SimplexStream` already implements `AsyncRead` and `AsyncWrite`, we can use already existing [`split(…)`](https://docs.rs/tokio/latest/tokio/io/fn.split.html) to create reader and writer half (where possibly the split reader and writer half can be [`unsplit`](https://docs.rs/tokio/latest/tokio/io/struct.ReadHalf.html#method.unsplit)).

What do you think of it now?

wutchzone:
I am not sure whether it makes sense to add more tests to test the `SimplexStream` because `DuplexStream` (which already uses `SimplexStream`) has an extensive test suite.

Darksonn (commented):
I think returning an already split `SimplexStream` makes sense as an API.

tokio/src/io/util/mem.rs
```diff
@@ -83,8 +99,10 @@ struct Pipe {
 /// written to a side before the write returns `Poll::Pending`.
 #[cfg_attr(docsrs, doc(cfg(feature = "io-util")))]
 pub fn duplex(max_buf_size: usize) -> (DuplexStream, DuplexStream) {
-    let one = Arc::new(Mutex::new(Pipe::new(max_buf_size)));
-    let two = Arc::new(Mutex::new(Pipe::new(max_buf_size)));
+    let (reader, writer) = simplex(max_buf_size);
+    let one = Arc::new(Mutex::new(reader.unsplit(writer)));
+    let (reader, writer) = simplex(max_buf_size);
+    let two = Arc::new(Mutex::new(reader.unsplit(writer)));
```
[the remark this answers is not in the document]
wutchzone: Should I make the `new` method public?

tokio/src/io/util/mem.rs
```diff
@@ -161,19 +179,24 @@ impl Drop for DuplexStream {
...
-            read_waker: None,
-            write_waker: None,
-        }
-    }
+/// Creates unidirectional buffer that acts like a pipe.
+///
+/// The `max_buf_size` argument is the maximum amount of bytes that can be
+/// written to a buffer before the it returns `Poll::Pending`.
+#[cfg_attr(docsrs, doc(cfg(feature = "io-util")))]
+pub fn simplex(max_buf_size: usize) -> (ReadHalf<SimplexStream>, WriteHalf<SimplexStream>) {
```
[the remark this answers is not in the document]
wutchzone: Great point, added the example.

tokio/src/io/util/mem.rs
```diff
@@ -47,15 +47,31 @@ use std::{
...
+/// let (mut receiver, mut sender) = tokio::io::simplex(64);
+///
+/// sender.write_all(b"ping").await?;
+///
+/// let mut buf = [0u8; 4];
+/// receiver.read_exact(&mut buf).await?;
+/// assert_eq!(&buf, b"ping");
+///
+/// # Ok(())
+/// # }
```
Darksonn, line 70: Please drop the empty line.

Darksonn (commented):
We could make `new` public, but then I'd call it `new_unsplit` or similar.

tokio/src/io/util/mem.rs
wutchzone: Dropped.

tokio/src/io/util/mem.rs
```diff
@@ -47,15 +47,30 @@ use std::{
...
+    write: Arc<Mutex<SimplexStream>>,
 }
 
-/// A unidirectional IO over a piece of memory.
+/// A unidirectional pipe to read and write bytes in memory.
 ///
-/// Data can be written to the pipe, and reading will return that data.
+/// # Example
+///
+/// ```
```
Darksonn, line 58: I think it would be nice to include a link to the `simplex` function that acts as the primary constructor.

Currently it looks like this: https://deploy-preview-6589--tokio-rs.netlify.app/tokio/io/struct.simplexstream

Darksonn (approved):
Overall LGTM. Just one more minor request below.

tokio/src/io/util/mem.rs
wutchzone: I extended the documentation with links & rebased the PR.

tokio/src/io/util/mem.rs
```diff
@@ -47,15 +47,33 @@ use std::{
...
-    write: Arc<Mutex<Pipe>>,
+    read: Arc<Mutex<SimplexStream>>,
+    write: Arc<Mutex<SimplexStream>>,
 }
 
-/// A unidirectional IO over a piece of memory.
+/// A unidirectional pipe to read and write bytes in memory. It can be constructed
+/// by [`simplex`] function which will create a pair of reader and writer or by
+/// calling [`SimplexStream::new_unsplit`] that will create a handle for both
+/// reading and writing.
```
Darksonn, line 57: Everything on the first markdown line is visible in the module docs.

image

Please move everything after the first period into a new paragraph.
```suggestion
/// A unidirectional pipe to read and write bytes in memory.
///
/// It can be constructed by [`simplex`] function which will create a pair of
/// reader and writer or by calling [`SimplexStream::new_unsplit`] that will
/// create a handle for both reading and writing.
```

tokio/src/io/util/mem.rs
Darksonn: Thanks!

tokio/src/io/util/mem.rs
wutchzone: Ops, should be fixed.

Darksonn:
Thanks!

CGamesPlay:
I notice that dropping a `SImplexStream` doesn't call `close_write`, so readers will be left hanging forever. I think the `Drop` impl for `DuplexStream` should be ported to `SimplexStream` instead?

wutchzone:
I agree that it should not hang if one of the halves is dropped. However, I don't think that it will be as easy as adding `Drop` to `SimplexStream`, because `fn duplex(…) -> (DuplexStream, DuplexStream)`, but `fn simplex(…) -> (ReadHalf<SimplexStream>, WriteHalf<SimplexStream>)`, and dropping just the read half or write half won't trigger the `Drop` of `SimplexStream` and specialized `Drop` traits are tricky.

PR #6590, opened 2024-05-26
Fix misleading NamedPipeServer example

Tacklebox:
## Motivation

The previous NamedPipeServer doc example seemed to imply that the return type of `NamedPipeServer::connect()` was a `Future<Result<some_kind_of_client>>>`, however, `connect()` returns a `Future<Result<()>>>`. The following line of code reopening the pipe would shadow the newly connected pipe immediately, making the following spawned task pointless.

## Solution

Instead, now it binds the original named pipe to a new variable after a client connects, allowing for the subsequent task to be able use the connection.


Darksonn (approved):
Thank you.

PR #6593, opened 2024-05-28
macros: allow setting `unhandled_panic` behavior in tokio::{main, test}

### LINKED ISSUE #6527: Allow setting `unhandled_panic` behavior as option on `tokio::test`
I have several unit tests that run some handler code that is under test in a `tokio::spawn`ed task, and sends/receives bytes to/from that handler code from the main task. My AsyncRead + AsyncWrite mock will panic if it sees unexpected bytes, and if this happens in the background task the test will hang. I'd prefer the test to shut down in this scenario, and so I'm using the `unhandled_panic` option introduced by #4516.

`#[tokio::test(unhandled_panic = ShutdownRuntime)`

Currently I manually set up a tokio runtime for my tests that require this behavior.

name1e5s:
## Motivation

Allow users to set `unhandled_panic` behavior in those macros directly.

Close #6527.

## Solution

This PR parses `unhandled_panic` in `tokio-macros`.

### DIFF
--- a/tests-build/tests/fail/macros_invalid_input.stderr
+++ b/tests-build/tests/fail/macros_invalid_input.stderr
@@ -76,40 +76,46 @@ error: Failed to parse value of `crate` as path: "456"
 41 | #[tokio::test(crate = "456")]
    |                       ^^^^^
 
-error: second test attribute is supplied, consider removing or changing the order of your test attributes
-  --> tests/fail/macros_invalid_input.rs:45:1
+error: The `unhandled_panic` option requires the `current_thread` runtime flavor. Use `#[tokio::test(flavor = "current_thread")]`
+  --> tests/fail/macros_invalid_input.rs:44:58
    |
-45 | #[test]
+44 | #[tokio::test(flavor = "multi_thread", unhandled_panic = "shutdown_runtime")]
+   |                                                          ^^^^^^^^^^^^^^^^^^
+
+error: second test attribute is supplied, consider removing or changing the order of your test attributes
+  --> tests/fail/macros_invalid_input.rs:48:1
+   |
+48 | #[test]
    | ^^^^^^^
 
 error: second test attribute is supplied, consider removing or changing the order of your test attributes
-  --> tests/fail/macros_invalid_input.rs:49:1
+  --> tests/fail/macros_invalid_input.rs:52:1
    |
-49 | #[::core::prelude::v1::test]
+52 | #[::core::prelude::v1::test]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 error: second test attribute is supplied, consider removing or changing the order of your test attributes
-  --> tests/fail/macros_invalid_input.rs:53:1
+  --> tests/fail/macros_invalid_input.rs:56:1
    |
-53 | #[core::prelude::rust_2015::test]
+56 | #[core::prelude::rust_2015::test]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 error: second test attribute is supplied, consider removing or changing the order of your test attributes
-  --> tests/fail/macros_invalid_input.rs:57:1
+  --> tests/fail/macros_invalid_input.rs:60:1
    |
-57 | #[::std::prelude::rust_2018::test]
+60 | #[::std::prelude::rust_2018::test]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-error: second test attribute is supplied, consider removing or changing the order of your test attributes
-  --> tests/fail/macros_invalid_input.rs:61:1
-   |
-61 | #[std::prelude::rust_2021::test]
-   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 error: second test attribute is supplied, consider removing or changing the order of your test attributes
   --> tests/fail/macros_invalid_input.rs:64:1
    |
-64 | #[tokio::test]
+64 | #[std::prelude::rust_2021::test]
+   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+error: second test attribute is supplied, consider removing or changing the order of your test attributes
+  --> tests/fail/macros_invalid_input.rs:67:1
+   |
+67 | #[tokio::test]
    | ^^^^^^^^^^^^^^
    |
    = note: this error originates in the attribute macro `tokio::test` (in Nightly builds, run with -Z macro-backtrace for more info)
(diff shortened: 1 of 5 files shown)


tokio/tests/macros_test.rs
```diff
@@ -92,3 +93,27 @@ pub mod issue_5243 {
...
+    use tokio_test::assert_err;
+
+    #[tokio::test(unhandled_panic = "shutdown_runtime")]
+    #[should_panic]
+    async fn unhandled_panic_shutdown_runtime() {
+        let rt = tokio::spawn(async {
+            panic!("This panic should shutdown the runtime.");
+        })
+        .await;
+        assert_err!(rt);
```
mox692, line 108:
```suggestion
        let _ = tokio::spawn(async {
            panic!("This panic should shutdown the runtime.");
        })
        .await;
```

tests-build/tests/fail/macros_invalid_input.stderr
```diff
@@ -1,113 +1,113 @@
 error: the `async` keyword is missing from the function declaration
- --> $DIR/macros_invalid_input.rs:6:1
+ --> tests/fail/macros_invalid_input.rs:6:1
```
mox692, line 2: Probably not a big deal, but do you know where these changes came from?

tokio-macros/src/lib.rs
```diff
@@ -202,6 +202,44 @@ use proc_macro::TokenStream;
...
+/// ```ignore
+/// #[tokio::main(unhandled_panic = "shutdown_runtime")]
+/// async fn main() {
+///     let _ = tokio::spawn(async {
+///         panic!("This panic will shutdown the runtime.");
+///     }).await;
+/// }
+/// ```
+///
+/// Equivalent code not using `#[tokio::test]`
```
mox692, line 220:
```suggestion
/// Equivalent code not using `#[tokio::main]`
```

tokio-macros/src/lib.rs
```diff
@@ -202,6 +202,44 @@ use proc_macro::TokenStream;
...
 /// }
 /// ```
+///
+/// ### Configure unhandled panic behavior
+///
+/// Available options are `shutdown_runtime` and `ignore`. For more details, see
+/// [`Builder::unhandled_panic`].
+///
+/// ```ignore
+/// #[tokio::main(unhandled_panic = "shutdown_runtime")]
```
mox692, line 212: Since unhandled_panic on multi thread runtime is not implemented yet, so for now it would be better to specify the runtime flavor as well.

```suggestion
/// #[tokio::main(flavor = "current_thread", unhandled_panic = "shutdown_runtime")]
```

mox692 (commented):
Overall looks good to me, I left some comments regarding the docs.

tests-build/tests/fail/macros_invalid_input.stderr
name1e5s: It seems that the changes came from `trybuild` version update. I modified the error text like [this line](https://github.com/tokio-rs/tokio/pull/6593/files#diff-3778ff38bebb2786ca2b45f4704af832e54cd75369b7b9215eae010cddc58a09R363) and updated reference compiler output with `TRYBUILD=overwrite`, then every `$DIR` changed. Same changes appear in other recent PRs like #5643.

name1e5s: > Probably not a big deal, but do you know where these changes came from?

It seems that the changes came from `trybuild` version update. I modified the error text like [this line](https://github.com/tokio-rs/tokio/pull/6593/files#diff-3778ff38bebb2786ca2b45f4704af832e54cd75369b7b9215eae010cddc58a09R363) and updated reference compiler output with `TRYBUILD=overwrite`, then every `$DIR` changed. Same changes appear in other recent PRs like #5643.

tokio-macros/src/lib.rs
Darksonn: The macro should probably emit an error if you try to use it with the multi-thread runtime.

name1e5s: > The macro should probably emit an error if you try to use it with the multi-thread runtime.

 Thanks for your advice, set `unhandled_panic` behavior on `multi_thread` runtime results in a compile error now.

tokio-macros/src/lib.rs
```diff
@@ -423,6 +463,47 @@ pub fn main_rt(args: TokenStream, item: TokenStream) -> TokenStream {
...
 /// ```
+///
+/// ### Configure unhandled panic behavior
+///
+/// Available options are `shutdown_runtime` and `ignore`. For more details, see
+/// [`Builder::unhandled_panic`].
+///
+/// This option is only compatible with the `current_thread` runtime.
+///
+/// ```ignore
```
mox692, line 474: For consistency with the other examples on this page, can you use `no_run` instead of `ignore`?
(The `ignore` annotation makes the doc a bit scary.)
Screen Shot 2024-06-03 at 19 57 29

name1e5s: > For consistency with the other examples on this page, can you use `no_run` instead of `ignore`? (The `ignore` annotation makes the doc a bit scary.) Screen Shot 2024-06-03 at 19 57 29

I tried and found it break those CI jobs running `cargo test` without `RUSTFLAGS="--cfg tokio_unstable"`, such as [this link](https://github.com/tokio-rs/tokio/actions/runs/9385991492/job/25845581971?pr=6593). Since we can't use conditional compilation tricks on examples, I think we can only use `ignore` here to make CI happy.

mox692 (approved):
Looks good to me, thanks.

tokio-macros/src/lib.rs
Darksonn: I'm pretty sure that `#[cfg(tokio_unstable)]` works perfectly fine inside an example.

name1e5s: > I'm pretty sure that `#[cfg(tokio_unstable)]` works perfectly fine inside an example.

Yes, it works fine. I mean we can not use words like
```
```#[cfg(tokio_unstable)]
```
to disable an example entirely. But your comment reminds me that we can use `#` lines to make the example works under `cargo test` without `tokio_unstable `, I will try it.

name1e5s: > > I'm pretty sure that `#[cfg(tokio_unstable)]` works perfectly fine inside an example.
> 
> Yes, it works fine. I mean we can not use words like
> 
> ```
> ```#[cfg(tokio_unstable)]
> ```
> 
> to disable an example entirely. But your comment reminds me that we can use `#` lines to make the example works under `cargo test` without `tokio_unstable `, I will try it.

All tests looks fine now.

PR #6607, opened 2024-06-01
test: make `Spawn` forward `size_hint`

thedjinn:
## Motivation

I have a need to use `Spawn` in conjunction with a `Stream` impl that has a custom `size_hint`, and need to assert on this size hint.

Currently the size hint of the wrapped stream can be accessed via `Deref`, but not by accessing the `Stream` impl on `Spawn` itself (it instead returns the default value of `(0, None)`), which is confusing.

## Solution

To reduce confusion this pull request adds a `size_hint` override in the `Stream` impl for `Spawn`. This override forwards the `size_hint` invocation to the wrapped stream, allowing its size hint to be accessed directly instead of having it replaced with the default value. A test was added to validate the behavior.

### DIFF
--- a/tokio-test/src/task.rs
+++ b/tokio-test/src/task.rs
@@ -148,6 +148,10 @@ impl<T: Stream> Stream for Spawn<T> {
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         self.future.as_mut().poll_next(cx)
     }
+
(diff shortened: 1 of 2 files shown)


mox692:
Hmm, this could break existing code. But at the same time, I don't think there are many codes that depend on the current non-overrideable `Spawn` behavior, so it could be acceptable.

What do you think about this change? @Darksonn

Darksonn (approved):
Looks good to me. I'm not worried about breaking changes in this case.

PR #6612, opened 2024-06-04
time: fix big time panic issue

wathenjiang:
## Motivation

The https://github.com/tokio-rs/tokio/pull/5710 tried to fix the panic issue when the instance is very big.

However, the `instant_to_tick` can still calculate a `u64` which is bigger than `MAX_SAFE_MILLIS_DURATION`.

## Solution

Even if a `u128` successfully try_into a `u64`, we still check whether it is bigger than `MAX_SAFE_MILLIS_DURATION`. If it is greater, we will return `MAX_SAFE_MILLIS_DURATION`.

### DIFF
--- a/tokio/src/runtime/time/source.rs
+++ b/tokio/src/runtime/time/source.rs
@@ -38,6 +38,7 @@ impl TimeSource {
     }
 
     #[cfg(test)]
+    #[allow(dead_code)]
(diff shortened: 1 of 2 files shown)


Darksonn (approved):
Thanks.

PR #6616, opened 2024-06-04
Update FreeBSD CI environment

asomers:
Use the newly released FreeBSD 14.1.

## Motivation

FreeBSD 14.0 will soon be EoL.

## Solution

Use 14.1 in CI.


Darksonn (approved):
Thank you for always being on top of our FreeBSD setup. I really appreciate it.

PR #6619, opened 2024-06-06
metrics: stabilize num_alive_tasks

rcoh:
## Motivation
Stabilize ~`num_active_tasks`~ `num_alive_tasks` so it can be used outside of `--cfg tokio_unstable`

# Solution
1. stabilize `num_alive_tasks`
2. Rename internal fields to match `num_alive_tasks`
3. Split `rt_metrics` into `rt_metrics` and `rt_unstable_metrics`.

Refs: #6546


mox692 (approved):

tokio/tests/rt_metrics.rs
Darksonn, line 1: We should probably split this file into `rt_metrics.rs` and `rt_metrics_unstable.rs`.

But it doesn't have to happen in this PR.

mox692:
Since https://github.com/tokio-rs/tokio/pull/6114 has been merged, we need to resolve conflict here.
Also, perhaps we should stabilize `num_active_tasks` instead of `active_tasks_count`?

tokio/tests/rt_metrics.rs
mox692: Yeah that will make it easer to manage code.

mox692 (approved):
Looks good to me.

tokio/src/runtime/metrics/runtime.rs
```diff
@@ -47,6 +47,29 @@ impl RuntimeMetrics {
         self.handle.inner.num_workers()
     }
 
+    /// Returns the current number of active tasks in the runtime.
+    ///
+    /// This value increases and decreases over time as tasks are spawned and as they are completed or cancelled.
+    ///
+    /// To see the total number of spawned tasks, see `spawned_tasks_count`. Note that this API currently requires using `--cfg tokio_unstable`.
```
Darksonn, line 54: If this is being stabilized, then the documentation should no longer say that it's unstable.

rcoh:
Ah that comment was attempting to refer to spawned_task_count

Darksonn:
Okay. It's a bit confusing like this. You don't need to say that it's unstable in the docs for other methods.

Darksonn:
I'm sorry for responding slowly here.

I'm wondering whether the name should be something else. The word "active" is the opposite of "idle", but the count includes idle tasks that have not yet exited. How about a name like `alive_tasks_count`?

rcoh:
runnable? That has some issues too. I wonder if this would be helped with a
TaskMetrics struct that could give some docs to explain the different task
states & group the metrics.

Darksonn:
I ended up renaming it in #6667.

Darksonn:
I don't think runnable works either. That also implies that idle tasks shouldn't be counted, but they are.

rcoh:
Sorry for the delay on this. @Darksonn I've rebased. I noticed there was a mix between `alive_tasks_count` and `num_alive_tasks` — would you like me to unify them or leave as is?

Darksonn:
That makes sense. The publicly used name is `num_alive_tasks`, so let's rename the internal things called `alive_tasks_count` to `num_alive_tasks`.

Darksonn (approved):
Thanks.

PR #6621, opened 2024-06-06
Implement Clone for AbortHandle

aschweig:
This is a small change so that AbortHandle implements Clone.

## Motivation

The [the module level docs](https://docs.rs/tokio/latest/tokio/task/index.html#cancellation) indicate

> Each task can only have one [JoinHandle](https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html), but it can have more than one [AbortHandle](https://docs.rs/tokio/latest/tokio/task/struct.AbortHandle.html).

Cloning is convenient for `AbortHandle`s given there can be more than one.

## Solution

The proposed solution is similar to JoinHandle's [`abort_handle`](https://github.com/tokio-rs/tokio/blob/8e15c234c60cf8132c490ccf03dd31738cfeaca8/tokio/src/runtime/task/join.rs#L299) method:
```
impl Clone for AbortHandle {
    /// Returns a cloned `AbortHandle` that can be used to remotely abort this task.
    fn clone(&self) -> Self {
        self.raw.ref_inc();
        Self::new(self.raw)
    }
}
```

### DIFF
--- a/tokio/src/runtime/task/abort.rs
+++ b/tokio/src/runtime/task/abort.rs
@@ -102,3 +102,11 @@ impl Drop for AbortHandle {
         self.raw.drop_abort_handle();
     }
 }
+
+impl Clone for AbortHandle {
+    /// Returns a cloned `AbortHandle` that can be used to remotely abort this task.
+    fn clone(&self) -> Self {
+        self.raw.ref_inc();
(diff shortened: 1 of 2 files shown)


Darksonn (approved):
Thank you.

PR #6630, opened 2024-06-12
rt: modify the type bound of LinkedList when call into_guarded

wathenjiang:
To modify the restrictions on the generic L in LinkedList when call `into_guarded`.

I believe that it can help to be less restrictive on the type of `Handle` when turning a linked list into a guarded linked list.


Darksonn (approved):
Thank you.

PR #6637, opened 2024-06-14
Add sync::watch::Sender::same_channel()

uklotzde:
Align the API with `sync::broadcast::Sender` by adding the missing `fn same_channel()` method.

## Motivation

Now that the `Sender` implements `Clone` this is useful.

## Solution

Duplicate the `Receivers`s implementation.


mox692 (approved):
Thanks!

Issue #6639, opened 2024-06-14
Improvement for `#[tokio::main]`


Jedsek:
`Importing main from other modules or crates.` has been stabilized in [rust 1.79](https://releases.rs/docs/1.79.0/).

Maybe we could improve it a little for `#[tokio::main]`?

This code was not allowed before:

```rust
#[tokio::main]
async fn hello_world() {
    println!("Hello, world!");
}
```

**note:**
This is `backwards compatible` because the previous code not allowed.

**how to solved:**
We could add prefix before any function and then `use prefix_{name} as main;`.

taiki-e:
> This code was not allowed before:

No, arbitrary function names are supported.

Jedsek:
> No, arbitrary function names are supported.

Um...really?

Jedsek:
Close it because It deviates from the semantics I thought.

PR #6640, opened 2024-06-14
Improvement for #[tokio::main]

Jedsek:
related issue: #6639 

Support for making functions whose names are not `main` to be `[tokio::main]` ed.


Jedsek:
Oh, I found that `#[tokio::main]` could be applied to any function, and what I want is `entry-point`), so, close this pr

PR #6663, opened 2024-06-27
io: implement `AsyncSeek` for `Empty`

zaeleus:
## Motivation

While porting a test from `std:io` to `tokio::io`, I discovered `AsyncSeek` is not implemented for `tokio::io::Empty`. This was unexpected, as [`std::io::Seek` was implemented for `std::io::Empty` in Rust 1.51.0](https://doc.rust-lang.org/stable/std/io/struct.Empty.html#impl-Seek-for-Empty) (2021-03-25).

## Solution

This implementation mimics the behavior of [the `Seek` implementation for `std::io::Empty`](https://github.com/rust-lang/rust/blob/036b38ced36b0ed16579f95b4647ba7424f6b1bc/library/std/src/io/util.rs#L81-L94).

### DIFF
--- a/tokio/tests/io_util_empty.rs
+++ b/tokio/tests/io_util_empty.rs
@@ -1,5 +1,6 @@
 #![cfg(feature = "full")]
-use tokio::io::{AsyncBufReadExt, AsyncReadExt};
+use tokio::io::{AsyncBufReadExt, AsyncReadExt, AsyncSeekExt};
+use tokio_test::assert_ok;
 
 #[tokio::test]
 async fn empty_read_is_cooperative() {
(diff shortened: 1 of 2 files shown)


tokio/tests/io_util_empty.rs
```diff
@@ -30,3 +31,16 @@ async fn empty_buf_reads_are_cooperative() {
         _ = tokio::task::yield_now() => {}
     }
 }
+
+#[tokio::test]
+async fn empty_seek() {
```
wathenjiang, line 36: Can we add more tests like the `empty_seeks` in  https://github.com/rust-lang/rust/commit/f1cd17961ccaac4bfaeeab81969cf36c56eec4a5?

wathenjiang (approved):
LGTM. Thanks!

PR #6666, opened 2024-06-28
Add `IntoFuture` support to `timeout` & `timeout_at`

### LINKED ISSUE #6665: tokio::time::timeout should support std::future::IntoFuture
Tokio APIs don't currently support `std::future::IntoFuture`. In particular `tokio::time::timeout` & `timeout_at`.

## Example: timeout
```rust
struct IntoFooture;

impl std::future::IntoFuture for IntoFooture {
    type Output = ();
    type IntoFuture = std::future::Ready<()>;

    fn into_future(self) -> Self::IntoFuture {
        std::future::ready(())
    }
}

fn foo() -> IntoFooture {
    IntoFooture
}

foo().await; // works

_ = tokio::time::timeout(Duration::from_secs(4), foo()).await; // !compile: `IntoFooture` is not a future
```
So currently the future must be explicitly converted here to use `timeout`.
```rust
_ = tokio::time::timeout(Duration::from_secs(4), foo().into_future()).await; // works
```

## Solution: Support `IntoFuture`
If we updated the timeout method to:
```rust
pub fn timeout<F>(duration: Duration, future: F) -> Timeout<F::IntoFuture>
where
    F: std::future::IntoFuture,
{...}
```
The original example would work without the `.into_future()` explicit call.

## MSRV
Note: This requires bumping the msrv to 1.64 as this version stabilised IntoFuture.

alexheretic:
## Motivation
Resolves #6665 

## Solution
* Use `IntoFuture` instead of `Future` for `timeout` & `timeout_at`.
* ~Bump msrv to 1.64 as this version stabilised IntoFuture.~ _Current msrv of 1.70 is sufficient_


conradludgate:
With #6645 the MSRV should bump to 1.70 anyway

alexheretic:
I've merged in latest `master`, which now has a modern enough msrv to support `IntoFuture`.

Darksonn (approved):
Thanks.

PR #6671, opened 2024-07-02
Doc: JoinSet::try_join_next() returns None when no tasks are completed

teor2345:
## Motivation

The documentation of `JoinSet`'s `try_join_next()` and `try_join_next_with_id()` doesn't say what happens if there are no completed tasks.

`join_next()` waits, but `try_join_next()` appears to return `None` immediately.

## Solution

Update the docs for both methods.


Darksonn (approved):
Thank you.

PR #6679, opened 2024-07-09
6536 - fix yield_calls_park_before_scheduling_again test

evanrittenhouse:
closes https://github.com/tokio-rs/tokio/issues/6536.

## Motivation

Fixes the failing test outlined in https://github.com/tokio-rs/tokio/issues/6536

## Solution
Since this is just a Loom-related change, I felt it fine to simply increment the counter if the lock call fails. If a more robust change is desired, please let me know and I will happily do so. It feels like such a change would probably involve attaching the conditionally-compiled Loom counter to some struct, but I haven't looked too deeply into it.

### DIFF
--- a/tokio/src/runtime/park.rs
+++ b/tokio/src/runtime/park.rs
@@ -35,7 +35,7 @@ tokio_thread_local! {
 // Bit of a hack, but it is only for loom
 #[cfg(loom)]
 tokio_thread_local! {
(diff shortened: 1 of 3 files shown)


mox692 (approved):
I thought it might be a good idea to make it a separate counter that counts `Parker::park_timeout()` calls since this counter no longer counts the actual `park`.

But I'm not sure if this PR should do it. I think I'm fine with it.

PR #6681, opened 2024-07-10
tests: pdfd-reaper: fix tests when '-' is absent from kernel version

hexagonal-sun:
## Motivation

On my machine, any pidfd reaper test that calls `is_pidfd_available` fails as my kernel version string does not contain a '-'; the code expects there to be one.

## Solution

Fix the test so that is works regardless on whether the kernel string contains a '-'.


mox692 (approved):
Thanks.

PR #6685, opened 2024-07-14
sync: add Sender<T>::closed future

evanrittenhouse:
## Motivation

Adds a `closed` Future to `broadcast::Sender<T>`, similar to the `oneshot` or `mpsc` variants, which completes when all subscribed receivers have been dropped.

## Solution
This is a simple `poll_fn` which wraps a check around `shared.tail.rx_cnt`, returning `Ready` if the remaining count is 0.

### DIFF
--- a/tokio/src/sync/broadcast.rs
+++ b/tokio/src/sync/broadcast.rs
@@ -804,6 +804,47 @@ impl<T> Sender<T> {
         Arc::ptr_eq(&self.shared, &other.shared)
     }
 
+    /// A future which completes when the number of [Receiver]s subscribed to this `Sender` reaches
+    /// zero.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use futures::FutureExt;
+    /// use tokio::sync::broadcast;
+    ///
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let (tx, mut rx1) = broadcast::channel::<u32>(16);
+    ///     let mut rx2 = tx.subscribe();
+    ///
+    ///     tokio::spawn(async move {
+    ///         assert_eq!(rx1.recv().await.unwrap(), 10);
+    ///     });
+    ///
+    ///     let _ = tx.send(10);
+    ///     assert!(tx.closed().now_or_never().is_none());
+    ///
+    ///     let _ = tokio::spawn(async move {
+    ///         assert_eq!(rx2.recv().await.unwrap(), 10);
+    ///     }).await;
+    ///
+    ///     assert!(tx.closed().now_or_never().is_some());
+    /// }
+    /// ```
+    pub async fn closed(&self) {
+        std::future::poll_fn(|_| {
+            let tail = self.shared.tail.lock();
+
+            if tail.closed || tail.rx_cnt == 0 {
+                Poll::Ready(())
+            } else {
+                Poll::Pending
+            }
+        })
+        .await;
+    }
+
     fn close_channel(&self) {
         let mut tail = self.shared.tail.lock();
         tail.closed = true;
--- a/tokio/src/loom/std/mutex.rs
+++ b/tokio/src/loom/std/mutex.rs
@@ -1,7 +1,7 @@
 use std::sync::{self, MutexGuard, TryLockError};
 
 /// Adapter for `std::Mutex` that removes the poisoning aspects
-/// from its api.
+/// from its API.
 #[derive(Debug)]
 pub(crate) struct Mutex<T: ?Sized>(sync::Mutex<T>);
 


tokio/src/sync/broadcast.rs
```diff
@@ -804,6 +804,47 @@ impl<T> Sender<T> {
...
+            let tail = self.shared.tail.lock();
+
+            if tail.closed || tail.rx_cnt == 0 {
+                Poll::Ready(())
+            } else {
+                Poll::Pending
+            }
+        })
+        .await;
+    }
```
tglane, line 846: I think this future will only get polled once because you are not waking up the task again.

Consider this example:
```
#[tokio::main]
async fn main() {
    let (tx, mut rx1) = broadcast::channel(16);

    tokio::spawn(async move {
        let r = rx1.recv().await.unwrap();
        println!("Rx1 received: {}", r);
        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
    });

    tx.send(20).unwrap();

    println!("Waiting for subscribers to drop");
    tx.closed().await;
    println!("Subscribers dropped");
}
```
Here the call to `closed` will block because it will only be polled once right when you call it. But the subscriber is not dropped at that point, so the `poll_fn` will return `Pending`. But because the future is never be polled again, this will not resolve after that.

Darksonn: Indeed, this is not correct. When returning `Pending`, you must register for wakeups so that you are polled again when progress can be made.

Compare with the mpsc channel:

https://github.com/tokio-rs/tokio/blob/4825c444eb29b05c9c0b6f4a4c89475cb2556c65/tokio/src/sync/mpsc/chan.rs#L58-L59

evanrittenhouse: True, and I guess `poll_fn` is flagged as unstable in CI anyway. I'll find another way around this - thanks both.

Darksonn: Tokio has a `poll_fn` implementation you can use if you find that you need it.

evanrittenhouse:
Work has been nuts - haven't had much time for anything else. This is on my backlog though

Darksonn:
No rush from my side. Just wanted to check in.

evanrittenhouse:
Pushed up what should be closer to the proper implementation. I'd like to add a `task.is_woken()` check after each rx is dropped in the integration test, but for some reason, dropping rx2 doesn't wake the task (though it does make it `Ready`). So I'll dig more into that when I have a bit more time, and fix the various CI issues

tokio/src/sync/broadcast.rs
```diff
@@ -163,6 +164,7 @@ use std::task::{Context, Poll, Waker};
 /// [`broadcast`]: crate::sync::broadcast
 pub struct Sender<T> {
     shared: Arc<Shared<T>>,
+    notify_rx_closed: Arc<Notify>,
```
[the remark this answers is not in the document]
evanrittenhouse: Yeah, I've put this in draft as it's not ready for prime time yet. Just wanted to give an update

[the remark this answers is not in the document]
Darksonn: Thanks! Sounds good. There's no rush. :)

tokio/src/sync/broadcast.rs
```diff
@@ -804,6 +817,50 @@ impl<T> Sender<T> {
...
+            let tail = self.shared.tail.lock();
+
+            if tail.closed || tail.rx_cnt == 0 {
+                return Poll::Ready(());
+            }
+
+            drop(tail);
+            return Poll::Pending;
+        })
+        .await;
```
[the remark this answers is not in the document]
evanrittenhouse: If we now notify only once, when the last receiver drops, I think we can just `await` the Notified directly, right? I originally had this block because I notified on every `Rx` drop, so had to include an additional check to ensure that we had just dropped the last `Rx`. 

Relatedly, just for my own understanding - would this not work because the internal `Future` state machine is progressed to the point where we're already in the `poll_fn`, so future `closed.await`s wouldn't actually `await` the notified again?

tokio/src/sync/broadcast.rs
```diff
@@ -1349,6 +1411,8 @@ impl<T> Drop for Receiver<T> {
 
         drop(tail);
 
+        self.shared.notify_rx_drop.notify_waiters();
```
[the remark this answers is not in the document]
evanrittenhouse: Much cleaner, thanks!

tokio/src/sync/broadcast.rs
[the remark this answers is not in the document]
Darksonn: Yes, `Notified` doesn't have spurious wakeups in the scenario we are using it, so it should work. Well, except for the fact that we probably want it to return immediately if it's already closed when we call it. We can do it like this:
```rs
loop {
    let notified = self.shared.notify_rx_drop.notified();
    if self.is_closed() { return; }
    notified.await;
}
```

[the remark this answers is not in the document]
evanrittenhouse: Sure, done. For my own curiosity, what's the advantage of the loop above vs. a simple `self.shared.notify_rx_drop.notified().await`, if `Notified` doesn't suffer from spurious wakeups?

Note: I used `tail.closed` instead of `self.is_closed()` as I couldn't find an `is_closed` on `Sender`, which means an extra lock

tokio/tests/sync_broadcast.rs
```diff
@@ -640,3 +640,20 @@ fn send_in_waker_drop() {
...
+    let mut task = task::spawn(tx.closed());
+    assert_pending!(task.poll());
+
+    drop(rx);
+    assert!(!task.is_woken());
+    assert_pending!(task.poll());
+
+    drop(rx2);
+    assert!(task.is_woken());
+    assert_ready!(task.poll());
```
Darksonn, line 658: Could you add another test that calls `subscribe` between dropping the last receiver and polling `closed()`, to see what happens in that case.

tokio/src/sync/broadcast.rs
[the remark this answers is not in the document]
Darksonn: If you don't use a loop, then it changes the behavior of the test I requested in https://github.com/tokio-rs/tokio/pull/6685#discussion_r1770601675. Which behavior we want is up for discussion.

tokio/tests/sync_broadcast.rs
evanrittenhouse: Yeah, will do - sorry for the delay!

evanrittenhouse: Done - I simply re-open the channel if the `rx_cnt` is `0` when calling `new_receiver`. This allows us to re-create the `closed()` future, which I do in the new ` broadcast_sender_closed_with_extra_subscribe` test

mox692: Hmm, it seems that this change made another test failing.

Now even after dropping all senders, the channel doesn't seem to be marked as closed when resubscribe is called afterwards.

tokio/src/sync/broadcast.rs
```diff
@@ -820,6 +876,11 @@ fn new_receiver<T>(shared: Arc<Shared<T>>) -> Receiver<T> {
     assert!(tail.rx_cnt != MAX_RECEIVERS, "max receivers");
 
     tail.rx_cnt = tail.rx_cnt.checked_add(1).expect("overflow");
+    if tail.closed {
+        // Potentially need to re-open the channel, if a new receiver has been added between calls
+        // to poll()
+        tail.closed = false;
+    }
```
tglane, line 883: This causes the test `tests/sync_broadcast::resubscribe_to_closed_channel` to fail because it also reopens the channel when the sender got dropped before.
Maybe instead of checking for `tail.closed` to reopen the channel we could check for `tail.rx_cnt == 0` before the call to `tail.rx_cnt.checked_add(1)`? In that case the channel would only be reopened if it got closed due to no more receivers being "alive". 
I only ran the sync tests locally real quick and got no failures, so maybe its something to work with?

evanrittenhouse: Thanks for the pointers! I've been meaning to get back to this and study the code a bit more, but have had some stuff crop up in my personal life so haven't had time, sorry if I'm causing unnecessary work for anyone

Darksonn: No worries, there isn't a rush from our side. Things take the time they take.

evanrittenhouse:
Last failure seems to just be docs-related, so I think we just have to kick the job

Darksonn (approved):
Thanks.

PR #6686, opened 2024-07-14
macros: fix doc format issue

paolobarbolini:
## Motivation

Docs for `#[tokio::test]` contained a formatting issue for one of the examples

## Solution

Fixes it by using the same format as the other examples


Darksonn (approved):
Thanks.

PR #6731, opened 2024-07-27
process: stabilize `Command::process_group`

niklasf:
With the recent MSRV bump from 1.63 to 1.70, this can now be stabilized.

The underlying [`std::os::unix::process::CommandExt::process_group()`](https://doc.rust-lang.org/stable/std/os/unix/process/trait.CommandExt.html#tymethod.process_group) was added in Rust 1.64.0.

Also adjust the documentation to match std, explaining why setting process groups may be useful.


Darksonn (approved):
Thanks.

PR #6774, opened 2024-08-13
macros: improve documentation for `select!`

barafael:
Closes: #6572

The original PR seems stale. I believe this PR fixes the problem surfaced in the discussion of the original PR.


Darksonn (approved):
Thanks.

PR #6778, opened 2024-08-15
tests: handle ECONNREFUSED in uds_stream::epollhup

hawkw:
## Motivation

Currently, the test `uds_stream::epollhup` expects that a `UdsStream::connect` future to a Unix socket which is closed by the accept side to always fail with `io::ErrorKind::ConnectionReset`. On illumos, and potentially other systems, it instead fails with `io::ErrorKind::ConnectionRefused`.

This was discovered whilst adding an illumos CI job in PR #6769. See: https://github.com/tokio-rs/tokio/pull/6769#issuecomment-2284753794

## Solution

This commit changes the test to accept either `ConenctionReset` or `ConnectionRefused`. This way, we are more tolerant of different operating systems which may decide to return slightly different errnos here. Both ECONNREFUSED and ECONNRESET seem reasonable to expect in this situation, although arguably, ECONNREFUSED is actually more correct: the acceptor did not accept the connection at all, which seems like "refusing" it to me...

This commit was cherry-picked from PR #6769.

### DIFF
--- a/tokio/tests/uds_stream.rs
+++ b/tokio/tests/uds_stream.rs
@@ -406,6 +406,17 @@ async fn epollhup() -> io::Result<()> {
     drop(listener);
 
     let err = connect.await.unwrap_err();
-    assert_eq!(err.kind(), io::ErrorKind::ConnectionReset);
+    let errno = err.kind();
+    assert!(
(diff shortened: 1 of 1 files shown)


Darksonn (approved):
Thanks.

PR #6780, opened 2024-08-15
rfc: propose an FFI API

Noah-Kennedy:
This RFC covers the process aspects of this and largely defers discussion of the technical aspects, preferring to allow those to be incrementally addressed over time.


Noah-Kennedy:
CC @Darksonn @davidbarsky

carllerche:
My initial thoughts that I posted in discord:

> My gut reaction is I don't want to pull in a FFI interface, even if unstable. It should be able to be maintained by a third party. If there is significant adoption and momentum behind a single FFI interface, we can consider upstreaming at that point. If there are technical reasons for Tokio to provide its own FFI interface, then that should be discussed as additions to Tokio. I expect there is not a single "FFI" strategy

PR #6789, opened 2024-08-17
Rb/clippy fixes

barafael:
A bunch of selected warnings produced by clippy. I organized the commits by lint group, so if one lint is not favored, we can hopefully just skip that commit.


tokio/src/sync/rwlock.rs
```diff
@@ -1094,7 +1094,7 @@ impl<T> From<T> for RwLock<T> {
     }
 }
 
-impl<T: ?Sized> Default for RwLock<T>
+impl<T> Default for RwLock<T>
```
barafael, line 1097: Default requires Sized, so the ?Sized makes no sense

PR #6790, opened 2024-08-18
fix explanation of 'zero bytes read' case

rosscanning:
As per this discussion:
https://users.rust-lang.org/t/will-tokio-ever-pass-a-readbuf-with-no-capacity-to-poll-read/116124/2
there are two possible reasons why zero bytes were added to the `ReadBuf`, but the docs only discuss one reason.


mox692 (approved):
Looks good to me, thank you!

PR #6808, opened 2024-08-30
rt: add LocalRuntime

Noah-Kennedy:
This change adds LocalRuntime, a new unstable runtime type which cannot be transferred across thread boundaries and supports spawn_local when called from the thread which owns the runtime.

The initial set of docs for this are iffy. Documentation is absent right now at the module level, with the docs for the LocalRuntime struct itself being somewhat duplicative of those for the `Runtime` type. This can probably be addressed later as stabilization nears.

This API has a few interesting implementation details:
- because it was considered beneficial to reuse the same Handle as the normal runtime, it is possible to call spawn_local from a runtime context while on a different thread from the one which drives the runtime and owns it. This forces us to check the thread ID before attempting a local spawn.
- An empty LocalOptions struct is passed into the build_local method in order to build the runtime. This will eventually have stuff in it like hooks.

Relates to #6739.

### DIFF
--- a/tokio/src/runtime/builder.rs
+++ b/tokio/src/runtime/builder.rs
@@ -3,11 +3,16 @@
 use crate::runtime::handle::Handle;
 #[cfg(tokio_unstable)]
 use crate::runtime::TaskMeta;
-use crate::runtime::{blocking, driver, Callback, HistogramBuilder, Runtime, TaskCallback};
+use crate::runtime::{
+    blocking, driver, Callback, HistogramBuilder, LocalOptions, LocalRuntime, Runtime, TaskCallback,
+};
 use crate::util::rand::{RngSeed, RngSeedGenerator};
 
+use crate::runtime::blocking::BlockingPool;
+use crate::runtime::scheduler::CurrentThread;
 use std::fmt;
 use std::io;
+use std::thread::ThreadId;
 use std::time::Duration;
 
 /// Builds Tokio Runtime with custom configuration values.
@@ -800,6 +805,29 @@ impl Builder {
         }
     }
 
+    /// Creates the configured `LocalRuntime`.
+    ///
+    /// The returned `LocalRuntime` instance is ready to spawn tasks.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use tokio::runtime::Builder;
+    ///
+    /// let rt  = Builder::new_current_thread().build_local(&mut Default::default()).unwrap();
+    ///
+    /// rt.block_on(async {
+    ///     println!("Hello from the Tokio runtime");
+    /// });
+    /// ```
+    #[allow(unused_variables)]
+    pub fn build_local(&mut self, options: &mut LocalOptions) -> io::Result<LocalRuntime> {
+        match &self.kind {
+            Kind::CurrentThread => self.build_current_thread_local_runtime(),
+            _ => panic!("Only current_thread is supported when building a local runtime"),
+        }
+    }
+
     fn get_cfg(&self, workers: usize) -> driver::Cfg {
         driver::Cfg {
             enable_pause_time: match self.kind {
@@ -1191,8 +1219,39 @@ impl Builder {
     }
 
     fn build_current_thread_runtime(&mut self) -> io::Result<Runtime> {
-        use crate::runtime::scheduler::{self, CurrentThread};
-        use crate::runtime::{runtime::Scheduler, Config};
+        use crate::runtime::runtime::Scheduler;
+
+        let (scheduler, handle, blocking_pool) =
+            self.build_current_thread_runtime_components(None)?;
+
+        Ok(Runtime::from_parts(
+            Scheduler::CurrentThread(scheduler),
+            handle,
+            blocking_pool,
+        ))
+    }
+
+    fn build_current_thread_local_runtime(&mut self) -> io::Result<LocalRuntime> {
+        use crate::runtime::local_runtime::LocalRuntimeScheduler;
+
+        let tid = std::thread::current().id();
+
+        let (scheduler, handle, blocking_pool) =
+            self.build_current_thread_runtime_components(Some(tid))?;
+
+        Ok(LocalRuntime::from_parts(
+            LocalRuntimeScheduler::CurrentThread(scheduler),
+            handle,
+            blocking_pool,
+        ))
+    }
+
+    fn build_current_thread_runtime_components(
+        &mut self,
+        local_tid: Option<ThreadId>,
+    ) -> io::Result<(CurrentThread, Handle, BlockingPool)> {
+        use crate::runtime::scheduler;
+        use crate::runtime::Config;
 
         let (driver, driver_handle) = driver::Driver::new(self.get_cfg(1))?;
 
@@ -1227,17 +1286,14 @@ impl Builder {
                 seed_generator: seed_generator_1,
                 metrics_poll_count_histogram: self.metrics_poll_count_histogram_builder(),
             },
+            local_tid,
         );
 
         let handle = Handle {
             inner: scheduler::Handle::CurrentThread(handle),
         };
 
-        Ok(Runtime::from_parts(
-            Scheduler::CurrentThread(scheduler),
-            handle,
-            blocking_pool,
-        ))
+        Ok((scheduler, handle, blocking_pool))
     }
 
     fn metrics_poll_count_histogram_builder(&self) -> Option<HistogramBuilder> {
--- a/tokio/src/runtime/builder.rs
+++ b/tokio/src/runtime/builder.rs
@@ -1,11 +1,9 @@
 #![cfg_attr(loom, allow(unused_imports))]
 
 use crate::runtime::handle::Handle;
+use crate::runtime::{blocking, driver, Callback, HistogramBuilder, Runtime, TaskCallback};
 #[cfg(tokio_unstable)]
-use crate::runtime::TaskMeta;
-use crate::runtime::{
-    blocking, driver, Callback, HistogramBuilder, LocalOptions, LocalRuntime, Runtime, TaskCallback,
-};
+use crate::runtime::{LocalOptions, LocalRuntime, TaskMeta};
 use crate::util::rand::{RngSeed, RngSeedGenerator};
 
 use crate::runtime::blocking::BlockingPool;
@@ -821,6 +819,8 @@ impl Builder {
     /// });
     /// ```
     #[allow(unused_variables)]
+    #[cfg(tokio_unstable)]
+    #[cfg_attr(docsrs, doc(cfg(tokio_unstable)))]
     pub fn build_local(&mut self, options: &mut LocalOptions) -> io::Result<LocalRuntime> {
         match &self.kind {
             Kind::CurrentThread => self.build_current_thread_local_runtime(),
@@ -1231,6 +1231,7 @@ impl Builder {
         ))
     }
 
+    #[cfg(tokio_unstable)]
     fn build_current_thread_local_runtime(&mut self) -> io::Result<LocalRuntime> {
         use crate::runtime::local_runtime::LocalRuntimeScheduler;
 
--- a/tokio/src/runtime/builder.rs
+++ b/tokio/src/runtime/builder.rs
@@ -818,7 +818,7 @@ impl Builder {
     ///     println!("Hello from the Tokio runtime");
     /// });
     /// ```
-    #[allow(unused_variables)]
+    #[allow(unused_variables, unreachable_patterns)]
     #[cfg(tokio_unstable)]
     #[cfg_attr(docsrs, doc(cfg(tokio_unstable)))]
     pub fn build_local(&mut self, options: &mut LocalOptions) -> io::Result<LocalRuntime> {
--- a/tokio/src/runtime/builder.rs
+++ b/tokio/src/runtime/builder.rs
@@ -807,6 +807,12 @@ impl Builder {
     ///
     /// The returned `LocalRuntime` instance is ready to spawn tasks.
     ///
+    /// # Panics
+    /// This will panic if `current_thread` is not the selected runtime flavor.
+    /// All other runtime flavors are unsupported by [`LocalRuntime`].
+    ///
+    /// [`LocalRuntime`]: [crate::runtime::LocalRuntime]
+    ///
     /// # Examples
     ///
     /// ```
--- a/tokio/src/runtime/builder.rs
+++ b/tokio/src/runtime/builder.rs
@@ -821,7 +821,7 @@ impl Builder {
     #[allow(unused_variables, unreachable_patterns)]
     #[cfg(tokio_unstable)]
     #[cfg_attr(docsrs, doc(cfg(tokio_unstable)))]
-    pub fn build_local(&mut self, options: &mut LocalOptions) -> io::Result<LocalRuntime> {
+    pub fn build_local(&mut self, options: &LocalOptions) -> io::Result<LocalRuntime> {
         match &self.kind {
             Kind::CurrentThread => self.build_current_thread_local_runtime(),
             _ => panic!("Only current_thread is supported when building a local runtime"),
--- a/tokio/src/runtime/handle.rs
+++ b/tokio/src/runtime/handle.rs
@@ -250,8 +250,8 @@ impl Handle {
     /// # Panics
     ///
     /// This function panics if the provided future panics, if called within an
-    /// asynchronous execution context, or if a timer future is executed on a
-    /// runtime that has been shut down.
+    /// asynchronous execution context, or if a timer future is executed on a runtime that has been
+    /// shut down.
     ///
     /// # Examples
     ///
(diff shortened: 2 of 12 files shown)


taikulawo:
Does the `#[tokio::main]` and `#[tokio::test]` support start as LocalRuntime?

Noah-Kennedy:
> Does the `#[tokio::main]` and `#[tokio::test]` support start as LocalRuntime?

No, not in this PR.

I scoped this change to adding the initial implementation of LocalRuntime and its associated set of core APIs.

I'd like to make these APIs available before we start looking at how to support `LocalRuntime` in the macros, and stabilize the runtime independently before we stabilize its usage semantics in our macros.

tokio/src/runtime/builder.rs
```diff
@@ -800,6 +803,31 @@ impl Builder {
...
+    /// let rt  = Builder::new_current_thread().build_local(&mut Default::default()).unwrap();
+    ///
+    /// rt.block_on(async {
+    ///     println!("Hello from the Tokio runtime");
+    /// });
+    /// ```
+    #[allow(unused_variables, unreachable_patterns)]
+    #[cfg(tokio_unstable)]
+    #[cfg_attr(docsrs, doc(cfg(tokio_unstable)))]
+    pub fn build_local(&mut self, options: &mut LocalOptions) -> io::Result<LocalRuntime> {
```
Darksonn, line 824: I think it's weird to take a `&mut LocalOptions` here. Why not `LocalOptions` by-value?

tokio/src/runtime/local_runtime/runtime.rs
```diff
@@ -0,0 +1,375 @@
...
+    {
+        // safety: spawn_local can only be called from `LocalRuntime`, which this is
+        unsafe {
+            if cfg!(debug_assertions) && std::mem::size_of::<F>() > BOX_FUTURE_THRESHOLD {
+                self.handle.spawn_local_named(Box::pin(future), None)
+            } else {
+                self.handle.spawn_local_named(future, None)
+            }
+        }
+    }
```
Darksonn, line 163: Hmm. These methods are identical. Perhaps just have a `spawn` with the signature of `spawn_local`?

tokio/src/runtime/scheduler/current_thread/mod.rs
```diff
@@ -48,6 +49,9 @@ pub(crate) struct Handle {
 
     /// User-supplied hooks to invoke for things
     pub(crate) task_hooks: TaskHooks,
+
+    /// If this is a `LocalRuntime`, flags the owning thread ID.
+    pub(crate) local_tid: Option<ThreadId>,
```
Darksonn, line 54: Don't we have our own thread id implementation?

tokio/src/runtime/task/list.rs
```diff
@@ -102,6 +102,26 @@ impl<S: 'static> OwnedTasks<S> {
         (join, notified)
     }
 
+    /// Bind a task that isn't safe to transfer across thread boundaries.
+    ///
+    /// # Safety
+    /// Only use this in `LocalRuntime` where the task cannot move
+    pub(crate) unsafe fn bind_local<T>(
```
Darksonn, line 109: Hmm. Is it too tricky to use `LocalOwnedTasks` here?

tokio/src/runtime/local_runtime/runtime.rs
```diff
@@ -0,0 +1,375 @@
...
+        F::Output: Send + 'static,
+    {
+        if cfg!(debug_assertions) && std::mem::size_of::<F>() > BOX_FUTURE_THRESHOLD {
+            self.handle.spawn_named(Box::pin(future), None)
+        } else {
+            self.handle.spawn_named(future, None)
+        }
+    }
+
+    /// Spawns a task which isn't `!Send + Sync` on the runtime.
```
yotamofek, line 148: I think you mean
```suggestion
    /// Spawns a task which isn't `Send + Sync` on the runtime.
```
?

altanozlu:
I was testing what is required for relaxing 'static requirement on LocalSet and I saw this PR & related issue.
I wonder if is it sound to enable creating tasks with lifetime other than 'static. 
Code below compiles and test works on master branch, if it make *sense* can this also be applied to LocalRuntime.
Since other callers to the function already uses 'static it shouldn't change anything for them.
```
#[tokio::test(flavor = "current_thread")]
async fn test_relaxed_life_time_for_local_set() {
    let mut value = 10;
    {
        LocalSet::new()
            .run_until(async {
                task::spawn_local(async {
                    value += 1;
                    println!("{value}");
                })
                .await
                .unwrap();
            })
            .await;
    }
}
```

here is diff:

```diff
diff --git a/tokio/src/runtime/task/list.rs b/tokio/src/runtime/task/list.rs
index 988d4228..041ced34 100644
--- a/tokio/src/runtime/task/list.rs
+++ b/tokio/src/runtime/task/list.rs
@@ -233,7 +233,7 @@ impl<S: 'static> LocalOwnedTasks<S> {
         }
     }
 
-    pub(crate) fn bind<T>(
+    pub(crate) fn bind<'a, T>(
         &self,
         task: T,
         scheduler: S,
@@ -241,8 +241,8 @@ impl<S: 'static> LocalOwnedTasks<S> {
     ) -> (JoinHandle<T::Output>, Option<Notified<S>>)
     where
         S: Schedule,
-        T: Future + 'static,
-        T::Output: 'static,
+        T: Future + 'a,
+        T::Output: 'a,
     {
         let (task, notified, join) = super::new_task(task, scheduler, id);
 
diff --git a/tokio/src/runtime/task/mod.rs b/tokio/src/runtime/task/mod.rs
index 6164bef9..c21de130 100644
--- a/tokio/src/runtime/task/mod.rs
+++ b/tokio/src/runtime/task/mod.rs
@@ -292,15 +292,15 @@ cfg_rt! {
     /// created. The first task reference is usually put into an `OwnedTasks`
     /// immediately. The Notified is sent to the scheduler as an ordinary
     /// notification.
-    fn new_task<T, S>(
+    fn new_task<'a, T, S>(
         task: T,
         scheduler: S,
         id: Id,
     ) -> (Task<S>, Notified<S>, JoinHandle<T::Output>)
     where
         S: Schedule,
-        T: Future + 'static,
-        T::Output: 'static,
+        T: Future + 'a,
+        T::Output: 'a,
     {
         let raw = RawTask::new::<T, S>(task, scheduler, id);
         let task = Task {
diff --git a/tokio/src/task/local.rs b/tokio/src/task/local.rs
index 5600f08e..f7b8e936 100644
--- a/tokio/src/task/local.rs
+++ b/tokio/src/task/local.rs
@@ -362,10 +362,10 @@ cfg_rt! {
     /// [`LocalSet`]: struct@crate::task::LocalSet
     /// [`tokio::spawn`]: fn@crate::task::spawn
     #[track_caller]
-    pub fn spawn_local<F>(future: F) -> JoinHandle<F::Output>
+    pub fn spawn_local<'a, F>(future: F) -> JoinHandle<F::Output>
     where
-        F: Future + 'static,
-        F::Output: 'static,
+        F: Future + 'a,
+        F::Output: 'a,
     {
         if cfg!(debug_assertions) && std::mem::size_of::<F>() > BOX_FUTURE_THRESHOLD {
             spawn_local_inner(Box::pin(future), None)
@@ -376,9 +376,9 @@ cfg_rt! {
 
 
     #[track_caller]
-    pub(super) fn spawn_local_inner<F>(future: F, name: Option<&str>) -> JoinHandle<F::Output>
-    where F: Future + 'static,
-          F::Output: 'static
+    pub(super) fn spawn_local_inner<'a, F>(future: F, name: Option<&str>) -> JoinHandle<F::Output>
+    where F: Future + 'a,
+          F::Output: 'a
     {
         match CURRENT.with(|LocalData { ctx, .. }| ctx.get()) {
             None => panic!("`spawn_local` called from outside of a `task::LocalSet`"),
@@ -949,10 +949,10 @@ impl Drop for LocalSet {
 
 impl Context {
     #[track_caller]
-    fn spawn<F>(&self, future: F, name: Option<&str>) -> JoinHandle<F::Output>
+    fn spawn<'a, F>(&'a self, future: F, name: Option<&str>) -> JoinHandle<F::Output>
     where
-        F: Future + 'static,
-        F::Output: 'static,
+        F: Future + 'a,
+        F::Output: 'a,
     {
         let id = crate::runtime::task::Id::next();
         let future = crate::util::trace::task(future, "local", name, id.as_u64());
diff --git a/tokio/tests/task_local_set.rs b/tokio/tests/task_local_set.rs
index ac46291a..6b5df42b 100644
--- a/tokio/tests/task_local_set.rs
+++ b/tokio/tests/task_local_set.rs
@@ -637,6 +637,23 @@ fn store_local_set_in_thread_local_with_runtime() {
     });
 }
 
+#[tokio::test(flavor = "current_thread")]
+async fn test_relaxed_life_time_for_local_set() {
+    let mut value = 10;
+    {
+        LocalSet::new()
+            .run_until(async {
+                task::spawn_local(async {
+                    value += 1;
+                    println!("{value}");
+                })
+                .await
+                .unwrap();
+            })
+            .await;
+    }
+}
+
 #[cfg(tokio_unstable)]
 mod unstable {
     use tokio::runtime::UnhandledPanic;

```

taikulawo:
I guess tokio use some unsafe to ignore lifetime after spawn call , but doesn't means it's sound.

Darksonn:
> I wonder if is it sound to enable creating tasks with lifetime other than 'static.

It is not sound. Whether or not the future is `Send` is irrelevant.

tokio/src/runtime/builder.rs
Noah-Kennedy: I decided to match the semantics of the builder itself, which uses a mutable reference to build.

I lean towards taking a reference of some sort rather than passing by-value, but I don't think it needs to be a mutable one necessarily.

Is there an advantage you can see in passing the options by-value instead of by reference?

tokio/src/runtime/scheduler/current_thread/mod.rs
Noah-Kennedy: We do, but our `ThreadId` type is meant more for uniquely identifying runtime threads than for identifying arbitrary threads. Storage and recording of an ID is managed by the `Context`, which is not workable when I am specifically trying to check and see if an arbitrary handle can be used to spawn things locally.

tokio/src/runtime/task/list.rs
Noah-Kennedy: I figured I'd be conservative for now and just map what we do in current_thread.

We can optimize this later.

tokio/src/runtime/builder.rs
Darksonn: With a mutable reference, the caller has to write `&mut` to pass their argument. This isn't an issue for the self parameter as it just uses the dot operator.

tokio/src/runtime/handle.rs
```diff
@@ -248,8 +248,8 @@ impl Handle {
     /// # Panics
     ///
     /// This function panics if the provided future panics, if called within an
-    /// asynchronous execution context, or if a timer future is executed on a
-    /// runtime that has been shut down.
+    /// asynchronous execution context, or if a timer future is executed on a runtime that has been
+    /// shut down.
```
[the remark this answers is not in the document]
Noah-Kennedy: Yup, I think my IDE tried to be "helpful" here.

tokio/src/runtime/local_runtime/options.rs
```diff
@@ -0,0 +1,9 @@
+/// `LocalRuntime`-only config options
+///
+/// Currently, there are no such options, but in the future, things like `!Send + !Sync` hooks may
+/// be added.
+#[derive(Default, Debug)]
+#[non_exhaustive]
+pub struct LocalOptions {
+    // todo add local hooks at a later point
+}
```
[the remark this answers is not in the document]
Noah-Kennedy: Oh shoot, good catch, that was sorta the whole point of this...

tokio/src/runtime/scheduler/current_thread/mod.rs
```diff
@@ -459,6 +465,33 @@ impl Handle {
         handle
     }
 
+    /// Spawn a task which isn't safe to send across thread boundaries onto the runtime.
+    ///
+    /// # Safety
+    pub(crate) unsafe fn spawn_local<F>(
```
ThinkChaos, line 471: Hi, just reading this out of interest :)

The safety section is empty here. Looks unintentional.

tokio/src/runtime/builder.rs
Noah-Kennedy: Done.

tokio/src/runtime/scheduler/current_thread/mod.rs
Noah-Kennedy: Thanks for the catch!

tokio/src/runtime/local_runtime/runtime.rs
```diff
@@ -0,0 +1,362 @@
...
+        unsafe {
+            if cfg!(debug_assertions) && std::mem::size_of::<F>() > BOX_FUTURE_THRESHOLD {
+                self.handle.spawn_local_named(Box::pin(future), None)
+            } else {
+                self.handle.spawn_local_named(future, None)
+            }
+        }
+    }
+
+    /// Runs the provided function on an executor dedicated to blocking operations.
```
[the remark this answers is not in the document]
Noah-Kennedy: I copied the wording from the non-local runtime. I've amended it though.

tokio/src/runtime/local_runtime/runtime.rs
```diff
@@ -0,0 +1,362 @@
...
+use std::future::Future;
+use std::marker::PhantomData;
+use std::time::Duration;
+
+/// A local Tokio runtime.
+///
+/// This runtime is identical to a `current_thread` [runtime], save for not being `!Send + !Sync`,
+/// and supporting `spawn_local`.
+///
+/// For more general information on how to use runtimes, see the [module] docs.
```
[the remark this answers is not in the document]
Noah-Kennedy: I have updated the documentation

tokio/src/task/local.rs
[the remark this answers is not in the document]
Noah-Kennedy: Done

tokio/src/runtime/local_runtime/runtime.rs
```diff
@@ -0,0 +1,376 @@
...
+
+    /// Enters the runtime context.
+    ///
+    /// This allows you to construct types that must have an executor
+    /// available on creation such as [`Sleep`] or [`TcpStream`]. It will
+    /// also allow you to call methods such as [`tokio::spawn`].
+    ///
+    /// [`Sleep`]: struct@crate::time::Sleep
+    /// [`TcpStream`]: struct@crate::net::TcpStream
+    /// [`tokio::spawn`]: fn@crate::spawn
```
Darksonn, line 257: Please explain whether it lets me call `spawn_local`.

tokio/src/runtime/local_runtime/runtime.rs
```diff
@@ -0,0 +1,376 @@
...
+        self.handle.enter()
+    }
+
+    /// Shuts down the runtime, waiting for at most `duration` for all spawned
+    /// work to stop.
+    ///
+    /// See the [struct level documentation](LocalRuntime#shutdown) for more details.
+    ///
+    /// # Examples
+    ///
```
Darksonn, line 295: It's probably worth pointing out that only `spawn_blocking` tasks can get left behind in this case. Ditto for `shutdown_background`.

tokio/src/runtime/local_runtime/runtime.rs
```diff
@@ -0,0 +1,376 @@
...
+
+/// A local Tokio runtime.
+///
+/// This runtime is capable of driving tasks which are not `Send + Sync` without the use of a
+/// `LocalSet`, and thus supports `spawn_local` without the need for a `LocalSet` context.
+///
+/// This runtime cannot be moved between threads or driven from different threads.
+///
+/// This runtime is incompatible with `LocalSet`. You should not attempt to drive a `LocalSet` within a
+/// `LocalRuntime`.
```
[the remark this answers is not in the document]
Noah-Kennedy: LocalRuntime will intercept `spawn_local` calls within the localset context, and get the events.

I didn't want to specify the exact behavior here, in case we make changes such that this panics or similar.

tokio/tests/rt_local.rs
```diff
@@ -0,0 +1,100 @@
...
+
+        tx.send(handle).unwrap();
+    });
+
+    let handle = rx.recv().unwrap();
+
+    let _guard = handle.enter();
+
+    spawn_local(async {});
+}
```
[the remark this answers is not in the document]
Noah-Kennedy: Yes, that works.

See the above test case, `test_spawn_local_from_guard`, which tests this.

tokio/src/runtime/handle.rs
```diff
@@ -345,6 +345,31 @@ impl Handle {
...
 
+    #[track_caller]
+    #[allow(dead_code)]
+    pub(crate) unsafe fn spawn_local_named<F>(
+        &self,
+        future: F,
+        _name: Option<&str>,
+    ) -> JoinHandle<F::Output>
+    where
+        F: Future + 'static,
```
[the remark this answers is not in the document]
Noah-Kennedy: Noted, will update when that happens if this lands after #6881.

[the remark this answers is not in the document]
hds: #6881 has been merged now, let me know if you have any issues with it.

Darksonn (approved):
LGTM assuming hds's comment is fixed.

tokio/src/runtime/handle.rs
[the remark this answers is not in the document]
Noah-Kennedy: I've pulled in its changes, had to derive Copy and Clone for the SpawnMeta, but otherwise not too bad.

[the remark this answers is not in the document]
hds: That sounds fine! Only didn't have it before because it's crate-internal and hadn't been needed.

PR #6818, opened 2024-09-03
Tokio-util: Enable the full feature when compiled for the playground

sarsko:
Like https://github.com/tokio-rs/tokio/pull/1960, but for tokio-util. Makes tokio-util usable in the Playground.

See https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=14e34737f9747022fcfaff8428e4eb53 for what happens currently.


Darksonn (approved):
Thank you.

PR #6852, opened 2024-09-17
doc: Fix a symbol issue

vxzyfx:
## Motivation

There is a punctuation error.

## Solution

Change “socket hijacking” to "socket hijacking".

### DIFF
--- a/tokio/src/net/tcp/socket.rs
+++ b/tokio/src/net/tcp/socket.rs
@@ -63,7 +63,7 @@ cfg_net! {
     ///     // previous one.
     ///     //
     ///     // On Windows, this allows rebinding sockets which are actively in use,
-    ///     // which allows “socket hijacking”, so we explicitly don't set it here.
+    ///     // which allows "socket hijacking", so we explicitly don't set it here.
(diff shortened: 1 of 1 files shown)


jofas:
> punctuation error

Just a nit, but the currently used quotation marks are known as [smart quotes](https://en.wikipedia.org/wiki/Quotation_marks_in_English#Smart_quotes). Given that the comment is written in English prose, I wouldn't say they are erroneously used from a typographical perspective (albeit it's rare to see them actually typed out, given that they are non-trivial to type on a QWERTY keyboard). So the question of whether one should use smart quotes or non-smart quotes here boils down to a matter of taste&mdash;rather than right or wrong&mdash;IMO.

vxzyfx:
Thank you for your explanation.

jofas:
My comment wasn't meant to be a rejection of your PR, I was just trying to contribute to the discussion. Whether your PR is an enhancement to the Tokio docs or not is not my decision to make. If you closed your PR just because of my comment, please reopen it and let the maintainers decide whether they want to merge this or not.

Darksonn (approved):
:woman_shrugging:

Issue #6865, opened 2024-09-24
`mpsc::Receiver::blocking_recv_many`: Blocking variant of `recv_many`


barafael:
## Problem
I'm working on a system where many messages are processed at an FFI boundary, which requires a blocking function call.

## Suggested Solution
I would like to receive more than one message if available from a blocking context. Like `Receiver::recv_many`, but in a thread or task spawned by `spawn_blocking`.

## Alternatives
So far, I've been offloading the blocking work via `spawn_blocking` and awaiting the handle. This works but is not ideal.

I'd be happy to give this a try.

PR #6867, opened 2024-09-24
sync: add mpsc::Receiver::blocking_recv_many

barafael:
Fixes: #6865

## Motivation

I would like to use an API like `recv_many`, but from a blocking context.

## Solution

Use crate::future::block_on in a similar fashion to how blocking_recv does it to call the async `recv_many` function.

## What's missing

Tests and docs. I'll complete them once it's clear if this entire change makes sense. Do you think referring to `blocking_recv` and `recv_many` from the docs is fine or should there be more specific docs?


Darksonn (approved):
Thanks.

PR #6870, opened 2024-09-26
sync, coop: apply cooperative scheduling to sync::broadcast::Receiver

### LINKED ISSUE #6855: make `broadcast::Receiver` cooperative
Currently, `broadcast::Receiver` does not participate in the [coop budget](https://docs.rs/tokio/latest/tokio/task/#cooperative-scheduling), so if there are a lot of messages available, a task will just keep receiving them and runtime may not be able to schedule other tasks.

So, it would make sense to have `broadcast::Receiver` participate in the coop budget as well.

tglane:
## Motivation
Similar to #6846 the function `sync::broadcast::Receiver::recv` utilize cooperative scheduling to prevent that a task could block the runtime. 

## Solution
I added the usage of `runtime::coop::Coop` that was added in #6846 to the function so that takes part in the coop budget.

Closes #6855.


Darksonn (approved):
Thanks, this is great!

PR #6894, opened 2024-10-10
process: add periodic polling in `Reaper`

ruihe774:
## Motivation

`Reaper` may oversleep giving overwritten SIGCHLD.

## Solution

Add periodic polling in `Reaper`.

*Please look at my comment in code for a description.*


ruihe774:
Sorry but giving `tokio::signal::unix::signal()` sends to all receivers, there is no such problem.

PR #6897, opened 2024-10-10
metrics: add H2 Histogram option to improve histogram granularity

rcoh:
## Motivation

Closes: #6896 

## Solution

This PR introduces a configurable new histogram for the poll-time histogram based on [H2 Histograms](https://iop.systems/blog/h2-histogram/). As part of this, three APIs are deprecated:
- `metrics_poll_count_histogram_scale`
- `metrics_poll_count_histogram_resolution`
- `metrics_poll_count_histogram_buckets`

Since I suspect these are commonly used, I thought it was worth a little added complexity to provide a graceful path to removing them.

They are replaced with `metrics_poll_count_histogram_configuration` that accepts a histogram configuration object. The intention of this change is to provide more knobs when configuring histograms (as customers will want with the H2 histogram).

The default histogram remains the linear histogram with 10 buckets. The default H2 histogram provides a 25% error bound from 100ns to 60s of poll time and uses about 1KB of storage. This is configurable by customers as well.

The `h2_histogram` code is fairly complex compared to the code for the other histogram modes, so I isolated it into its own module. I brought back proptest to test invariants across a wide range of possible parameters–since there isn't any actual branching in this code, we don't get any value from making them fuzz tests (other than that they are a little more annoying to run).


tokio/Cargo.toml
```diff
@@ -96,7 +96,7 @@ mio = { version = "1.0.1", optional = true, default-features = false }
 parking_lot = { version = "0.12.0", optional = true }
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
-socket2 = { version = "0.5.5", optional = true, features = [ "all" ] }
```
carllerche, line 99: What happened here (in this file)?

carllerche (approved):
I'm fine w/ this. Thanks.

tokio/src/runtime/metrics/histogram.rs
```diff
@@ -112,52 +224,62 @@ impl HistogramBatch {
...
+    }
+
+    pub(crate) fn legacy_mut(&mut self, f: impl Fn(&mut LegacyBuilder)) {
+        if let Some(legacy) = &mut self.legacy {
+            f(legacy)
+        } else {
+            let mut legacy = LegacyBuilder::default();
+            f(&mut legacy);
+            self.legacy = Some(legacy)
         }
```
hds, line 249: `Option` has a function that does handles the None case for you:

```suggestion
        let legacy = self.legacy.get_or_insert_with(|| LegacyBuilder::default());
        f(legacy);
```

tokio/src/runtime/metrics/histogram.rs
```diff
@@ -112,52 +224,62 @@ impl HistogramBatch {
...
+                match legacy.scale {
+                    HistogramScale::Linear => HistogramType::Linear(LinearHistogram {
+                        num_buckets: legacy.num_buckets,
+                        bucket_width: resolution,
+                    }),
+                    HistogramScale::Log => HistogramType::LogLegacy(LegacyLogHistogram {
+                        num_buckets: legacy.num_buckets,
+                        first_bucket_width: resolution,
+                    }),
+                }
```
hds, line 271: Could these match statements be combined? We don't seem to use resolution after the second one. I think perhaps we wouldn't need the `mut` on `resolution` after that either.

```suggestion
                match legacy.scale {
                    HistogramScale::Linear => HistogramType::Linear(LinearHistogram {
                        num_buckets: legacy.num_buckets,
                        bucket_width: resolution,
                    }),
                    HistogramScale::Log => HistogramType::LogLegacy(LegacyLogHistogram {
                        num_buckets: legacy.num_buckets,
                        first_bucket_width: resolution.next_power_of_two(),
                    }),
                }
```

hds (approved):
I have a couple of suggestions on code style and some formatting nits, but this look good to me.

There is one place I mentioned where I think some more description on what the examples are doing would help people understand the configuration options.

tokio/src/runtime/metrics/histogram/h2_histogram.rs
```diff
@@ -0,0 +1,479 @@
...
+    }
+}
+
+/// Configuration for a [`LogHistogram`]
+///
+/// The log-scaled histogram implements an H2 histogram where the first bucket covers
+/// the range from 0 to [`LogHistogramBuilder::min_value`] and the final bucket covers
+/// [`LogHistogramBuilder::max_value`] to infinity. The precision is bounded to the specified
+/// [`LogHistogramBuilder::max_error`]. Specifically, the precision is the next smallest value
+/// of `2^-p` such that it is smaller than the requested precision.
```
hds, line 124: This sentence has 2 `precision`s in it, should one of them be `maximum error`? Otherwise perhaps the first `precision` could be `actual precision` to contract `requested precision`?

tokio/src/runtime/builder.rs
```diff
@@ -1142,14 +1149,84 @@ impl Builder {
...
+        ///             .max_error(0.1)
+        ///             .max_buckets(1024)
+        ///             .expect("configuration uses 488 buckets")
+        ///         )
+        ///     )
+        ///     .build()
+        ///     .unwrap();
+        /// ```
+        ///
+        ///
```
hds, line 1225: Remove extra line.

```suggestion
        ///
```

hds (commented):
Couple more small things.

PR #6899, opened 2024-10-11
stabilize worker_total_busy_duration

Owen-CH-Leung:
## Motivation

Stabilize worker_total_busy_duration so it can be used outside of --cfg tokio_unstable

## Solution

Move the API impl out of tokio_unstable flag

Ref: #6546


tokio/src/runtime/metrics/histogram.rs
```diff
@@ -35,31 +35,34 @@ pub(crate) struct HistogramBatch {
...
+/// logarithmic scale.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[non_exhaustive]
+#[allow(unreachable_pub)]
+pub enum HistogramScale {
+    /// Linear bucket scale
+    Linear,
+
+    /// Logarithmic bucket scale
+    #[allow(dead_code)]
```
rcoh, line 48: is this change required for the mean time stabilization or just a coincidental change?

tokio/src/runtime/metrics/runtime.rs
```diff
@@ -96,6 +99,54 @@ impl RuntimeMetrics {
...
+        /// ```
+        pub fn worker_total_busy_duration(&self, worker: usize) -> Duration {
+        let nanos = self
+            .handle
+            .inner
+            .worker_metrics(worker)
+            .busy_duration_total
+            .load(Relaxed);
+        Duration::from_nanos(nanos)
+        }
```
rcoh, line 147: formatting issues in here—autoformat doesn't run because of the parent macro

tokio/src/runtime/scheduler/mod.rs
```diff
@@ -9,6 +9,8 @@ cfg_rt! {
     pub(crate) use inject::Inject;
 
     use crate::runtime::TaskHooks;
+
+    use crate::runtime::{WorkerMetrics};
```
rcoh, line 13:
```suggestion
    use crate::runtime::WorkerMetrics;
```

tokio/src/runtime/scheduler/mod.rs
```diff
@@ -193,10 +195,15 @@ cfg_rt! {
...
+
+        #[allow(dead_code)]
+        pub(crate) fn worker_metrics(&self, worker: usize) -> &WorkerMetrics {
+            match_flavor!(self, Handle(handle) => handle.worker_metrics(worker))
+        }
     }
 
     cfg_unstable_metrics! {
-        use crate::runtime::{SchedulerMetrics, WorkerMetrics};
+        use crate::runtime::{SchedulerMetrics};
```
rcoh, line 206:
```suggestion
        use crate::runtime::SchedulerMetrics;
```

tokio/src/runtime/scheduler/multi_thread/handle/metrics.rs
```diff
@@ -1,7 +1,8 @@
 use super::Handle;
+use crate::runtime::WorkerMetrics;
 
 cfg_unstable_metrics! {
-    use crate::runtime::{SchedulerMetrics, WorkerMetrics};
+    use crate::runtime::{SchedulerMetrics};
```
rcoh, line 5:
```suggestion
    use crate::runtime::SchedulerMetrics;
```

Owen-CH-Leung: Thanks & fixed.

tokio/src/runtime/scheduler/mod.rs
Owen-CH-Leung: Thanks & fixed.

tokio/src/runtime/scheduler/mod.rs
Owen-CH-Leung: Thanks & fixed.

tokio/src/runtime/metrics/histogram.rs
Owen-CH-Leung: I believe this change is required.

At master we have implemented another `struct WorkerMetrics` at `tokio/src/runtime/metrics/mock.rs` which didn't use `struct Histogram` defined at `tokio/src/runtime/metrics/histogram.rs`. 

In order to stabilize the API, I have to removed this mock WorkerMetrics and instead point to the "real" `WorkerMetrics` at `tokio/src/runtime/metrics/worker.rs`

The "real" `WorkerMetrics` has a field `poll_count_histogram` which is `Option<Histogram>`, and thus will attempt to parse `tokio/src/runtime/metrics/histogram.rs`. From there, you can see Histogram and HistogramBuilder both refers to `HistogramScale` which is hidden inside `tokio_unstable`. Clearly this wouldn't compile.

hds: I need to read through this PR in more detail so bare with me, but we shouldn't be making this stable and public if it isn't actually needed to to stabilize `worker_total_busy_duration` (and I'm mostly sure that it isn't).

In general, all the `#[allow(dead_code)]` that are required for this chnage give me the impression that we're exposing something that is only really used in `tokio_unstable` and so we should find a way to expose it only unstable `tokio_unstable`.

hds (changes requested):
Please bare with me as I haven't finished the review, but my first impression is that this PR is making more things public than is strictly necessary for the stabilization of `worker_total_busy_duration` and we should try to avoid that.

This is especially true since #6897 is also touching the histogram (although not taking it out of `tokio_unstable`) and that would be a breaking change if this PR is released first.

tokio/src/runtime/metrics/histogram.rs
Owen-CH-Leung: Thanks for your comment!

I think the core issue here is that by exposing `worker_total_busy_duration`, we are also exposing the "real" `WorkerMetrics` at `tokio/src/runtime/metrics/worker.rs`, which in turn will attempt to parse `HistogramScale`.

Currently at master branch, when no `tokio_unstable` flag is passed in, the `WorkerMetrics` at `tokio/src/runtime/metrics/mock.rs` will be parsed (which is just an empty struct). And if the flag is passed in, the `WorkerMetrics` at `tokio/src/runtime/metrics/worker.rs` will be parsed

https://github.com/tokio-rs/tokio/blob/master/tokio/src/runtime/metrics/mod.rs#L27-L40

Owen-CH-Leung: Yeah becoz `HistogramScale` is still behind `cfg_unstable_metrics`:

https://github.com/tokio-rs/tokio/pull/6899/files#diff-15ebfca6124144b3cffd9845fa3ce5596c342ce94d84dd406ee84f856db4c66cR23-R25

hds: OK, now I believe I understand.

I would suggest that instead of compiling with the entire worker metrics in order to access only the `busy_duration_total` field, we should gate all the fields that we won't be using on `cfg_unstable_metrics`. Otherwise users will still be paying the price for metrics which they don't have access to - and that is something that we would really like to avoid.

Have a look at the runtime Builder for an example of how we have fields and implementations that touch them gated on a cfg flag:

https://github.com/tokio-rs/tokio/blob/512e9decfb683d22f4a145459142542caa0894c9/tokio/src/runtime/builder.rs#L125-L126

As a general rule, if you need `#[allow(dead_code)]` then there is probably something that should be gated on a cfg flag instead.

Owen-CH-Leung: Thanks @hds . I've reverted changes to `histogram.rs`. Also the real `Histogram`, `HistogramBuilder` and `HistogramBatch` are gated behind unstable flag now, and instead I've created a mocked version of these for compilation. 

For `WorkerMetrics`,  as we target to stabilize more metrics, I'd suggest exposing all fields instead of stabilizing only `busy_duration_total` and putting other fields behind unstable.

Let me know what you think

hds: @Owen-CH-Leung the problem with stabilizing all of `WorkerMetrics` is that when `tokio_unstable` is **not** enabled, the runtime will pay the price for all those metrics, but there will be no way to access them. For this reason I think that it would be better to only stabilize what we're exposing.

Owen-CH-Leung: Thanks @hds . I've made changes to hide most of the fields of `WorkerMetrics` behind unstable flag, except for `busy_duration_total`, `queue_depth` and `thread_id`. The latter 2 fields are needed in order for `set_queue_depth` and `set_thread_id` to work properly.

I've also enriched the mock `MetricsBatch` to have minimal implementation of `batch::MetricsBatch` so that the `worker_total_busy_duration` API can function properly under stable build

Let me know your thoughts!

tokio/src/runtime/scheduler/current_thread/mod.rs
```diff
@@ -533,7 +533,7 @@ impl Handle {
         self.shared.inject.len()
     }
 
-    #[allow(dead_code)]
+    // #[allow(dead_code)]
```
rcoh:
```suggestion
```

tokio/src/runtime/scheduler/multi_thread/handle/metrics.rs
```diff
@@ -18,7 +18,7 @@ impl Handle {
         self.shared.injection_queue_depth()
     }
 
-    #[allow(dead_code)]
+    // #[allow(dead_code)]
```
rcoh:
```suggestion
```

hds (changes requested):

tokio/src/runtime/metrics/mock.rs
```diff
@@ -20,32 +39,31 @@ impl SchedulerMetrics {
...
+        let now = Instant::now();
+
+        MetricsBatch {
+            busy_duration_total: 0,
+            processing_scheduled_tasks_started_at: now,
+        }
     }
 
-    pub(crate) fn submit(&mut self, _to: &WorkerMetrics, _mean_poll_time: u64) {}
+    pub(crate) fn submit(&mut self, worker: &WorkerMetrics, _mean_poll_time: u64) {
```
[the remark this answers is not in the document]
Owen-CH-Leung: Indeed spliting `metrics::MetricsBatch` is a much viable way of stabilising. I've adopted your suggestion and split it into stable & unstable (and group unstable functions into a single unstable block. Thanks a lot for reviewing!

tokio/src/runtime/metrics/batch.rs
```diff
@@ -39,87 +52,107 @@ pub(crate) struct MetricsBatch {
...
     }
 
     /// The worker was unparked.
     pub(crate) fn unparked(&mut self) {
-        self.park_unpark_count += 1;
+        #[cfg(tokio_unstable)]
+        {
+            self.park_unpark_count += 1;
+        }
     }
```
hds, line 156: In the rest of the Tokio code base, we do this a different way, so let's stick to that convension. Instead of gating functionality within a function, we have a separate empty function definition when the cfg flag isn't enabled. So this function would become:

```rust
cfg_unstable_metrics! {
    /// The worker was unparked.
    pub(crate) fn unparked(&mut self) {
        self.park_unpark_count += 1;
    }
}

cfg_not_unstable_metrics! {
    /// The worker was unparked.
    pub(crate) fn unparked(&mut self) {}
}
```

Please do the same here. Keep a single `cfg_unstable_metrics` block (and a single `cfg_not_unstable_metrics` block) for all the functions that require this behavior, so that they're grouped together.

For the more complex functions above that have a mix of stablized and unstablized implementation, split the unstablized part out into a separate function with an impl in each of the macro blocks (see example in the comment on `submit`).

tokio/src/runtime/metrics/batch.rs
```diff
@@ -39,87 +52,107 @@ pub(crate) struct MetricsBatch {
...
-            .park_unpark_count
-            .store(self.park_unpark_count, Relaxed);
-        worker.noop_count.store(self.noop_count, Relaxed);
-        worker.steal_count.store(self.steal_count, Relaxed);
-        worker
-            .steal_operations
-            .store(self.steal_operations, Relaxed);
-        worker.poll_count.store(self.poll_count, Relaxed);
-
+    pub(crate) fn submit(&mut self, worker: &WorkerMetrics, _mean_poll_time: u64) {
```
hds, line 105: As per the comment on `unparked`, let's follow the convension elsewhere in the Tokio codebase where we need different implementations for stablized vs. unstablized functionality. Here that would mean the following:

```rust
pub(crate) fn submit(&mut self, worker: &WorkerMetrics, mean_poll_time: u64) {
    worker
        .busy_duration_total
        .store(self.busy_duration_total, Relaxed);

    self.submit_unstable(worker, mean_poll_time);
}

cfg_not_unstable_metrics! {
    #[inline(always)]
    fn submit_unstable(&mut self, _worker: &WorkerMetrics, _mean_poll_time: u64) {}
}

cfg_unstable_metrics! {
    #[inline(always)]
    fn submit_unstable(&mut self, worker: &WorkerMetrics, mean_poll_time: u64){
        worker.mean_poll_time.store(_mean_poll_time, Relaxed);
        worker.park_count.store(self.park_count, Relaxed);
        worker
            .park_unpark_count
            .store(self.park_unpark_count, Relaxed);
        worker.noop_count.store(self.noop_count, Relaxed);
        worker.steal_count.store(self.steal_count, Relaxed);
        worker
            .steal_operations
            .store(self.steal_operations, Relaxed);
        worker.poll_count.store(self.poll_count, Relaxed);

        worker
            .local_schedule_count
            .store(self.local_schedule_count, Relaxed);
        worker.overflow_count.store(self.overflow_count, Relaxed);

        if let Some(poll_timer) = &self.poll_timer {
            let dst = worker.poll_count_histogram.as_ref().unwrap();
            poll_timer.poll_counts.submit(dst);
        }
    }
}
```

Use the same `cfg_unstable_metrics` and `cfg_not_unstable_metrics` blocks as for the other functions that need them (so there is only one of each in `impl MetricsBatch`.

tokio/src/runtime/metrics/batch.rs
```diff
@@ -39,87 +52,107 @@ pub(crate) struct MetricsBatch {
...
-    /// Instant when the most recent task started polling.
-    poll_started_at: Instant,
+        /// Instant when the most recent task started polling.
+        poll_started_at: Instant,
+    }
 }
 
 impl MetricsBatch {
-    pub(crate) fn new(worker_metrics: &WorkerMetrics) -> MetricsBatch {
+    pub(crate) fn new(_worker_metrics: &WorkerMetrics) -> MetricsBatch {
```
hds, line 71: Split this into 2 implementations each gated by `cfg_(not_)unstable_metrics!`.

tokio/src/runtime/metrics/mod.rs
```diff
@@ -34,7 +35,12 @@ cfg_unstable_metrics! {
 }
 
 cfg_not_unstable_metrics! {
-    mod mock;
+    mod worker;
+    pub(crate) use worker::WorkerMetrics;
```
hds, line 39: Please move the modules and imports that are in both the macro blocks out above them, we don't need to gate them at all.

tokio/src/runtime/metrics/worker.rs
```diff
@@ -57,32 +70,40 @@ pub(crate) struct WorkerMetrics {
...
 
 impl WorkerMetrics {
-    pub(crate) fn from_config(config: &Config) -> WorkerMetrics {
-        let mut worker_metrics = WorkerMetrics::new();
-        worker_metrics.poll_count_histogram = config
-            .metrics_poll_count_histogram
-            .as_ref()
-            .map(|histogram_builder| histogram_builder.build());
-        worker_metrics
+    cfg_not_unstable_metrics! {
```
hds, line 73: Let's move this down to be right above the `cfg_unstable_metrics!` block so that we keep the conditionally compiled implementations together.

hds (changes requested):
The implementation is looking good now. Just a few organizational things so that we do conditional compilation the same as the rest of the code base.

Owen-CH-Leung:
@hds Thanks for your detailed review!

I've revised the PR to gate code based on `cfg_(not_)unstable_metrics!` macros, and also re-organize code in the `mod.rs` and `worker.rs`. Let me know if you have any other feedback.

rcoh (approved):
Overall this looks good to me. I think incremental stabilization is inherently annoying to do—I think it might be better if we invested in some primitives. I think we should also endeavor to standardize the way functions are made to be no-ops.

tokio/src/runtime/metrics/batch.rs
```diff
@@ -133,40 +193,66 @@ impl MetricsBatch {
...
 }
 
 cfg_rt_multi_thread! {
     impl MetricsBatch {
-        pub(crate) fn incr_steal_count(&mut self, by: u16) {
-            self.steal_count += by as u64;
+        pub(crate) fn incr_steal_count(&mut self, _by: u16) {
+            #[cfg(tokio_unstable)] {
+                self.steal_count += _by as u64;
+            }
```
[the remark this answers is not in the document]
Owen-CH-Leung: Thanks @rcoh . I've added a new macro and put all the ab code implementation into this macro. I also recorganize fields so that stable fields come first. Please can I have your review again ? Thanks!

tokio/src/runtime/metrics/batch.rs
```diff
@@ -133,40 +197,84 @@ impl MetricsBatch {
...
-        pub(crate) fn incr_steal_count(&mut self, by: u16) {
-            self.steal_count += by as u64;
+        cfg_metrics_variant! {
+            stable: {
+                pub(crate) fn incr_steal_count(&mut self, _by: u16) {}
+            },
+            unstable: {
+                pub(crate) fn incr_steal_count(&mut self, _by: u16) {
+                    self.steal_count += _by as u64;
+                }
```
rcoh, line 254:
```suggestion
                pub(crate) fn incr_steal_count(&mut self, by: u16) {
                    self.steal_count += by as u64;
                }
```

tokio/src/runtime/metrics/worker.rs
```diff
@@ -15,40 +18,50 @@ use std::thread::ThreadId;
 #[derive(Debug, Default)]
 #[repr(align(128))]
 pub(crate) struct WorkerMetrics {
```
rcoh, line 20: same—please move stable fields to top

tokio/src/runtime/metrics/worker.rs
```diff
@@ -15,40 +18,50 @@ use std::thread::ThreadId;
...
     /// Number of tasks scheduled for execution on the worker's local queue.
     pub(crate) local_schedule_count: MetricAtomicU64,
 
+    #[cfg(tokio_unstable)]
     /// Number of tasks moved from the local queue to the global queue to free space.
     pub(crate) overflow_count: MetricAtomicU64,
 
     /// Number of tasks currently in the local queue. Used only by the
     /// current-thread scheduler.
     pub(crate) queue_depth: MetricAtomicUsize,
```
rcoh, line 62: is this intentionally stabilized?

tokio/src/runtime/metrics/batch.rs
Owen-CH-Leung: Thanks & revised

tokio/src/runtime/metrics/worker.rs
Owen-CH-Leung: I believe `queue_depth` is already stabilized at master and is used by the current thread scheduler : 

https://github.com/tokio-rs/tokio/blob/master/tokio/src/runtime/scheduler/current_thread/mod.rs#L341

tokio/src/runtime/metrics/batch.rs
```diff
@@ -1,125 +1,189 @@
...
-            self.poll_count_on_last_park = self.poll_count;
+    cfg_metrics_variant! {
+        stable: {
+            /// The worker is about to park.
+            pub(crate) fn about_to_park(&mut self) {}
+        },
+        unstable: {
+            /// The worker is about to park.
+            pub(crate) fn about_to_park(&mut self) {
+                #[cfg(tokio_unstable)]
```
rcoh, line 161: don't need a double config here

tokio/src/runtime/metrics/batch.rs
```diff
@@ -133,40 +197,84 @@ impl MetricsBatch {
...
-            poll_timer.poll_counts.measure(elapsed, 1);
+    cfg_metrics_variant! {
+        stable: {
+            /// Stop polling an individual task
+            pub(crate) fn end_poll(&mut self) {}
+        },
+        unstable: {
+            /// Stop polling an individual task
+            pub(crate) fn end_poll(&mut self) {
+                #[cfg(tokio_unstable)]
```
rcoh, line 224: another double cfg

tokio/src/macros/cfg.rs
```diff
@@ -602,3 +602,15 @@ macro_rules! cfg_is_wasm_not_wasi {
         )*
     }
 }
+
+macro_rules! cfg_metrics_variant {
```
rcoh, line 606: please add a doc comment here

rcoh (approved):
This LGTM (other than three small notes inline).

I am fine with it as is—one option that other reviewers may consider is using the metrics variant macro you defined on the function bodies instead—unless for macro reasons that's impossible? In any case, I found it plenty readable since the function bodies are all quite short.

Thanks for the iterations on this?

tokio/src/runtime/metrics/batch.rs
Owen-CH-Leung: Thanks & removed double cfg

tokio/src/macros/cfg.rs
Owen-CH-Leung: Added. Thanks

rcoh:
LGTM, thanks!

Darksonn:
Thanks for doing this, and thanks @rcoh for the review. 

It isn't great that stabilizing a metric takes a change whose diff is 600+ lines long. Why is the change so large? I'm not saying there's a problem with your PR; if our codebase makes it difficult, then that's on us, but I'd like to understand the reason.

cc @Noah-Kennedy @hds

rcoh:
minimal stabilizations are usually OK. The problem with this (and the
reason it ended up so large) is that it was the first stabilization of a
component of the metrics system inside the workers.

Previously, on stable, the entire struct was subbed with a mock/stub and
the actual impl was only enabled on unstable. This metric was the first
metric producing this to be enabled.

There are a couple of ways that we could make this smaller:
- Accept that the metrics will be written to, even on stable (potential
performance cost)
- allow more dead code to exist (which I think would mitigate the level of
cfg gating required)

Ultimately, the heavily nested module structure where everything is a layer
of re-exports (not sure why this is, some historical reason perhaps?)
definitely makes stabilizing these metrics more of a hassle

hds (approved):
Looks good to me, just need to get the FreeBSD tests working.

Thank you for all your work on this one!

Darksonn:
Ok, thanks, that's fine then. The FreeBSD tests are already fixed on master.

Owen-CH-Leung:
> Ok, thanks, that's fine then. The FreeBSD tests are already fixed on master.

Shall we merge this PR ? I'd like to file a few more PRs to stabilize more metrics after this

PR #6905, opened 2024-10-16
ci: Print out the correct sort order for the dictionary on failure

rcoh:
## Motivation

When adding words to the dictionary, it can be hard to actually get the en_US.UTF8 locale to work on all platforms. This prints it in CI so you can copy-paste it back.


Darksonn (approved):
Thank you.

PR #6957, opened 2024-11-07
metrics: Fix bug where num_buckets became inconsistent

rcoh:
## Motivation

In Tokio 1.41, #6897 refactored histograms and their construction. Unfortunately, this change included a bug where `num_buckets` was not properly preserved when the legacy builder methods were used. This could cause an inconsistency where even if the customer set max buckets, we would still use the default.

**This can result in a runtime panic because the histogram thinks it has `N` buckets, but in reality, it only has 10.**

## Solution

Fix the bug.


Darksonn (approved):
Thanks.

PR #6999, opened 2024-11-29
runtime: fixed yield_now lost wakeup

quininer:
## Motivation

Fixes https://github.com/tokio-rs/tokio/issues/6996

## Solution

The PR makes wake defer only occur in the enter runtime state. this avoid lost wakeup due to defer during exit runtime.


Darksonn (approved):
Thank you.

PR #7023, opened 2024-12-08
Run spawn_pinned test on miri

tiif:
This test previously blocked on a miri bug, but now it can run successfully. :tada:

Tracking issue #6812


Darksonn (approved):
Thanks.

PR #7035, opened 2024-12-13
Minor simplification in lib.rs doc string

barafael:


Darksonn (commented):
Thanks! Can you fix these too?

* README.md
* tokio/README.md
* tokio/src/runtime/mod.rs

Darksonn (approved):
Thanks.

PR #7066, opened 2025-01-04
Change `AsyncReadExt::read` docs formatting to clarify requirements

AeonSolstice:
## Motivation

I was reading the docs for the method, and in its current state, it read to me as:
> No guarantees are provided about the contents of `buf` when this function is called, implementations cannot rely on [any] ~property~ [of] the contents of `buf` being `true`.

With the inline block formatting on *true*, it makes it seem that you must not rely on `buf` containing the value `true`.

## Solution

Remove the inline block formatting on *true*, as we are not talking about the boolean value `true` in code.


Darksonn (approved):
Thanks.

PR #7067, opened 2025-01-04
Fix ambiguity in `TcpStream::try_write_vectored` docs

AeonSolstice:
## Motivation

The buffer cannot be possible.

## Solution

Make it clear that the last buffer that is read from by the method, may have only been partially read.

As a side note, does the method override the buffers' contents? If not, then "consumed" should be replaced with "written".


Darksonn (approved):
Thanks.

PR #7081, opened 2025-01-09
sync: add `CancellationToken::run_until_cancelled_owned`

tysen:
Pretty natural addition to match the `cancelled_owned` fn.

## Motivation

```rust
let token = CancellationToken::new();
tokio::spawn(token.run_until_cancelled(my_async_fn())); // doesn't compile
tokio::spawn(async move { // current workaround
    token.run_until_cancelled(my_async_fn())).await
});
tokio::spawn(token.run_until_cancelled_owned(my_async_fn())); // compiles, looks nicer
```

## Solution
Just a carbon copy of the existing fn switching to the appropriate cancellation future.

### DIFF
--- a/tokio-util/src/sync/cancellation_token.rs
+++ b/tokio-util/src/sync/cancellation_token.rs
@@ -287,6 +287,22 @@ impl CancellationToken {
         }
         .await
     }
+
+    /// Runs a future to completion and returns its result wrapped inside of an `Option`
+    /// unless the `CancellationToken` is cancelled. In that case the function returns
+    /// `None` and the future gets dropped.
+    ///
+    /// The function takes self by value and returns a future that owns the token.
+    ///
+    /// # Cancel safety
+    ///
+    /// This method is only cancel safe if `fut` is cancel safe.
+    pub async fn run_until_cancelled_owned<F>(self, fut: F) -> Option<F::Output>
+    where
+        F: Future,
+    {
+        self.run_until_cancelled(fut).await
+    }
 }
 
 // ===== impl WaitForCancellationFuture =====
--- a/tokio-util/tests/sync_cancellation_token.rs
+++ b/tokio-util/tests/sync_cancellation_token.rs
@@ -493,3 +493,58 @@ fn run_until_cancelled_test() {
         );
     }
 }
+
+#[test]
+fn run_until_cancelled_owned_test() {
+    let (waker, _) = new_count_waker();
+
+    {
+        let token = CancellationToken::new();
+        let to_cancel = token.clone();
+
+        let takes_ownership = move |token: CancellationToken| {
+            token.run_until_cancelled_owned(std::future::pending::<()>())
+        };
+
+        let fut = takes_ownership(token);
+        pin!(fut);
+
+        assert_eq!(
+            Poll::Pending,
+            fut.as_mut().poll(&mut Context::from_waker(&waker))
+        );
+
+        to_cancel.cancel();
+
+        assert_eq!(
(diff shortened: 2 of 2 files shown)


tokio-util/src/sync/cancellation_token.rs
```diff
@@ -253,42 +266,55 @@ impl CancellationToken {
...
+            cancellation: self.cancelled_owned(),
             future: fut,
         }
         .await
     }
 }
 
+// ===== impl RunUntilCancelledFuture =====
+
+impl<C, F: Future> Future for RunUntilCancelledFuture<C, F>
```
tglane, line 299:
```suggestion
impl<C, F> Future for RunUntilCancelledFuture<C, F>
```

The bound on `Future` for `F` is already specified in the where clause.

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -99,6 +99,19 @@ pin_project! {
     }
 }
 
+pin_project! {
+    /// A Future that is resolved once the corresponding [`CancellationToken`]
+    /// is cancelled or a given Future gets resolved. It is biased towards the
+    /// Future completion.
+    #[must_use = "futures do nothing unless polled"]
+    struct RunUntilCancelledFuture<C, F: Future> {
```
tglane, line 107: Know this type basically is a biased select future I think because `C` technically is not required to be some sort of `CancellationToken`. I think we could replace the usage of this type with a `tokio::select!` if we add the macros feature to `tokio-util`s `tokio` dependency. @Darksonn do you have a optionion on that?

tglane (changes requested):

Darksonn (commented):
Is there a reason that this PR is not just this?
```rs
pub async fn run_until_cancelled_owned<F>(self, fut: F) -> Option<F::Output>
where
    F: Future,
{
    self.run_until_cancelled(fut).await
}
```
I believe it should be sufficient - the async/await syntax takes care of making it owned.

tokio-util/tests/sync_cancellation_token.rs
```diff
@@ -493,3 +493,51 @@ fn run_until_cancelled_test() {
         );
     }
 }
+
+#[test]
+fn run_until_cancelled_owned_test() {
```
tglane, line 498: I think it would be nice to add a case to this test that would only compile with the `run_until_cancelled_owned` and not with `run_until_cancelled` like you mentioned in the PR description as this was the reason behind this PR.

tglane (changes requested):

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -287,6 +287,22 @@ impl CancellationToken {
...
+
+    /// Runs a future to completion and returns its result wrapped inside of an `Option`
+    /// unless the `CancellationToken` is cancelled. In that case the function returns
+    /// `None` and the future gets dropped.
+    ///
+    /// The function takes self by value.
+    ///
+    /// # Cancel safety
+    ///
+    /// This method is only cancel safe if `fut` is cancel safe.
```
tglane, line 299: The docs should mention the difference to `run_until_cancelled` in some way.
Maybe add something like this to the docs:
``` 
///
/// The function takes self by value and returns a future that owns the
/// token.
```

tglane (changes requested):
Apart from this nit it looks good to me know.

tglane (approved):

Darksonn (approved):
Thanks.

PR #7090, opened 2025-01-11
sync: fix sync::broadcast::Sender<T>::closed() doctest

evanrittenhouse:
The test's previous iteration could sometimes flake since we didn't await the completion of the first task. Since the tasks only existed to `move` the relevant `rx`'s in, to force a drop, we can omit them entirely and drop the `rx`s via `drop()`. This prevents any scheduling-related flakes.


Darksonn (approved):
Thank you.

Issue #7077, opened 2025-01-08
Add `.len()` method to `UnboundedSender`


tqwewe:
**Is your feature request related to a problem? Please describe.**
The number of buffered messages can be derived from a **bounded** mpsc channel sender with `len = tx.max_capacity() - tx.capacity()`. However there's no methods available to get the number of buffered messages from an **unbounded** mpsc channel sender.

**Describe the solution you'd like**
A `UnboundedSender::len()` method which returns how many messages are buffered in the channel and pending to be received.

```rust
// Currently possible
let (boundex_tx, bounded_rx) = mpsc::channel(10);
let bounded_len = bounded_tx.max_capacity() - bounded.capacity();

// Proposed `len` method
let (unbounded_tx, unbounded_rx) = mpsc::unbounded_channel();
let unbounded_len = unbounded_tx.len();
```

unikol-salm70900:
Hello, we tried to solve the issue.

This is what we did:

Implement a len() method for UnboundedSender to return the number of buffered messages in the channel. This involves modifying the UnboundedSender struct and adding the new method.

You can review changes in this commit: https://github.com/unikol-salm70900/tokio-rs-tokio-7077/commit/3bebf4976c506e10dd3c8308bb5434a73292ee07.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

Latta AI seeks to solve problems in open source projects as part of its mission to support developers around the world. Learn more about our mission at https://latta.ai/ourmission . If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

tqwewe:
Another related issue is <https://github.com/tokio-rs/tokio/issues/6314>.

> I do wonder what the usecase is for knowing the length from the sender.

In my case, I'm implementing a least-loaded kind of scheduler for a pool of actors. The pool stores the sender of each actor, and in order to forward a message to the actor with the least amount of load I could pick the actor in the pool with the smallest `.len()` value.

The alternative of course would be that the actor pool keeps a counter for each actor, but this brings some restrictions, particularly when the worker actors are generic actors defined by the user.

Archibajl:
So I tried something here https://github.com/tokio-rs/tokio/pull/7092 
Let me know if this is what you want, if this fulfils your needs, or if it's even an acceptable update for this repo.

[...]

PR #7092, opened 2025-01-12
7707 updated Tx struct and UnboundedSender to return message length

Archibajl:
update recommended by https://github.com/tokio-rs/tokio/issues/7077
simply they requested that we produce a len method that returns the length of messages in que for UnboundedSender struct

This should return the length of messages in que, though it will get it at an instance, thus as a threaded application may be inaccurate when the instance is updated simultaneously, though should be approximately correct for the time/instance of the method call.

no tests for this as there is not a test class for this file and associated structs (which needs to be rectified), same tests pass/ & fail as the regular master branch.

I've been looking for something to help contribute to this repo, though I'm new to rust, this repo, and my ide, so let me know what's wrong, what can be better, and what I'm missing.

hope this works for you and I hope to be a more useful contributor in the future.

### DIFF
--- a/tokio/src/sync/mpsc/chan.rs
+++ b/tokio/src/sync/mpsc/chan.rs
@@ -1,6 +1,6 @@
 use crate::loom::cell::UnsafeCell;
 use crate::loom::future::AtomicWaker;
-use crate::loom::sync::atomic::AtomicUsize;
+use crate::loom::sync::atomic::{AtomicUsize, Ordering};
 use crate::loom::sync::Arc;
 use crate::runtime::park::CachedParkThread;
 use crate::sync::mpsc::error::TryRecvError;
@@ -207,6 +207,10 @@ impl<T, S: Semaphore> Tx<T, S> {
             return;
         }
         notified.await;
+    }
+
+    pub(crate) fn len(&self) -> usize {
+        self.inner.tx_count.load(Ordering::SeqCst)
     }
 }
 
--- a/tokio/src/sync/mpsc/chan.rs
+++ b/tokio/src/sync/mpsc/chan.rs
@@ -208,10 +208,6 @@ impl<T, S: Semaphore> Tx<T, S> {
         }
         notified.await;
     }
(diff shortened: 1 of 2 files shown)


tokio/src/sync/mpsc/unbounded.rs
```diff
@@ -675,6 +675,11 @@ impl<T> UnboundedSender<T> {
     pub fn weak_count(&self) -> usize {
         self.chan.weak_count()
     }
+
+    //Returns length of [`UnboundedSender`] semaphores /2, supposidly the number of messages.
```
tqwewe, line 679: Code docs start with `/// `, and it probably shouldn't reveal the inner workings of the channel by refering to the semaphore, since its not really relvant to the end user.

Also "supposidly" should be "supposedly".

What @conradludgate said in their comment https://github.com/tokio-rs/tokio/issues/7077#issuecomment-2577091067, this probably implies that the len function itself should do the division itself. But I'd probably put this division in the chan.rs `len` function

tokio/src/sync/mpsc/chan.rs
```diff
@@ -208,6 +208,10 @@ impl<T, S: Semaphore> Tx<T, S> {
         }
         notified.await;
     }
+
+    pub(crate) fn len(&self) -> usize {
+        self.inner.tx_count.load(Ordering::SeqCst)
```
conradludgate, line 213: This `tx_count` counts how many `Sender`s are active, not how many messages are waiting

tokio/src/sync/mpsc/unbounded.rs
Archibajl: updated comment and moved division into the chan.rs classes len function

tokio/src/sync/mpsc/chan.rs
Archibajl: wouldn't the number of senders imply the same number of semaphores -> 2x number of messages?

PR #7107, opened 2025-01-17
Rough implementation of task poll callbacks

rcoh:
_Just throwing this up for initial discussion, I need to do work to clean up internal names & get all the right `tokio_unstable` gating.

- [ ] Implement on current thread worker
- [ ] Implement on alt runtime
- [ ] fix unstable
- [ ] more tests and docs

## Motivation

We would like to be able to store metadata before polls start and stop to allow us to instrument all polls, not just polls where you have spawned the task.

## Solution

Add callbacks for poll start and stop


rcoh:
closing in favor of #7120

PR #7112, opened 2025-01-18
runtime: clean up magic number usage

maminrayej:
This PR contains a minor clean up regarding the magic number `16` in relation to the `pending_release` buffer capacity.


Darksonn (approved):
Thanks.

PR #7117, opened 2025-01-22
build: upgrade `windows-sys` 0.52 → 0.59

ErichDonGubler:
## Motivation

We at Mozilla would like to upgrade WGPU and crates we otherwise use in `mozilla-central` to `windows` 0.59. 😀 This will let us handle breaking changes, and stay in sync. with the current ecosystem for Windows targets.

## Solution

Migrate `tokio` to `windows` 0.59.

## Remaining work

- [ ] Figure out why other stuff is failing to compile (hit a timebox).

### DIFF
--- a/tokio/Cargo.toml
+++ b/tokio/Cargo.toml
@@ -117,7 +117,7 @@ libc = { version = "0.2.168" }
 nix = { version = "0.29.0", default-features = false, features = ["aio", "fs", "socket"] }
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = "0.52"
+version = "0.59"
 optional = true
 
 [target.'cfg(windows)'.dev-dependencies.windows-sys]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,3 +17,10 @@ members = [
 
 [workspace.metadata.spellcheck]
 config = "spellcheck.toml"
+
+
+[patch.crates-io]
+# TODO: Consume a release of `mio` that contains <https://github.com/tokio-rs/mio/pull/1857>.
+mio = { git = "https://github.com/erichdongubler-contrib/mio", rev = "1ef9c035f9c147621a6aa1d7b80a14d7acb09555" }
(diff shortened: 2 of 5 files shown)


Darksonn:
Thanks for the PR! Happy to merge this once the dependencies are fixed as well.

elsnerpaul:
Hi, any update on this? Would love to see this :)

Darksonn:
This is still waiting for dependencies.

ErichDonGubler:
Update on dependencies: `mio` is now ready to go, just waiting on `socket2` now. 😤

atouchet:
socket2 0.6 is finally out.

Tokio update PR: #7443

Darksonn:
Okay socket2 has landed. We can merge this now if it's updated.

ErichDonGubler:
Removed the `patch` entry from `Cargo.toml`. LFG! 😀

tokio/Cargo.toml
```diff
@@ -106,7 +106,7 @@ tracing = { version = "0.1.29", default-features = false, features = ["std"], op
 [target.'cfg(all(tokio_uring, target_os = "linux"))'.dependencies]
 io-uring = { version = "0.7.6", default-features = false }
 libc = { version = "0.2.168" }
-mio = { version = "1.0.1", default-features = false, features = ["os-poll", "os-ext"] }
+mio = { version = "1.0.4", default-features = false, features = ["os-poll", "os-ext"] }
```
Darksonn: There's no need for this change. Tokio continues to be compatible with 1.0.1, 1.0.2, and 1.0.3.

ErichDonGubler: 🤔 This would be good for forcing users to unify their `windows` crate ecosystem to the same versions, but I suppose it's not absolutely necessary to make things compile. Reverted with da52dfb1.

Darksonn: You're going to get the newest version unless you do something special, so I'm not worried about that kind of thing.

Darksonn (approved):
Thank you.

PR #7123, opened 2025-01-25
ci: test AArch64/Armv7hf Linux on ubuntu-22.04-arm runner

taiki-e:
## Motivation

Currently, we are testing `aarch64-unknown-linux-gnu` and `armv7-unknown-linux-gnueabihf` using QEMU, but QEMU is slow than native execution.

GitHub [now supports free AArch64 Linux runners as public preview](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/), so this PR use it for them.

## Solution

As for `aarch64-unknown-linux-gnu`, it is enough to change the OS to run on, and as for `armv7-unknown-linux-gnueabihf`, it is necessary to change the OS to run on and install some additional packages, but the [latter is handled by the setup-cross-toolchain-action](https://github.com/taiki-e/setup-cross-toolchain-action/releases/tag/v1.27.0) currently already in use, so in practice it is also enough to change the OS to run on.

(As for `armv5te-unknown-linux-gnueabi`, Ubuntu does not provide packages for `armel`, so unlike `armv7-unknown-linux-gnueabihf` (`armhf`), we still need to use QEMU for now.)

This PR also removes CircleCI config (`.circleci/config.yml`) because native AArch64 test is now covered in GitHub Actions.

Note that this PR uses `ubuntu-22.04-arm` instead of `ubuntu-24.04-arm` due to upstream bug: https://github.com/rust-lang/rust/issues/135867
(I leaved TODO comments about it.)


.github/workflows/ci.yml
```diff
@@ -532,15 +532,19 @@ jobs:
 
   cross-test-with-parking_lot:
```
taiki-e, line 533: Test for aarch64 is no longer "cross-test", but I leave it as is for now, the new runner is still in public preview, and if something goes wrong we can revert this PR with less-conflicts.

Darksonn (approved):
Thanks!

PR #7125, opened 2025-01-25
Stabilize `runtime::id::Id`

chinedufn:
This commit stabilizes the `runtime::id::Id`.

`runtime::id::Id` uniquely identifies a runtime relative to the parent process' other currently running runtimes.

It was introduced in July 2023 and has only received one modification since, where `impl From<NonZeroU32> for Id` was added in December 2023 in 3a4aef17b2c70d255affa51eb473efcf703896e4 .

## `std::from::From` implementations

This commit removes the runtime `Id`'s `From` implementations.

Removing unnecessary publc APIs makes stabilizing the `Id` less risky. Instead, a crate private `Id::new` method has been introduced.

This means that as of this commit, users can no longer safely construct
 an `Id`.
Instead, they must get a runtime `Id` from a method such as `Handle::current().id()`.

The `Id`'s documentation explains that it is intended to be an opaque handle. Therefore, we should not expose a way to construct one.

### DIFF
--- a/tokio/src/runtime/handle.rs
+++ b/tokio/src/runtime/handle.rs
@@ -404,36 +404,27 @@ impl Handle {
         }
     }
 
-    cfg_unstable! {
-        /// Returns the [`Id`] of the current `Runtime`.
-        ///
-        /// # Examples
-        ///
-        /// ```
-        /// use tokio::runtime::Handle;
-        ///
-        /// #[tokio::main(flavor = "current_thread")]
-        /// async fn main() {
-        ///   println!("Current runtime id: {}", Handle::current().id());
-        /// }
-        /// ```
-        ///
-        /// **Note**: This is an [unstable API][unstable]. The public API of this type
-        /// may break in 1.x releases. See [the documentation on unstable
-        /// features][unstable] for details.
-        ///
-        /// [unstable]: crate#unstable-features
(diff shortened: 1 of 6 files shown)


chinedufn:
Closes https://github.com/tokio-rs/tokio/issues/5545

ADD-SP:
Hi @chinedufn, sorry for the late reply. Do you still want to work on this PR? If so, could you resolve the merge conflicts? If not, may I take over this work?

chinedufn:
Hi. I've rebased this PR onto the latest master branch. Merge conflicts are resolved.

tokio/src/task/local.rs
```diff
@@ -936,7 +936,7 @@ cfg_unstable! {
         /// [unstable]: crate#unstable-features
         /// [`Id`]: struct@crate::runtime::Id
         pub fn id(&self) -> runtime::Id {
```
martin-g, line 938: Should this method also be stable now ? Or this will be done sometime later ?

chinedufn: I haven't thought about whether this method should be stable.

Can we land this PR as is and consider stabilizing this separately?

ADD-SP: Follow up PR: <https://github.com/tokio-rs/tokio/pull/7776>

PR #7143, opened 2025-02-06
fs: align symlink parameter names with std

florian-g2:
## Motivation
Using `src` and `dst` as parameter names in the context of symlinks is confusing, especially in their order.

I usually think from left to right: a symlink points to a target. 
So I assumed that `src` would point to `dst`, but that isn’t the case here.

## Solution
Align to the standard library and use `original` and `link` as the parameter names.

### DIFF
--- a/tokio/src/fs/hard_link.rs
+++ b/tokio/src/fs/hard_link.rs
@@ -7,7 +7,7 @@ use std::path::Path;
 ///
 /// This is an async version of [`std::fs::hard_link`].
 ///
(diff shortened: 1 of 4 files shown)


maminrayej (approved):

maminrayej:
LGTM. But for consistency, please also rename the parameters for `hard_link` as well.

florian-g2:
Right. Missed that.

maminrayej:
Thanks.

PR #7152, opened 2025-02-13
sync: `oneshot::Receiver::is_terminated()`

cratelyn:
this commit introduces a new method to `tokio::sync::oneshot::Receiver<T>`.

broadly speaking, users of the oneshot channel are encouraged to `.await` the `Receiver<T>` directly, as it will only yield a single value.

users implementing their own `std::future::Future`s directly may instead poll the receiver via `<Receiver<T> as Future<Output = Result<T, RecvError>::poll(..)`. note that the contract of `Future::poll()` states that clients should not poll a future after it has yielded `Poll::Ready(value)`.

this commit provides a way to inspect the state of a receiver, to avoid violating the contact of `Future::poll(..)`, or requiring that a oneshot channel users track this state themselves externally via mechanisms like `futures::future::FusedFuture`, or wrapping the receiver in an `Option<T>`.

NB: this makes a small behavioral change to the implementation of `<Receiver<T> as Future<Output = Result<T, RecvError>::poll(..)`. this change is acceptable, per the `Future::poll()` documentation regarding panics:

> Once a future has completed (returned Ready from poll), calling its
poll method again may panic, block forever, or cause other kinds of problems; the Future trait places no requirements on the effects of such a call.

the upside of this is that it means a broken or closed channel, e.g. when the sender is dropped, will settle as "finished" after it yields an error.

see:
* https://github.com/tokio-rs/tokio/pull/7137#discussion_r1940242052
* https://doc.rust-lang.org/stable/std/future/trait.Future.html#panics

### DIFF
--- a/tokio/src/sync/oneshot.rs
+++ b/tokio/src/sync/oneshot.rs
@@ -941,12 +941,9 @@ impl<T> Receiver<T> {
     /// Sending a value and polling it.
     ///
     /// ```
-    /// use futures::task::noop_waker_ref;
     /// use tokio::sync::oneshot;
     ///
-    /// use std::future::Future;
-    /// use std::pin::Pin;
-    /// use std::task::{Context, Poll};
+    /// use std::task::Poll;
     ///
     /// #[tokio::main]
     /// async fn main() {
@@ -956,7 +953,7 @@ impl<T> Receiver<T> {
     ///     assert!(!rx.is_terminated());
     ///
     ///     // A receiver is not terminated it is polled and is still pending.
-    ///     let poll = Pin::new(&mut rx).poll(&mut Context::from_waker(noop_waker_ref()));
+    ///     let poll = futures::poll!(&mut rx);
     ///     assert_eq!(poll, Poll::Pending);
     ///     assert!(!rx.is_terminated());
     ///
--- a/tokio/src/sync/oneshot.rs
+++ b/tokio/src/sync/oneshot.rs
@@ -931,6 +931,67 @@ impl<T> Receiver<T> {
         }
     }
 
+    /// Checks if this receiver is terminated.
+    ///
+    /// This function returns true if this receiver has already yielded a [`Poll::Ready`] result.
+    /// If so, this receiver should no longer be polled.
+    ///
+    /// # Examples
+    ///
+    /// Sending a value and polling it.
+    ///
+    /// ```
+    /// use futures::task::noop_waker_ref;
+    /// use tokio::sync::oneshot;
+    ///
+    /// use std::future::Future;
+    /// use std::pin::Pin;
+    /// use std::task::{Context, Poll};
+    ///
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let (tx, mut rx) = oneshot::channel();
+    ///
+    ///     // A receiver is not terminated when it is initialized.
+    ///     assert!(!rx.is_terminated());
+    ///
+    ///     // A receiver is not terminated it is polled and is still pending.
+    ///     let poll = Pin::new(&mut rx).poll(&mut Context::from_waker(noop_waker_ref()));
+    ///     assert_eq!(poll, Poll::Pending);
+    ///     assert!(!rx.is_terminated());
+    ///
+    ///     // A receiver is not terminated if a value has been sent, but not yet read.
+    ///     tx.send(0).unwrap();
+    ///     assert!(!rx.is_terminated());
+    ///
+    ///     // A receiver *is* terminated after it has been polled and yielded a value.
+    ///     assert_eq!((&mut rx).await, Ok(0));
+    ///     assert!(rx.is_terminated());
+    /// }
+    /// ```
+    ///
+    /// Dropping the sender.
+    ///
+    /// ```
+    /// use tokio::sync::oneshot;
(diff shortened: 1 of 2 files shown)


cratelyn:
i see issues in ci at the moment, but i am marking this as ready for review because i believe these errors to be unrelated.

```
⚠️ Failed to start an instance: INVALID_ARGUMENT: Not Found 404 Not Found
POST [https://compute.googleapis.com:443/compute/v1/projects/cirrus-ci-community/zones/us-central1-c/instances](https://compute.googleapis.com/compute/v1/projects/cirrus-ci-community/zones/us-central1-c/instances)
```

tokio/tests/sync_oneshot.rs
```diff
@@ -292,3 +292,74 @@ fn sender_changes_task() {
...
+
+    let mut task = task::spawn(());
+    let poll = task.enter(|cx, _| Pin::new(&mut rx).poll(cx));
+    assert_ready_err!(poll);
+
+    assert!(
+        rx.is_terminated(),
+        "channel IS terminated after value is read"
+    );
+}
```
maminrayej, line 365: You could also add a test for `is_terminated` before and after `try_recv` as well.

cratelyn: great idea! 8fc0759 adds a test case that exercises `try_recv` as well.

tokio/src/sync/oneshot.rs
```diff
@@ -931,6 +931,67 @@ impl<T> Receiver<T> {
...
+    ///
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let (tx, mut rx) = oneshot::channel();
+    ///
+    ///     // A receiver is not terminated when it is initialized.
+    ///     assert!(!rx.is_terminated());
+    ///
+    ///     // A receiver is not terminated it is polled and is still pending.
+    ///     let poll = Pin::new(&mut rx).poll(&mut Context::from_waker(noop_waker_ref()));
```
maminrayej, line 959: I think using something like `futures::poll!(&mut rx)` makes the example easier to follow.

tokio/src/sync/oneshot.rs
```diff
@@ -1106,18 +1167,18 @@ impl<T> Future for Receiver<T> {
 
         let ret = if let Some(inner) = self.as_ref().get_ref().inner.as_ref() {
             #[cfg(all(tokio_unstable, feature = "tracing"))]
-            let res = ready!(trace_poll_op!("poll_recv", inner.poll_recv(cx)))?;
+            let res = ready!(trace_poll_op!("poll_recv", inner.poll_recv(cx))).map_err(Into::into);
```
maminrayej, line 1170: Out of curiosity, what is the motivation for this change?

tokio/src/sync/oneshot.rs
cratelyn: sure thing! see f7df23c1.

tokio/src/sync/oneshot.rs
cratelyn: this is described in more detail in e18318be62cefc99700923131fac7b47511c2dea.

> NB: this makes a small behavioral change to the implementation of
`<Receiver<T> as Future<Output = Result<T, RecvError>::poll(..)`. this
change is acceptable, per the `Future::poll()` documentation regarding
panics:
>
> > _Once a future has completed (returned Ready from poll), calling its
poll method again may panic, block forever, or cause other kinds of
problems; the Future trait places no requirements on the effects of such
a call._
>
> the upside of this is that it means a broken or closed channel, e.g.
when the sender is dropped, will settle as "finished" after it yields an
error.

the idea here is that we should set `self.inner = None` once we yield either a `Poll::Ready(Ok(T))` _or_ an error `Poll::Ready(Err(RecvError))`, so that we can detect when the receiver has terminated.

maminrayej: Ah you're right.

cratelyn: thankfully, i don't believe this would qualify as a breaking change. while there is a visible change in behavior (see [this](https://github.com/cratelyn/tokio/commit/b7d1b3eacfc1c722a64f70a4cc46ceaeea60dd15#diff-3d58acf0561e7ae958c437e513cc513c34d8917cbb51906d97319699beb9db7cR1044-R1057) example in b7d1b3e), the only callers affected would be in violation of the `Future` contract stating that `poll()` shall not be called again once `Ready(Output)` is yielded.

thanks for asking, it _is_ a change worth being intentional and explicit about. :slightly_smiling_face:

maminrayej (approved):

maminrayej:
Thanks.

mox692:
@rushilmehra Can you describe it in a separate issue? I think this change is not supposed to affect the existing code.

rushilmehra:
> @rushilmehra Can you describe it in a separate issue? I think this change is not supposed to affect the existing code.

We were accidentally polling a oneshot receiver after it was complete, honestly just misuse on our side. I can open a separate issue if you still think it's useful though

kylezs:
Fortunately for us it only appeared as a flaky test 😄 

Thanks for the nice PR description btw, helped finding the issue quickly :)

PR #7153, opened 2025-02-13
sync: `oneshot::Receiver::is_empty()`

cratelyn:
this commit introduces a new method to `tokio::sync::oneshot::Receiver<T>`.

this method returns true if the channel has no messages waiting to be received. this is similar to the existing `tokio::sync::mpsc::Receiver::is_empty()` and `tokio::sync::broadcast::Receiver::is_empty()` methods.

see:
* https://docs.rs/tokio/latest/tokio/sync/broadcast/struct.Receiver.html#method.is_empty
* https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.is_empty
* https://github.com/tokio-rs/tokio/pull/7137#discussion_r1940242052

### DIFF
--- a/tokio/src/sync/oneshot.rs
+++ b/tokio/src/sync/oneshot.rs
@@ -989,6 +989,51 @@ impl<T> Receiver<T> {
         self.inner.is_none()
     }
 
+    /// Checks if a channel is empty.
+    ///
+    /// This method returns `true` if the channel has no messages.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use tokio::sync::oneshot;
+    ///
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let (tx, mut rx) = oneshot::channel();
+    ///     assert!(rx.is_empty());
+    ///
+    ///     tx.send(0).unwrap();
+    ///     assert!(!rx.is_empty());
+    ///
+    ///     let _ = (&mut rx).await;
+    ///     assert!(rx.is_empty());
+    /// }
+    /// ```
+    pub fn is_empty(&self) -> bool {
+        if let Some(inner) = self.inner.as_ref() {
+            let state = State::load(&inner.state, Acquire);
+            if state.is_complete() {
+                // SAFETY: If `state.is_complete()` returns true, then the
+                // `VALUE_SENT` bit has been set and the sender side of the
+                // channel will no longer attempt to access the inner
+                // `UnsafeCell`. Therefore, it is now safe for us to access the
+                // cell.
+                //
+                // The channel is empty if it does not have a value.
+                unsafe { !inner.has_value() }
+            } else if state.is_closed() {
+                // The receiver closed the channel...
+                true
+            } else {
+                // No value has been sent yet.
+                true
+            }
+        } else {
+            true
+        }
+    }
+
     /// Attempts to receive a value.
     ///
     /// If a pending value exists in the channel, it is returned. If no value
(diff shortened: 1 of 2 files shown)


tokio/src/sync/oneshot.rs
```diff
@@ -931,6 +931,51 @@ impl<T> Receiver<T> {
...
+            } else if state.is_closed() {
+                // The receiver closed the channel...
+                true
+            } else {
+                // No value has been sent yet.
+                true
+            }
+        } else {
+            true
+        }
```
maminrayej, line 976: The branching makes it a little bit difficult to understand. Maybe it could be simplified a bit like this:
```suggestion
        let Some(inner) = self.inner.as_ref() else {
            return true;
        };

        let state = State::load(&inner.state, Acquire);
        if state.is_complete() {
            // SAFETY: If `state.is_complete()` returns true, then the
            // `VALUE_SENT` bit has been set and the sender side of the
            // channel will no longer attempt to access the inner
            // `UnsafeCell`. Therefore, it is now safe for us to access the
            // cell.
            //
            // The channel is empty if it does not have a value.
            unsafe { !inner.has_value() }
        } else {
            // The receiver closed the channel or no value has been sent yet.
            true
        }
```

tokio/tests/sync_oneshot.rs
```diff
@@ -292,3 +292,41 @@ fn sender_changes_task() {
...
+    assert!(rx.is_empty(), "channel IS empty after value is read");
+}
+
+#[test]
+fn receiver_is_empty_rx_close() {
+    let (_tx, mut rx) = oneshot::channel::<i32>();
+    assert!(rx.is_empty());
+    rx.close();
+    assert!(rx.is_empty());
+}
```
maminrayej, line 332: You could also add a test for `is_empty` before and after `try_recv` as well.

maminrayej (commented):
If I understand correctly, it's not safe to poll the receiver even if `is_empty` returns `true`, since you can consume the message with `try_recv` and incorrectly try to poll the receiver later on. I wonder whether it's worth it to document this fact on the `is_empty` method and redirect users to `is_terminated` implemented in #7152.

tokio/src/sync/oneshot.rs
cratelyn: that's a great idea, thank you! 87a0070 applies this change.

tokio/tests/sync_oneshot.rs
cratelyn: this additional test coverage is introduced in abcff8f

cratelyn:
> If I understand correctly, it's not safe to poll the receiver even if `is_empty` returns `true`, since you can consume the message with `try_recv` and incorrectly try to poll the receiver later on. I wonder whether it's worth it to document this fact on the `is_empty` method and redirect users to `is_terminated` implemented in #7152.

that's right, and i agree that this would be a good signpost to provide in order to mitigate misuse and clarify purpose.

note that this branch is now based upon #7152, to reflect that the documentation of `is_empty()` now references an interface proposed in that pull request.

b7d1b3ea introduces additional documentation and accompanying examples to help draw attention to what exactly "empty" means, semantically.

cratelyn:
85323ff addresses the same feedback as offered in #7152, [here](https://github.com/tokio-rs/tokio/pull/7152#discussion_r1957289039).

cratelyn:
this branch has been rebased, now that the prerequisite #7152 has been merged.

maminrayej:
Thanks.

PR #7155, opened 2025-02-13
mpsc::peek

srxg:
This PR introduces a `peek` method to the `mpsc::Receiver`. This allows consumers to inspect the next message in the queue without removing it, enabling use cases that require lookahead behavior. The `recv` method has been modified to utilise peek: if a message is available, `recv` reads the value, advances the internal queue, and updates the semaphore accordingly.

### Motivation
Currently, the `mpsc::Receiver` does not provide a way to inspect the next message in the queue without consuming it. Users may want to make decisions based on the upcoming message before actually removing it from the channel.

### Solution
The peek method retrieves the next message without modifying the queue state. It works by:

1. Calling try_peeking_ahead() to determine the correct block to peek from, handling block transitions if needed.
2. Returning the peeked value via block.peek(self.index).
3. block.peek(slot_index) checks if the slot at slot_index is ready using ready_offset_read. If ready, it returns a reference to the value without modifying the queue.

### DIFF
--- a/tokio-util/Cargo.toml
+++ b/tokio-util/Cargo.toml
@@ -35,7 +35,7 @@ __docs_rs = ["futures-util"]
 
 [dependencies]
 tokio = { version = "1.28.0", path = "../tokio", features = ["sync"] }
-bytes = "1.0.0"
+bytes = "1.2.1"
 futures-core = "0.3.0"
 futures-sink = "0.3.0"
 futures-io = { version = "0.3.0", optional = true }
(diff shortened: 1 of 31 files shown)


srxg:
Although the tests pass, there is unexpected behavior in the `contention/bounded_full` benchmark in `benches/sync_mpsc.rs`. Occasionally, tx.send() fails with `SendError`, and the sender reports as closed. However, `recv()` never appears to return None. Further investigation is needed.

PR #7159, opened 2025-02-15
docs: fix nesting of next sections under examples

edev:
Promote the "Feature flags" and "Supported platforms" sections out from under "Examples", as they are not examples. Adjust their subsections accordingly. Expose these subsections via navigation sidebar.

## Motivation

I believe the documentation subheadings under "Examples" might have been nested by mistake. It seems like they were meant to follow "Examples", and I suspect that no one noticed this mistake because the default page styling makes the nesting structure hard to see.

This matters for at least two reasons:
1. Accessibility, e.g. to users of screen readers.
2. The "Sections" navigation menu in the sidebar only renders `#` and `##` headings. The headings "Unstable features", "`WASM` support", and "Unstable `WASM` support" are absent from this menu, and I believe they were meant to be present.

## Solution

Promote each subsection under `Examples` by one level.

### DIFF
--- a/tokio/src/lib.rs
+++ b/tokio/src/lib.rs
@@ -302,7 +302,7 @@
 //! }
 //! ```
 //!
-//! ## Feature flags
+//! # Feature flags
 //!
 //! Tokio uses a set of [feature flags] to reduce the amount of compiled code. It
 //! is possible to just enable certain features over others. By default, Tokio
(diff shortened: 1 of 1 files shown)


maminrayej:
Using heading level 2 is consistent with "Working With Tasks", "CPU-bound tasks and blocking code", and "Asynchronous IO". I guess "Examples" is using heading level 1 since it's an overall example of using tokio.

Darksonn (approved):
I guess those are part of the "tour of Tokio"

scr

This layout makes sense to me, thanks.

PR #7161, opened 2025-02-15
sync: add `OnceCell::wait_initialized` method

wyfo:
## Motivation

Such a method was mentioned in #7151 as nice to have, so here it is.

## Solution

The previous `OnceCell` implementation was using a semaphore, but if it was convenient to synchronize initialization, it is not suited to wait for it. That's why the synchronization has completely been rewritten to handle it.

There is now 3 possible states: *unset*, *locked*, and *set*. The cell is initialized to *unset*. When it is initialized, the cell is first locked, i.e. its state set to *locked*, then the value is written, and finally the state is set to *set*. As the other synchronization primitives, it uses a mutex-protected linked list of waiters, which is drained when initialization succeeds. If initialization fails, the state is rolled back to *unset* and one awaiting initializer is wakened up. 

The new implementation takes one less word in memory, and it has been optimized for ZST case. In fact, another goal of this new implementation is to serve as base for the `Event` primitive of #7151, so basically a `OnceCell<()>`.
By the way, a proper tracing implementation has been added to `OnceCell`.

### DIFF
--- a/tokio/src/sync/once_cell.rs
+++ b/tokio/src/sync/once_cell.rs
@@ -452,39 +452,43 @@ impl<T> OnceCell<T> {
         }
     }
 
-    fn poll_wait(&self, cx: &mut Context<'_>, node: Pin<&mut Waiter>, queued: bool) -> Poll<u8> {
-        // Checks if there is no need to wait, using acquire ordering
-        // as value may be accessed after in case of success.
-        let state = self.state.load(Ordering::Acquire);
-        if state == STATE_SET || node.is_initializer && state == STATE_UNSET {
-            return Poll::Ready(state);
-        }
+    /// Wait for the state to have the required value.
+    /// If the node is pushed in the waiter
+    fn poll_wait(&self, cx: &mut Context<'_>, node: Pin<&mut Waiter>) -> Poll<u8> {
         // Clone the waker before locking, a waker clone can be triggering arbitrary code.
         let waker = cx.waker().clone();
         let mut waiters = self.waiters.lock();
-        // Checks the state, this time with the lock held, so the state access
-        // is synchronized by the lock. If the waiter has been notified, then the
-        // state is ensured to have been modified.
+        // SAFETY: node is not moved out of pinned reference,
+        // and it is accessed with the lock held
+        let node = unsafe { node.get_unchecked_mut() };
+        // Checks the state with the lock held; if the waiter has been notified,
+        // then the state is ensured to have been modified.
+        // Uses acquire ordering as value may be accessed after in case of success.
         let state = self.state.load(Ordering::Acquire);
         if state == STATE_SET || node.is_initializer && state == STATE_UNSET {
-            return Poll::Ready(state);
-        }
-        // SAFETY: node is not moved out of pinned reference
-        unsafe {
-            let node = node.get_unchecked_mut();
-            // waker is only accessed with the waiters lock acquired
-            node.waker = Some(waker);
-            // If the waiter is not already in the wait queue, enqueue it.
-            if !queued {
+            // If there was a waker, it means the node was queued,
+            // so it is removed from the list.
+            if node.waker.take().is_some() {
+                // SAFETY: The node is contained in the list.
+                unsafe { waiters.remove(NonNull::from(node)) };
+            }
+            Poll::Ready(state)
+        } else {
+            // If there was no waker, it means the node was not queued,
+            // so it is pushed in the list.
+            if node.waker.replace(waker).is_none() {
                 waiters.push_front(NonNull::from(node));
             }
+            Poll::Pending
         }
-        Poll::Pending
     }
 
     /// Waits for the `OnceCell` to be initialized, and returns a reference
     /// to the value stored.
     pub async fn wait_initialized(&self) -> &T {
+        if let Some(value) = self.get() {
+            return value;
+        }
         let state = WaitFuture::new(self, false).await;
         debug_assert_eq!(state, STATE_SET);
         // SAFETY: the cell has been initialized
@@ -630,10 +634,14 @@ unsafe impl<T: Sync + Send> Sync for OnceCell<T> {}
 unsafe impl<T: Send> Send for OnceCell<T> {}
 
 /// An entry in the wait queue.
+///
+/// `Waiter` should be accessed with the wait queue lock acquired.
 struct Waiter {
     /// If the waiter is a cell initializer.
     is_initializer: bool,
-    /// Waiting task.
+    /// The waker of the awaiting task. A waker is present
+    /// if and only if the waiter is queued in the wait queue;
+    /// the invariant must be upheld when the queue lock is released.
     waker: Option<Waker>,
     /// Intrusive linked-list pointers.
     pointers: linked_list::Pointers<Waiter>,
@@ -683,7 +691,9 @@ generate_addr_of_methods! {
 struct WaitFuture<'a, T> {
     node: Waiter,
     cell: &'a OnceCell<T>,
-    queued: bool,
+    /// If `true`, the node *may* be queued in the wait queue.
+    /// If `false`, the node is not in the wait queue.
+    maybe_queued: bool,
 }
 
 impl<'a, T> WaitFuture<'a, T> {
@@ -695,7 +705,7 @@ impl<'a, T> WaitFuture<'a, T> {
         Self {
             cell,
             node: Waiter::new(is_initializer),
-            queued: false,
+            maybe_queued: false,
         }
     }
 
@@ -709,7 +719,7 @@ impl<'a, T> WaitFuture<'a, T> {
             (
                 Pin::new_unchecked(&mut this.node),
                 this.cell,
-                &mut this.queued,
+                &mut this.maybe_queued,
             )
         }
     }
@@ -717,9 +727,8 @@ impl<'a, T> WaitFuture<'a, T> {
 
 impl<T> Drop for WaitFuture<'_, T> {
     fn drop(&mut self) {
-        // If the future has not been queued, there is no node in the wait list, so we
-        // can skip acquiring the lock.
-        if !self.queued {
+        // If the node is not in the queue, we can skip acquiring the lock.
+        if !self.maybe_queued {
             return;
         }
         // This is where we ensure safety. The future is being dropped,
@@ -738,16 +747,17 @@ impl<T> Future for WaitFuture<'_, T> {
     type Output = u8;
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        let (node, cell, queued) = self.project();
+        let (node, cell, maybe_queued) = self.project();
         let coop = ready!(crate::task::coop::poll_proceed(cx));
 
-        match cell.poll_wait(cx, node, *queued) {
+        match cell.poll_wait(cx, node) {
             Poll::Ready(state) => {
                 coop.made_progress();
+                *maybe_queued = true;
                 Poll::Ready(state)
             }
             Poll::Pending => {
-                *queued = true;
+                *maybe_queued = false;
                 Poll::Pending
             }
         }
--- a/tokio/src/sync/once_cell.rs
+++ b/tokio/src/sync/once_cell.rs
@@ -466,7 +466,7 @@ impl<T> OnceCell<T> {
         // Uses acquire ordering as value may be accessed after in case of success.
         let state = self.state.load(Ordering::Acquire);
         if state == STATE_SET || node.is_initializer && state == STATE_UNSET {
-            // If there was a waker, it means the node was queued,
+            // If there is a waker, it means the node is queued,
             // so it is removed from the list.
             if node.waker.take().is_some() {
                 // SAFETY: The node is contained in the list.
@@ -474,7 +474,7 @@ impl<T> OnceCell<T> {
             }
             Poll::Ready(state)
         } else {
-            // If there was no waker, it means the node was not queued,
+            // If there is no waker, it means the node is not queued,
             // so it is pushed in the list.
             if node.waker.replace(waker).is_none() {
                 waiters.push_front(NonNull::from(node));
--- a/tokio/src/sync/once_cell.rs
+++ b/tokio/src/sync/once_cell.rs
@@ -15,6 +15,7 @@ use std::ptr;
 use std::ptr::NonNull;
 use std::sync::atomic::Ordering;
 use std::task::{ready, Context, Poll, Waker};
+
 // This file contains an implementation of an OnceCell. The principle
 // behind the safety of the cell is that any thread with an `&OnceCell` may
 // access the `value` field according the following rules:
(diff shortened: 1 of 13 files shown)


tokio/src/sync/once_cell.rs
```diff
@@ -1,13 +1,20 @@
...
+use std::convert::Infallible;
 use std::error::Error;
 use std::fmt;
 use std::future::Future;
+use std::marker::PhantomPinned;
 use std::mem::MaybeUninit;
 use std::ops::Drop;
+use std::pin::Pin;
 use std::ptr;
-use std::sync::atomic::{AtomicBool, Ordering};
```
Darksonn, line 9: The fact that this used to use atomics without going through `crate::loom` tells me that the `OnceCell` is not loom-tested at all. An implementation as complex as this will require the addition of [loom tests](https://github.com/tokio-rs/tokio/tree/master/tokio/src/sync/tests).

Darksonn (commented):
Thanks for the PR. I agree that this approach is viable, but it significantly increases the complexity. Be prepared that this will take a while to review.

tokio/src/sync/once_cell.rs
wyfo: I'm working on it right now, and found a non-trivial synchronization issue that I'm fixing. Will push the tests soon.

wyfo: It should be good now. As usual, thanks to `loom` for reminding me how hard correct synchronization is.

tokio/src/sync/once_cell.rs
```diff
@@ -68,11 +76,18 @@ use std::sync::atomic::{AtomicBool, Ordering};
...
-    semaphore: Semaphore,
+    waiters: Mutex<LinkedList<Waiter, <Waiter as linked_list::Link>::Target>>,
+    /// The current number of available permits in the semaphore.
+    #[cfg(all(tokio_unstable, feature = "tracing"))]
+    resource_span: tracing::Span,
 }
 
+const STATE_UNSET: u8 = 0;
+const STATE_LOCKED: u8 = 1;
+const STATE_SET: u8 = 2;
```
Darksonn, line 89: If you swap these, then a ZST oncecell will only have states 0 and 1 instead of states 0 and 2. That seems nicer.
```suggestion
const STATE_UNSET: u8 = 0;
const STATE_LOCKED: u8 = 2;
const STATE_SET: u8 = 1;
```

tokio/src/sync/once_cell.rs
```diff
@@ -304,30 +370,141 @@ impl<T> OnceCell<T> {
...
+    /// `MaybeUninit::assume_init` is indeed always safe and valid for ZSTs.
+    fn set_zst(&self, value: T) -> Result<(), SetError<T>> {
+        assert_eq!(std::mem::size_of::<T>(), 0);
+        // There is no memory to synchronize with ZSTs, hence relaxed ordering.
+        match self.state.compare_exchange(
+            STATE_UNSET,
+            STATE_SET,
+            Ordering::Relaxed,
+            Ordering::Relaxed,
+        ) {
```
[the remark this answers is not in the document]
wyfo: Fair enough, I will fix it.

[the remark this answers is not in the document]
wyfo: Done in https://github.com/tokio-rs/tokio/pull/7161/commits/3571889dd95d3f8a2c231a0c80ad7c21f7a40445

tokio/src/sync/once_cell.rs
```diff
@@ -1,12 +1,20 @@
...
 use std::mem::MaybeUninit;
 use std::ops::Drop;
+use std::pin::Pin;
 use std::ptr;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::ptr::NonNull;
+use std::sync::atomic::Ordering;
+use std::task::{ready, Context, Poll, Waker};
 
 // This file contains an implementation of an OnceCell. The principle
```
wyfo, line 19: I forgot to update the internal documentation, working on it.

tokio/src/sync/once_cell.rs
```diff
@@ -464,6 +629,131 @@ unsafe impl<T: Sync + Send> Sync for OnceCell<T> {}
...
+        }
+    }
+}
+
+/// Wait for the cell to be initialized, or the state to be unlocked
+/// in the case of an initializer.
+struct WaitFuture<'a, T> {
+    node: Waiter,
+    cell: &'a OnceCell<T>,
+    queued: bool,
```
Darksonn, line 686: This `queued` field is just for the optimization in `Drop` to skip the mutex? I don't think you can actually ever make use of that optimization as currently implemented, so you can just drop the field.

tokio/src/sync/once_cell.rs
```diff
@@ -304,30 +352,144 @@ impl<T> OnceCell<T> {
...
+        }
+        // Try to lock the state. Relaxed ordering is enough for success,
+        // as the value is written but not read after. In case of failure,
+        // acquire ordering is used to have happens-before the successful
+        // initialization.
+        match self.state.compare_exchange(
+            STATE_UNSET,
+            STATE_LOCKED,
+            Ordering::Relaxed,
+            Ordering::Acquire,
```
Darksonn, line 367: I would expect the code below to be okay:
```rs
static mut GLOBAL: i32 = 0;
static GLOBAL_INIT: OnceCell<()> = OnceCell::new();
```
Imagine you have 10 threads do this in parallel:
```rs
let is_leader = cell.get_or_try_init(|| unsafe {
    GLOBAL += 1;
    if GLOBAL == 10 {
    	Ok(())
    } else {
    	Err(())
    }
}).is_ok();

if is_leader {
    assert_eq!(unsafe { GLOBAL }, 10);
}
```
But we need `Acquire` on `UNSET`->`LOCKED` and `Release` on `LOCKED`->`SET` for the above to be synchronized correctly. With the current relaxed orderings, two different initializers would have a data race on `GLOBAL`.

tokio/src/sync/once_cell.rs
```diff
@@ -304,30 +352,144 @@ impl<T> OnceCell<T> {
...
+            Ok(_) => {
+                #[cfg(all(tokio_unstable, feature = "tracing"))]
+                self.resource_span.in_scope(|| {
+                    tracing::trace!(
+                        target: "runtime::resource::state_update",
+                        state = "set",
+                    );
+                });
+                // Forget the value, as `T` could implement `Drop`.
+                std::mem::forget(value);
```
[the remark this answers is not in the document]
wyfo: Good catch!

tokio/src/sync/once_cell.rs
wyfo: Could develop about why the fields would never been used?
If `wait_initialized` is called while the cell is already initialized, `poll_wait` will return `Poll::Ready` and `queued` field will never be set to `true`.
Tracing the early return of `drop` in tests shows that it is executed.

tokio/src/sync/once_cell.rs
wyfo: I would not have thought to this use case ... To make it work, the memory order in the `DropGuard` should also be changed to `Release`. And in the same spirit, success memory order to the first `set` CAS should also be `Acquire`, to have an happens-before relation with potential previously failed `get_or_try_init`.

tokio/src/sync/once_cell.rs
Darksonn: I guess you are right.

tokio/src/sync/once_cell.rs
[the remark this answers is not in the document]
wyfo: Fixed in https://github.com/tokio-rs/tokio/pull/7161/commits/9a17bcd093e7e572bcd13ce5d11143898afd62bd

tokio/src/sync/once_cell.rs
wyfo: Fixed in https://github.com/tokio-rs/tokio/pull/7161/commits/7a79f29d90876f87d059f6176f737c2876c41657

wyfo:
https://github.com/tokio-rs/tokio/pull/7161/commits/cc8c284e06ddecdc9897e0ec792848ddaca7b42a fixes a bug of a waiter not being able to be enqueued a second time.
In the commit, I've clarify the invariant of `Waiter`: there is a waker if and only if the node is queued, and I use it to remove the node from the queue when the future is ready, or to (re)queue it when pending.
I've also removed the first state check in `poll_wait` without the lock acquired. In fact, saving a lock here would prevent the `maybe_queued` flag to be updated, so dropping the future would have to lock, so there was no gain in the end.

There was no loom test catching the bug, but it found it while writing another one about happens-before between initialization attempts. I'm still working on it, so I will push it after.

tokio/src/sync/once_cell.rs
wyfo: I've added a loom test for this use case in https://github.com/tokio-rs/tokio/pull/7161/commits/33754c2bf3e2d9f58da8c15f96693d4b53364d96

tokio/src/sync/once_cell.rs
```diff
@@ -182,11 +191,39 @@ impl<T> OnceCell<T> {
...
+                loc.col = location.column(),
+            )
+        };
+        #[cfg(all(tokio_unstable, feature = "tracing"))]
+        resource_span.in_scope(|| {
+            tracing::trace!(
+                target: "runtime::resource::state_update",
+                state = "unset",
+            );
+        });
```
Darksonn, line 216: Although it's nice to add tracing support, this PR is already complex enough that I would prefer to add it in a follow-up PR.

wyfo: I've removed it in https://github.com/tokio-rs/tokio/pull/7161/commits/7409540229dede17601e4480b60ef5af7e5dc84c

tokio/src/sync/once_cell.rs
```diff
@@ -304,32 +315,135 @@ impl<T> OnceCell<T> {
...
+        // then the state is ensured to have been modified.
+        // Uses acquire ordering as value may be accessed after in case of success.
+        let state = self.state.load(Ordering::Acquire);
+        if state == STATE_SET || node.is_initializer && state == STATE_UNSET {
+            // If there is a waker, it means the node is queued,
+            // so it is removed from the list.
+            if node.waker.take().is_some() {
+                // SAFETY: The node is contained in the list.
+                unsafe { waiters.remove(NonNull::from(node)) };
+            }
```
Darksonn, line 423: You don't need this condition, you can just call `remove` unconditionally. It's okay to call `remove` if the `node` is not in any list afair.

tokio/src/sync/once_cell.rs
```diff
@@ -304,32 +315,135 @@ impl<T> OnceCell<T> {
...
+            Poll::Ready(state)
+        } else {
+            // If there is no waker, it means the node is not queued,
+            // so it is pushed in the list.
+            if node.waker.replace(waker).is_none() {
```
Darksonn, line 428: Here (and above) you run the destructor of the waker with the lock held.

tokio/src/sync/once_cell.rs
```diff
@@ -304,32 +315,135 @@ impl<T> OnceCell<T> {
...
+    }
+
+    /// Notify one initializer task.
+    fn notify_unlocked(&self) {
+        let mut waiters = self.waiters.lock();
+        if let Some(mut waiter) = waiters.drain_filter(|w| w.is_initializer).next() {
+            // SAFETY: we hold the lock, so we can access the waker.
+            let waker = unsafe { waiter.as_mut().waker.take() };
+            drop(waiters);
+            waker.unwrap().wake();
```
Darksonn, line 400: Hmm .. this could be slow if there are many waiting for completion. It's O(n) every time we fail, which could become O(n^2) with many failures.

tokio/src/sync/once_cell.rs
wyfo: Fixed in https://github.com/tokio-rs/tokio/pull/7161/commits/cac4ee0ca2354d2fa30c041ea00dca72da80418f

tokio/src/sync/once_cell.rs
wyfo: Done in https://github.com/tokio-rs/tokio/pull/7161/commits/cac4ee0ca2354d2fa30c041ea00dca72da80418f

tokio/src/sync/once_cell.rs
wyfo: Indeed, but it only happens when `get_or[_try]_init` and `wait_initialized` are mixed together. The current cell doesn't provide `wait_initialized`, so existing workflows using it should not be impacted. 
I would expect `wait_initialized` to be more used in combination with `set` (as in the `Event` primitive from #7151), so this kind of workflow would also not be impacted.

To address the issue of the "mixed" workflow, we could consider the current solution as an acceptable tradeoff, at least to begin with, for the reasons above (no regression). Otherwise, i see several solutions:
- add a second linked list (behind the same lock), simple but would increase the size of the struct by 2 words, so one more than the current struct, which seems still acceptable
- use a single linked list, but keep an additional node reference, let's name it "middle node", which would be the first "initializer" node;
  * initializer node are pushed to the front, setting the middle node if it is unset
  * `wait_initialized` node are pushed to the back
  * awaking a initializer removes the middle node from the list, the node on its left becoming the new middle node
  * awaking all nodes just drain the list
this implementation is obviously more complex, but reduce the additional size to 1 word only (same as current)
- another slightly more complex implementation would be to use the state as tagged pointer to store the "middle node" above, so there would no need to increase the struct size. I admit I like this solution, but that will be for another time.

By the way, I've noticed that my use of `drain_filter` is bad, because it breaks the fairness of the initialization locking, as it starts from the front, while I'm using `push_front` (as there is no `push_back`). If we keep a single linked list, I will then write a new method to pop back with a predicate.

Darksonn: Hmm. If we were okay with breaking the fairness, we could push initializers to the front and waiters to the back. This would mean that we could just pop the first element of the list to find an initializer. But we probably should not break fairness. Can you think of any similar tricks that don't break fairness?

Also, it sounds like we're missing a test for the fairness.

wyfo: Yes, we should preserve fairness. But I don't have any other tricks than the ones I described above, with a "middle node". Do you think I should try it? Or go with two lists? Or keep it like this for now, just fixing fairness?

Darksonn: To be honest ... I think we should just introduce a new separate type rather than try to extend `OnceCell`. By having one type that only provides `set` and `wait`, and another type that only provides `get_or[_try]_init`, this problem just disappears.

Or does your use-case actually need a single type with both features?

wyfo: Thank you for your honesty, I understand. My use case doesn't need both, so I can close this one and open a new PR with a new `Event<T>` primitive using a `Notify` internally, as it was the initial plan. 
I got a little bit too much enthusiastic opening this PR, motivated by the memory size reduction and by the alignment with `OnceLock`, which does provide `wait`, but it was obviously harder than I expected. Thank you a lot for your review by the way, it was a really good experience for me on a topic as complex as memory ordering.

So, if you agree, let's close the PR. I will still open a new one to add loom tests to `OnceCell`, to not have written them for nothing. Then, if you also agree, I will open one for `Event`.

Darksonn: Sounds good, and thanks for the PR!

PR #7166, opened 2025-02-20
net: debug_assert on creating a tokio socket from a blocking one

Noah-Kennedy:
See #5595 and #7172.

This adds a debug assertion that checks that a supplied underlying std socket is set to nonblocking mode when constructing a tokio socket object from such an object.

This only works on unix.

### DIFF
--- a/tokio/src/net/tcp/listener.rs
+++ b/tokio/src/net/tcp/listener.rs
@@ -210,7 +210,7 @@ impl TcpListener {
     /// will block the thread, which will cause unexpected behavior.
     /// Non-blocking mode can be set using [`set_nonblocking`].
     ///
-    /// Passing a listener in blocking mode is always errornous, and the behavior in that case may change in the future. For example, it could panic.
+    /// Passing a listener in blocking mode is always erroneous, and the behavior in that case may change in the future. For example, it could panic.
     ///
     /// [`set_nonblocking`]: std::net::TcpListener::set_nonblocking
     ///
--- a/tokio/src/net/tcp/listener.rs
+++ b/tokio/src/net/tcp/listener.rs
@@ -9,6 +9,7 @@ use std::fmt;
 use std::io;
 use std::net::{self, SocketAddr};
 use std::task::{ready, Context, Poll};
+use crate::util::blocking_check::check_socket_for_blocking;
 
 cfg_net! {
     /// A TCP socket server, listening for connections.
@@ -208,6 +209,8 @@ impl TcpListener {
     /// non-blocking mode. Otherwise all I/O operations on the listener
     /// will block the thread, which will cause unexpected behavior.
     /// Non-blocking mode can be set using [`set_nonblocking`].
+    /// 
+    /// Tokio's handling of blocking sockets may change in the future.
     ///
     /// [`set_nonblocking`]: std::net::TcpListener::set_nonblocking
     ///
@@ -236,6 +239,8 @@ impl TcpListener {
     /// explicitly with [`Runtime::enter`](crate::runtime::Runtime::enter) function.
     #[track_caller]
     pub fn from_std(listener: net::TcpListener) -> io::Result<TcpListener> {
+        check_socket_for_blocking(&listener)?;
+        
         let io = mio::net::TcpListener::from_std(listener);
         let io = PollEvented::new(io)?;
         Ok(TcpListener { io })
--- a/tokio/src/net/tcp/listener.rs
+++ b/tokio/src/net/tcp/listener.rs
@@ -210,7 +210,9 @@ impl TcpListener {
     /// will block the thread, which will cause unexpected behavior.
     /// Non-blocking mode can be set using [`set_nonblocking`].
     ///
-    /// Passing a listener in blocking mode is always erroneous, and the behavior in that case may change in the future. For example, it could panic.
+    /// Passing a listener in blocking mode is always erroneous,
+    /// and the behavior in that case may change in the future.
+    /// For example, it could panic.
     ///
     /// [`set_nonblocking`]: std::net::TcpListener::set_nonblocking
     ///
--- a/tokio/src/net/tcp/listener.rs
+++ b/tokio/src/net/tcp/listener.rs
@@ -1,11 +1,11 @@
 use crate::io::{Interest, PollEvented};
 use crate::net::tcp::TcpStream;
+use crate::util::check_socket_for_blocking;
 
 cfg_not_wasi! {
     use crate::net::{to_socket_addrs, ToSocketAddrs};
 }
 
-use crate::util::blocking_check::check_socket_for_blocking;
 use std::fmt;
 use std::io;
 use std::net::{self, SocketAddr};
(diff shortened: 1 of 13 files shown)


Noah-Kennedy:
Windows does not support this, so it only works on unix for now.

tokio/src/util/blocking_check.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[cfg(unix)]
+#[allow(unused_variables)]
+pub(crate) fn check_socket_for_blocking<S: AsFd>(s: &S) -> crate::io::Result<()> {
+    #[cfg(debug_assertions)]
+    {
+        let sock = socket2::SockRef::from(s);
+
+        if !sock.nonblocking()? {
+            eprintln!("Warning: registering a blocking socket, this may be a bug!");
```
jeff-hiner, line 12: Is it possible/likely that users are already using stderr for other things? Emitting stderr from within a library feels kinda gross; it might be easier to spot if we just panic here.

Noah-Kennedy: I actually feel similarly, the request to do it this way specifically came from @carllerche.

I'll leave it up to him to voice why he wants this approach.

Ten0 (approved):
Thanks for getting this to move forward!

Is `from_std_set_nonblocking`/`from_std_assume_nonblocking` [still](https://github.com/tokio-rs/tokio/pull/5597#issuecomment-1671414799) the long term plan?
image 
Or will it instead be something where tokio sets it to nonblocking?

Noah-Kennedy:
> Thanks for getting this to move forward!
> 
> Is `from_std_set_nonblocking`/`from_std_assume_nonblocking` [still](https://github.com/tokio-rs/tokio/pull/5597#issuecomment-1671414799) the long term plan? image Or will it instead be something where tokio sets it to nonblocking, [like axum_server currently does](https://github.com/programatik29/axum-server/blob/master/src/server.rs#L250)? Or will the panic be considered to be enough?

I think there's a bit that needs figured out still about what we do going forward. The trouble with the prior suggestion is that we actually still have the issue of `TryFrom` to figure out - we can't exactly deprecate that, and we need to figure out whether what it's behavior will be.

I think our two long-term options here are going to be to either panic, or set the socket option silently, and there's good arguments to be made for both. There should also be something along the lines of `from_std_unchecked` or `from_std_assume_nonblocking`, of course.

Given that we either need to leave the footgun in `TryFrom` or make a behavioral change anyways, I don't think leaving `from_std` as is makes much sense.

tokio/src/util/mod.rs
```diff
@@ -10,6 +10,9 @@ pub(crate) mod metric_atomics;
 #[cfg(any(feature = "rt", feature = "signal", feature = "process"))]
 pub(crate) mod once_cell;
 
+#[cfg(feature = "net")]
+pub(crate) mod blocking_check;
```
Darksonn, line 14: Consider re-exporting `check_socket_for_blocking` here to make the imports more nice.

tokio/src/util/blocking_check.rs
Darksonn, line 12: The warning should say that it comes from Tokio. Perhaps include `Location::caller` via `#[track_caller]` to print the caller location.

tokio/src/net/tcp/listener.rs
```diff
@@ -209,6 +210,8 @@ impl TcpListener {
     /// will block the thread, which will cause unexpected behavior.
     /// Non-blocking mode can be set using [`set_nonblocking`].
     ///
+    /// Tokio's handling of blocking sockets may change in the future.
```
Darksonn, line 213: Can we improve this wording? I don't think it very clearly explains the situation. You could say something along the lines of "Passing a listener in blocking mode is always errornous, and the behavior in that case may change in the future. For example, it could panic."

Darksonn (commented):
This also needs review from Carl before it can be merged.

tokio/src/util/blocking_check.rs
```diff
@@ -0,0 +1,25 @@
+#[cfg(unix)]
+use std::os::fd::AsFd;
+
+#[cfg(unix)]
+#[allow(unused_variables)]
+pub(crate) fn check_socket_for_blocking<S: AsFd>(s: &S) -> crate::io::Result<()> {
```
carllerche, line 6: Why does this return `io::Result`? It doesn't seem like the fn needs a return argument.

tokio/src/net/tcp/listener.rs
```diff
@@ -5,6 +5,7 @@ cfg_not_wasi! {
     use crate::net::{to_socket_addrs, ToSocketAddrs};
 }
 
+use crate::util::blocking_check::check_socket_for_blocking;
```
carllerche, line 8: Very minor and not a needed change to merge, but IMO this check makes more sense in the `net` module.

tokio/src/util/blocking_check.rs
carllerche: I spent some more time thinking about this question. I think that I am OK starting with a `debug_assert`. I would like there to be a way to disable the assertion (in case a user has a legitimate case) **, and** the panic message links to an issue for the user to find out how to disable the panic **and** report their legitimate use case.

To disable the panic, we can use a `tokio_allow_from_blocking_fd` `cfg` flag or something similar. The linked issue can include steps for setting it.

cc/ @Darksonn wdyt?

carllerche (changes requested):

tokio/src/util/blocking_check.rs
Darksonn: Using a debug assertion with a `--cfg` to disable it seems preferable to the `eprintln!`.

tokio/src/util/blocking_check.rs
Noah-Kennedy: If the nonblocking check fails, propating the error seems the most straightforward response.

Noah-Kennedy:
@carllerche @Darksonn I have updated the PR and addressed the feedback. This is ready for re-review.

Darksonn (approved):

Noah-Kennedy:
from offline discussion with @Darksonn: need to wrap the lines on the docs to not be super long

this is good to merge after that is done

tokio/src/util/blocking_check.rs
```diff
@@ -0,0 +1,29 @@
...
+    #[cfg(not(tokio_allow_from_blocking_fd))]
+    {
+        let sock = socket2::SockRef::from(s);
+
+        debug_assert!(
+            sock.nonblocking()?,
+            "Registering a blocking socket with the tokio runtime is unsupported. \
+            If you wish to do anyways, please add `--cfg tokio_allow_from_blocking_fd` to your \
+            RUSTFLAGS. See 7172 for more details."
```
carllerche, line 16:
```suggestion
            RUSTFLAGS. See #7172 for more details (github.com/tokio-rs/tokio/issues/7172)."
```

How about including the link for those who don't realize 7172 references an issue?

carllerche (changes requested):
Small proposal for the panic message.

tokio/src/util/blocking_check.rs
carllerche: Many modern terminals will make URLs clickable as well.

carllerche: I don't think the `https://` is needed. It should end up getting parsed without it and default to https.

carllerche (approved):
LGTM thanks

PR #7170, opened 2025-02-23
Optimize and Refactor Benchmark Code

NeoByteXx:
for _ in 0..(NUM_WORKERS) { → for _ in 0..NUM_WORKERS {
Improves readability and follows Rust conventions.

Fix loop condition in Spinner (signal_benchmark.rs)

if self.count > 3 { → if self.count >= 3 {
Ensures correct iteration count.


maminrayej:
> if self.count > 3 { → if self.count >= 3 {
Ensures correct iteration count.

I don't understand why the previous iteration count was wrong and what this change fixes. Could you explain the motivation behind this change?

PR #7186, opened 2025-02-27
Consistently use worker_threads instead of core_threads

aaamourao:
## Motivation

The builder parameter `core_threads` was renamed to `worker_threads` but it wasn't consistently renamed inside the code, what might be confusing.

## Solution

Renamed `core_threads` local variables to `worker_threads` in `builder.rs`.


Darksonn (approved):
Thank you.

PR #7188, opened 2025-03-02
util: optimize buffer reserve for AnyDelimiterCodec::encode

dlzht:
reserve enough capacity for `chunk` and `self.sequence_writer`

related issue: #7169


maminrayej:
Thanks.

maminrayej (approved):

PR #7195, opened 2025-03-05
Implement Extend for JoinSet

elichai:
## Motivation

`JoinSet` Implements `FromIterator` it's only natural that it also implements `Extend`

## Solution

Implement `Extend` for `JoinSet`


martin-g (approved):

tokio/src/task/join_set.rs
```diff
@@ -656,6 +656,8 @@ where
 /// # Examples
 ///
 /// ```
+/// # #[cfg(not(target_family = "wasm"))]
+/// # {
```
Darksonn, line 660: Use ! to avoid block:
```suggestion
/// # #![cfg(not(target_family = "wasm"))]
```

tokio/src/task/join_set.rs
```diff
@@ -649,6 +649,51 @@ where
...
+///     while let Some(res) = set.join_next().await {
+///         let idx = res.unwrap();
+///         seen[idx] = true;
+///     }
+///
+///     for i in 0..10 {
+///         assert!(seen[i]);
+///     }
+/// }
+/// # }
```
martin-g:
```suggestion
```

tokio/src/task/join_set.rs
martin-g: it seems `wasm-pack` does not like it. 
`cargo test --doc` passes OK locally.
I don't find any other usage of `/// # #![cfg(...)]` in the codebase.

martin-g: I reverted the last two commits (git reset --hard) to fix the build.

PR #7208, opened 2025-03-10
Remove an old custom `OnceCell` implementation in favor of `std`

a1phyr:
This is not needed anymore, our MSRV is high enough to use `std::sync::OnceLock`.


Darksonn (approved):
Thanks.

PR #7212, opened 2025-03-11
test: Helpful panic messages for tokio_test::io

vi:
Introduce `tokio_test::io::Builder::name` to configure name of the mock object, to include in panic messages.

Also show number of remaining actions or action index in some cases to help debugging failed tests.

## Motivation

When using `tokio_test::io::Mock`, it is tricky to debug problems, especially if multiple mocks are used. For example, if it panics with "unexpected write", it is hard to determine which of the mock objects faced the unexpected write and at what position of the scenario.

## Solution

`Builder` gained `name` method to set make `Mock` remember a string.

Panic messages are adjusted to include that name (if set), also some additional information (like the number of remaining actions) to aid hack-try-print-repeat-based debugging sessions.

Before:

    unexpected write

After:

    unexpected write (name my_mock_name, 0 actions remain)

### DIFF
--- a/tokio-test/src/io.rs
+++ b/tokio-test/src/io.rs
@@ -130,8 +130,8 @@ impl Builder {
     }
 
     /// Set name of the mock IO object to include in panic messages and debug output
-    pub fn name(&mut self, name: String) -> &mut Self {
-        self.name = name;
+    pub fn name(&mut self, name: impl Into<String>) -> &mut Self {
+        self.name = name.into();
         self
     }
 
--- a/tokio-test/src/io.rs
+++ b/tokio-test/src/io.rs
@@ -52,6 +52,7 @@ pub struct Handle {
 pub struct Builder {
     // Sequence of actions for the Mock to take
     actions: VecDeque<Action>,
+    name: String,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


tokio-test/src/io.rs
```diff
@@ -127,6 +129,12 @@ impl Builder {
         self
     }
 
+    /// Set name of the mock IO object to include in panic messages and debug output
+    pub fn name(&mut self, name: String) -> &mut Self {
```
mox692, line 133: How about accepting `Into<String>` for more flexibility?

For example: https://github.com/tokio-rs/tokio/blob/8507e28f89916662d2f61af823993483169d912c/tokio/src/runtime/builder.rs#L488

vi: Adjusted.

mox692 (approved):
Looks good to me!

vi:
Shall I submit further additions to `io::Mock` (e.g. ability to ignore unexpected things after certain points or explicitly inject EOF in reading part while still expecting more writes to come) or it would better suit in a separate crate?

mox692:
If that feature can be implemented outside of tokio, then it would generally be better to implement it elsewhere.

vi:
Published forked `tokio_test::io` with multiple additional features: [`tokio-io-mock-fork`](https://crates.io/crates/tokio-io-mock-fork).

If desirable, I can prepare a pull request with some of the features to upstream `tokio-test`.

PR #7230, opened 2025-03-24
Mark runtime::Handle as unwind-safe

akonradi-signal:
This type is used internally by the `Runtime`, which is already marked unwind-safe. This allows using types that contain one or more `Handle`s with the `catch_unwind` function without resorting to `AssertUnwindSafe`.

Related to #4657


Darksonn (approved):
Thank you.

PR #7249, opened 2025-04-09
Feat(process): Implement `Command::spawn_with`

PaulDance:
## Motivation

#7238: under Windows, the spawn sometimes has to be configured in a specific way in order to achieve a specific behavior or to stay compatible with some other API. In our case, it is about spawning child processes as PPL when the parent (Rust) process currently runs as PPL.

In the standard library, it is currently supported through the nightly-only [`windows_process_extensions_raw_attribute`](https://doc.rust-lang.org/unstable-book/library-features/windows-process-extensions-raw-attribute.html) feature. It was previously implemented by configuring the `Command` object using its [`raw_attribute`](https://doc.rust-lang.org/1.84.0/std/os/windows/process/trait.CommandExt.html#tymethod.raw_attribute) extension method and then the regular [`spawn`](https://doc.rust-lang.org/stable/std/process/struct.Command.html#method.spawn) added the attributes on the Windows process when appropriate. This meant that simply creating and configuring a `std::process::Command` object was sufficient to spawn Tokio processes since there is [`impl From<std::process:Command> for tokio::process::Command`](https://docs.rs/tokio/latest/tokio/process/struct.Command.html#impl-From%3CCommand%3E-for-Command).

However, the unstable API has recently been changed and the attribute setting must now be explicitly performed at spawn time through a separate spawning method: [`Command::spawn_with_attributes`](https://doc.rust-lang.org/stable/std/os/windows/process/trait.CommandExt.html#tymethod.spawn_with_attributes). This means that the previous solution does not work anymore.

## Solution

Since nightly APIs cannot currently be consumed from Tokio in any way, including when put behind some sort of cfg gate such as `tokio_unstable`, this proposes to add a broader API that enables customizing what is used in order to actually perform the synchronous process spawn while letting the rest stay as is.

This makes for a relatively small change and straightforward API addition. In particular, it allows keeping the nightly-only code outside of Tokio and more generally any kind of desired spawn customization, although not strictly necessary for the initial use case.

However, it could be considered too vague for such a niche use case while it could be replaced with a more specific `spawn_with_attributes` when the upstream feature is stabilized, even though it could take a long time, but this is the best comprise found for now. This is why the linked issue is not marked here as to be closed: it circumvents the problem at hand, but does not more directly solve it.

## Testing

No tests covering the new API have been added. Indeed, the standard library's side has landed a bit too recently (https://github.com/rust-lang/rust/pull/123604, 1.85) it seems compared to the nightly distribution seen in CI so the Windows-specific part could be tested, while there does not seem to be nightly-only tests yet either. Also, considering the current [`spawn`](https://docs.rs/tokio/latest/tokio/process/struct.Command.html#method.spawn) API has been internally rewritten in terms similar to `spawn_with`'s, the current tests should already cover this adequately.

### DIFF
--- a/tokio/src/process/mod.rs
+++ b/tokio/src/process/mod.rs
@@ -249,7 +249,7 @@ use std::future::Future;
 use std::io;
 use std::path::Path;
 use std::pin::Pin;
-use std::process::{Command as StdCommand, ExitStatus, Output, Stdio};
+use std::process::{Child as StdChild, Command as StdCommand, ExitStatus, Output, Stdio};
 use std::task::{ready, Context, Poll};
 
 #[cfg(unix)]
@@ -860,8 +860,98 @@ impl Command {
     /// On Unix platforms this method will fail with `std::io::ErrorKind::WouldBlock`
     /// if the system process limit is reached (which includes other applications
     /// running on the system).
+    #[inline]
     pub fn spawn(&mut self) -> io::Result<Child> {
-        imp::spawn_child(&mut self.std).map(|spawned_child| Child {
+        // On two lines to circumvent a mutable borrow check failure.
+        let child = self.std.spawn()?;
+        self.build_child(child)
+    }
+
+    /// Executes the command as a child process with a custom spawning function,
+    /// returning a handle to it.
+    ///
+    /// This is identical to [`Self::spawn`] in every aspect except the spawn:
+    /// here, it is customizable through the `with` parameter instead of
+    /// defaulting to the usual spawn. In fact, [`Self::spawn`] is just
+    /// [`Self::spawn_with`] with [`StdCommand::spawn`].
+    ///
+    /// This is useful mostly under Windows for now, since the platform exposes
+    /// special APIs to configure child processes when spawning them with various
+    /// attributes that customize the exact behavior of the spawn operation.
+    ///
+    /// # Examples
+    ///
+    /// Basic usage:
+    ///
+    /// ```no_run
+    /// # async fn test() { // allow using await
+    /// use std::process::Stdio;
+    ///
+    /// let output = tokio::process::Command::new("ls")
+    ///     .stdin(Stdio::null())
+    ///     .stdout(Stdio::piped())
+    ///     .stderr(Stdio::piped())
+    ///     .spawn_with(std::process::Command::spawn)
+    ///     .unwrap()
+    ///     .wait_with_output()
+    ///     .await
+    ///     .unwrap();
+    /// # }
+    /// ```
+    ///
+    /// Actually customizing the spawn under Windows:
+    ///
+    /// ```ignore
+    /// #![feature(windows_process_extensions_raw_attribute)]
+    /// # #[cfg(windows)]   // Windows-only nightly APIs are used here.
+    /// # async fn test() { // Allow using await.
+    /// use std::os::windows::process::{CommandExt, ProcThreadAttributeList};
+    /// use std::process::Stdio;
+    /// use tokio::process::Command;
+    ///
+    /// let parent = Command::new("cmd").spawn().unwrap();
+    /// let parent_process_handle = parent.raw_handle();
+    ///
+    /// const PROC_THREAD_ATTRIBUTE_PARENT_PROCESS: usize = 0x00020000;
+    /// let attribute_list = ProcThreadAttributeList::build()
+    ///     .attribute(PROC_THREAD_ATTRIBUTE_PARENT_PROCESS, &parent_process_handle)
+    ///     .finish()
+    ///     .unwrap();
+    ///
+    /// let _output = Command::new("ls")
+    ///     .stdin(Stdio::null())
+    ///     .stdout(Stdio::piped())
+    ///     .stderr(Stdio::piped())
+    ///     .spawn_with(|cmd| cmd.spawn_with_attributes(&attribute_list))
+    ///     .unwrap()
+    ///     .wait_with_output()
+    ///     .await
+    ///     .unwrap();
+    /// # }
+    /// ```
+    #[cfg(tokio_unstable)]
+    #[cfg_attr(docsrs, doc(cfg(tokio_unstable)))]
+    #[inline]
+    pub fn spawn_with(
+        &mut self,
+        with: impl Fn(&mut StdCommand) -> io::Result<StdChild>,
+    ) -> io::Result<Child> {
+        // On two lines to circumvent a mutable borrow check failure.
+        let child = with(&mut self.std)?;
+        self.build_child(child)
+    }
+
+    /// Small indirection for the spawn implementations.
+    ///
+    /// This is introduced for [`Self::spawn`] and [`Self::spawn_with`] to use:
+    /// [`Self::spawn`] cannot depend directly on on [`Self::spawn_with`] since
+    /// it is behind `tokio_unstable`. It also serves as a way to reduce
+    /// monomorphization bloat by taking in an already-spawned child process
+    /// instead of a command and custom spawn function.
+    fn build_child(&self, child: StdChild) -> io::Result<Child> {
+        let spawned_child = imp::build_child(child)?;
+
+        Ok(Child {
             child: FusedChild::Child(ChildDropGuard {
                 inner: spawned_child.child,
                 kill_on_drop: self.kill_on_drop,
--- a/tokio/src/process/unix/mod.rs
+++ b/tokio/src/process/unix/mod.rs
@@ -115,8 +115,7 @@ impl fmt::Debug for Child {
     }
 }
 
-pub(crate) fn spawn_child(cmd: &mut std::process::Command) -> io::Result<SpawnedChild> {
-    let mut child = cmd.spawn()?;
+pub(crate) fn build_child(mut child: StdChild) -> io::Result<SpawnedChild> {
     let stdin = child.stdin.take().map(stdio).transpose()?;
     let stdout = child.stdout.take().map(stdio).transpose()?;
     let stderr = child.stderr.take().map(stdio).transpose()?;
--- a/tokio/src/process/windows.rs
+++ b/tokio/src/process/windows.rs
@@ -27,7 +27,7 @@ use std::io;
 use std::os::windows::prelude::{AsRawHandle, IntoRawHandle, OwnedHandle, RawHandle};
 use std::pin::Pin;
 use std::process::Stdio;
-use std::process::{Child as StdChild, Command as StdCommand, ExitStatus};
+use std::process::{Child as StdChild, ExitStatus};
 use std::sync::Arc;
 use std::task::{Context, Poll};
 
@@ -66,8 +66,7 @@ struct Waiting {
 unsafe impl Sync for Waiting {}
 unsafe impl Send for Waiting {}
 
-pub(crate) fn spawn_child(cmd: &mut StdCommand) -> io::Result<SpawnedChild> {
-    let mut child = cmd.spawn()?;
+pub(crate) fn build_child(mut child: StdChild) -> io::Result<SpawnedChild> {
     let stdin = child.stdin.take().map(stdio).transpose()?;
     let stdout = child.stdout.take().map(stdio).transpose()?;
     let stderr = child.stderr.take().map(stdio).transpose()?;
--- a/spellcheck.dic
+++ b/spellcheck.dic
@@ -1,4 +1,4 @@
-300
+302
 &
 +
 <
@@ -70,6 +70,7 @@ connectionless
 coroutines
 cpu
 cpus
+customizable
 Customizable
 datagram
 Datagram
@@ -169,6 +170,7 @@ mio's
 miri
 misconfigured
 mock's
+monomorphization
 mpmc
 mpsc
 multi


Darksonn:
cc @ipetkov

tokio/src/process/mod.rs
```diff
@@ -860,8 +860,78 @@ impl Command {
...
+    ///         .unwrap()
+    ///         .wait_with_output()
+    ///         .await
+    ///         .unwrap();
+    /// # }
+    /// ```
+    pub fn spawn_with(
+        &mut self,
+        with: impl Fn(&mut StdCommand) -> io::Result<StdChild>,
+    ) -> io::Result<Child> {
```
Darksonn, line 933: We should probably gate this method behind `--cfg tokio_unstable`. Once `spawn_with_attributes` is stabilized, we will want to get rid of it again, but we can't do that if we stabilize this method.

PaulDance: Yes, I thought about this, but since `spawn` was made to internally call it for simplification, if the feature is disabled then it cannot depend on it. Should we let it stay fully duplicated or a cfg-gated mix?

PaulDance: I went ahead and pushed the fully duplicated version of simplicity.

PaulDance: Ah yes, I didn't think about this due to the cfg gate. It seems better in every way, so even though the duplication is acceptable, I'll still do this.

tokio/src/process/unix/mod.rs
```diff
@@ -115,8 +115,11 @@ impl fmt::Debug for Child {
     }
 }
 
-pub(crate) fn spawn_child(cmd: &mut std::process::Command) -> io::Result<SpawnedChild> {
-    let mut child = cmd.spawn()?;
+pub(crate) fn spawn_child_with(
+    cmd: &mut StdCommand,
+    with: impl Fn(&mut StdCommand) -> io::Result<StdChild>,
+) -> io::Result<SpawnedChild> {
+    let mut child = with(cmd)?;
```
ipetkov, line 122: Might be worth reworking this slightly to avoid monomorphization bloat (basically the entire rest of the function body does not need to be re-monormorphized for every unique `with` that is passed in):

Something like:

```rust
pub(crate) fn spawn_child_with(
    cmd: &mut StdCommand,
    with: impl Fn(&mut StdCommand) -> io::Result<StdChild>,
) -> io::Result<SpawnedChild> {
    spawn_child(&mut with(cmd)?)
}

fn spawn_child(cmd: &mut StdCommand) -> io::Result<SpawnedChild> {
  // ...
}
```

tokio/src/process/windows.rs
```diff
@@ -66,8 +66,11 @@ struct Waiting {
 unsafe impl Sync for Waiting {}
 unsafe impl Send for Waiting {}
 
-pub(crate) fn spawn_child(cmd: &mut StdCommand) -> io::Result<SpawnedChild> {
-    let mut child = cmd.spawn()?;
+pub(crate) fn spawn_child_with(
```
ipetkov, line 69: Same comment (and same trick) to cut down on the code bloat

tokio/src/process/mod.rs
```diff
@@ -860,8 +860,90 @@ impl Command {
...
+    }
+
+    /// Small indirection for the implementation of [`Self::spawn`] and
+    /// [`Self::spawn_with`] to use: [`Self::spawn`] cannot depend directly on
+    /// on [`Self::spawn_with`] since it is behind `tokio_unstable`.
+    fn spawn_with_inner(
+        &mut self,
+        with: impl Fn(&mut StdCommand) -> io::Result<StdChild>,
+    ) -> io::Result<Child> {
+        imp::spawn_child_with(&mut self.std, with).map(|spawned_child| Child {
```
ipetkov, line 946: Might be worth also moving the body of `map` into a helper function to avoid the generics bloat

ipetkov (approved):
Overall the changes seem sensible! Just have a few minor comments around cutting down unnecessary monomorphization

PaulDance:
Something like this?

tokio/src/process/unix/mod.rs
PaulDance: Here, you probably meant that the helper function takes an `std::process::Child` in since `with: impl Fn(&mut StdCommand) -> io::Result<StdChild>`, right? This is what I've used in the proposed update.

tokio/src/process/mod.rs
PaulDance: I went ahead and removed the `map` entirely for safety with regards to monomorphization bloat here.

tokio/src/process/mod.rs
```diff
@@ -860,8 +860,97 @@ impl Command {
...
+    ///         .await
+    ///         .unwrap();
+    /// # }
+    /// ```
+    #[cfg(tokio_unstable)]
+    #[inline]
+    pub fn spawn_with(
+        &mut self,
+        with: impl Fn(&mut StdCommand) -> io::Result<StdChild>,
+    ) -> io::Result<Child> {
```
Darksonn, line 937: This is missing the "Available on `tokio_unstable` only" marker in the docs.
```rs
```

PaulDance: Woops, thanks! Done.

ipetkov (approved):

tokio/src/process/mod.rs
```diff
@@ -860,8 +860,98 @@ impl Command {
...
+    ///         .unwrap()
+    ///         .wait_with_output()
+    ///         .await
+    ///         .unwrap();
+    /// # }
+    /// ```
+    ///
+    /// Actually customizing the spawn under Windows:
+    ///
+    /// ```ignore
```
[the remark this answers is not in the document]
PaulDance: Not sure: are the doc tests compiled with nightly?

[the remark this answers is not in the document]
Darksonn: Ah, we probably don't use nightly to run tests. Please confirm that you have successfully compiled the test locally with a sufficiently-recent nightly toolchain and `no_run`. Then we can leave it as `ignore`.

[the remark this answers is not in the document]
PaulDance: Just checked and if I read https://github.com/tokio-rs/tokio/blob/c3037adac90fbbf92a160b970f38e50b66936b4c/.github/workflows/ci.yml#L808 correctly, then it does not run using nightly, so unless a few changes are done there, which could be considered a bit overkill only for this single API, I don't think it would be possible currently, right?

tokio/src/process/mod.rs
```diff
@@ -860,8 +860,98 @@ impl Command {
...
+    ///
+    /// let output = Command::new("ls")
+    ///         .stdin(Stdio::null())
+    ///         .stdout(Stdio::piped())
+    ///         .stderr(Stdio::piped())
+    ///         .spawn_with(|cmd| cmd.spawn_with_attributes(&attribute_list))
+    ///         .unwrap()
+    ///         .wait_with_output()
+    ///         .await
+    ///         .unwrap();
```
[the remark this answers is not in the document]
PaulDance: Woops, this is copy-pasta. You're right, done.

tokio/src/process/mod.rs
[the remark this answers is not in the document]
PaulDance: Yes, I read the CI configuration wrong, so thought nightly was not yet used for the docs, but it actually is, which seems to work in our case: https://github.com/tokio-rs/tokio/actions/runs/14490788847/job/40646574822?pr=7249.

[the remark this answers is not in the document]
Darksonn: I believe that's only for generating the html docs (which does require nightly). I would expect that *running* docs tests still happens on stable.

Anyway, to clarify, I meant that I just wanted you to test it locally on your machine once.

(But if it works in CI, even better.)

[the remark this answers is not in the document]
PaulDance: Indeed, it actually failed later on.

In order to test this locally, I copied the example to a temporary test:

```diff
diff --git a/tokio/tests/process_smoke.rs b/tokio/tests/process_smoke.rs
index e7fec6d6..e0106e69 100644
--- a/tokio/tests/process_smoke.rs
+++ b/tokio/tests/process_smoke.rs
@@ -1,3 +1,4 @@
+#![cfg_attr(windows, feature(windows_process_extensions_raw_attribute))]
 #![warn(rust_2018_idioms)]
 #![cfg(all(feature = "full", not(target_os = "wasi"), not(miri)))] // Wasi cannot run system commands
 
@@ -32,3 +33,30 @@ async fn simple() {
     assert_eq!(child.id(), None);
     drop(child.kill());
 }
+
+#[cfg(windows)]
+#[tokio::test]
+async fn test_cmd_spawn_with() {
+    use std::os::windows::process::{CommandExt, ProcThreadAttributeList};
+    use std::process::Stdio;
+    use tokio::process::Command;
+
+    let parent = Command::new("cmd").spawn().unwrap();
+    let parent_process_handle = parent.raw_handle();
+
+    const PROC_THREAD_ATTRIBUTE_PARENT_PROCESS: usize = 0x00020000;
+    let attribute_list = ProcThreadAttributeList::build()
+        .attribute(PROC_THREAD_ATTRIBUTE_PARENT_PROCESS, &parent_process_handle)
+        .finish()
+        .unwrap();
+
+    let _output = Command::new("ls")
+        .stdin(Stdio::null())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn_with(|cmd| cmd.spawn_with_attributes(&attribute_list))
+        .unwrap()
+        .wait_with_output()
+        .await
+        .unwrap();
+}
```

and then:

```console
$ RUSTFLAGS='--cfg tokio_unstable' cargo +nightly check --tests --all-features --target x86_64-pc-windows-msvc -p tokio
    Checking tokio v1.44.2 (/home/pauldance/git/pkgs/tokio/tokio)
    Checking tokio-stream v0.1.17 (/home/pauldance/git/pkgs/tokio/tokio-stream)
    Checking tokio-test v0.4.4 (/home/pauldance/git/pkgs/tokio/tokio-test)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.61s
```

A few errors and warnings needed fixing, which I did in the latest patch.

tokio/src/process/mod.rs
```diff
@@ -860,8 +860,98 @@ impl Command {
...
+    ///     .wait_with_output()
+    ///     .await
+    ///     .unwrap();
+    /// # }
+    /// ```
+    ///
+    /// Actually customizing the spawn under Windows:
+    ///
+    /// ```ignore
+    /// # #![feature(windows_process_extensions_raw_attribute)]
```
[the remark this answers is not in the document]
PaulDance: Yes, done.

PaulDance:
Ah yes, I didn't know I had this inconsistency. I now disabled this proxying, so it should be better?

Darksonn:
Now it wants to use your hotmail rather than the harfanglab.fr address you put in the SoB.

PaulDance:
Eh, whatever, fine by me :shrug:

Darksonn:
Thanks for the PR!

PaulDance:
Many thanks!

piscisaureus:
This is very useful, thank you!
🙏 for this to become stable.

Darksonn:
This was added as unstable as a temporary measure until stdlib stabilizes `spawn_with_attributes`. Did you have a different use-case in mind?

Pistonight:
@Darksonn  I had a different use case for this: I wanted to take the stdout/stderr of a child and pipe it to another command like the example here: https://doc.rust-lang.org/std/process/struct.Stdio.html#impl-From%3CChildStdout%3E-for-Stdio

(Maybe there's another way to do it other than my approach below, please let me know)

I could take it from `tokio::process::ChildStdout`, but that involves converting it from sync to async and back to sync, which seems unnecessary:
```rust
#[tokio::main]
async fn main() {
    let hello = Command::new("echo")
        .arg("Hello, world!")
        .stdout(Stdio::piped())
        .spawn() // converted to nonblocking io here
        .expect("failed echo command");
    
    let std_io: Result<Stdio, _> = hello.stdout.unwrap().try_into(); // extra try_into() and unwrap() compared to the std example
    
    let reverse = Command::new("rev")
        .stdin(std_io.unwrap())
        .output().await.unwrap();
    // works, but not ideal
    assert_eq!(reverse.stdout, b"!dlrow ,olleH\n");
}
```

What I really want to do:
```rust
    let mut std_io: Option<Stdio> = None;
    let hello = Command::new("echo")
        .arg("Hello, world!")
        .stdout(Stdio::piped())
        .spawn_with(|cmd| {
            let mut child = cmd.spawn()?;
            std_io = child.stdout.take().map(Stdio::from); // take control of the stdout handle
          // ^ does not compile due to being inside Fn closure
            Ok(child)
        })
        .expect("failed echo command");
    
    let reverse = Command::new("rev")
        .stdin(std_io.take().unwrap()) // put it in directly without converting it to non-blocking
        .output().await.unwrap();
    
    assert_eq!(reverse.stdout, b"!dlrow ,olleH\n");
```
Note that this doesn't compile and currently has to be worked around with `RefCell` unless the closure is upgraded to `FnMut` or `FnOnce`, but it works once I do that

Darksonn:
If you have change requests, then please file a new issue. 

As for upgrading the closure to `FnOnce`, I would be happy to see a PR for that.

PR #7252, opened 2025-04-11
(tokio-util JoinMap) Remove raw-entry feature in favour of HashTable API.

conradludgate:
## Motivation

While rebasing #7075, I noticed the change from #7219 which updates hashbrown, requiring the addition of the `raw-entry` feature. That inspired me to take a look at the actual JoinMap impl and I thought it could be improved by using `HashTable` instead.

## Solution

Replacing the `tasks_by_key` map with a `HashTable`, we unlock a few benefits:
1. When searching by ID, we don't need to store the id in `tasks_by_key`, since we can fetch the ID from the `AbortHandle`.
2. We don't need to duplicate the `S: BuildHasher`, as we are in-control over the hash-function, we can fetch the existing hasher from `hashes_by_task`.

However, removing the ID from `tasks_by_key` will slow down those lookups (only in `join_next()`), as they now need a vtable lookup, rather than comparing against the id already in the current cache line.

---

I don't feel strongly about this PR, I'm just quite fond of the HashTable API and thought I'd try it here.

### DIFF
--- a/tokio-util/src/task/join_map.rs
+++ b/tokio-util/src/task/join_map.rs
@@ -1,5 +1,5 @@
-use hashbrown::hash_map::RawEntryMut;
-use hashbrown::HashMap;
+use hashbrown::hash_table::Entry;
+use hashbrown::{HashMap, HashTable};
 use std::borrow::Borrow;
 use std::collections::hash_map::RandomState;
 use std::fmt;
@@ -102,6 +102,9 @@ use tokio::task::{AbortHandle, Id, JoinError, JoinSet, LocalSet};
 /// [contains]: fn@Self::contains_key
 #[cfg_attr(docsrs, doc(cfg(all(feature = "rt", tokio_unstable))))]
 pub struct JoinMap<K, V, S = RandomState> {
+    /// The hasher to use with the following hashtables.
+    hash_builder: S,
+
     /// A map of the [`AbortHandle`]s of the tasks spawned on this `JoinMap`,
     /// indexed by their keys and task IDs.
     ///
@@ -109,7 +112,7 @@ pub struct JoinMap<K, V, S = RandomState> {
     /// spawning tasks, and the task's IDs. The IDs are stored here to resolve
     /// hash collisions when looking up tasks based on their pre-computed hash
     /// (as stored in the `hashes_by_task` map).
-    tasks_by_key: HashMap<Key<K>, AbortHandle, S>,
+    tasks_by_key: HashTable<(Key<K>, AbortHandle)>,
 
     /// A map from task IDs to the hash of the key associated with that task.
     ///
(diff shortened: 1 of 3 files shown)


tokio-util/src/task/join_map.rs
```diff
@@ -416,26 +396,26 @@ where
...
+                .entry(hash, |(k, _)| *k == key, |(k, _)| hash_one(hash_builder, k));
         match entry {
-            RawEntryMut::Occupied(mut occ) => {
+            Entry::Occupied(occ) => {
                 // There was a previous task spawned with the same key! Cancel
                 // that task, and remove its ID from the map of hashes by task IDs.
-                let Key { id: prev_id, .. } = occ.insert_key(map_key);
-                occ.insert(abort).abort();
-                let _prev_hash = self.hashes_by_task.remove(&prev_id);
+                let (_, abort) = std::mem::replace(occ.into_mut(), (key, abort));
```
Darksonn, line 411: This drops the key. If the destructor of the key panics, we don't call `abort` or adjust `hashes_by_task`.

tokio-util/src/task/join_map.rs
```diff
@@ -716,7 +698,9 @@ where
     #[inline]
     pub fn shrink_to_fit(&mut self) {
         self.hashes_by_task.shrink_to_fit();
-        self.tasks_by_key.shrink_to_fit();
+        let hash_builder = self.hashes_by_task.hasher();
+        self.tasks_by_key
+            .shrink_to_fit(|(k, _)| hash_one(hash_builder, k));
```
Darksonn, line 703: I'm confused. Why is `hashes_by_task` needed inside `tasks_by_key` here?

conradludgate: We only need the hasher inside `hashes_by_task`. It saves us from having to store a separate instance of the hasher.

An alternative solution could be to have a separate `hasher: S`, and then make the `HashMap<Id, u64>` use a cheaper hash function like a simple fibonacci hash (so no state is needed). That might be overkill though.

tokio-util/src/task/join_map.rs
conradludgate: Good catch

tokio-util/src/task/join_map.rs
conradludgate: For reference: https://github.com/conradludgate/tokio/commit/5086eeb183c7214e443c4897eac2be848ef1565b

tokio-util/tests/task_join_map.rs
```diff
@@ -343,3 +345,34 @@ async fn duplicate_keys2() {
 
     assert!(map.join_next().await.is_none());
 }
+
+#[cfg(not(target_arch = "wasm32"))]
```
Darksonn: There are downstream users of Tokio who compile with panic=abort in cases other than wasm.
```suggestion
#[cfg_attr(not(panic = "unwind"), ignore)]
```

Darksonn: Also I believe this fixes the unused import warnings since `#[cfg(ignore)]` just means the test doesn't run - it's still compiled and can use imports.

Issue #7250, opened 2025-04-10
Improve Consistency for Echo Example Programs


nskins:
**Problem**
I believe there is a slight issue with the presentation of the "echo" examples: namely, `echo.rs` and `echo-udp.rs`. These examples create servers that listen on TCP and UDP (respectively). However, they interact with a "client" program called `connect.rs` that handles the logic for sending both TCP and UDP payloads. I find it confusing that the client logic for TCP and UDP would be bundled into one file while the server logic would be split across two separate programs. It would be easier to understand if both the client and the server had a similar interface for the end user.

**Suggested Solution**
I see two ways of resolving this:
1. We could combine the `echo.rs` and `echo-udp.rs` examples into a single file (say, `echo.rs`) with a command-line argument to indicate which protocol to use (`connect.rs` does this with `--udp`). This might make more sense for maintainability (i.e. reusing common code).
2. We could keep the echo examples separate (as `echo-tcp.rs` and `echo-udp.rs`) and separate the client program `connect.rs` into `connect-tcp.rs` and `connect-udp.rs`. This might make more sense for ease of learning (since it keeps the separate pairs of examples simple).

Curious to hear what others think. I'd be happy to submit a PR if we determine a change should be made!

nskins:
Sounds good. I have started a [branch](https://github.com/nskins/tokio/tree/improve-echo-example-consistency) for this and should have a PR ready soon.

PR #7256, opened 2025-04-11
Examples: Improve Echo Example Consistency

nskins:
## Motivation

This addresses the suggestion brought up in https://github.com/tokio-rs/tokio/issues/7250.

## Solution

This PR splits up `connect.rs` into two separate programs: `connect-tcp.rs` and `connect-udp.rs`. This matches the echo examples which are split up by protocol already. This should improve readability and ease of understanding so end users only need to consider one protocol at a time.

Additionally, documentation referencing these new example names has been updated appropriately. I tested these changes by running all examples that interact with these files and verifying that the behavior is the same as before.

Closes: https://github.com/tokio-rs/tokio/issues/7250


examples/connect-tcp.rs
```diff
@@ -0,0 +1,77 @@
...
+                }
+            })
+            .map(Ok);
+
+        match future::join(sink.send_all(&mut stdin), stdout.send_all(&mut stream)).await {
+            (Err(e), _) | (_, Err(e)) => Err(e.into()),
+            _ => Ok(()),
+        }
+    }
+}
```
maminrayej, line 77: Now that the code for TCP and UDP are moved to separate files, I think it makes sense to remove the `mod tcp` and move the `use` statements to the top of the file.

examples/connect-udp.rs
```diff
@@ -46,50 +36,11 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-            .map(Ok);
-
-        match future::join(sink.send_all(&mut stdin), stdout.send_all(&mut stream)).await {
-            (Err(e), _) | (_, Err(e)) => Err(e.into()),
-            _ => Ok(()),
-        }
-    }
-}
-
 mod udp {
```
maminrayej, line 44: Same thing as `connect-tcp.rs` here.

examples/connect-tcp.rs
nskins: Good point! I made the requested changes, and the examples are definitely a bit more readable now.

examples/connect-udp.rs
nskins: Made the requested change.

maminrayej (approved):

maminrayej:
Thanks.

PR #7260, opened 2025-04-12
Benches: Add Helper Functions for Building Runtimes to `benches/spawn.rs`

nskins:
## Motivation

Other benchmarks build the runtime via the use of helper functions (for example `benches/sync_semaphore.rs` and `benches/fs.rs`). We could reduce duplicate code by doing the same in `benches/spawn.rs`.

## Solution

This PR adds helper functions for building single and multi-threaded runtimes in `benches/spawn.rs`.

### DIFF
--- a/benches/spawn.rs
+++ b/benches/spawn.rs
@@ -10,10 +10,21 @@ async fn work() -> usize {
     black_box(val)
 }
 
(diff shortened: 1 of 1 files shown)


maminrayej (approved):

maminrayej:
Thanks.

Issue #7270, opened 2025-04-16
Code comment inconsistency


YichiZhang0613:
In tokio-master/tokio/src/runtime/metrics/histogram/h2_histogram.rs, the code indicate max_error will be panic if < 0.0 or > 1.0, while the code will panic if max_error <= 0.0 or max_error >= 1.0, which will panic when max_error = 0.0 or 1.0 and that is inconsistent with comment. 
```rust
/// # Panics
    /// - `max_error` < 0
    /// - `max_error` > 1
    pub fn max_error(mut self, max_error: f64) -> Self {
        assert!(max_error > 0.0, "max_error must be greater than 0");
        assert!(max_error < 1.0, "max_error must be less than 1");
        ...
   }
```

mox692:
Feel free to open a PR for that.

PR #7273, opened 2025-04-17
Fix Code Comment Inconsistency

YichiZhang0613:
## Motivation

Solve code comment inconsistency problem in #7270.

## Solution

Fix comments to be consistent with code.

Closes https://github.com/tokio-rs/tokio/issues/7270.


mox692:
Thanks, can you update here as well?

https://github.com/tokio-rs/tokio/blob/7a6c424f6e07d79e7c029866ed601bb948aba10a/tokio/src/runtime/metrics/histogram/h2_histogram.rs#L201

PR #7274, opened 2025-04-17
Introduce `Timed` to track how long a future is alive for

narwhaljs:
This aims to add a simple feature that allows users to easily track how long a future was alive for.  It attempts to follow all standards similar to `Timeout`. It can be constructed manually through the `tokio::time::timed()` function, or through the extension trait allowing _any_future_ to call `any_future.timed()`. 
 

## Motivation

Are you tired of writing:
```
let now = std::time::Instant::now();
do_some_async_work().await;
let elapsed = now.elapsed();
println!("Finished some async work in {elapsed:?}");
```

## Solution

Now you can write:
```
let (_, elapsed) = do_some_async_work().timed().await;
println!("Finished some async work in {elapsed:?}");
```


mox692:
Hmm, I think other crates can provide similar features, so I'm not sure if we should include this in tokio.

narwhaljs:
> Hmm, I think other crates can provide similar features, so I'm not sure if we should include this in tokio.

Yeah i think futures is probably a better spot for it, or any other ideas? Or is there something that does this already? Fine to close this.

mox692:
I believe it's generally better not to include functionality in tokio that can be implemented in other crates.

ADD-SP:
I'm also not inclined to add this feature into tokio, This feature is more related to observability things, not the async runtime itself.

ADD-SP:
https://crates.io/crates/future-timing/0.1.0

There is a crate provides similar features.

I closed this PR based on our discussion, thanks for your contribution.

ADD-SP:
T

PR #7276, opened 2025-04-18
Stabilize worker_park_count and worker_unpark_count

Owen-CH-Leung:
## Motivation

Further to #6899 , this PR further stabilises 2 more API `worker_unpark_count` and `worker_park_count`.

## Solution

Move the API impl out of tokio_unstable flag

Ref: https://github.com/tokio-rs/tokio/issues/6546

### DIFF
--- a/tokio/src/runtime/metrics/batch.rs
+++ b/tokio/src/runtime/metrics/batch.rs
@@ -14,11 +14,9 @@ pub(crate) struct MetricsBatch {
     /// Instant at which work last resumed (continued after park).
     processing_scheduled_tasks_started_at: Instant,
 
-    #[cfg(tokio_unstable)]
     /// Number of times the worker parked.
     park_count: u64,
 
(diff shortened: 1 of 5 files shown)


hds (approved):
This looks good to me!

I think it's unfortunate that we're duplicating code between stable and unstable functions, but at least they're right next to one another, and I don't think that this PR is the place to rework the stable/unstable code split (I know we went over that a lot in previous PRs).

Owen-CH-Leung:
> This looks good to me!
> 
> I think it's unfortunate that we're duplicating code between stable and unstable functions, but at least they're right next to one another, and I don't think that this PR is the place to rework the stable/unstable code split (I know we went over that a lot in previous PRs).

Thanks! I agree about the code duplicaton. Once we fully stabilize the metrics API, the duplication issue should naturally resolve itself as we'll no longer need separate code paths for stable and unstable features

PR #7290, opened 2025-04-26
Issue 4665/tcp stream  shutdown on mac os on localhost yields error

soundofspace:


tokio/tests/tcp_shutdown.rs
```diff
@@ -26,4 +28,46 @@ async fn shutdown() {
...
+        let mut stream = assert_ok!(TcpStream::connect(&addr).await);
+        sleep(Duration::from_millis(2)).await;
+        assert_ok!(AsyncWriteExt::shutdown(&mut stream).await);
+    });
+
+    let (stream, _) = assert_ok!(srv.accept().await);
+    // By setting linger to 0 we will trigger a TCP reset
+    stream.set_linger(Some(Duration::new(0, 0))).unwrap();
+
+    sleep(Duration::from_millis(1)).await;
```
mox692, line 50: Is it possible to use a channel instead of `sleep` to avoid making this test flaky?

mox692 (commented):
Overall looks good to me.

PR #7291, opened 2025-04-26
streamlined some unsafe code

nevakrien:
## Motivation
I read some of the source code and I have seen a very weird construction of an uninit array.

## Solution

rust actually provides [new_uninit_slice](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.new_uninit_slice) and since unsafe cell is gurnteed to have the same layout as T the pointer cast kind of works out naturally.


nevakrien:
oh seems like u guys are on 2021 addition so I will need to fix my code to use layout directly which kinda sucks... 
nvm I dont think its worth the hassle at that point its harder to read

PR #7297, opened 2025-05-01
Tokio: issue-6374: update `inner.pos` for `tokio::fs::File`

Suryakant-Soni:


mox692:
You may want to refer to #6381 too, which tries to resolve the same issue.

Suryakant-Soni:
> You may want to refer to #6381 too, which tries to resolve the same issue.
Thanks a lot for your comment, Yes I went through the PR, will need review discussion, will write it up to start with 🙌

Suryakant-Soni:
Hi @Darksonn, @mox692  can you please help reviewing the PR, thanks.

tokio/src/fs/file.rs
```diff
@@ -113,7 +113,7 @@ enum State {
 #[derive(Debug)]
 enum Operation {
     Read(io::Result<usize>),
-    Write(io::Result<()>),
+    Write(io::Result<usize>),
```
Suryakant-Soni, line 116: changing Write enum variant from Write(io::Result<()>) to Write(io::Result<usize>) so that the pos usize value can be propagated out from spawn_blocking closure

tokio/src/fs/file.rs
```diff
@@ -759,8 +765,6 @@ impl AsyncWrite for File {
                     })?;
 
                     inner.state = State::Busy(blocking_task_join_handle);
-
-                    return Poll::Ready(Ok(n));
```
Suryakant-Soni, line 763: Returning Poll::Ready(Ok(n)) is now deferred to the second polling iteration, allowing inner.pos to be updated before the result is returned.

mox692:
I'll take a look in a few days!

mox692:
Sorry, I'm not quite following yet:

1. Could you elaborate what problem this PR is trying to address? Specifically, what problems were caused because the `inner.pos` was not updated?
2. Are there any tests that catch the problem? (i.e. the tests that would fail without this changes)

Suryakant-Soni:
> Sorry, I'm not quite following yet:
> 
> 1. Could you elaborate what problem this PR is trying to address? Specifically, what problems were caused because the `inner.pos` was not updated?
> 2. Are there any tests that catch the problem? (i.e. the tests that would fail without this changes)


### **What problem this PR addresses**
There are two distinct but related issues:

**### 1. `inner.pos` not being updated causes state inconsistency and represents a potential latent bug*

Currently, `tokio::fs::File` does not update` inner.pos `after a read or write operation completes. This means the internal cursor (inner.pos) can become out of sync with the actual OS file position. While the AsyncSeek trait can query the true file offset via a seek, doing so repeatedly is inefficient and breaks abstraction in cases when inner.pos is being assumed to be updated and and may break future designs where consistent file position tracking is essential (e.g. resumable I/O, file streaming, or caching strategies).
By not updating inner.pos, the internal state becomes inaccurate, leading to:

- Logical bugs if future operations assume pos is correct.
- Inefficiencies if users must explicitly seek after every I/O.

### **2. Poll::Ready(Ok(n)) returned too early — async contract violation**

The existing implementation may return `Poll::Ready(Ok(n)) ` from poll_write (and similar) before the `spawn_blocking` task has completed. This violates the async contract, because:

`Poll::Ready` must only be returned once the blocking task has completed and the result is available.
Returning early breaks async polling logic, and causes tests (and possibly real-world callers) to assume the operation has completed when it hasn't.

Testing these issues - 

1. testing of feature that depends on inner.pos
Currently `inner.pos` is not exposed, but if it were, it would be easy to test and verify that it's out of sync. Any consumer might reasonably assume that inner.pos is usable, as it represents an abstraction over the low-level file object.

2. for 2nd issue this is a [demo test](https://github.com/tokio-rs/tokio/pull/7331/files#diff-819a978ed02740f846c00a8d7cd0c30f1012ed9cd97c2a98005349aa1c67b5e0R982), `open_write_new ` demonstrates the failure if the correct async flow is enforced

Suryakant-Soni:
Hi @mox692 , @Darksonn , @carllerche , 
Any update or view on this, thanks.

Suryakant-Soni:
Thanks a lot @mox692  for taking time to provide your comments 🙌 ,
I understand - 
1. `tokio::fs::File`'s poll_seek internally delegates to std::fs::File::seek, so there shouldn't be any disparity between the api responses. While we could cache the position to avoid repeated syscalls, it seems not much necessary for any change at this point.👍 

2. I think [poll_write contract ](https://docs.rs/tokio/latest/tokio/io/trait.AsyncWrite.html#tymethod.poll_write) says that `On success, returns Poll::Ready(Ok(num_bytes_written)).`  
but I see  `poll::ready` is returned early in [poll_write](https://github.com/tokio-rs/tokio/blob/master/tokio/src/fs/file.rs#L763) irrespective of the result of spawning task, unlike [poll_read](https://github.com/tokio-rs/tokio/blob/master/tokio/src/fs/file.rs#L614) which waits for the task to complete before returning readiness

3. That said, when a subsequent async operation like flush is invoked, the previous write task is guaranteed to complete due to the `ready!(Pin::new(&mut rx).poll(cx))?` call. So practically, the system behaves correctly in most scenarios.

4. Let me know if I’m missing anything conceptually — I’d appreciate the clarification since I am new to the project 🙌 

5. Since I don’t currently have a concrete failing case or regression tied to this, I agree it makes sense to hold off on changes for now. We can always revisit if a stronger motivation emerges.

mox692:
I have the same thought. I assumed there was an problem mentioned in the original issue, but AFAICT the code looks fine.

Issue #7304, opened 2025-05-04
Feature request: Add `biased` support to `join!` and `try_join!`


jlizen:
**Is your feature request related to a problem? Please describe.**
I have two long-running futures that I need to drive locally in a task.

Specifically, the futures are a AWS Lambda function handler listener, and secondary event listener that does some extra cleanup operations after the main event processing is complete. Something along [these lines](https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/examples/extension-internal-flush/src/main.rs).

Normally I would reach for `join!` or `try_join!` for this. However, I don't actually want to randomize which future I poll first. I want to always poll my primary server if it is ready to make progress, over the 'cleanup' event listener. Context here is that an AWS Lambda only receives one request a time per node.

Instead, I want to invoke `join!` or `try_join!` in a way that always polls sequentially based on the order of futures in the macro body. 

**Describe the solution you'd like**
Tokio added `biased` support to `select!` some time back - https://github.com/tokio-rs/tokio/issues/2181

It would be great to have the same API for `join!` or `try_join!`

**Describe alternatives you've considered**
You can accomplish the same thing using `select!` in a loop, and `select!` does support `biased`.  

This gets very verbose though since you need to explicitly continue polling subsequent futures after one of them finishes. `join!` and `try_join!` are designed to be a friendlier APIs for driving multiple futures to completion in the same task, so I'd prefer to use them if possible to do so with `biased` polling.

maminrayej:
Adding `biased;` makes sense, but note that Tokio does not randomize among futures passed to `join!` and `try_join!`. It rotates which future is polled first, so for example if you have three futures, they will be polled like:
```
(f1, f2, f3) -> (f2, f3, f1) -> (f3, f1, f2) -> (f1, f2, f3) -> ...
```
adding `biased;` would effectively remove the rotation and keep the order as `(f1, f2, f3)`.

jlizen:
> Tokio does not randomize among futures passed to join! and try_join!. It rotates which future is polled first

Thanks for the clarification! Yes, I am interested in always polling in the order they are passed in, rather than cycling which is polled first per wake.

jlizen:
Oh, I didn't state it explicitly, but I'm open to doing the implementation work for this - sharing for feedback though on whether a contribution would be accepted.

maminrayej:
> ... whether a contribution would be accepted.

Of course.

PR #7307, opened 2025-05-05
feat(tokio/macros): add biased mode to join! and try_join!

jlizen:
Closes: #7304 

Adding a `biased;` mode to the `join!` and `try_join!` macros that always polls the wrapped futures in their order of declaration, rather than rotating each poll.

This parallels the `select!` API (though is simpler under the hood, since these macros rotate rather than randomize which branch is polled).

## Motivation

This is useful when the contained futures interact in a way where polling order is significant. 

My specific case is that I want to `join!` or `try_join!` two separate listener futures in an AWS Lambda. Lambdas receive one request at a time. One of my listeners is my primary function handler, which is higher priority. The second is an extension that does post-request 'telemetry flushing'. So, there is never a need for the second future to be polled first.

Here is a rough example: [ref](https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/examples/extension-internal-flush/src/main.rs).

Something 

## Solution

I added an optional `biased;` token to each macro's entrypoint. This resolves to a literal variable, `$rotate_poll_order`, as either `true` or `false` (with no `biased;` declaration resulting in `true`).

Then, inside the generated code, we only increment the skip counter used to rotate polling, if `$rotate_poll_order` is `true`.

I initially changed the skip order counter from `u32` to `Option<u32>`, but that would have resulted in a bit muddier code. Adding an if clause around the single line incrementing the skip counter, was much simpler.

The performance impact of this change on cases that do NOT use the `biased` mode is:
- one additional boolean stored on the stack
- one extra branch every poll (checking the boolean), which branch prediction should handle cleanly


jlizen:
Here's the PR related to our discussion yesterday @maminrayej - thanks for the guidance!

tokio/src/macros/try_join.rs
```diff
@@ -142,13 +196,18 @@ doc! {macro_rules! try_join {
...
+        // a different future will be polled first
         // to ensure every future passed to join! gets a chance to make progress even if
         // one of the futures consumes the whole budget.
         //
         // This is number of futures that will be skipped in the first loop
         // iteration the next time.
+        //
+        // If running in biased mode, this value will always be 0.
         let mut skip_next_time: u32 = 0;
+        let rotate_skip = $rotate_poll_order;
```
Darksonn, line 210: I would like to avoid having this variable at all when using biased. I don't think it can be optimized out because it gets stored as a field in the closure type.

tokio/tests/macros_try_join.rs
```diff
@@ -179,6 +179,37 @@ async fn a_different_future_is_polled_first_every_time_poll_fn_is_polled() {
...
+
+    tokio::join!(
+        biased;
+        fut(1, Arc::clone(&poll_order)),
+        fut(2, Arc::clone(&poll_order)),
+        fut(3, Arc::clone(&poll_order)),
+    );
+
+    // Each time the future created by join! is polled, it should start
+    // by polling a different future first.
```
Darksonn, line 206: Isn't this comment looks wrong. This is a biased `join!`, so it's *not* a different future each time.

Please make this test run both with and without `biased` to show both behaviors.

tokio/tests/macros_join.rs
Darksonn, line 1: Please include some tests for edge cases similar to the existing `empty_join` test at the bottom and the four `sync_*` tests at the top.

jlizen:
Appreciate the eyes, Darksonn@ - I'll take a stab at implementing the requested feedback.

tokio/tests/macros_join.rs
jlizen: my previous implementation didn't support usage of `biased;` in the empty input, but I can add that in and add the corresponding test

jlizen: My current approach is just to extend the additional tests with more test cases, to avoid exploding into a `*_biased` version of every test (with the exception of the ordering and size ones). But glad to split them.

tokio/src/macros/join.rs
```diff
@@ -68,6 +111,10 @@ doc! {macro_rules! join {
 #[cfg(not(doc))]
 doc! {macro_rules! join {
```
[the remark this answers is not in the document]
jlizen: Thanks, updating to:
```
    ($(biased;)? $($future:expr),*) => { unimplemented!() }
```

Let me know if you'd prefer me to tweak the formatting, `rustfmt` seems to be agnostic to it.

tokio/tests/macros_try_join.rs
Darksonn: I did not see that. If it's already there in a separate test, then that's fine.

tokio/src/macros/try_join.rs
jlizen: My newest commit removes this variable, along with `skip_next_time`, from the biased branch's `poll_fn()` (so it will now be smaller than the unbiased equivalent). 

It's a bit verbose... since each branch of `match $rotate_poll_order {}` needs to generate its own unnameable poll function. I initially was pulling out common logic into a closure, but I felt it was ultimately less readable.

Glad to take another stab at simplifying if you have any feedback.

jlizen: Ah, that's clever. Thanks for the advice! Will take a crack at it.

jlizen: Got it working, thanks!

jlizen:
@Darksonn latest refactor is up.

I noticed our existing `try_join!` poll sequencing test was actually using `join!`, I assume by accident. Updated that as well.

tokio/src/macros/support.rs
```diff
@@ -28,3 +28,34 @@ cfg_macros! {
 pub use std::future::{Future, IntoFuture};
 pub use std::pin::Pin;
 pub use std::task::{Context, Poll};
+
+#[doc(hidden)]
+#[derive(Default, Debug)]
+pub struct Rotator<const COUNT: u32> {
```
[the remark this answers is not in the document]
jlizen: Sure, makes sense. Will do.

tokio/src/macros/try_join.rs
```diff
@@ -100,6 +119,37 @@ macro_rules! doc {
...
+        /// # Ok(())
+        /// }
+        ///
+        /// #[tokio::main]
+        /// async fn main() {
+        ///     let res = tokio::try_join!(
+        ///         //do_stuff_async() will always be polled first when woken
+        ///         biased;
+        ///         do_stuff_async(),
+        ///         more_async_work());
```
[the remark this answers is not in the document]
jlizen: Will do on both counts. Sorry, was overly reliant on `rustfmt` here, will manually clean up.

tokio/tests/macros_try_join.rs
```diff
@@ -163,13 +171,15 @@ async fn a_different_future_is_polled_first_every_time_poll_fn_is_polled() {
 
             tokio::task::yield_now().await;
         }
+        Ok::<(), Infallible>(())
     };
 
-    tokio::join!(
+    tokio::try_join!(
```
jlizen, line 177: this was a mistake in existing test, i think

jlizen:
@Darksonn latest changes are up per your comments, thanks for your time on this PR review.

jlizen:
@Darksonn latest changes are up per your comments, thanks for your time on this PR review.

tokio/src/macros/join.rs
```diff
@@ -32,6 +32,25 @@ macro_rules! doc {
...
+        ///
+        /// You may want this if your futures may interact in a way where known polling order is significant.
+        ///
+        /// But there is an important caveat to this mode. It becomes your responsibility
+        /// to ensure that the polling order of your futures is fair. If for example you
+        /// are joining a stream and a shutdown future, and the stream has a
+        /// huge volume of messages that takes a long time to finish processing per poll, you should
+        /// place the shutdown future earlier in the `join!` list to ensure that it is
+        /// always polled, and will not be delayed due to the stream future taking a long time to return
+        /// `Poll::Pending`.
```
Darksonn, line 52: I don't think that a shutdown future is an amazing example here. That makes sense for `select!`, but you wouldn't really have a shutdown future in `join!`. Not sure what a better example would be, though.

jlizen: Here is a real-world example usage involving a shutdown signal: https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/lambda-runtime/src/lib.rs#L227

The distinction here is that the other task is a _server_ future, not a stream future, meaning it will generally keep running and generating new request futures. So we need to continue driving it. Even if the shutdown future will only ever come up once and then resolve, we still need to check it first every time.

Anyway I'll tweak wording from stream -> server and finesse it a bit more.

jlizen: Also renaming the section `Fairness` -> `Poll Ordering`, it's not really about fairness anymore.

jlizen: Actually, I changed it back. My initial change was basically restating that deterministic ordering is useful. But it's important for readers to consider the case that one future might take a long time to poll.

Based on the above lambda runtime link, I do think there are cases where you would be joining a shutdown future, so it's probably still relevant? Anyway I think it does a good job of illustrating the fairness risk?

Glad to keep thinking about it though if not sold and try to come up with another example.

tokio/src/macros/join.rs
```diff
@@ -21,7 +21,7 @@ macro_rules! doc {
         /// The supplied futures are stored inline and do not require allocating a
         /// `Vec`.
         ///
-        /// ### Runtime characteristics
+        /// ### Fairness
```
Darksonn: Note sure why this uses three hashes, but it's a sub-section of `Notes` so let's use two.
```suggestion
        /// ## Fairness
```

tokio/src/macros/join.rs
```diff
@@ -32,6 +32,25 @@ macro_rules! doc {
         ///
         /// [`tokio::spawn`]: crate::spawn
         ///
+        /// # Poll Ordering
```
Darksonn: Let's make this a sub-section together with fairness.
```suggestion
        /// ## Poll Ordering
```

tokio/src/macros/try_join.rs
```diff
@@ -30,6 +30,25 @@ macro_rules! doc {
         ///
         /// [`tokio::spawn`]: crate::spawn
         ///
+        /// # Fairness
```
Darksonn: Please make sure that the section names for `try_join!` match those we have for `join!`.

tokio/src/macros/join.rs
```diff
@@ -54,6 +73,30 @@ macro_rules! doc {
         ///     // do something with the values
         /// }
         /// ```
+        ///
+        ///  Using the `biased;` mode to control polling order.
```
Darksonn:
```suggestion
        /// Using the `biased;` mode to control polling order.
```

tokio/src/macros/join.rs
```diff
@@ -96,25 +143,21 @@ doc! {macro_rules! join {
...
         let mut futures = &mut futures;
 
-        // Each time the future created by poll_fn is polled, a different future will be polled first
+        const COUNT: u32 = $($total)*;
+
+        // Each time the future created by poll_fn is polled,
+        // if not running in biased mode,
+        // a different future will be polled first
         // to ensure every future passed to join! gets a chance to make progress even if
         // one of the futures consumes the whole budget.
```
Darksonn: Please reflow this text so that the line breaks are at 80 chars. (or whatever line length the rest of this file uses)

tokio/src/macros/join.rs
```diff
@@ -164,15 +207,51 @@ doc! {macro_rules! join {
...
-        $crate::join!(@{ () (0) } $($e,)*)
+        $crate::join!(@{ rotator=$crate::macros::support::Rotator<COUNT>; () (0) } $($e,)*)
     };
 
+    (biased;) => { async {}.await };
+
     () => { async {}.await }
 }}
+
+/// Rotates by one every [`Self::num_skip`] call up to COUNT - 1.
```
Darksonn:
```suggestion
/// Rotates by one each [`Self::num_skip`] call up to `COUNT - 1`.
```

tokio/src/macros/join.rs
```diff
@@ -164,15 +207,51 @@ doc! {macro_rules! join {
...
+#[derive(Default, Debug)]
+pub struct Rotator<const COUNT: u32> {
+    next: u32,
+}
+
+impl<const COUNT: u32> Rotator<COUNT> {
+    /// Rotates by one every [`Self::num_skip`] call up to COUNT - 1
```
Darksonn:
```suggestion
    /// Rotates by one each [`Self::num_skip`] call up to `COUNT - 1`.
```

tokio/src/macros/join.rs
```diff
@@ -164,15 +207,51 @@ doc! {macro_rules! join {
...
+        num_skip
+    }
+}
+
+/// [`Self::num_skip`] always returns 0.
+#[derive(Default, Debug)]
+pub struct BiasedRotator {}
+
+impl BiasedRotator {
+    /// Always returns 0
```
Darksonn:
```suggestion
    /// Always returns 0.
```

Darksonn (commented):
I'm okay with the docs. A few nits:

jlizen:
@Darksonn updated to match your suggestions, except that I tweaked to unify section titles with `select!`

tokio/src/macros/try_join.rs
```diff
@@ -30,6 +30,25 @@ macro_rules! doc {
...
+        /// macro usage. See the examples for details. This will cause `try_join` to poll
+        /// the futures in the order they appear from top to bottom.
+        ///
+        /// You may want this if your futures may interact in a way where known polling order is significant.
+        ///
+        /// But there is an important caveat to this mode. It becomes your responsibility
+        /// to ensure that the polling order of your futures is fair. If for example you
+        /// are joining a stream and a shutdown future, and the stream has a
+        /// huge volume of messages that takes a long time to finish processing per poll, you should
+        /// place the shutdown future earlier in the `join!` list to ensure that it is
```
Darksonn:
```suggestion
        /// place the shutdown future earlier in the `try_join!` list to ensure that it is
```

jlizen: Thanks, sorry to miss thay

Darksonn (approved):
Thanks.

PR #7329, opened 2025-05-12
time: eliminate `UnsafeCell` around the `TimerShared`

ADD-SP:
## Summary

This PR eliminates the unnecessary `UnsafeCell`  to tighten the risky area.

```bash
$ cargo bench --bench time_timeout
```

I ran the benchmark multiple times and the results were often conflicting, sometimes regressing and sometimes not.

Given that the time scale of the benchmark is in the 10ns range, I don't think this change has a significant performance impact.

## Background

`UnsafeCell` always expose the risk of data race, which brings extra mental burden while reading the source.

I drew a complicated flowchart to check data races.

timer_flow

After finish the above chart, I think the `UnsafeCell` is not needed for the current design of timer,
so I opened this PR to make it looks like more safe, and make it more convenient for others.

By the way, the following `unsafe` can also be eliminated by `pin_project`.
I will open another PR for this to keep the diff small in each PR.

https://github.com/tokio-rs/tokio/blob/bdd64cc9d3561115d125584484404e9d1f7d7cca/tokio/src/runtime/time/entry.rs#L535

### DIFF
--- a/tokio/src/runtime/time/entry.rs
+++ b/tokio/src/runtime/time/entry.rs
@@ -63,7 +63,6 @@ use crate::sync::AtomicWaker;
 use crate::time::Instant;
 use crate::util::linked_list;
 
-use std::cell::UnsafeCell as StdUnsafeCell;
 use std::task::{Context, Poll, Waker};
 use std::{marker::PhantomPinned, pin::Pin, ptr::NonNull};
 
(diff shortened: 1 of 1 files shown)


carllerche:
Thanks for the detailed PR. It has been a while since I looked at the timer code in depth, but the change looks correct to me. Since I am not fresh here, I will also let @Darksonn look as well before merging.

Darksonn (approved):
The reason this is okay is that fields of `TimerShared` have additional wrappers of `UnsafeCell` or use atomics, so even if we remove the cell here, there is a second cell in cases where it matters.

PR #7336, opened 2025-05-16
docs: socket.rs document of method recv_buffer_size

jscptman:
## Motivation

When I read the document of method "recv_buffer_size" I found this issue that this method shoud associate to "set_recv_buffer_size" not "set_send_buffer_size", so I think there is a describtion problem.

## Solution
change "set_send_buffer_size" -> "set_recv_buffer_size" description that in recv_buffer_size method document


Darksonn (approved):
Thank you.

PR #7340, opened 2025-05-17
runtime: eliminate unnecessary lfence while operating on `queue::Local<T>`

ADD-SP:
<del>*Blocked by #7339 which fixes the CI failures caused by the new compilation error message of the new stable toolchain.*</del>

## Motivation

`queue::Local<T>` is never been send to another thread (except while shutting down), the `tail` is only updated by the worker thread who owned this local queue, so any change to `tail` is immediately visible to the current thread.

This means the `Acquire` ordering is not required for loading `tail` while operating on the local queue.

### Unsync load

Since the lfence is not required for this scenario, the atomic loading was replaced by simple pointer deref.

### Remove all methods of the `queue::Inner<T>`

Because of the following reasons:

* All method on the `Inner<T>` can be implemented in differently way for `Local<T>` and `Steal<T>`
* Adding method like `len_local` or `unsync_len` creates possibilities for incorrect usage in the future which causes ABA problems in critical path.

I removed all methods on the `Inner<T>` and re-implement it for both `Local<T>` and `Steal<T>`.

### Remove the feature gate of `Steal<T>::len()`

The feature gate `cfg_unstable_metrics` is removed for `Steal<T>::len()` to avoid duplicating the same code.

This doesn't seem to break anything, but if it does please let me know and I'll switch to something else.

## Anything else?

Ideally, the `queue::Local<T>` should be marked as `!Send` to lower the risk that unintentionally send it to another thread, but it required the changes of the starting and shutdown logic, I will try to open another PR for this improvement.

### DIFF
--- a/tokio/src/runtime/scheduler/multi_thread/queue.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/queue.rs
@@ -107,12 +107,19 @@ pub(crate) fn local<T: 'static>() -> (Steal<T>, Local<T>) {
 impl<T> Local<T> {
     /// Returns the number of entries in the queue
     pub(crate) fn len(&self) -> usize {
-        self.inner.len() as usize
+        let (_, head) = unpack(self.inner.head.load(Acquire));
+        // safety: this is the **only** thread that updates this cell.
+        let tail = unsafe { self.inner.tail.unsync_load() };
+        len(head, tail)
     }
 
     /// How many tasks can be pushed into the queue
     pub(crate) fn remaining_slots(&self) -> usize {
-        self.inner.remaining_slots()
+        let (steal, _) = unpack(self.inner.head.load(Acquire));
+        // safety: this is the **only** thread that updates this cell.
+        let tail = unsafe { self.inner.tail.unsync_load() };
+
+        LOCAL_QUEUE_CAPACITY - len(steal, tail)
     }
 
     pub(crate) fn max_capacity(&self) -> usize {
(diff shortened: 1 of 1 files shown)


mox692:
I just merged https://github.com/tokio-rs/tokio/pull/7339, so merging master should fix the ci issue 👍

tokio/src/runtime/scheduler/multi_thread/queue.rs
```diff
@@ -107,12 +107,19 @@ pub(crate) fn local<T: 'static>() -> (Steal<T>, Local<T>) {
...
     }
 
     /// How many tasks can be pushed into the queue
     pub(crate) fn remaining_slots(&self) -> usize {
-        self.inner.remaining_slots()
+        let (steal, _) = unpack(self.inner.head.load(Acquire));
+        // safety: this is the **only** thread that updates this cell.
+        let tail = unsafe { self.inner.tail.unsync_load() };
+
+        LOCAL_QUEUE_CAPACITY - len(steal, tail)
```
Darksonn, line 122: Why not just `LOCAL_QUEUE_CAPACITY - self.len()`?

ADD-SP: pr

Thanks for your review, I might be mistaken, but I thought the slots that are being stolen was not available for pushing new task.

Darksonn: Oh I see now. I did not notice that it was using `steal` instead of `head` for the first parameter to `len`.

Toby-Shi-cloud:
I am learning lock-free structure right now. Currently I just read the implementation of this steal-queue.
This PR is great and eliminates unnecessary acquire atomic loads, but I still have a small question:
I wonder is there any reason to use unsync_load instead of relaxed atomic load, since in this implementation, the only use of unsync_load is to read `tail` immediately after using acquire atomic load to read `head` (which, I think, already prevents the optimization).

PR #7358, opened 2025-05-24
time&runtime: cumulative minor improvements

ADD-SP:
## Summary

Just a few minor improvements while reading the source.

*It would be convenient to enable the 'Hide whitespace' in diff view.*

### runtime: remove outdated clippy attributes

These `clippy` attributes were introduced by #6808 and it looks like they were forgotten to be removed before merging into master.

***

### time: eliminate unsafe blocks using `pin_project`

Follow-up of #7329.

Usually, the `Pin::get_unchecked_mut` can be replaced with `pin_project`, this reduce the unsafe area without sacrificing readability and performance.

*Note: The `#[doc = "..."]` attributes have been replaced with `//` due to the limitation of `pin_project_lite` (ref: https://github.com/taiki-e/pin-project-lite/issues/41).*

***

### time: remove outdated comments

These comments were introduced by the [initial commit](https://github.com/tokio-rs/tokio/commit/f052f85315936a685f36c99dbceae0de541e0e4e) of #3080, but not be removed in the final version before merging it into the master.

***

### time: add comment to `TimerEntry::is_elapsed` for better readability

https://github.com/tokio-rs/tokio/blob/d7d4f7d08bf75daf23bec676b03cb4c88993e5d0/tokio/src/runtime/time/entry.rs#L502-L508

The logic here doesn't seem very intuitive, and I paused for a while while reading to understand it, so I've added some comments to make try to make it a bit more obvious.

### DIFF
--- a/tokio/src/runtime/local_runtime/runtime.rs
+++ b/tokio/src/runtime/local_runtime/runtime.rs
@@ -376,7 +376,6 @@ impl LocalRuntime {
     }
 }
 
-#[allow(clippy::single_match)] // there are comments in the error branch, so we don't want if-let
 impl Drop for LocalRuntime {
     fn drop(&mut self) {
         if let LocalRuntimeScheduler::CurrentThread(current_thread) = &mut self.scheduler {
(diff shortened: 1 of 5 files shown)


Darksonn (approved):
Thanks.

PR #7375, opened 2025-05-30
macros: add "local" runtime flavor

Twey:
## Motivation

We would like to be able to create a local runtime from the `tokio::main` attribute in order to be able to use `spawn_local`.

## Solution

Add a `local` flavor to `tokio::main` which behaves like `current-thread`, but calls `build_local(Default::default())` instead of `build()`.

### DIFF
--- a/tests-integration/tests/macros_main.rs
+++ b/tests-integration/tests/macros_main.rs
@@ -16,6 +16,13 @@ async fn spawning() -> usize {
     join.await.unwrap()
 }
 
+#[cfg(tokio_unstable)]
+#[tokio::main(flavor = "local")]
+async fn local_main() -> usize {
+    let join = tokio::task::spawn_local(async { 1 });
+    join.await.unwrap()
+}
+
 #[test]
 fn main_with_spawn() {
     assert_eq!(1, spawning());
@@ -24,5 +31,8 @@ fn main_with_spawn() {
 #[test]
 fn shell() {
     assert_eq!(1, basic_main());
-    assert_eq!(bool::default(), generic_fun::<bool>())
+    assert_eq!(bool::default(), generic_fun::<bool>());
+
+    #[cfg(tokio_unstable)]
+    assert_eq!(1, local_main());
 }
--- a/tokio-macros/src/entry.rs
+++ b/tokio-macros/src/entry.rs
@@ -10,6 +10,7 @@ type AttributeArgs = syn::punctuated::Punctuated<syn::Meta, syn::Token![,]>;
 enum RuntimeFlavor {
     CurrentThread,
     Threaded,
+    Local,
 }
 
 impl RuntimeFlavor {
@@ -17,6 +18,7 @@ impl RuntimeFlavor {
         match s {
             "current_thread" => Ok(RuntimeFlavor::CurrentThread),
             "multi_thread" => Ok(RuntimeFlavor::Threaded),
+            "local" => Ok(RuntimeFlavor::Local),
             "single_thread" => Err("The single threaded runtime flavor is called `current_thread`.".to_string()),
             "basic_scheduler" => Err("The `basic_scheduler` runtime flavor has been renamed to `current_thread`.".to_string()),
             "threaded_scheduler" => Err("The `threaded_scheduler` runtime flavor has been renamed to `multi_thread`.".to_string()),
@@ -178,14 +180,14 @@ impl Configuration {
 
         let flavor = self.flavor.unwrap_or(self.default_flavor);
         let worker_threads = match (flavor, self.worker_threads) {
-            (F::CurrentThread, Some((_, worker_threads_span))) => {
+            (F::CurrentThread | F::Local, Some((_, worker_threads_span))) => {
                 let msg = format!(
                     "The `worker_threads` option requires the `multi_thread` runtime flavor. Use `#[{}(flavor = \"multi_thread\")]`",
                     self.macro_name(),
                 );
                 return Err(syn::Error::new(worker_threads_span, msg));
             }
-            (F::CurrentThread, None) => None,
+            (F::CurrentThread | F::Local, None) => None,
             (F::Threaded, worker_threads) if self.rt_multi_thread_available => {
                 worker_threads.map(|(val, _span)| val)
             }
@@ -207,7 +209,7 @@ impl Configuration {
                 );
                 return Err(syn::Error::new(start_paused_span, msg));
             }
-            (F::CurrentThread, Some((start_paused, _))) => Some(start_paused),
+            (F::CurrentThread | F::Local, Some((start_paused, _))) => Some(start_paused),
             (_, None) => None,
         };
 
@@ -219,7 +221,7 @@ impl Configuration {
                 );
                 return Err(syn::Error::new(unhandled_panic_span, msg));
             }
-            (F::CurrentThread, Some((unhandled_panic, _))) => Some(unhandled_panic),
+            (F::CurrentThread | F::Local, Some((unhandled_panic, _))) => Some(unhandled_panic),
             (_, None) => None,
         };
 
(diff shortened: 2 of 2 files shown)


tests-integration/tests/macros_main.rs
ADD-SP, line 1: image

It looks like the file permissions were accidentally changed?

Twey: Oops! Thanks :)

Darksonn:
Can you please make sure that CI passes? There's a rustfmt failure.

Twey:
Thanks for taking the time to review @ADD-SP @Darksonn! I think I've addressed your comments. Would you mind taking another look?

tokio-macros/src/entry.rs
```diff
@@ -178,14 +180,14 @@ impl Configuration {
...
+            (F::CurrentThread | F::Local, Some((_, worker_threads_span))) => {
                 let msg = format!(
                     "The `worker_threads` option requires the `multi_thread` runtime flavor. Use `#[{}(flavor = \"multi_thread\")]`",
                     self.macro_name(),
                 );
                 return Err(syn::Error::new(worker_threads_span, msg));
             }
-            (F::CurrentThread, None) => None,
+            (F::CurrentThread | F::Local, None) => None,
             (F::Threaded, worker_threads) if self.rt_multi_thread_available => {
```
Darksonn, line 192: Please add logic to `rt_multi_thread_available` to detect whether `--cfg tokio_unstable` is set and emit a hard error if it is not.

Darksonn (commented):
Sorry about the delay in review.

carllerche:
This looks fine to me. I'm happy to merge this when @Darksonn is. Are there docs? My main concern is explaining to users the difference between flavor local and flavor current_thread and how to pick between them.

Twey:
Is there a reason to use `current_thread` over `local` when it's available?

Darksonn:
@Twey I'm pretty reluctant to change the default runtime. I do not want a repeat of #4142 where we had to revert something.

Twey:
I'm happy enough to not change the default runtime — but I'm not really clear on what I should write in the documentation about when to choose `current_thread` over `local`.

tokio-macros/src/entry.rs
Twey: I'm not sure I understand this comment.  Did you mean to add the logic to `RuntimeFlavor::from_str`?

Darksonn: The error should be similar to this error:

https://github.com/tokio-rs/tokio/blob/2506c9fa9916a1bdffbc762f7eb2ae5c2fd23836/tokio-macros/src/entry.rs#L193-L197

Twey: Maybe you meant https://github.com/tokio-rs/tokio/pull/7375/commits/a0ca77ee1cc33d1993d1f85aae6bf66641192fa0; if I've misinterpreted please feel free to suggest changes :)

Twey:
I'm not certain I've done the right thing but I've done my best :)  Please feel free to suggest further changes if I didn't understand something correctly.

tokio-macros/src/lib.rs
```diff
@@ -125,7 +141,36 @@ use proc_macro::TokenStream;
...
+/// ```ignore
+/// #[tokio::main(flavor = "local")]
+/// async fn main() {
+///     println!("Hello world");
+/// }
+/// ```
+///
+/// Equivalent code not using `#[tokio::main]`
+///
+/// ```ignore
```
Darksonn: Does this work?
```suggestion
/// ```
/// #![cfg(tokio_unstable)]
```

Twey: No, it will be executed.  We could use `text` but then we lose Rust highlighting.

Darksonn: If we can execute it, then I think we should. Why can't we?

tokio-macros/src/entry.rs
Twey: Sorry, I just saw your comment.  Is https://github.com/tokio-rs/tokio/commit/4363203e67bb39edf8f959d88fd4940a0771c688 more what you had in mind?

Darksonn: I think both are okay.

tokio-macros/src/lib.rs
Twey: We can do b040de5d053649147e2210fa3d0f9688ffe7fbb8.

Twey: We can't execute it because the `LocalExecutor` doesn't exist if we don't have `tokio_unstable`, which we don't usually have in tests.  But maybe it's clearer this way anyway.

Darksonn: What if we add a `cfg(tokio_unstable)` to the test so it only actually runs under `tokio_unstable`? I believe we run tests both with and without the flag.

tokio-macros/src/entry.rs
Twey: This is harder than it appears because [`RUSTFLAGS` is not passed to proc macros](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#target-applies-to-host).  I guess we have to _emit_ a `cfg` check, so it actually can't live in either of these places.

tokio-macros/src/lib.rs
Twey: Right, that's https://github.com/tokio-rs/tokio/commit/b040de5d053649147e2210fa3d0f9688ffe7fbb8 :)

tokio-macros/src/entry.rs
Twey: 44a31b1be51de957d5d4661599e3b9850bf4b108 seems to work in my local tests.

Darksonn: Doesn't that also trigger other unrelated errors when it fails?

What we do today is essentially define the `main` macro two times to pass different values to [the `rt_multi_thread` boolean](https://github.com/tokio-rs/tokio/blob/master/tokio-macros/src/entry.rs#L488), and have `lib.rs` do this:
```rs
#[cfg(feature = "rt-multi-thread")]
pub use tokio_macros::main;
#[cfg(not(feature = "rt-multi-thread"))]
pub use tokio_macros::main_rt as main;
```
This way, the macro knows based on the cfgs that apply in the main Tokio crate. We could use the same approach and have four macros.

tokio-macros/src/lib.rs
mox692, line 1: The permission of this file is updated too.

tokio-macros/src/entry.rs
Twey: The quadratic explosion is a little bit scary to me.  We can avoid emitting additional code if checks fail like 213598d0995db300d8323cf74b1f6cbecd7751fa.

tokio-macros/src/lib.rs
Twey: Thanks :sweat: My editor thought it was a shell script because of the `#!`.  Fixed (and misbehaving editor firmly corrected).

jobafr:
Hi, I found this PR because I was trying to figure out why `spawn_local` would not work when using the `current-thread` runtime from `tokio::main`. Maybe you could add something to the [docs of `spawn_local`](https://docs.rs/tokio/1.46.1/src/tokio/task/local.rs.html#327-368) to indicate that this is going to work when using a `main` with `#[tokio::main(flavor = "local")]`?

Maybe something like

> Spawns a `!Send` future on the current [`LocalSet`] or [`LocalRuntime`].
> 
> This is possible when either using one of these types explicitly, or by opting to use `RuntimeFlavor::Local` for the calling `main` function like so: `#[tokio::main(flavor = "local")]`.

Twey:
> Maybe you could add something to the [docs of spawn_local](https://docs.rs/tokio/1.46.1/src/tokio/task/local.rs.html#327-368)

In general it doesn't seem like we refer to the macros in the core API docs. I don't have anything against it myself, but I'll let the others weigh in on whether that's something we want to start doing. :)

Darksonn:
Referring to the macros in core API docs is perfectly fine.

Twey:
Okay! https://github.com/tokio-rs/tokio/pull/7375/commits/8c3d83804fa76fc1d4b85c27876d5ff07b185a65

I slightly reworded it @jobafr to avoid mention of the internal `RuntimeFlavor::Local` enum, and to be explicit (again) that it will only work with `tokio_unstable`.  I hope that's okay.

tokio-macros/src/entry.rs
```diff
@@ -437,14 +457,21 @@ fn parse_knobs(mut input: ItemFn, is_test: bool, config: FinalConfig) -> TokenSt
...
-                .build()
+                .#build
                 .expect("Failed building the Runtime")
                 .block_on(#body_ident);
         }
+
+        #[allow(unreachable_code)]
+        {
+            panic!("fell through checks")
+        }
```
mox692: Is this code needed? (since this is after the return statement, I think it will never be executed.)

mox692: Or, does this work as a safeguard for future macro changes?

Twey: If the checks fail, the function body (which contains ill-formed code) won't be called.  This trailer exists to provide a value of the function return type in the case that we aren't returning the real return value earlier so that we don't surface an unhelpful type mismatch error to the user.

Twey: c.f. https://github.com/tokio-rs/tokio/pull/7375#discussion_r2164249335

Darksonn: As far as I can tell, the addition of this panic is the cause of the worse error messages seen in `tests-build/tests/fail/macros_type_mismatch.stderr`. Having reasonably good error messages when used incorrectly is a goal of the macro.

Can you try annotating the fall-through case with `#[cfg(not(tokio_unstable))]` and see if that avoids the regression on the error messages.

Twey: Done in https://github.com/tokio-rs/tokio/pull/7375/commits/99335095677c14b7bf76c5a40f9de2804e6014bc.

Darksonn (approved):
Thanks.

PR #7391, opened 2025-06-06
Introduce `notified_owned` for thread-safe notification futures.

Sagebati:
Allows returning an owned notification future that can be moved across threads. Inspired by https://github.com/tokio-rs/tokio/issues/7231

## Motivation

A common pattern when using Notify is to launch tasks and then notify them at the same time using `notify_waiters`. In the current API it can be quite challenging because you don't know if the task got the time to register before notifying.  A good way of doing it would be to register then launch the task. Problem to do that you would need the result future from `notified()` (aka `Notified<'_>`) to be Send.

Ex: 

```rust 
let n = Arc::new(Notify::new()); // We can't use notified here because we need to send it.

tokio::spawn({
  let n = n.clone();
  async move {
    // if we need the code to be correct we need to register the waiter here
    tokio::sleep(10).await;
    n.notified().await;
    println!("event received");
}})

n.notify_waiters();
```

Ex: Workaround using a oneshot and Shared<_>

```rust 
let (sx, rx) = oneshot();
let rx = rx.shared();

for _ in 0..10 {
  let rx = rx.clone();
  tokio::spawn(async move { rx.clone().await; println!("event received"); });
}

sx.send(());
```

## Solution

Added a new method `notified_owned` that returns `Send` `Notified<'static>`.

### DIFF
--- a/tokio/src/sync/notify.rs
+++ b/tokio/src/sync/notify.rs
@@ -10,6 +10,7 @@ use crate::loom::sync::atomic::AtomicUsize;
 use crate::loom::sync::Mutex;
 use crate::util::linked_list::{self, GuardedLinkedList, LinkedList};
 use crate::util::WakeList;
+use std::sync::Arc;
 
 use std::future::Future;
 use std::marker::PhantomPinned;
@@ -374,6 +375,12 @@ impl Drop for NotifyWaitersList<'_> {
     }
 }
 
+#[derive(Debug)]
+enum NotifyRef<'a> {
+    Owned(Arc<Notify>),
+    Ref(&'a Notify),
+}
+
 /// Future returned from [`Notify::notified()`].
 ///
 /// This future is fused, so once it has completed, any future calls to poll
@@ -382,7 +389,7 @@ impl Drop for NotifyWaitersList<'_> {
 #[must_use = "futures do nothing unless you `.await` or poll them"]
 pub struct Notified<'a> {
     /// The `Notify` being received on.
-    notify: &'a Notify,
+    notify: NotifyRef<'a>,
 
     /// The current state of the receiving process.
     state: State,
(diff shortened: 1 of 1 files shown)


ADD-SP (commented):
I think #7231 is a reasonable use case , and this issue is similar to #1998.

Could you add some tests against this new feature? Also we'd better aligning the naming pattern of [`tokio::sync::OwnedSemaphorePermit`](tokio::sync::OwnedSemaphorePermit) instead of `Notified<'static>`, how about `OwnedNotified`?

tokio/src/sync/notify.rs
```diff
@@ -374,6 +375,12 @@ impl Drop for NotifyWaitersList<'_> {
     }
 }
 
+#[derive(Debug)]
+enum NotifyRef<'a> {
+    Owned(Arc<Notify>),
+    Ref(&'a Notify),
+}
```
Darksonn, line 382: I'm not sure if we want more future types, but if we do, it has to be a new type. An enum pays a cost for everyone using `notified`, and we don't want to go that route.

Sagebati: Fair, I didn't take into account the cost of the access, any workaround for not repeating the implementation between Notified and NotifiedOwned ?  I suppose using an union is cursed, maybe a macro ?

Darksonn: Let's avoid unions or macros. How about extracting common logic into shared functions?

ariaandika:
Hi, any update on this ? if you’re no longer working on it, id be happy to take it over.

Sagebati:
Sorry, you can take over. I don't have time right now to work on it

Le ven. 18 juil. 2025, 13:44, Aria Andika ***@***.***> a
écrit :

PR #7400, opened 2025-06-10
util: Add `AbortOnDropHandle::detach`

goffrie:
## Motivation

Sometimes after creating an `AbortOnDropHandle` I want to be able to change my mind and allow the task to live indefinitely. In particular I'd like to make the default spawn method in my codebase wrap the task in an `AbortOnDropHandle` to avoid task leaks, but the caller can still decide to detach the task if it's known to be OK.

## Solution

Add `AbortOnDropHandle::detach` which cancels the abort-on-drop. The JoinHandle is returned for flexibility but I'd expect the typical caller to just drop it immediately.

The use of unsafe is unfortunate but AFAIK there is no way to safely cancel a `Drop` impl (and recover the fields) without extra overhead (eg `Option`).

### DIFF
--- a/tokio-util/src/task/abort_on_drop.rs
+++ b/tokio-util/src/task/abort_on_drop.rs
@@ -5,6 +5,7 @@ use tokio::task::{AbortHandle, JoinError, JoinHandle};
 
 use std::{
     future::Future,
+    mem::ManuallyDrop,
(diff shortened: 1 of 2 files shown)


Darksonn (approved):
Thanks.

PR #7416, opened 2025-06-24
Fix documentation link for RawFd in TcpSocket docs

VolodymyrBg:
Update the documentation link for [RawFd] in the TcpSocket docs to point to the correct and up-to-date location in the Rust standard library. This ensures that users are directed to the proper definition of RawFd when reading the documentation.


VolodymyrBg:
> Did you mean https://doc.rust-lang.org/std/os/fd/type.RawFd.html?
> 
> This PR proposes https://doc.rust-lang.org/std/os/unix/io/index.html#raw-file-descriptors, but it is not the page of `RawFd`.
> 
> 

Ops, yes

ADD-SP (approved):
Thanks!

PR #7418, opened 2025-06-24
sync: Introduce the `SetOnce` struct

Daksh14:
The `SetOnce` data structure allows you to contain a T value and wait until the initialization of that T value in the Cell. It can only be initialized once and can be destroyed using `into_inner` or by dropping it.

The wait() allows you to block on a future which wakes up when the `SetOnce` is initalized

## Motivation

After #7161 failed and was messy, we needed a new type that is like OnceCell but with tigter restrictions around T 
Something similar in async_io in python https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event

## Solution

Proposal of the implementation of `SetOnce` data structure.


tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,343 @@
...
+    /// initialized the wakers are notified and the Future returned from this
+    /// function completes.
+    ///
+    /// If this function is called after the `SetOnce` is initialized then
+    /// empty future is returned which completes immediately.
+    pub async fn wait(&self) {
+        if !self.initialized() {
+            let _ = self.notify.notified().await;
+        }
+    }
```
ADD-SP:
```rust
pub async fn wait(&self) -> &T;
```

I would suggest this interface, because it is inconvenient to `await` first and then invoke `SetOnce::get`.

ADD-SP:
```rust
pub fn get(&self) -> Option<&T>
```

Other two reasons.

1. `Option<T>` always create possibilities for runtime panic.
2. It makes not sense to invoke `SetOnce::get` without preceding `SetOnce::wait().await`.

So I think `SetOnce::get` is unnecessary, let `SetOnce::wait` return `&T` is a better interface.

ADD-SP (commented):
I haven't looked at entire implementation yet, so I just left interface design comment.

tokio/src/sync/set_once.rs
Daksh14: @ADD-SP I agree that wait should return the T so we dont have to invoke get but shouldn't the wait return `Option<&T>` because what if you call wait on `SetOnce` which isn't initialized, should we just return a `futures::empty()`?

Daksh14: /cc @Darksonn

ADD-SP: > shouldn't the wait return Option<&T> because what if you call wait on SetOnce which isn't initialized.

I think the `SetOnce::wait().await` should only be resolved if any of the following requirements is met:

1. Its value has been initialized.
2. (OR) All other `SetOnce` instances has been dropped but the value is still uninitialized.

I guess you meant that we should return `None` on (2). This should be a reasonable point which I didn't realized before.

Now, there are two options.

```rust
pub async fn wait(&self) -> Option<&T>;

// OR

// Error means all other instances
// has been dropped and value
// is still uninitialized.
pub async fn wait(&self) -> Result<&T, Error>
```

Selecting  `Option<&T>` indicates that we consider it a normal use case that `SetOnce` are never initialized. `Result<&T, Error`> indicates it is not.

What do you think?

Daksh14: we can have a `try_wait` and `wait` so we have both, `wait` should return `None` if memory is uninitialized but at that point you can just turn the result into an option with `.ok()` so I think we should keep the `-> Result<&T, Error>` to be the most descriptive as a library standpoint

Darksonn: Case 2 cannot happen since `SetOnce` isn't a channel that you clone. The return value can be a bare `&T`.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,343 @@
...
+
+impl<T> Drop for SetOnce<T> {
+    fn drop(&mut self) {
+        if self.initialized() {
+            // SAFETY: We're inside the drop implementation of SetOnce
+            // AND we're also initalized. This is the best way to ensure
+            // out data gets dropped
+            unsafe {
+                let _ = self.value.with_mut(|ptr| ptr::read(ptr).assume_init());
+            }
```
Darksonn: Please use `drop_in_place` here.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,343 @@
...
+
+        // Using release ordering so any threads that read a true from this
+        // atomic is able to read the value we just stored.
+        if !self.value_set.swap(true, Ordering::Release) {
+            // SAFETY: We are swapping the value_set AtomicBool from FALSE to
+            // TRUE with it being previously false and followed by that we are
+            // initializing the unsafe Cell field with the value
+            unsafe {
+                self.set_value(value);
+            }
```
Darksonn: This logic isn't enough. You are writing the value *after* changing the atomic, so other threads can see the atomic holding the value `true` before the value is written.

You will need an `AtomicU8` with three states: EMPTY, SETTING, DONE. The extra state "SETTING" means "set is currently running". So the logic will be:

* `compare_exchange` from EMPTY -> SETTING.
* Write the value.
* Release store DONE.

Other threads will only make the value accessible if the state is DONE.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,377 @@
...
+///     let arc = Arc::new(once);
+///     let first_cl = Arc::clone(&arc);
+///     let second_cl = Arc::clone(&arc);
+///
+///     tokio::spawn(async move { first_cl.set(20) });
+///
+///     tokio::spawn(async move { second_cl.set(10) });
+///
+///     let res = arc.get(); // returns None
+///     arc.wait().await; // lets wait until the value is set
```
ADD-SP: > let res = arc.get(); // returns None

We cannot assume spawned task will be not be polled immediately.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,377 @@
...
+}
+
+#[repr(u8)]
+#[derive(PartialEq, Eq)]
+pub(crate) enum AtomicState {
+    Empty = 0,
+    // SETTING means "set is currently running"
+    Setting = 1,
+    Done = 2,
+}
```
ADD-SP: The common style in tokio is declare there const values. I don't strongly against this `enum` approach, but it would be better aligning to existing style.

https://github.com/tokio-rs/tokio/blob/3e890cc0171ddb210acdcfec831b7c7bcbb0d2d9/tokio/src/sync/task/atomic_waker.rs#L131-L138

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,377 @@
...
+    // AtomicState::Done.
+    unsafe fn set_value(&self, value: T) {
+        self.value.with_mut(|ptr| (*ptr).as_mut_ptr().write(value));
+
+        // Store that we're done in Release so any `Acquire` can see
+        self.value_set
+            .store(AtomicState::Done as _, Ordering::Release);
+
+        // notify the waiting wakers that the value is set
+        self.notify.notify_waiters();
```
ADD-SP: This might not notify all waiters, consider this execution timeline

1. Thread A: `SetOnce::wait().await`
2. Thread B: `SetOnce::set()`
3. Thread C: `SetOnce::wait().await`.

In this case, thread C will not be notified.

> If a task is currently waiting, that task is notified. Unlike with `notify_one()`, no permit is stored to be used by the next call to `notified().await`. The purpose of this method is to notify all already registered waiters. Registering for notification is done by acquiring an instance of the Notified future via calling `notified()`.
>
> *[`tokio::sync::Notify::notify_waiters`]*


[`tokio::sync::Notify::notify_waiters`]: https://docs.rs/tokio/1.45.1/tokio/sync/struct.Notify.html#method.notify_waiters

tokio/src/sync/set_once.rs
ADD-SP: > Case 2 cannot happen since SetOnce isn't a channel that you clone. The return value can be a bare &T.

Ah, that's true. I would suggest `async fn wait(&self) -> &T`.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,377 @@
...
+    ///
+    /// If the `SetOnce` already has a value, this call will fail with an
+    /// [`SetError::AlreadyInitializedError`].
+    ///
+    /// If the `SetOnce` is empty, but some other task is currently trying to
+    /// set the value, this call will fail with [`SetError::InitializingError`].
+    ///
+    /// [`SetError::AlreadyInitializedError`]: crate::sync::SetError::AlreadyInitializedError
+    /// [`SetError::InitializingError`]: crate::sync::SetError::InitializingError
+    pub fn set(&self, value: T) -> Result<(), SetError<T>> {
```
ADD-SP: I'm a little bit uncomfortable to use the `OnceCell`'s error type for another different type, this is because every changes to `SetOnce` or `OnceCell` or `SetError` might affect both three types.

It is possible to make a compatible change to one type, but its a breaking change for another type due to coupled Error type. I haven't come up with a specific example for my concern, but I still suggest a new error type.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,377 @@
...
+            Err(true_val) => {
+                // The value is being currently set in another thread
+                if true_val == state_setting {
+                    Err(SetError::InitializingError(value))
+                } else {
+                    // here means true_val is either Done or higher
+                    Err(SetError::AlreadyInitializedError(value))
+                }
+            }
+        }
```
ADD-SP:
```rust
Err(actual) if actual == state_setting => {},
Err(actual) if actual == state_done => {},
Err(actual) => { /* consider panic in this branch */},
```

This might be more readable.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,377 @@
...
+    fn eq(&self, other: &SetOnce<T>) -> bool {
+        self.get() == other.get()
+    }
+}
+
+impl<T: Eq> Eq for SetOnce<T> {}
+
+impl<T> Drop for SetOnce<T> {
+    fn drop(&mut self) {
+        if self.initialized() {
```
ADD-SP: `AtomicU8::get_mut` should be a better choice.

tokio/src/sync/set_once.rs
Daksh14: @ADD-SP I will change the comment to say "sometimes returns None" point is to call wait() to wait until the value is set but we're talking about a API change now

tokio/src/sync/set_once.rs
Daksh14: Then this style shall be adhered 👍  I was debating which method to use

tokio/src/sync/set_once.rs
Darksonn: To fix this, we must update `wait` to use the following logic:

1. Create a `notified()` future and pin it.
2. Perform an `AcqRel` load of the state. If the value is `DONE`, then return `self.get_unchecked()`.
3. Await the `notified()` future.
4. Go back to step 1.

Note that ordinary loads cannot use the `AcqRel` ordering, so you will need to use `fetch_add(0)` or `compare_exchange(DONE, DONE)` or similar so that the operation is a write operation even though it doesn't actually change the value.

We will also need to change `set_value` to use `AcqRel` for the store instead of just `Release`. This will similarly need to use a read/write operation instead of just a store; probably a `swap` is sufficient.

Here is the analysis of why this works:

Since the read in `wait` and the write in `set_value` are both operations on the same atomic, one of them must come first in the modification order of that atomic. There are two cases:

* If the store in `set_value` comes first in the modification order, then the load in `wait` will see the value `DONE` and return `self.get_unchecked()`. The call to `get_unchecked()` is guaranteed to see the write to `self.value` because writing the value to `self.value` happens-before the release store to `self.value_set` happens-before the acquire load of `self.value_set` happens-before the call to `self.get_unchecked()`. Note that in this case we rely on the store being release and the load being acquire.
* If the load in `wait` comes first in the modification order, then the "read" part of the store in `set_value` will see the value "written" by the load in `wait`. This means that creating the `notified()` future happens-before the `AcqRel` load of the state happens-before the `AcqRel` store of `DONE` happens-before the call to `notify_waiters()`. Therefore, it's guaranteed that `notify_waiters()` will wake up the `notified()` future.

There are some optimizations one could perform on top of this, but let's start with this logic so that it's correct. We can talk about performance afterwards.

ADD-SP: I would propose an alternative for this issue.

Replacing `Notify` with `Semaphore`, and the `wait` logic should be:

1. `Semaphore::acquire().await`
2. `assert_eq!(self.state.fetch_add(0, Acquire), DONE)`
3. `self.get_unchecked()`

ADD-SP: The second alternative proposed by me is get #7391 merged, and then we can register the waiter on `SetOnce::new`.

Darksonn: Using a `Semaphore` would work, and you would not need an atomic at all in that case. You would be able to use [`Semaphore::close`](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#method.close) to indicate that the value has been written.

I don't think it makes sense to register the waiter in `SetOnce::new`. When you create a new `SetOnce` it is independent from any existing ones, and it may receive any number of waiters from future calls to `wait`.

tokio/src/sync/set_once.rs
Darksonn: I don't think we want two separate kinds of errors here. With `OnceCell` initialization can fail and can take a long time, so it's important to consider it to be a separate state, but with `SetOnce` the initialization state is always very short-lived and infallible, so I don't think we want to expose it to the end-user.

tokio/src/sync/set_once.rs
ADD-SP: > it may receive any number of waiters from future calls to `wait`

You are right. Pre-registering doesn't work.

ADD-SP: > Using a Semaphore would work, and you would not need an atomic at all in that case.

Since the official docs of [`tokio::sync::Semaphore`] doesn't guarantee the memory ordering, I'm worried about the risk  of depending on technical implementation details of `Semaphore` in `SetOnce`, also worried about other tokio developers will follow this pattern in the future development.

What do you think?

[`tokio::sync::Semaphore`]: https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html

tokio/src/sync/set_once.rs
ADD-SP: > but with SetOnce the initialization state is always very short-lived and infallible, so I don't think we want to expose it to the end-user.

I agree with this, it makes a lot of sense.

tokio/src/sync/set_once.rs
Darksonn: The semaphore does provide the necessary orderings, and it's not something that's going to change. That said, it could make sense to mention this in the documentation.

tokio/src/sync/set_once.rs
Darksonn: Closing this thread as it's covered by https://github.com/tokio-rs/tokio/pull/7418#discussion_r2177105406

tokio/src/sync/set_once.rs
ADD-SP: > The semaphore does provide the necessary orderings, and it's not something that's going to change. That said, it could make sense to mention this in the documentation.

I moved this topic into #7435 so we can focus on PR review instead of documentation discussion.

Darksonn: After looking more into this, I don't think the `Semaphore` is the way to go. It doesn't have a mechanism to wait until the semaphore is closed, but that's what `wait` needs.

Instead, I suggest adding a mutex and having `set` just lock the mutex. That should be a rather simple way to fix this.

ADD-SP: > It doesn't have a mechanism to wait until the semaphore is closed

Will [`tokio::sync::Semaphore::acquire().await`] returns error on close?

[`tokio::sync::Semaphore::acquire().await`]: https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#method.acquire

ADD-SP: However, relying on this kind of error report for implementation is not a ideal pattern.

Darksonn: Yes, but we don't want `wait()` to actually acquire the permit. The permit should only be acquired by `set` - that way we know that a call to `set` should initialize the value if it successfully acquires the permitl.

tokio/src/sync/set_once.rs
Daksh14: Resolved as we're back to using my original AtomicBool approach

tokio/src/sync/set_once.rs
Daksh14: using `AtomicBool::get_mut`

tokio/src/sync/set_once.rs
Daksh14: Closing as we're back to using `AtomicBool` and Mutex

tokio/src/sync/set_once.rs
Daksh14: This is closed as a swap is enough now as we are doing the writes after locking the mutex to ensure one caller of set() at a time

tokio/src/sync/set_once.rs
Daksh14: Changed to `async fn wait(&self) -> &T` and added  `async fn get_wait(&self) -> Option<&T>`

tokio/src/sync/set_once.rs
ADD-SP: Is *[`tokio_util::sync::CancellationToken`]* a viable approach?

[`tokio_util::sync::CancellationToken`]: https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html

ADD-SP: We can cancel the `CancellationToken` after setting the value, and `await` the token in `SetOnce::wait`.

Also the implementation details of `CancellationToken` provides necessary orderings.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,392 @@
...
+
+        // look at the state again
+        if self.value_set.load(Ordering::Acquire) {
+            // SAFETY: the state is initialized
+            return Some(unsafe { self.get_unchecked() });
+        }
+
+        None
+    }
+}
```
ADD-SP: What is the use case of this method?

ADD-SP: If we want a sync method to check if a `SetOnce` is initialized, `SetOnce::try_get` is a more readable name.

Daksh14: This methods waits on the notified future and checks the state, I didn't choose `try_get` because we have `set` that returns a `Result` and `get_wait()` reutrning `Option`  and here we don't need really return a `Error` but I can just replace it with `SetError<T>`

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,392 @@
+use super::{Notify, SetError};
+use crate::{loom::cell::UnsafeCell, pin};
+use std::fmt;
+use std::mem::MaybeUninit;
+use std::ops::Drop;
+use std::ptr;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Mutex;
```
Darksonn: Please use synchronization primitives from `crate::loom` instead.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,392 @@
...
+            Ok(_) => {
+                // Using release ordering so any threads that read a true from this
+                // atomic is able to read the value we just stored.
+                if !self.value_set.swap(true, Ordering::Release) {
+                    // SAFETY: We are swapping the value_set AtomicBool from FALSE to
+                    // TRUE with it being previously false and followed by that we are
+                    // initializing the unsafe Cell field with the value
+                    unsafe {
+                        self.set_value(value);
+                    }
```
Darksonn: Please set the value first, and then write to the atomic. Also, using `store` is sufficient. With the mutex, we no longer need `swap`.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,392 @@
...
+                    Ok(())
+                } else {
+                    Err(SetError::AlreadyInitializedError(value))
+                }
+            }
+            Err(_) => {
+                // If we failed to lock the mutex, it means some other task is
+                // trying to set the value, so we return an error.
+                Err(SetError::InitializingError(value))
+            }
```
Darksonn: No this error is for mutex poisoning. We don't use that in Tokio. When you switch to the mutex in `crate::loom`, it will go away.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,392 @@
...
+        } else {
+            None
+        }
+    }
+
+    /// Waits until set is called. The future returned will keep blocking until
+    /// the `SetOnce` is initialized.
+    ///
+    /// If the `SetOnce` is already initialized, it will return the value
+    // immediately.
```
Darksonn:
```suggestion
    /// If the `SetOnce` is already initialized, it will return the value
    /// immediately.
```

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,392 @@
...
+        // wait until the value is set
+        (&mut notify_fut).await;
+
+        // look at the state again
+        if self.value_set.load(Ordering::Acquire) {
+            // SAFETY: the state is initialized
+            return Some(unsafe { self.get_unchecked() });
+        }
+
+        None
```
Darksonn: No need to check this. This also lets you get rid of `get_wait` and just have `wait`. It's not possible for the state to fail to initialize.
```suggestion
        Some(unsafe { self.get_unchecked() })
```

tokio/src/sync/set_once.rs
Daksh14: this is removed and just `wait()` exists

tokio/tests/sync_set_once.rs
ADD-SP, line 1: Please use [`loom::model`] for testing.

[`loom::model`]: https://docs.rs/loom/0.7.2/loom/fn.model.html

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,371 @@
...
+        // SAFETY: lock the mutex to ensure only one caller of set
+        // can run at a time.
+        let guard = self.lock.lock();
+
+        unsafe {
+            self.set_value(value);
+        }
+
+        drop(guard);
+
```
ADD-SP, line 312:
```suggestion
        drop(guard);

```

This may not be necessary.

Daksh14: Without the drop CI was complaining so I added it

ADD-SP: This looks strange, what was the CI complain about?

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,371 @@
...
+        unsafe {
+            self.value.with_mut(|ptr| (*ptr).as_mut_ptr().write(value));
+        }
+
+        // Using release ordering so any threads that read a true from this
+        // atomic is able to read the value we just stored.
+        self.value_set.store(true, Ordering::Release);
+
+        // notify the waiting wakers that the value is set
+        self.notify.notify_waiters();
```
Darksonn: Please call `notify_waiters()` after releasing the `self.lock` mutex. This will probably require you to move everything in `set_once` into `set`, but I think that's okay when `set` is so simple already.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,371 @@
...
+    /// [`SetError::AlreadyInitializedError`].
+    ///
+    /// If the `SetOnce` is empty, but some other task is currently trying to
+    /// set the value, this call will fail with [`SetError::InitializingError`].
+    ///
+    /// [`SetError::AlreadyInitializedError`]: crate::sync::SetError::AlreadyInitializedError
+    /// [`SetError::InitializingError`]: crate::sync::SetError::InitializingError
+    pub fn set(&self, value: T) -> Result<(), SetError<T>> {
+        if self.initialized() {
+            return Err(SetError::AlreadyInitializedError(value));
```
Darksonn: Please make a new error type, since we never fail with `InitializingError`.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,371 @@
...
+        }
+
+        // SAFETY: lock the mutex to ensure only one caller of set
+        // can run at a time.
+        let guard = self.lock.lock();
```
Darksonn, line 292: You must check `self.initialized()` after taking the lock. Otherwise, another thread could set the value in between checking initialized and taking the lock.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,371 @@
...
+
+        drop(guard);
+
+        Ok(())
+    }
+
+    /// Takes the value from the cell, destroying the cell in the process.
+    /// Returns `None` if the cell is empty.
+    pub fn into_inner(mut self) -> Option<T> {
+        if self.initialized() {
```
Darksonn: This can use `self.value.get_mut()` too.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,371 @@
...
+    /// avoid this, use `get_wait()` which returns an `Option<&T>` instead of
+    /// `&T`.
+    pub async fn wait(&self) -> &T {
+        let notify_fut = self.notify.notified();
+        pin!(notify_fut);
+
+        if self.value_set.load(Ordering::Acquire) {
+            // SAFETY: the state is initialized
+            return unsafe { self.get_unchecked() };
+        }
```
Darksonn: Please take the mutex when checking `value_set`.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,371 @@
+use super::{Notify, SetError};
+use crate::{
+    loom::{cell::UnsafeCell, sync::Mutex},
+    pin,
+};
+use std::fmt;
+use std::mem::MaybeUninit;
+use std::ops::Drop;
+use std::ptr;
+use std::sync::atomic::{AtomicBool, Ordering};
```
Darksonn: Please use atomics from `crate::loom`. This is required for writing loom tests.

Daksh14: Fixed using loom atomics now

tokio/src/sync/set_once.rs
Daksh14: Changed to use get_mut()

tokio/src/sync/set_once.rs
Daksh14: Made a new error type

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,412 @@
...
+
+        if self.value_set.load(Ordering::Acquire) {
+            // SAFETY: the state is initialized
+            return unsafe { self.get_unchecked() };
+        }
+
+        drop(guard);
+
+        // wait until the value is set
+        (&mut notify_fut).await;
```
Darksonn:
```suggestion
        notify_fut.await;
```

tokio/tests/sync_set_once.rs
Daksh14: Hey @ADD-SP I have added two tests `set_once_drop_test` and `set_once_wait_test` using `loom::model`. @Darksonn Mentioned these are good for now but let me know if you'd like different style of tests one with multiple threads

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,403 @@
...
+        self.notify.notify_waiters();
+
+        Ok(())
+    }
+
+    /// Takes the value from the cell, destroying the cell in the process.
+    /// Returns `None` if the cell is empty.
+    pub fn into_inner(self) -> Option<T> {
+        // TODO: Use get_mut()
+        let value_set = self.value_set.load(Ordering::Relaxed);
```
ADD-SP, line 323: We may have forgotten this todo.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,403 @@
...
+///
+///     let arc = Arc::new(once);
+///     let first_cl = Arc::clone(&arc);
+///     let second_cl = Arc::clone(&arc);
+///
+///     tokio::spawn(async move { first_cl.set(20) });
+///
+///     tokio::spawn(async move { second_cl.set(10) });
+///
+///     let res = arc.get(); // sometimes returns None
```
ADD-SP: Since our example should be simple enough and less confusing, I suggest that we remove this line or let it be on the top of `tokio::spawn(async move { first_cl.set(20) });`

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,403 @@
...
+    ///
+    /// # Note
+    ///
+    /// This will keep waiting until the `SetOnce` is initialized, so it
+    /// should be used with care to avoid blocking the current task
+    /// indefinitely.
+    pub async fn wait(&self) -> &T {
+        loop {
+            let notify_fut = self.notify.notified();
+            pin!(notify_fut);
```
ADD-SP: Maybe a dumb question, why do we need to `pin!` this future?

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,403 @@
...
+                return unsafe { self.get_unchecked() };
+            }
+
+            drop(guard);
+
+            // wait until the value is set
+            notify_fut.await;
+        }
+    }
+}
```
ADD-SP: It is hard for me to eyeball the correctness of this logic. I spent half the time I dedicated to this PR just trying to grasp it.

I'm not doubting the correctness of this implementation. I just want to explore a simpler implementation.

My feeling is that [`CancellationToken`] could help us to simplify the implementation, and its semantic also matches our requirements here.

Not sure if you have already tried this, if not, would you mind taking a try?

[`CancellationToken`]: https://docs.rs/tokio-util/0.7.15/tokio_util/sync/struct.CancellationToken.html

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,403 @@
...
+        self.get() == other.get()
+    }
+}
+
+impl<T: Eq> Eq for SetOnce<T> {}
+
+impl<T> Drop for SetOnce<T> {
+    fn drop(&mut self) {
+        // TODO: Use get_mut()
+        if self.value_set.load(Ordering::Relaxed) {
```
ADD-SP, line 129: We may have forgotten this todo.

tokio/src/sync/tests/loom_set_once.rs
```diff
@@ -0,0 +1,67 @@
...
+        });
+
+        let foo = Foo::new(Arc::clone(&drop_counter));
+
+        let set = tx.set(foo).is_ok();
+        let res = thread.join().unwrap();
+
+        drop(tx);
+
+        assert_eq!(drop_counter.load(Ordering::Acquire), 2);
```
ADD-SP: For reviewers: we don't need RMW here because the

1. `Arc` invokes `Foo::drop` in the current thread.
2. AND the `Foo::drop` performs a `Release` RMW.

tokio/tests/sync_set_once.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    struct Foo {}
+
+    let fooer = Foo {};
+
+    impl Drop for Foo {
+        fn drop(&mut self) {
+            NUM_DROPS.fetch_add(1, Ordering::Release);
+        }
+    }
```
ADD-SP:
```suggestion
    struct Foo;

    impl Drop for Foo {
        fn drop(&mut self) {
            NUM_DROPS.fetch_add(1, Ordering::Release);
        }
    }

    let fooer = Foo;
```

tokio/tests/sync_set_once.rs
```diff
@@ -0,0 +1,229 @@
...
```
ADD-SP:
```suggestion
    struct Foo;

    impl Drop for Foo {
        fn drop(&mut self) {
            NUM_DROPS.fetch_add(1, Ordering::Release);
        }
    }

    let fooer = Foo;
```

tokio/tests/sync_set_once.rs
```diff
@@ -0,0 +1,229 @@
...
```
ADD-SP: Why did we repeat the same `Foo` implementation in several tests?

tokio/src/sync/set_once.rs
Daksh14: This todo exists because we were previously using [`std::sync::atomic::AtomicBool::get_mut()`](https://doc.rust-lang.org/std/sync/atomic/index.html) but after switching to [`loom::sync::atomic::AtomicBool`](https://docs.rs/loom/latest/loom/sync/atomic/struct.AtomicBool.html) we don't have it. We can make a PR in loom

tokio/src/sync/set_once.rs
Daksh14: Okay I wanna keep the comment to explain my reasoning, its abstract in the sense you can have as many threads and as many layers of complexity behind and invoke the `wait()` to confirm that the value has been set and return that value 

>  And wait in some other thread until the value has been set.

tokio/tests/sync_set_once.rs
Daksh14: Checkout `https://github.com/tokio-rs/tokio/blob/master/tokio/tests/sync_once_cell.rs` I copied it from there because this was initially tried to be attributed to OnceCell but it was a mess and it was inspired from what `OnceCell` is. Let me clean it up for you but for OnceCell it should be done in a seperate PR.

tokio/tests/sync_set_once.rs
Daksh14: What type of change is this sorry, is there something wrong with my editor?

tokio/src/sync/set_once.rs
ADD-SP: Ah, i see, this is not a blocker since `into_inner` is not a hot method.

tokio/src/sync/set_once.rs
ADD-SP: Resolved, please checkout https://github.com/tokio-rs/tokio/pull/7418#discussion_r2202272516 for details.

tokio/tests/sync_set_once.rs
ADD-SP: This is just a style change, I moved the `let fooer = Foo;` to the bottom of the `impl Drop for Foo`.

tokio/tests/sync_set_once.rs
ADD-SP: > but for OnceCell it should be done in a seperate PR.

I opened a new PR (#7458) for existing tests.

tokio/src/sync/set_once.rs
Darksonn: We can't use `CancellationToken` here because it's defined by tokio-util. Also, the token has *a lot* of extra complexity to handle child tokens that we don't need.

tokio/src/sync/set_once.rs
Darksonn: Even if it was, it probably has no perf impact in practice.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,403 @@
+use super::Notify;
+
+use crate::{
+    loom::{
+        cell::UnsafeCell,
+        sync::{atomic::AtomicBool, Mutex},
+    },
+    pin,
+};
```
Darksonn: We don't use this import style in Tokio. Generally we have one `use` statement per module.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,403 @@
...
+//      Some(&T)
+//
+// The value cannot be changed after set() is called. Subsequent calls to set()
+// will return a `SetOnceError`
+
+/// A thread-safe cell that can be written to only once.
+/// A `SetOnce` is inspired from python's
+/// [`asyncio.Event`](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event)
+/// type. It can be used to wait until the value of the `SetOnce` is set like
+/// a "Event" mechanism
```
Darksonn: This formatting could be improved.

* Please have an empty line after the summary. Otherwise markdown will not treat it as a line break.
* Please separate out the link url.
* Missing period.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,403 @@
...
+
+impl<T: Eq> Eq for SetOnce<T> {}
+
+impl<T> Drop for SetOnce<T> {
+    fn drop(&mut self) {
+        // TODO: Use get_mut()
+        if self.value_set.load(Ordering::Relaxed) {
+            // SAFETY: We're inside the drop implementation of SetOnce
+            // AND we're also initalized. This is the best way to ensure
+            // out data gets dropped
```
Darksonn: Safety comments must explain why an unsafe operation is ok, and nothing else. In this case, it drops a value, so it should explain why there is a value to drop. It does not need to discuss whether this is the best way of doing something, or whether there are other strategies to do it. If you want to discuss that, it should go in a normal comment.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,403 @@
...
+/// #[tokio::main]
+/// async fn main() -> Result<(), SetOnceError<u32>> {
+///     ONCE.set(2)?;
+///     let result = ONCE.get();
+///     assert_eq!(result, Some(&2));
+///
+///     Ok(())
+/// }
+/// ```
+pub struct SetOnce<T> {
```
Darksonn, line 89: Since this defines a new type and adds new async methods, they should be listed in `tests/assert_send_sync.rs`.

tokio/src/sync/tests/loom_set_once.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn drop(&mut self) {
+        self.num_drops.fetch_add(1, Ordering::Release);
+    }
+}
+
+#[test]
+fn set_once_drop_test() {
+    loom::model(|| {
+        let tx = Arc::new(SetOnce::new());
+        let rx = Arc::clone(&tx);
```
Darksonn: I don't think rx/tx are good names. Rx means receiver and tx means transmitter, but you are also calling `set` on `rx`, which doesn't make sense.

tokio/src/sync/tests/loom_set_once.rs
```diff
@@ -0,0 +1,67 @@
...
+
+        let drop_counter = Arc::new(AtomicU32::new(0));
+        let counter_cl = Arc::clone(&drop_counter);
+
+        let thread = thread::spawn(move || {
+            return rx.set(Foo::new(counter_cl)).is_ok();
```
Darksonn: Unnecessary return.
```suggestion
            rx.set(Foo::new(counter_cl)).is_ok()
```

tokio/tests/sync_set_once.rs
```diff
@@ -0,0 +1,229 @@
...
+fn is_some_initializing() {
+    let rt = runtime::Builder::new_current_thread()
+        .enable_time()
+        .build()
+        .unwrap();
+
+    static ONCE: SetOnce<u32> = SetOnce::const_new();
+
+    rt.block_on(async {
+        time::pause();
```
Darksonn: If you need to pause time, use `start_paused()` in the runtime builder. But you don't use time here, so you don't need it.

tokio/src/sync/set_once.rs
ADD-SP: This makes sense.

tokio/src/sync/set_once.rs
Daksh14: Let me know if this one is better

tokio/src/sync/set_once.rs
Daksh14: Fixed all of the problems, let me know

tokio/src/sync/set_once.rs
Daksh14: Changed to have use statement per line

tokio/src/sync/tests/loom_set_once.rs
Daksh14: Changed variable named to be more descriptive

tokio/tests/sync_set_once.rs
Daksh14: removed time here

tokio/tests/sync_set_once.rs
Daksh14: I re-emplemented this to keep one Foo struct

tokio/tests/sync_set_once.rs
Daksh14: Resolved as I re-emplemented this

tokio/src/sync/set_once.rs
Daksh14: @Darksonn can explain best but I think the future returned from notified needs to be pinned while we lock the mutex and do a AtomicRead on `value_set`

tokio/src/sync/set_once.rs
ADD-SP: It is not easy to cover all purposes using a single example, I suggest to write separated example for each purpose.

I think [`std::sync::OnceLock`] is good example, it is pretty straight forward, and no words like 'maybe' or 'sometimes', everything is deterministic.



For `OnceLock::wait`, there is a separated example to show the typical usage.

[`std::sync::OnceLock`]: https://doc.rust-lang.org/std/sync/struct.OnceLock.html

I think there are somethings we'd like to cover by exmaples

* Handling the multiple call of the `SetOnce::set`.
* `SetOnce::get` returns `Option<&T>` and never blocks.
* `SetOnce::wait` blocks and returns `&T`.

We may not cover everything, but please make sure the example is straightforward if we are going to write it.

ADD-SP: Some people (like me) just want to take a few seconds to scan through the examples and start using them, this is the reason I'm still focusing on our example.

I also saw someone copy-paste-modify the provided example for their project, we also hope our example doesn't make this case too difficult.

Daksh14: @ADD-SP I hope you like the new examples I added, check them out and lmk

ADD-SP: These are great examples, thank you!

tokio/src/sync/set_once.rs
Darksonn: Well, the important part is creating the future before checking `value_set`. It's possible that pinning it doesn't have much effect.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,409 @@
...
+    /// immediately.
+    ///
+    /// # Note
+    ///
+    /// This will keep waiting until the `SetOnce` is initialized, so it
+    /// should be used with care to avoid blocking the current task
+    /// indefinitely.
+    pub async fn wait(&self) -> &T {
+        loop {
+            let notify_fut = self.notify.notified();
```
Darksonn: It could make sense to add a fast-path here. That way, we don't need to access the `Notify` at all if the value is already set.
```suggestion
    pub async fn wait(&self) -> &T {
        loop {
            if let Some(value) = self.get() {
	        return value;
	    }

            let notify_fut = self.notify.notified();
```

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,409 @@
...
+/// set was called
+///
+/// [`SetOnce::set`]: crate::sync::SetOnce::set
+#[derive(Debug, PartialEq, Eq)]
+pub struct SetOnceError<T> {
+    /// The cell was already initialized when [`SetOnce::set`] was called.
+    ///
+    /// [`SetOnce::set`]: crate::sync::OnceCell::set
+    val: T,
+}
```
Darksonn: I think it's fine to just do this:
```suggestion
pub struct SetOnceError<T>(pub T);
```
That lets you get rid of the `From` impl and write `SetOnceError(value)` instead of `SetOnceError::from(value)`. It also allows the end-use to get the value out of the error, which is the primary reason we return the value in the error.

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,409 @@
...
+
+impl<T> fmt::Display for SetOnceError<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "AlreadyInitializedError")
```
Darksonn:
```suggestion
        write!(f, "SetOnceError")
```

tokio/tests/sync_set_once.rs
```diff
@@ -0,0 +1,209 @@
...
+    static ONCE: SetOnce<u32> = SetOnce::const_new();
+
+    let first = ONCE.set(5);
+    assert_eq!(first, Ok(()));
+    let second = ONCE.set(6);
+    assert!(second.is_err());
+}
+
+#[test]
+fn set_while_initializing_or_already_init() {
```
Darksonn: This looks like it just tests the same thing as `set_twice`. No need to test the same thing twice.

tokio/tests/sync_set_once.rs
```diff
@@ -0,0 +1,209 @@
...
+    let rt = runtime::Builder::new_current_thread()
+        .enable_time()
+        .build()
+        .unwrap();
+
+    static ONCE: SetOnce<u32> = SetOnce::const_new();
+
+    rt.block_on(async {
+        let handle1 = rt.spawn(async {
+            time::sleep(Duration::from_millis(1)).await;
```
Darksonn: It's best to avoid sleeping in tests. Even short sleeps can add up over thousands of tests to a long time spent doing nothing.

tokio/tests/sync_set_once.rs
```diff
@@ -0,0 +1,209 @@
...
+    rt.block_on(async {
+        time::pause();
+
+        tokio::spawn(async { ONCE.set(20) });
+
+        tokio::spawn(async { ONCE.set(10) });
+
+        let res = ONCE.get();
+
+        assert_eq!(res, None);
```
Darksonn: This test is just a complex way to write this, so I think we can get rid of the test as I don't think it tests something that other tests don't already test.
```rs
static ONCE: SetOnce<u32> = SetOnce::const_new();
assert_eq!(ONCE.get(), None);
ONCE.set(20);
ONCE.set(10);
```
(well, technically the two last calls could happen in either order, but still)

tokio/tests/sync_set_once.rs
```diff
@@ -0,0 +1,209 @@
...
+    });
+}
+
+#[test]
+fn is_some_initializing() {
```
Darksonn: I think this test makes sense, but you can simplify it by just using `#[tokio::test]`.

tokio/tests/sync_set_once.rs
Darksonn, line 1: The `into_inner` method has slightly complex logic, so I think it makes sense to add some tests for it as well. It can just be two tests:

* What happens if you call `into_inner` on an empty `OnceCell`?
* What happens if you call `into_inner` on a non-empty `OnceCell`? (Using `Foo` to check drop count.)

tokio/tests/sync_set_once.rs
```diff
@@ -0,0 +1,209 @@
...
+    pub fn new(drops: Arc<AtomicU32>) -> Self {
+        Foo { drops }
+    }
+}
+
+impl Drop for Foo {
+    fn drop(&mut self) {
+        self.drops.fetch_add(1, Ordering::Release);
+    }
+}
```
Darksonn: We can simplify a lot of these tests by moving some common logic into `Foo` like this:
```rs
#[derive(Clone)]
struct DropCounter {
    drops: Arc<AtomicU32>,
}

impl DropCounter {
    fn new() -> Self {
        DropCounter {
	    drops: Arc::new(AtomicU32::new(0)),
	}
    }

    fn assert_num_drops(&self, value: u32) {
        assert_eq!(value, self.drops.load(Ordering::Relaxed));
    }
}

impl Drop for DropCounter {
    fn drop(&mut self) {
        self.drops.fetch_add(1, Ordering::Relaxed);
    }
}
```
Here you can clone `DropCounter` instead of cloning the `Arc` in tests.

tokio/tests/sync_set_once.rs
Darksonn: I just realized it's the same tests as `set_and_wait`.

tokio/src/sync/set_once.rs
ADD-SP: Another reason we want to remove the `impl<T> From<T> for SetOnceError<T>` is that downstream should not construct this error manually, so the `From` sugar is not needed.

tokio/src/sync/set_once.rs
Daksh14: added the fast path

tokio/src/sync/set_once.rs
Daksh14: Removed From impl and added the tuple struct

tokio/tests/sync_set_once.rs
Daksh14: removed the test with sleep

tokio/tests/sync_set_once.rs
Daksh14: removed the test with sleep

tokio/tests/sync_set_once.rs
Daksh14: Using DropCounter now and cloning it

tokio/tests/sync_set_once.rs
Daksh14: This is already checked by `drop_into_inner` and `drop_into_inner_new_with`

tokio/tests/sync_set_once.rs
Daksh14: yes but you mentioned its good to have the same tests so miri can also check it along with Loom

tokio/tests/sync_set_once.rs
Daksh14: rewrote this test, let me know

tokio/tests/sync_set_once.rs
ADD-SP, line 1: Would mind covering this case?

1. New an empty `SetOnce`.
2. Poll the `SetOnce::wait()`, this makes sure the `notify_fut` was created and we are waiting on it.
    https://github.com/tokio-rs/tokio/blob/25e0574f7b305abdd98cff14cf4f87241aa0edca/tokio/src/sync/set_once.rs#L357
3. Invoke `SetOnce::set()`.

Please make sure (2) always occurs before (3), so that we can cover the concern mentioned in https://github.com/tokio-rs/tokio/pull/7418#discussion_r2176268671.

tokio/src/sync/set_once.rs
ADD-SP: Could you remove the `pin!(notify_fut)`? I think this should be ok as long as there is no CI failures.

tokio/tests/sync_set_once.rs
Darksonn: I mean that it's good to have the same test both in `loom_set_once.rs` and in this file. I don't think you should have the same test twice when both are in this file.

tokio/tests/sync_set_once.rs
Darksonn: Both of those tests are the second case, because the `OnceCell` is empty in both tests. But I see that the `into_inner_int_empty_setonce` test takes care of that.

tokio/tests/sync_set_once.rs
```diff
@@ -0,0 +1,151 @@
...
+
+#[test]
+fn drop_into_inner_new_with() {
+    let fooer = DropCounter::new();
+
+    let once_cell = SetOnce::new_with(Some(fooer.clone()));
+    let val = once_cell.into_inner();
+    fooer.assert_num_drops(0);
+    mem::drop(val);
```
Darksonn:
```suggestion
    drop(val);
```

tokio/tests/sync_set_once.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+#[tokio::test]
+async fn set_and_wait() {
+    static ONCE: SetOnce<u32> = SetOnce::const_new();
+
+    let _ = tokio::spawn(async { ONCE.set(5) }).await;
+    let value = ONCE.wait().await;
+    assert_eq!(*value, 5);
+}
```
Darksonn: Can we also have a version of this test using `std::thread::spawn`?

And also a test that spawns two real threads and calls `set` from both, asserting that exactly one succeeds. Then we have tests that correspond to both loom tests.

tokio/src/sync/tests/loom_set_once.rs
```diff
@@ -0,0 +1,65 @@
...
+
+        let foo = Foo::new(Arc::clone(&drop_counter));
+
+        let set = set_once.set(foo).is_ok();
+        let res = thread.join().unwrap();
+
+        drop(set_once);
+
+        assert_eq!(drop_counter.load(Ordering::Acquire), 2);
+        assert!(res || set);
```
Darksonn: This succeeds even if *both* threads return true, but we want exactly one thread to return true. This will make the test more precise:
```suggestion
        assert!(res != set);
```

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,402 @@
...
+//      future will keep on waiting.
+//  2. When `value_set` is true, the wait() future completes, get() will return
+//      Some(&T)
+//
+// The value cannot be changed after set() is called. Subsequent calls to set()
+// will return a `SetOnceError`.
+
+/// A thread-safe cell that can be written to only once. A `SetOnce`
+/// is inspired from python's [`asyncio.Event`] type. It can be used to wait
+/// until the value of the `SetOnce` is set like a "Event" mechanism.
```
Darksonn: A newline here looks nicer in the docs.
```suggestion
/// A thread-safe cell that can be written to only once. 
///
/// A `SetOnce` is inspired from python's [`asyncio.Event`] type. It can be
/// used to wait until the value of the `SetOnce` is set like a "Event"
/// mechanism.
```

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,402 @@
...
+/// in Tokio allows the initialization procedure to be asynchronous.
+///
+/// # Example
+///
+/// ```
+/// use tokio::sync::{SetOnce, SetOnceError};
+///
+/// static ONCE: SetOnce<u32> = SetOnce::const_new();
+///
+/// #[tokio::main]
```
Darksonn: I would swap the examples so that this one is first.

tokio/tests/sync_set_once.rs
Darksonn: The `set_and_wait` test does exactly this. Tests use a single-threaded runtime, so it's guaranteed that `set` does not get called until `wait` yields.

tokio/tests/sync_set_once.rs
Daksh14: Same test with `std::thread::spawn` is now added called `set_and_wait_threads`  and test that spawns two real threads and calls `set` from both asserts one of them succeeds is `set_and_wait_multiple_threads`

tokio/src/sync/set_once.rs
Daksh14: @ADD-SP in my latest commit I removed `pin` and the CI passes so indeed it works without pin, I'm not sure about if somewhere there might be an edge case but the implementation being simple enough should be good

tokio/tests/sync_set_once.rs
ADD-SP: Emm, there is an `.await` at the line 101, the spawned task might be polled before the `ONCE.wait().await`.

https://github.com/tokio-rs/tokio/blob/3f8b1c669d70a833ee2a0b91198b7044a9135564/tokio/tests/sync_set_once.rs#L97-L104

Darksonn: You would be correct on the multi-threaded runtime. But not on the current-thread runtime, which `#[tokio::test]` uses by default.

tokio/tests/sync_set_once.rs
```diff
@@ -0,0 +1,180 @@
...
+#[tokio::test]
+#[cfg_attr(target_family = "wasm", ignore)]
+async fn set_and_wait_threads() {
+    static ONCE: SetOnce<u32> = SetOnce::const_new();
+
+    std::thread::spawn(|| {
+        ONCE.set(4).unwrap();
+    })
+    .join()
+    .unwrap();
```
Darksonn: Please call `.join()` after the call to `wait()`.

tokio/tests/sync_set_once.rs
```diff
@@ -0,0 +1,180 @@
...
+    ONCE.set(5).unwrap();
+    let value = ONCE.get().unwrap();
+    assert_eq!(*value, 5);
+}
+
+#[tokio::test]
+async fn set_and_wait() {
+    static ONCE: SetOnce<u32> = SetOnce::const_new();
+
+    let _ = tokio::spawn(async { ONCE.set(5) }).await;
```
Darksonn: Do not await the task. Otherwise you might as well do `ONCE.set(5)` without a task at all.

tokio/tests/sync_set_once.rs
Darksonn: Oh, you are right there is a bug in this test. I left https://github.com/tokio-rs/tokio/pull/7418#discussion_r2222444770

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,402 @@
...
+            let notify_fut = self.notify.notified();
+            // take the lock because we're reading value_set
+            let guard = self.lock.lock();
+
+            if self.value_set.load(Ordering::Relaxed) {
+                // SAFETY: the state is initialized
+                return unsafe { self.get_unchecked() };
+            }
+
+            drop(guard);
```
Darksonn: I want to add a scope here for the guard. There are some cases where using `drop(guard)` on a non-Send value causes Rust to still think that the future is non-Send when it should be Send. It looks like that may not be the case here as `async_send_sync.rs` seems to be okay, but I feel better with the scope.

I also took the opportunity to improve the comment.
```suggestion
            {
                // Taking the lock here ensures that a concurrent call to `set`
                // will see the creation of `notify_fut` in case the check
                // fails.
                let guard = self.lock.lock();

                if self.value_set.load(Ordering::Relaxed) {
                    // SAFETY: the state is initialized
                    return unsafe { self.get_unchecked() };
                }
            }
```

Darksonn (approved):

tokio/src/sync/set_once.rs
```diff
@@ -0,0 +1,404 @@
...
+            if let Some(val) = self.get() {
+                return val;
+            }
+
+            let notify_fut = self.notify.notified();
+            {
+                // Taking the lock here ensures that a concurrent call to `set`
+                // will see the creation of `notify_fut` in case the check
+                // fails.
+                let _ = self.lock.lock();
```
Darksonn:
```suggestion
                let _guard = self.lock.lock();
```

Issue #7421, opened 2025-06-25
impl Clone for tokio::net::unix::SocketAddr


xumaple:
Given that `tokio::net::unix::SocketAddr` is a wrapper type around std, and the std version does `impl Clone`, I'd have to imagine that the tokio version doesn't `impl Clone` simply because nobody has requested this feature yet.

I have a use case in which I would like for this to be implemented, seems like a simple derive call will get the trick done, and I'm happy to open a PR for it. 

Is there any reason I'm missing here for why this shouldn't/can't be done?

ADD-SP:
Implementing `Clone` for [`tokio::net::unix::SocketAddr`](https://docs.rs/tokio/1.45.1/tokio/net/unix/struct.SocketAddr.html) is reasonable to me.

PR #7422, opened 2025-06-26
impl Clone for net::unix::SocketAddr

xumaple:
## Motivation

`tokio::net::unix::SocketAddr` wraps a `std::os::unix::net::SocketAddr`. The latter implements `Clone`, but the former does not. This can be a bit annoying for certain niche use cases, such as when we want to use the `tokio` version in relation to `axum::extract::connect_info::Connected`. More than that, it simply makes sense to `impl Clone` here if the wrapped inner is already `impl Clone`.

## Solution

added a `#[derive(Clone)]`.

Closes https://github.com/tokio-rs/tokio/issues/7421.


ADD-SP (approved):
Since the std version impls the `Clone`, I think doing this for tokio version makes sense.

PR #7426, opened 2025-06-27
net: allow `pipe::OpenOptions::read_write` on Android

marshallpierce:
## Motivation

I'd like to read from a fifo forever on Android. However, `OpenOptions.read_write()` isn't present on Android.

## Solution

Tweaked the `cfg` to enable `read_write` on both Linux and Android.

I'm not sure if CI is set up for Android, but I enabled Android for the tests that I saw used read_write as well.


marshallpierce:
I should note that I tried this on an Android device. Before the change, `echo foo > fifo` caused reading to terminate (via `.lines().next_line()`), and after the change, reading continues.

Darksonn (approved):
Thank you.

PR #7431, opened 2025-06-27
[WIP] runtime: steal from the LIFO slot

hawkw:
Still a draft, wanted to get all the Loom CI to run.

### DIFF
--- a/tokio/src/runtime/scheduler/multi_thread/queue.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/queue.rs
@@ -451,6 +451,12 @@ impl<T> Steal<T> {
         // tasks in `dst`.
         let mut n = self.steal_into2(dst, dst_tail);
 
+        // Ooh, there's another task just sitting there? Grab that one, too!
+        let lifo = self.0.lifo.take();
+        if lifo.is_some() {
+            n += 1;
+        }
+
         if n == 0 {
             // No tasks were stolen
             return None;
@@ -462,22 +468,31 @@ impl<T> Steal<T> {
         // We are returning a task here
         n -= 1;
 
-        let ret_pos = dst_tail.wrapping_add(n);
-        let ret_idx = ret_pos as usize & MASK;
+        let ret = if lifo.is_some() {
+            // If we took the task from the LIFO slot, just return it as the
+            // next task to run, rather than messing around with the buffer.
+            lifo
+        } else {
+            // The LIFO slot was was empty, so take the last task we squirted
+            // into `dst` instead.
+            let ret_pos = dst_tail.wrapping_add(n);
+            let ret_idx = ret_pos as usize & MASK;
 
-        // safety: the value was written as part of `steal_into2` and not
-        // exposed to stealers, so no other thread can access it.
-        let ret = dst.inner.buffer[ret_idx].with(|ptr| unsafe { ptr::read((*ptr).as_ptr()) });
+            // safety: the value was written as part of `steal_into2` and not
+            // exposed to stealers, so no other thread can access it.
+            let task = dst.inner.buffer[ret_idx].with(|ptr| unsafe { ptr::read((*ptr).as_ptr()) });
+            Some(task)
+        };
 
         if n == 0 {
             // The `dst` queue is empty, but a single task was stolen
-            return Some(ret);
+            return ret;
         }
 
         // Make the stolen items available to consumers
         dst.inner.tail.store(dst_tail.wrapping_add(n), Release);
 
-        Some(ret)
+        ret
     }
 
     // Steal tasks from `self`, placing them into `dst`. Returns the number of
--- a/tokio/src/runtime/scheduler/multi_thread/queue.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/queue.rs
@@ -52,6 +52,13 @@ pub(crate) struct Inner<T: 'static> {
     /// Only updated by producer thread but read by many threads.
     tail: AtomicUnsignedShort,
 
+    /// When a task is scheduled from a worker, it is stored in this slot. The
+    /// worker will check this slot for a task **before** checking the run
+    /// queue. This effectively results in the **last** scheduled task to be run
+    /// next (LIFO). This is an optimization for improving locality which
+    /// benefits message passing patterns and helps to reduce latency.
+    lifo: task::AtomicNotified<T>,
+
     /// Elements
     buffer: Box<[UnsafeCell<MaybeUninit<task::Notified<T>>>; LOCAL_QUEUE_CAPACITY]>,
 }
@@ -92,6 +99,7 @@ pub(crate) fn local<T: 'static>() -> (Steal<T>, Local<T>) {
     let inner = Arc::new(Inner {
         head: AtomicUnsignedLong::new(0),
         tail: AtomicUnsignedShort::new(0),
+        lifo: task::AtomicNotified::empty(),
         buffer: make_fixed_size(buffer.into_boxed_slice()),
     });
 
@@ -110,7 +118,9 @@ impl<T> Local<T> {
         let (_, head) = unpack(self.inner.head.load(Acquire));
         // safety: this is the **only** thread that updates this cell.
         let tail = unsafe { self.inner.tail.unsync_load() };
-        len(head, tail)
+        // XXX(eliza): the `is_some` here might be racy? do we need to pack a
+        // LIFO-presence bit into one of the atomics?
+        len(head, tail) + (self.inner.lifo.is_some() as usize)
     }
 
     /// How many tasks can be pushed into the queue
@@ -387,6 +397,18 @@ impl<T> Local<T> {
         };
 
         Some(self.inner.buffer[idx].with(|ptr| unsafe { ptr::read(ptr).assume_init() }))
+    }
+
+    /// Pushes a task to the LIFO slot, returning the task previously in the
+    /// LIFO slot (if there was one).
+    pub(crate) fn push_lifo(&self, task: task::Notified<T>) -> Option<task::Notified<T>> {
+        self.inner.lifo.swap(Some(task))
+    }
+
+    /// Pops the task currently held in the LIFO slot, if there is one;
+    /// otherwise, returns `None`.
+    pub(crate) fn pop_lifo(&self) -> Option<task::Notified<T>> {
+        self.inner.lifo.take()
     }
 }
 
--- a/tokio/src/runtime/scheduler/multi_thread/worker.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/worker.rs
@@ -106,7 +106,7 @@ struct Core {
     /// queue. This effectively results in the **last** scheduled task to be run
     /// next (LIFO). This is an optimization for improving locality which
     /// benefits message passing patterns and helps to reduce latency.
-    lifo_slot: Option<Notified>,
+    lifo_slot: task::AtomicNotified<Arc<Handle>>,
 
     /// When `true`, locally scheduled tasks go to the LIFO slot. When `false`,
     /// they go to the back of the `run_queue`.
@@ -257,7 +257,7 @@ pub(super) fn create(
 
         cores.push(Box::new(Core {
             tick: 0,
-            lifo_slot: None,
+            lifo_slot: task::AtomicNotified::empty(),
             lifo_enabled: !config.disable_lifo_slot,
             run_queue,
             is_searching: false,
@@ -1084,17 +1084,15 @@ impl Handle {
             true
         } else {
             // Push to the LIFO slot
-            let prev = core.lifo_slot.take();
-            let ret = prev.is_some();
+            let prev = core.lifo_slot.swap(Some(task));
 
             if let Some(prev) = prev {
                 core.run_queue
                     .push_back_or_overflow(prev, self, &mut core.stats);
+                true
+            } else {
+                false
             }
-
-            core.lifo_slot = Some(task);
-
-            ret
         };
 
         // Only notify if not currently parked. If `park` is `None`, then the
--- a/tokio/src/runtime/scheduler/multi_thread/worker.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/worker.rs
@@ -101,13 +101,6 @@ struct Core {
     /// Used to schedule bookkeeping tasks every so often.
     tick: u32,
 
-    /// When a task is scheduled from a worker, it is stored in this slot. The
-    /// worker will check this slot for a task **before** checking the run
-    /// queue. This effectively results in the **last** scheduled task to be run
-    /// next (LIFO). This is an optimization for improving locality which
-    /// benefits message passing patterns and helps to reduce latency.
-    lifo_slot: task::AtomicNotified<Arc<Handle>>,
-
     /// When `true`, locally scheduled tasks go to the LIFO slot. When `false`,
     /// they go to the back of the `run_queue`.
     lifo_enabled: bool,
@@ -257,7 +250,6 @@ pub(super) fn create(
 
         cores.push(Box::new(Core {
             tick: 0,
-            lifo_slot: task::AtomicNotified::empty(),
             lifo_enabled: !config.disable_lifo_slot,
             run_queue,
             is_searching: false,
@@ -405,7 +397,7 @@ where
         // If we heavily call `spawn_blocking`, there might be no available thread to
         // run this core. Except for the task in the lifo_slot, all tasks can be
         // stolen, so we move the task out of the lifo_slot to the run_queue.
-        if let Some(task) = core.lifo_slot.take() {
+        if let Some(task) = core.run_queue.pop_lifo() {
             core.run_queue
                 .push_back_or_overflow(task, &*cx.worker.handle, &mut core.stats);
         }
@@ -617,7 +609,7 @@ impl Context {
                 };
 
                 // Check for a task in the LIFO slot
-                let task = match core.lifo_slot.take() {
+                let task = match core.run_queue.pop_lifo() {
                     Some(task) => task,
                     None => {
                         self.reset_lifo_enabled(&mut core);
@@ -852,7 +844,7 @@ impl Core {
     }
 
     fn next_local_task(&mut self) -> Option<Notified> {
-        self.lifo_slot.take().or_else(|| self.run_queue.pop())
+        self.run_queue.pop_lifo().or_else(|| self.run_queue.pop())
     }
 
     /// Function responsible for stealing tasks from another worker
@@ -908,7 +900,7 @@ impl Core {
     }
 
     fn has_tasks(&self) -> bool {
-        self.lifo_slot.is_some() || self.run_queue.has_tasks()
+        self.run_queue.has_tasks()
     }
 
     fn should_notify_others(&self) -> bool {
@@ -917,7 +909,7 @@ impl Core {
         if self.is_searching {
             return false;
         }
-        self.lifo_slot.is_some() as usize + self.run_queue.len() > 1
+        self.run_queue.len() > 1
     }
 
     /// Prepares the worker state for parking.
@@ -1084,7 +1076,7 @@ impl Handle {
             true
         } else {
             // Push to the LIFO slot
-            let prev = core.lifo_slot.swap(Some(task));
+            let prev = core.run_queue.push_lifo(task);
 
             if let Some(prev) = prev {
                 core.run_queue
--- a/tokio/src/runtime/task/atomic_notified.rs
+++ b/tokio/src/runtime/task/atomic_notified.rs
@@ -1,0 +1,58 @@
+use crate::loom::sync::atomic::AtomicPtr;
+use crate::runtime::task::{Header, Notified, RawTask};
+
+use std::marker::PhantomData;
+use std::ptr;
+use std::ptr::NonNull;
+use std::sync::atomic::Ordering::{AcqRel, Acquire};
+
+/// An atomic cell which can contain a pointer to a [`Notified`] task.
+///
+/// This is similar to the [`crate::util::AtomicCell`] type, but specialized to
+/// hold a task pointer --- this type "remembers" the task's scheduler generic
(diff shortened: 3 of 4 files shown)


hawkw:
(shoutout to @mkeeter for nerd-sniping me into actually doing this)

tokio/src/runtime/scheduler/multi_thread/queue.rs
```diff
@@ -429,33 +452,48 @@ impl<T> Steal<T> {
...
+
+        if n == 0 && lifo.is_none() {
             // No tasks were stolen
             return None;
         }
 
-        dst_stats.incr_steal_count(n as u16);
+        // If we also grabbed the task from the LIFO slot, include that in the
+        // steal count as well.
+        dst_stats.incr_steal_count(n as u16 + lifo.is_some() as u16);
```
ADD-SP: For reviewers: Rust has defined the casting from `bool` to `u16`.

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -1090,17 +1082,15 @@ impl Handle {
...
                     .push_back_or_overflow(prev, self, &mut core.stats);
+                true
+            } else {
+                false
             }
-
-            core.lifo_slot = Some(task);
-
-            ret
         };
```
ADD-SP: I took a longer time than the old code to understand the new one. Is `match core.run_queue.push_lifo(task)` more readable?

tokio/src/runtime/scheduler/multi_thread/queue.rs
mkeeter: It's also possible to use `u16::from(bool)`, if you care about [`cast_lossless`](https://rust-lang.github.io/rust-clippy/stable/index.html#cast_lossless) cleanliness.

tokio/src/runtime/scheduler/multi_thread/queue.rs
```diff
@@ -429,33 +452,48 @@ impl<T> Steal<T> {
...
+            // The LIFO slot was was empty, so take the last task we squirted
+            // into `dst` instead.
+            let ret_pos = dst_tail.wrapping_add(n);
+            let ret_idx = ret_pos as usize & MASK;
+
+            // safety: the value was written as part of `steal_into2` and not
+            // exposed to stealers, so no other thread can access it.
+            let task = dst.inner.buffer[ret_idx].with(|ptr| unsafe { ptr::read((*ptr).as_ptr()) });
+            Some(task)
+        };
```
mkeeter: Nit: the diff could be minimized by keeping `ret` as a `task::Notified<T>`, instead of an `Option<..>` which is always `Some(..)`.

```suggestion
        let ret = if let Some(lifo) = lifo {
            // If we took the task from the LIFO slot, just return it as the
            // next task to run, rather than messing around with tasks from the
            // queue.
            lifo
        } else {
            // We are returning a task from the queue here.
            n -= 1;

            // The LIFO slot was was empty, so take the last task we squirted
            // into `dst` instead.
            let ret_pos = dst_tail.wrapping_add(n);
            let ret_idx = ret_pos as usize & MASK;

            // safety: the value was written as part of `steal_into2` and not
            // exposed to stealers, so no other thread can access it.
            dst.inner.buffer[ret_idx].with(|ptr| unsafe { ptr::read((*ptr).as_ptr()) })
        };
```

(This eliminates changes at the return sites later on)

hawkw: Thanks, that's definitely a little tidier!

tokio/src/runtime/scheduler/multi_thread/worker.rs
hawkw: How about https://github.com/tokio-rs/tokio/pull/7431/commits/adf146f2fec6dff680cd8ed03d8a0cf06da80960 ?

tokio/src/runtime/tests/loom_multi_thread/queue.rs
```diff
@@ -116,32 +175,59 @@ fn steal_overflow() {
 fn multi_stealer() {
```
ADD-SP, line 175: For reviewers: This test is not changed, except the inner `fn steal_tasks(..) -> usize` is moved out of this test for reusing in test `multi_stealer_lifo`.

tokio/tests/rt_threaded.rs
```diff
@@ -692,6 +692,77 @@ fn mutex_in_block_in_place() {
...
+
+        let result = tokio::time::timeout(Duration::from_secs(30), blocked_task_joined).await;
+        println!("[main] result: {result:?}");
+        // Before possibly panicking, make sure that we wake up the blocked task
+        // so that it doesn't stop the runtime from shutting down.
+        block_thread_tx.send(()).unwrap();
+        churn.abort();
+        result
+            .expect("task in LIFO slot should complete within 30 seconds")
+            .expect("task in LIFO slot should not panic");
```
ADD-SP:
```suggestion
        // 30 seconds is an arbitrary number, we add timeout here to avoid hanging
        // this test forever if something goes wrong.
        let result = tokio::time::timeout(Duration::from_secs(30), blocked_task_joined).await;
        println!("[main] result: {result:?}");
        // Before possibly panicking, make sure that we wake up the blocked task
        // so that it doesn't stop the runtime from shutting down.
        block_thread_tx.send(()).unwrap();
        churn.abort();
        result
            .expect("task in LIFO slot should complete within a reasonable time")
            .expect("task in LIFO slot should not panic");
```

Is my understanding correct?

tokio/tests/rt_threaded.rs
```diff
@@ -692,6 +692,77 @@ fn mutex_in_block_in_place() {
...
+            }
+        });
+
+        let blocked_task_joined = tokio::spawn(async move {
+            println!("[LIFO] task started");
+            task_started_tx.send(()).unwrap();
+            println!("[LIFO] task waiting for wakeup...");
+            unblock_rx.await.unwrap();
+            println!("[LIFO] task running after wakeup");
+        });
```
ADD-SP:
```suggestion
        // this task will be pushed into the LIFO slot
        let blocked_task_joined = tokio::spawn(async move {
            // this task was taken out the LIFO slot
            println!("[LIFO] task started");
            task_started_tx.send(()).unwrap();
            println!("[LIFO] task waiting for wakeup...");
            unblock_rx.await.unwrap();
            println!("[LIFO] task running after wakeup");
        });
```

tokio/tests/rt_threaded.rs
```diff
@@ -692,6 +692,77 @@ fn mutex_in_block_in_place() {
...
+        // Wait for the blocked task to have been polled once and have yielded
+        // before we spawn the task that will notify it.
+        task_started_rx.await.unwrap();
+        println!("[main] LIFO slot task start acked!");
+
+        // Now, spawn a task that will notify the blocked task before going
+        // blocking forever.
+        tokio::spawn(async move {
+            println!("[blocker] sending wakeup");
+            unblock_tx.send(()).unwrap();
```
ADD-SP:
```suggestion
            // this wakes up the `blocked_task_joined ` and pushes
            // it into the LIFO slot
            unblock_tx.send(()).unwrap();
```

tokio/tests/rt_threaded.rs
```diff
@@ -692,6 +692,77 @@ fn mutex_in_block_in_place() {
...
+
+            println!("[blocker] blocking the worker thread...");
+            // Block the worker thread indefinitely by waiting for a message on
+            // a blocking channel. Using a channel rather than e.g. `loop {}`
+            // allows us to terminate the task cleanly when the test finishes.
+            let _ = block_thread_rx.recv();
```
ADD-SP, line 775:
```suggestion
            // Block the worker thread indefinitely by waiting for a message on
            // a blocking channel, so `blocked_task_joined` will never be
            // taken by this worker, so other workers have a chance to
            // steal it.
            // 
            // Using a channel rather than e.g. `loop {}`
            // allows us to terminate the task cleanly when the test finishes.
            let _ = block_thread_rx.recv();
```

ADD-SP (commented):
It's not easy to understand test `lifo_stealable`, so I've proposed some comments to help with understanding.

I also have a question, did we have some performance regression test that covers the relevant code path? It would be better to confirm no significant regression as this PR changes the hot path a lot.

tokio/tests/rt_threaded.rs
```diff
@@ -692,6 +692,77 @@ fn mutex_in_block_in_place() {
...
+        // I/O driver and another can be parked running the timer wheel and
+        // there's still at least one worker free to steal the blocked task.
+        .worker_threads(4)
+        .enable_time()
+        .build()
+        .unwrap();
+
+    rt.block_on(async {
+        // Keep the runtime busy so that the workers that might steal the
+        // blocked task don't all park themselves forever.
```
ADD-SP, line 734:
```suggestion
        // Keep the runtime busy so that the workers that might steal the
        // blocked task don't all park themselves forever.
        //
        // Since multi-thread scheduler doesn't push to the LIFO slot if it
        // it not empty, so this task will be pushed into the global inject
        // queue eventually, and unpark another worker thread that will steal
        // the `blocked_task_joined` from the LIFO slot.
```

I added some comments to help me to understand this test, but it looks too long, feel free to rewrite it.

hawkw: As I understand it, the reason this task doesn't go in the LIFO slot isn't because of the injector queue behavior, it's because it's being woken up by a timer,  and is therefore woken by whatever worker thread ends up running the time driver, rather than the worker that was last running this task.

hawkw:
@ADD-SP thanks for your suggested comment improvements for the test! I ended up adding a bunch of comments in 0889f08f27ee6f92582d6b4ad4cd10b1d4ae51a1 --- I wanted to also add a higher-level overview at the beginning of the test explaining what the different channels are used for, and I changed the names of some things. Let me know what you think? Hopefully it's a bit easier to understand now.

tokio/src/runtime/tests/loom_multi_thread/queue.rs
hawkw: Yeah, that's correct.

hawkw:
@ADD-SP:
> I also have a question, did we have some performance regression test that covers the relevant code path? It would be better confirming no significant regression as this PR changes the hot path a lot.
> 
> If there is a performance regression, we can discuss if this is acceptable for supporting steal-able LIFO slot.

I think the existing `rt_multi_threaded.rs` benchmarks should exercise the code changed here --- in particular, I think it's [the `rt_multi_ping_pong` benchmark](https://github.com/tokio-rs/tokio/blob/171cd148a37da40dcbb8b06bf2c67634b2ba1f87/benches/rt_multi_threaded.rs#L174-L217) in that file exercises the performance improvement from the LIFO slot optimization, since it represents a situation where two tasks send messages back and forth.

I'll go run these benchmarks and report my results. However, I think it might be worthwhile to also compare performance on an ARM system. I would guess that if there's a noticeable performance regression from changing the LIFO slot from a mutable local pointer to a shared `AtomicPointer` would be more noticeable with ARM's weaker memory semantics. On x86 systems, changing accesses to the LIFO slot pointer to atomic accesses with `AcqRel` ordering should be basically the same as a non-atomic access, but these will result in different instructions on an ARM system.

Unfortunately, I don't have easy access to a multi-core ARM machine to run these benchmarks on. If someone else with an ARM machine, like an Apple Silicon Mac, wants to run the benches on ARM, I'd love to see the results.

hawkw:
This [FreeBSD CI run failure](https://github.com/tokio-rs/tokio/pull/7431/checks?check_run_id=45315301706) looks a bit weird --- it looks like maybe something timed out? I'm re-running it.

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -1090,17 +1082,16 @@ impl Handle {
...
-                    .push_back_or_overflow(prev, self, &mut core.stats);
+            match core.run_queue.push_lifo(task) {
+                Some(prev) => {
+                    // There was a previous task in the LIFO slot which needs
+                    // to be pushed to the back of the run queue.
+                    core.run_queue
+                        .push_back_or_overflow(prev, self, &mut core.stats);
+                    true
+                }
+                None => false,
```
Darksonn: This boolean indicates whether other worker threads are notified that there is work available to steal. Since the lifo slot is now stealable, it should no longer be false in this scenario.

Darksonn: Is this something we could have caught with a loom test? Maybe put something in the lifo slot and then block the thread, concurrently with another thread doing nothing?

tokio/src/runtime/scheduler/multi_thread/queue.rs
martin-g, line 600: Should this be updated to also check the lifo ?

hawkw:
Okay, I'm trying to page this one back in, as I'd really like to be able to get it fixed.

Unfortunately, [the test failure on `aarch64-pc-windows-msvc` with `parking_lot`](https://github.com/tokio-rs/tokio/actions/runs/22974412770/job/66699541085) is kinda sketching me out. The test that fails is `rt_unstable_metrics::worker_local_schedule_count`:

```
        FAIL [   0.095s] ( 667/1280) tokio::rt_unstable_metrics worker_local_schedule_count
  stdout ───

    running 1 test
    test worker_local_schedule_count ... FAILED

    failures:

    failures:
        worker_local_schedule_count

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 23 filtered out; finished in 0.07s
    
  stderr ───

    thread 'worker_local_schedule_count' (2232) panicked at tokio\tests\rt_unstable_metrics.rs:598:5:
      left: 2
     right: 1
    stack backtrace:

```

This test fails only on the ARM builds with `parking_lot` (or at least, on the `aarch64-pc-windows-msvc` one, the other ARM parking lot builds were cancelled). The ARM builds without `parking_lot` (`armv5te-unknown-linux-gnueabi`, `armv7-unknown-linux-gnueabihf`, `aarch64-unknown-linux-gnu`, and `aarch64-pc-windows-msvc`) all pass.

I do think this failure is _probably_ indicative of a race that actually results in us missing a schedule someplace, which is uncomfortable. Since the worker metrics are only written from the worker and are only read across threads in this test _after_ everything on the runtime has happened and the workers have shut down, I don't _think_ it's just a case of missing the counter increment. 

Changing the `AtomicNotified` cell to use `SeqCst` rather than `AcqRel` ordering in 3a6043d7adadf7103cb459efeb9cc7039306ffac fixes this test. I don't _like_ that this fixes it, since it suggests to me that the `SeqCst` is necessary to synchronize with...something else...somewhere else in the runtime, and I haven't figured out what that is. It's extra sketchy that this doesn't happen with `std::sync::Mutex` but that does at least explain why none of the `loom` tests catch this --- they wouldn't catch any races specific to `parking_lot`, and loom also doesn't simulate sequential consistency anyway.

I don't _really_ want to move forwards with just blindly changing the operations to `SeqCst` without an understanding of what precisely requires that additional guarantee, but...well, it does fix it.

tokio/src/runtime/scheduler/multi_thread/queue.rs
```diff
@@ -429,23 +452,37 @@ impl<T> Steal<T> {
         // tasks in `dst`.
         let mut n = self.steal_into2(dst, dst_tail);
 
-        if n == 0 {
+        // Ooh, there's another task just sitting there? Grab that one, too!
+        let lifo = self.0.lifo.take();
+
+        if n == 0 && lifo.is_none() {
             // No tasks were stolen
             return None;
```
Darksonn: We could simplify the implementation a lot by doing this:
```suggestion
        if n == 0 {
            // No tasks were stolen, try the LIFO slot.
            return self.0.lifo.take();
```
and then just remove all the changes below. This may even be desirable anyway?

Darksonn:
What probably happened in the test is this:

1. The outer tasks spawns the inner task into LIFO slot.
2. Inner task stolen by other thread.
3. Other thread finishes inner task.
4. Outer task checks inner task for completion, and finds it was already completed, and does not yield.

This will eliminate the second call to 'schedule' when the inner task completes, which normally wakes up the outer task, as the outer task has normally already yielded at that point. But now the LIFO slot can be stolen, so the inner task may complete before the outer task yields, eliminating a yield (and hence schedule) in outer task.

tokio/src/runtime/scheduler/multi_thread/queue.rs
hawkw: Hm, that's an interesting thought, this way the LIFO task would only be stolen if there are no other tasks in the queue. That might, indeed, be desirable, as it means the LIFO task is more likely to be polled again from the thread it was notified on, and only gets stolen by another thread if that thread's queue is empty...which could be a reasonable heuristic for the case I actually care about, the one where the worker is very busy or permanently stuck; if the queue is totally empty that would capture the case where the worker has been stuck long enough that someone else already came along and stole all its tasks. I'll have to think it through a little longer since I want to make sure the lifo slot always gets stolen if the worker is in a wedged state.

hawkw: I did this in https://github.com/tokio-rs/tokio/pull/7431/changes/c47b9c87f91303c2494fd5d6778084e57576c679, it certainly makes the diff a lot tidier, and I think it _may_ reduce the amount of cross-thread synchronization a bit by only stealing LIFO tasks the *second* time a worker gets stolen from if nothing else has been woken on it since the last time. Which I _think_ should mean that in the case where the worker is blocked on something for a long time, the LIFO task will still eventually get rescued from the stuck worker, but if the worker is not blocked for long enough to get stolen from twice, the LIFO task is less likely to move across threads, which is...probably good?

Darksonn: A task with n pending tasks will have to get stolen from log(n) times before the lifo slot is taken, because it always steals half of the tasks, and you need that many halvings to take everything.

tokio/src/runtime/tests/loom_multi_thread/queue.rs
Darksonn, line 1: How long do these loom tests take? They look really complex and I'm concerned about increasing CI by hours.

tokio/tests/rt_threaded.rs
```diff
@@ -692,6 +692,112 @@ fn mutex_in_block_in_place() {
...
+        //
+        // Since this task will always be woken by whichever worker is holding
+        // the time driver, rather than a worker that's executing tasks, it
+        // shouldn't ever kick the victim task out of its worker's LIFO slot.
+        let churn = tokio::spawn(async move {
+            loop {
+                tokio::time::sleep(Duration::from_millis(64)).await;
```
Darksonn: No reason for this test to take that long.
```suggestion
                tokio::time::sleep(Duration::from_millis(4)).await;
```

Darksonn:
We need some docs updates. Here are a few places:

* https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.disable_lifo_slot
* https://docs.rs/tokio/latest/tokio/runtime/index.html#multi-threaded-runtime-behavior-at-the-time-of-writing

Please also grep for 'lifo' to see if there are others.

tokio/src/runtime/tests/loom_multi_thread/queue.rs
hawkw: On my machine, at least, these really don't take that long: with a preemption bound of 2 and 10000 max branches, I can get through the whole `queue` module in 6.6 seconds.

```console
eliza@hekate ~/Code/tokio $ LOOM_MAX_PREEMPTIONS=2 LOOM_MAX_BRANCHES=10000 RUSTFLAGS="--cfg loom -C debug_assertions" LOOM_LOCATION=true LOOM_LOG=trace \
    cargo nextest run --lib --release --features full loom_multi_thread::queue
shell  0.54s user 0.19s system 0% cpu 2:46.62 total
children  45.49s user 4.97s system 30% cpu 2:46.62 total
    Finished `release` profile [optimized] target(s) in 0.05s
────────────
 Nextest run ID 8bd7c7ec-4cdb-4cd4-8231-09d490cd0f59 with nextest profile: default
    Starting 6 tests across 7 binaries (112 tests skipped)
        PASS [   0.222s] tokio runtime::tests::loom_multi_thread::queue::chained_steal
        PASS [   0.407s] tokio runtime::tests::loom_multi_thread::queue::multi_stealer
        PASS [   0.571s] tokio runtime::tests::loom_multi_thread::queue::basic_lifo
        PASS [   0.658s] tokio runtime::tests::loom_multi_thread::queue::steal_overflow
        PASS [   4.609s] tokio runtime::tests::loom_multi_thread::queue::multi_stealer_lifo
        PASS [   6.598s] tokio runtime::tests::loom_multi_thread::queue::basic
────────────
     Summary [   6.658s] 6 tests run: 6 passed, 112 skipped
```

hawkw: of course, this is running on:

```console
eliza@hekate ~/Code/tokio $ inxi -C
CPU:
  Info: 24-core model: AMD Ryzen Threadripper 9960X s bits: 64 type: MT MCP cache: L2: 24 MiB
  Speed (MHz): avg: 3080 min/max: 400/5487 cores: 1: 3080 2: 3080 3: 3080 4: 3080 5: 3080
    6: 3080 7: 3080 8: 3080 9: 3080 10: 3080 11: 3080 12: 3080 13: 3080 14: 3080 15: 3080 16: 3080
    17: 3080 18: 3080 19: 3080 20: 3080 21: 3080 22: 3080 23: 3080 24: 3080 25: 3080 26: 3080
    27: 3080 28: 3080 29: 3080 30: 3080 31: 3080 32: 3080 33: 3080 34: 3080 35: 3080 36: 3080
    37: 3080 38: 3080 39: 3080 40: 3080 41: 3080 42: 3080 43: 3080 44: 3080 45: 3080 46: 3080
    47: 3080 48: 3080
```
so i'm not sure if the CI workers are clocking that high, or how many cores they're given. but, i don't think these tests are going to run for hours?

PR #7436, opened 2025-07-01
runtime: fix handling of cancelled Ops

mox692:
## Motivation
As mentioned in https://github.com/tokio-rs/tokio/pull/7321#pullrequestreview-2973357507, there was a deadlock scenario in the previous uring-related code under the following scenario:

1. The user submits an io_uring operation (Op).
2. The Op completes, the driver [gets an notification](https://github.com/tokio-rs/tokio/blob/3e890cc0171ddb210acdcfec831b7c7bcbb0d2d9/tokio/src/runtime/io/driver.rs#L205), and the driver [updates](https://github.com/tokio-rs/tokio/blob/3e890cc0171ddb210acdcfec831b7c7bcbb0d2d9/tokio/src/runtime/io/driver/uring.rs#L91) the lifecycle of the corresponding Op to `Completed`. 
3. Before the Op is polled again, (due to runtime shutdown or similar reasons) it gets [cancelled](https://github.com/tokio-rs/tokio/blob/3e890cc0171ddb210acdcfec831b7c7bcbb0d2d9/tokio/src/runtime/driver/op.rs#L65-L81).
    - At this point, the lifecycle is [updated from `Completed` to `Cancelled`](https://github.com/tokio-rs/tokio/blob/3e890cc0171ddb210acdcfec831b7c7bcbb0d2d9/tokio/src/runtime/io/driver/uring.rs#L289).
4. Next, the `UringContext` gets dropped (due to runtime shutdown).
5. In `UringContext::drop()`, it waits for the completion of all in-flight operations. 
	* Note that the cancelled Op mentioned above still remains in `cancel_ops` with its lifecycle, so we run [this while-loop](https://github.com/tokio-rs/tokio/blob/3e890cc0171ddb210acdcfec831b7c7bcbb0d2d9/tokio/src/runtime/io/driver/uring.rs#L169).
6. However, the Op had actually already completed, and the `IoDriver` had already consumed its notification in step 2. Therefore, the [`submit_and_wait(1)`](https://github.com/tokio-rs/tokio/blob/3e890cc0171ddb210acdcfec831b7c7bcbb0d2d9/tokio/src/runtime/io/driver/uring.rs#L172C18-L172C33) call here ends up hanging indefinitely.

This issue can be reproduced (relatively frequently) by running the test in [dff8711](https://github.com/tokio-rs/tokio/pull/7321/commits/dff87112ac7475b55ac81171d29fcd9acb543128) of #7321. 

## Solution
For Ops that have completed but are cancelled before being polled again, we now drop the ops within `cancel_op()`.  
This fixes the issue where already completed Ops would be waited on indefinitely.

### DIFF
--- a/tokio/src/net/tcp/socket.rs
+++ b/tokio/src/net/tcp/socket.rs
@@ -81,7 +81,7 @@ cfg_net! {
     ///
     /// [`RawFd`]: https://doc.rust-lang.org/std/os/fd/type.RawFd.html
     /// [`RawSocket`]: https://doc.rust-lang.org/std/os/windows/io/type.RawSocket.html
-    /// [`AsRawFd`]: https://doc.rust-lang.org/std/os/unix/io/trait.AsRawFd.html
+    /// [`AsRawFd`]: https://doc.rust-lang.org/std/os/fd/trait.AsRawFd.html
     /// [`AsRawSocket`]: https://doc.rust-lang.org/std/os/windows/io/trait.AsRawSocket.html
     /// [`socket2`]: https://docs.rs/socket2/
     #[cfg_attr(docsrs, doc(alias = "connect_std"))]
--- a/tokio/src/net/unix/pipe.rs
+++ b/tokio/src/net/unix/pipe.rs
@@ -120,7 +120,7 @@ pub fn pipe() -> io::Result<(Sender, Receiver)> {
 /// ```
 #[derive(Clone, Debug)]
 pub struct OpenOptions {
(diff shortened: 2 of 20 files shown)


mox692:
The commit [de2da05](https://github.com/mox692/tokio/pull/72/commits/de2da0598407a52e336894d695c9a6f7943868f7) will be reverted and this PR will be re-targeted to master once we've verified the ci gets green.

mox692:
The CI on de2da05 has passed , so this should be ready for review.

PR #7440, opened 2025-07-02
runtime(unstable): fix task hook spawn locations for `tokio::spawn`

hawkw:
## Motivation

Unfortunately, due to an oversight on my part, the capturing of spawn
locations was only tested with the `Runtime::spawn` method, and *not*
with `tokio::spawn`/`tokio::task::spawn`, which is how most tasks are
spawned in Real Life. And, it turned out that because this was not
tested...well, it was broken. Agh. My bad.

## Solution

Although the whole call chain for spawning tasks using `tokio::spawn`
was correctly annotated with `#[track_caller]`, the location wasn't
propagated correctly because of the `context::with_current(|handle| {
... })` closure that accesses the current runtime. Because the call to
spawn the task occurs inside a closure, the *closure*'s location is
captured instead of the caller. This means any task spawned by
`tokio::spawn` records its location as being in
`tokio/src/task/spawn.rs`, which is not what we'd like. This commit
fixes that by capturing the spawn location outside the `with_current`
closure and passing it in explicitly.

I've updated the tests to also spawn a task with `tokio::spawn`, so that
we ensure this works correctly.

### DIFF
--- a/tokio/src/runtime/handle.rs
+++ b/tokio/src/runtime/handle.rs
@@ -329,7 +329,7 @@ impl Handle {
     }
 
     #[track_caller]
-    pub(crate) fn spawn_named<F>(&self, future: F, _meta: SpawnMeta<'_>) -> JoinHandle<F::Output>
+    pub(crate) fn spawn_named<F>(&self, future: F, meta: SpawnMeta<'_>) -> JoinHandle<F::Output>
     where
         F: Future + Send + 'static,
         F::Output: Send + 'static,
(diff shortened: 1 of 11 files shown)


hds:
@hawkw How about we include tests for these cases using tracing-mock as well?

This provides us with pretty clear integrations tests of this functionality.

I had only covered one case up until now (for the spawn location), but adding more should be easy enough. Like this one:

https://github.com/tokio-rs/tokio/blob/master/tokio/tests/tracing_task.rs#L79

hawkw:
@hds I'd be happy to add tests for the tracing integration as well! However, that was previously using a different code path and wouldn't actually have caught this issue --- the locations for `tracing` spans were being captured in a different place than where the one that's put in the task for the runtime hooks. Now, we're using the same code path for both, which is the only reason the tracing  bits were touched in this branch.

Do you have a preference on whether to add those tests in this PR or a separate one?

hds:
In this case I think that another PR works fine.

PR #7453, opened 2025-07-09
chore(runtime): reorder metrics to be grouped by cfg-gates

jlizen:
## Motivation

https://github.com/tokio-rs/tokio/pull/7449 fixed the generated cfg gates in our rustdocs for runtime metrics. In the interest of reviewability, we avoided reordering the metric methods. 

However this resulted in interwoven groups of cfg-gates and multiple macro declarations per each. This hurts maintainability and readability.

## Solution

I reorganized the module to group all like-gated methods together.

### DIFF
--- a/tokio/src/runtime/metrics/runtime.rs
+++ b/tokio/src/runtime/metrics/runtime.rs
@@ -342,6 +342,225 @@ impl RuntimeMetrics {
                 .worker_metrics(worker)
                 .thread_id()
         }
+
+        /// Renamed to [`RuntimeMetrics::global_queue_depth`]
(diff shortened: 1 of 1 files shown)


ADD-SP (commented):
It would be better splitting a huge commit into serval commit, one method, one commit. This could make the code review easier.

jlizen:
@ADD-SP  I'm probably not going to have time to split each moved metric into its own commit. I can go ahead and close this PR out? Want an issue to consolidate the existing metrics into one macro invocation feature combo, for maintainability? Or do you not mind so much the current state?

PR #7454, opened 2025-07-11
task: opens `Id::as_u64` to clients

aaronjeline:
Allows clients to observe the contents of Task IDs, with no guarantees on contents.

Fixes: #7430

## Motivation

Solves the ask in the referenced issue (#7430) to allow a lightweight/cheap way to see the contents of a task Id. 

## Solution

Simply opens `task::Id::as_u64` to `pub` instead of `pub(crate)`

### DIFF
--- a/tokio/src/runtime/task/id.rs
+++ b/tokio/src/runtime/task/id.rs
@@ -74,7 +74,21 @@ impl Id {
         }
     }
 
-    pub(crate) fn as_u64(&self) -> u64 {
+    /// Retrieves the underlying [`u64`] for this [`Id`]
+    /// There are no guarantees about the order or contents of this [`u64`].
+    ///
+    /// # Example
+    ///
+    /// ```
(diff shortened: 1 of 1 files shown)


aaronjeline:
If maintainers would prefer this being on the `tokio_unstable` feature happy to change that

ADD-SP (commented):
#7430 wants to expose this method for getting rid of the ugly unsafe hack of tokio-dtrace, would you mind talking about your requirements?

We currently has no enough motivation to expose a stable representation of task id.

An alternative is gating this method by `tokio_unstable`.

However, my feeling is that once we open up the this unstable method, it will end up being used so heavily in the observability infrastructure, leading to a lot of resistance when we try to break this interface in the future, even this is `tokio_unstable`.

aaronjeline:
Looks like the big requirement is being able to observe a task id without allocating

PR #7456, opened 2025-07-11
Added composition decoder and encoder

tomDev5:
Motivation detailed in #7455 , put the code from the POC as part of `tokio_util`.


ADD-SP:
Thanks for you contribution! I believe that third-party crates are a more appropriate place to implement such features. See also <https://github.com/tokio-rs/tokio/issues/7455#issuecomment-3210588372>.

PR #7461, opened 2025-07-16
time: delay the `Arc::clone` of the `scheduler::Handle` until registering timer

ADD-SP:
*blocks #7467*

## Background

This improvement was found while working on the delayed cancellation (#7384),
Since I don't like to include a un-relevant change into a big patch, I made it a separate commit

This might be a low-hanging fruit.

## Motivation

https://github.com/tokio-rs/tokio/blob/0a3fe460862720d005b8fa39cdde30575a063785/tokio/src/time/sleep.rs#L250-L256

The current implementation always clone the `scheduler::Handle` for each timer, even this timer is not registered.

There are two usage of this handle for timer:

1. Ensure the time driver is enabled.
    https://github.com/tokio-rs/tokio/blob/0a3fe460862720d005b8fa39cdde30575a063785/tokio/src/runtime/time/entry.rs#L480-L484
2. Registering or clear the entry from the global wheel.
    https://github.com/tokio-rs/tokio/blob/0a3fe460862720d005b8fa39cdde30575a063785/tokio/src/runtime/time/entry.rs#L590-L595

For (1), A `&Handle` is enough, no need to make a clone.

For (2), we can delay the `.clone()` until we are about to register the entry.

Delaying the `Arc::clone` improves the performance on multi-core machine.

## Solution

* Storing the `&schedule:::Handle` in the `TimerShared`.

[`TimerShared`]: https://github.com/tokio-rs/tokio/blob/f6b9d0cd94c0b7f3fd4cd5d58f151f8f97862e58/tokio/src/runtime/time/entry.rs#L334

### DIFF
--- a/tokio/src/runtime/time/tests/mod.rs
+++ b/tokio/src/runtime/time/tests/mod.rs
@@ -1,14 +1,13 @@
 #![cfg(not(target_os = "wasi"))]
 
-use std::future::poll_fn;
-use std::sync::Barrier;
-use std::task::{Context, Waker};
-use std::time::Duration;
+use std::{future::poll_fn, task::Context, time::Duration};
 
-use tokio_test::assert_pending;
+#[cfg(not(loom))]
+use futures::task::noop_waker_ref;
 
 use crate::loom::sync::atomic::{AtomicBool, Ordering};
 use crate::loom::sync::Arc;
+use crate::task::yield_now;
 
 use super::TimerEntry;
 
@@ -18,23 +17,6 @@ fn model(f: impl Fn() + Send + Sync + 'static) {
 
     #[cfg(not(loom))]
     f();
-}
-
-fn block_on<T>(f: impl std::future::Future<Output = T>) -> T {
-    #[cfg(loom)]
-    return loom::future::block_on(f);
-
-    #[cfg(not(loom))]
-    {
-        let rt = crate::runtime::Builder::new_current_thread()
-            .build()
-            .unwrap();
-        rt.block_on(f)
-    }
-}
-
-fn cx() -> Context<'static> {
-    Context::from_waker(Waker::noop())
 }
 
 fn rt(start_paused: bool) -> crate::runtime::Runtime {
@@ -50,25 +32,18 @@ fn single_timer() {
     model(|| {
         let rt = rt(false);
         let handle = rt.handle();
-        let barrier = Arc::new(Barrier::new(2));
-        let barrier_clone = barrier.clone();
 
         rt.block_on(async move {
             let handle_ = handle.clone();
-            let jh = handle.spawn_blocking(move || {
+            let jh = handle.spawn(async move {
                 let entry =
                     TimerEntry::new(handle_.inner.driver().clock().now() + Duration::from_secs(1));
                 pin!(entry);
-                assert_pending!(entry.as_mut().poll_elapsed(&mut cx()));
-                // Make sure the previous poll_elapsed was called, so the runtime handle
-                // was cloned and stored in the entry.
-                barrier_clone.wait();
-                block_on(poll_fn(|cx| entry.as_mut().poll_elapsed(cx))).unwrap();
+
+                poll_fn(|cx| entry.as_mut().poll_elapsed(cx)).await.unwrap();
             });
 
-            // Making sure the first poll_elapsed was called, so the runtime handle
-            // was cloned and stored in the entry.
-            barrier.wait();
+            yield_now().await;
 
             let time = handle.inner.driver().time();
             let clock = handle.inner.driver().clock();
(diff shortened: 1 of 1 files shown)


tokio/src/runtime/time/tests/mod.rs
ADD-SP, line 1: ## For reviewers

Hiding whitespace reduces diffs.

tokio/src/runtime/time/tests/mod.rs
```diff
@@ -46,26 +34,26 @@ fn single_timer() {
...
-                handle_.inner.clone(),
-                handle_.inner.driver().clock().now() + Duration::from_secs(1),
-            );
-            pin!(entry);
+        rt.block_on(async move {
+            let handle_ = handle.clone();
+            let jh = handle.spawn(async move {
+                let entry =
+                    TimerEntry::new(handle_.inner.driver().clock().now() + Duration::from_secs(1));
+                pin!(entry);
```
ADD-SP, line 41: ## For reviewers

Since the `thread::spawn` runs the closure out of the tokio context (which means that we cannot get the current scheduler handle), I changed to the `tokio::spawn`.

Maybe `tokio::spawn_blocking` is better choice as it actually runs closure in a separate thread, which is more close to the original behavior.

tokio/src/runtime/time/tests/mod.rs
```diff
@@ -46,26 +34,26 @@ fn single_timer() {
...
 
-            block_on(std::future::poll_fn(|cx| entry.as_mut().poll_elapsed(cx))).unwrap();
-        });
+                poll_fn(|cx| entry.as_mut().poll_elapsed(cx)).await.unwrap();
+            });
 
-        thread::yield_now();
+            yield_now().await;
```
ADD-SP, line 46: ## For reviewers

We should change this to `thread::yield_now` if we are going to use `tokio::spawn_blocking` ([context is here](https://github.com/tokio-rs/tokio/pull/7461/files#r2210451027)).

tokio/src/runtime/time/tests/mod.rs
```diff
@@ -144,47 +132,48 @@ fn reset_future() {
...
-                .instant_to_tick(start + Duration::from_millis(1500)),
-        );
+            handle.process_at_time(
+                handle
+                    .time_source()
+                    .instant_to_tick(start + Duration::from_millis(1500)),
+            );
 
-        assert!(!finished_early.load(Ordering::Relaxed));
+            assert!(!finished_early.load(Ordering::Relaxed));
```
ADD-SP, line 164: ## For reviewers

We should use RMW here if we are going to use `tokio::spawn_blocking` here.

tokio/src/runtime/time/tests/mod.rs
```diff
@@ -144,47 +132,48 @@ fn reset_future() {
...
+                handle
+                    .time_source()
+                    .instant_to_tick(start + Duration::from_millis(2500)),
+            );
 
-        jh.join().unwrap();
+            jh.await.unwrap();
 
-        assert!(finished_early.load(Ordering::Relaxed));
+            assert!(finished_early.load(Ordering::Relaxed));
```
ADD-SP, line 174: ## For reviewers

We should use RMW here if we are going to use `tokio::spawn_blocking` here.

tokio/src/time/sleep.rs
```diff
@@ -251,9 +252,14 @@ impl Sleep {
...
-        use crate::runtime::scheduler;
-        let handle = scheduler::Handle::current();
-        let entry = TimerEntry::new(handle, deadline);
+        // ensure both scheduler handle and time driver are available,
+        // otherwise panic
+        let is_time_enabled = scheduler::Handle::with_current(|hdl| hdl.driver().time.is_some());
+        assert!(
+            is_time_enabled,
+            "A Tokio 1.x context was found, but timers are disabled. Call `enable_time` on the runtime builder to enable timers."
+        );
```
ADD-SP: ## For reviewers

Triggering panic inside the closure doesn't report the panic location of the caller, so we need to panic in the outside.

tokio/src/runtime/time/tests/mod.rs
ADD-SP: outdated

tokio/src/runtime/time/tests/mod.rs
ADD-SP: outdated

tokio/src/runtime/time/tests/mod.rs
ADD-SP: outdated

tokio/src/runtime/time/tests/mod.rs
ADD-SP: done

tokio/src/time/sleep.rs
```diff
@@ -251,9 +253,11 @@ impl Sleep {
         deadline: Instant,
         location: Option<&'static Location<'static>>,
     ) -> Sleep {
-        use crate::runtime::scheduler;
-        let handle = scheduler::Handle::current();
-        let entry = TimerEntry::new(handle, deadline);
+        // ensure both scheduler handle and time driver are available,
+        // otherwise panic
+        let is_time_enabled = scheduler::Handle::with_current(|hdl| hdl.driver().time.is_some());
+        assert!(is_time_enabled, "{TIME_DISABLED_ERROR}");
```
ADD-SP, line 259: ## For reviewers

Triggering panic inside the closure doesn't report the panic location of the caller, so we need to panic outside the closure.

tokio/src/time/sleep.rs
ADD-SP: outdated

tokio/src/runtime/time/tests/mod.rs
```diff
@@ -24,12 +33,8 @@ fn block_on<T>(f: impl std::future::Future<Output = T>) -> T {
...
 }
 
-fn model(f: impl Fn() + Send + Sync + 'static) {
-    #[cfg(loom)]
-    loom::model(f);
-
-    #[cfg(not(loom))]
-    f();
+fn cx() -> Context<'static> {
+    Context::from_waker(Waker::noop())
```
ADD-SP: ## For reviewers

The loom reports `futures::task::noop_waker_ref` due to `NULL` function pointer, `Waker::noop` uses real function pointer.

ADD-SP: outdated

ADD-SP:
superseded by #7473

PR #7462, opened 2025-07-16
sync: fix `CancellationToken::run_until_cancelled`

stepantubanov:
When token is already cancelled, provided future should not be polled.

## Motivation

Example: We have a loop where we get something from a stream but want to exit in case of shutdown:
```rust
loop {
  let Some(item) = shutdown.run_until_cancelled(stream.next()).await else {
     info!("Exit due to shutdown");
     break;
  };
  ...
}
```
With current logic, as long as the stream is yielding new items, the loop is not terminating even after the token is cancelled, which seems like a bug to me.

## Solution

I've changed the order of `if` branches to make it check the token before polling the future.


tokio-util/tests/sync_cancellation_token.rs
```diff
@@ -492,6 +492,21 @@ fn run_until_cancelled_test() {
             fut.as_mut().poll(&mut Context::from_waker(&waker))
         );
     }
+
+    // Does not poll the future when token is already cancelled.
```
ADD-SP:
```suggestion
    // Do not poll the future when token is already cancelled.
```

stepantubanov: Done

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -293,10 +293,10 @@ impl CancellationToken {
 
             fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
                 let this = self.project();
-                if let Poll::Ready(res) = this.future.poll(cx) {
-                    Poll::Ready(Some(res))
-                } else if this.cancellation.poll(cx).is_ready() {
+                if this.cancellation.poll(cx).is_ready() {
                     Poll::Ready(None)
+                } else if let Poll::Ready(res) = this.future.poll(cx) {
+                    Poll::Ready(Some(res))
```
ADD-SP: ## For reviewers

This affects the performance of the happy path as `this.cancellation.poll(cx)` internally locks the waiter list, this is especially visible for the use case mentioned in this PR.

```rust
loop {
  let Some(item) = shutdown.run_until_cancelled(stream.next()).await else {
     info!("Exit due to shutdown");
     break;
  };
  ...
}
```

However, reliability over performance.

lucab:
Casual observer here, but I noticed that here is an explicit comment here about the bias of this polling-order:
https://github.com/tokio-rs/tokio/blob/0a3fe460862720d005b8fa39cdde30575a063785/tokio-util/src/sync/cancellation_token.rs#L274-L283

And I think this bias was explicitly requested here: https://github.com/tokio-rs/tokio/pull/6618#discussion_r1630125807

As an additional random thought, if the problem is related to calling `run_until_cancelled()` in a loop on a token that is already cancelled, does it maybe make sense to keep the current order bias but check for `is_cancelled()` upfront before (conditionally) constructing the `RunUntilCancelledFuture`?

ADD-SP:
> And I think this bias was explicitly requested here: https://github.com/tokio-rs/tokio/pull/6618#discussion_r1630125807

This is a good finding, I didn't noticed this, and I think we should also change this comment because:

1. The [docs] doesn't mention this biased behavior.
2. This biased behavior is likely surprising the downstream.

I'm also open for keeping the existing biased behavior for better performance of happy path, but writing a warning about this behavior and suggesting a recommended approach to deal with this.

[docs]: https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html

lucab:
> As an additional random thought, if the problem is related to calling `run_until_cancelled()` in a loop on a token that is already cancelled, does it maybe make sense to keep the current order bias but check for `is_cancelled()` upfront before (conditionally) constructing the `RunUntilCancelledFuture`?

I tried to implement this and check it against the newly added test. I posted my changes to https://github.com/stepantubanov/tokio/pull/1/files in case you want to go that way.

stepantubanov:
@lucab Thanks, seems like a good way to preserve the bias while correcting the behavior for this scenario.

ADD-SP (approved):
@stepantubanov @lucab Thank you both for your contributions, this is a better solution that I didn't realize before.

PR #7464, opened 2025-07-17
time: Introduce advance_stable

kstrafe:
#7463 

## Motivation

Current time manipulation methods may not guarantee all futures are consistently polled to pending before the clock advances, potentially resulting in missed wakes or "unstable" advancing behavior.

What is meant by unstable behavior is that running `M*advance(N)` does not yield the same results as `advance(M*N)`. I think this behavior is not intuitive and also not helpful in tests, especially when performing SUT-like testing, where we want to advance a system's time by a certain amount. This should mimic how time advances in a real program.

Additionally, if we want to create a program that uses logical time steps, then every call to advance should awake timers in the order of their expiration. Any subsequent timers created from polling a future that would fall within the current logical time step should also be polled and driven further. This way, we can test and run the program in the exact same way, eliminating one source of non-determinism and closing the distance between test and release code.

## Solution

This patch implements `advance_stable`, a function that:

1. Polls all tasks from the task queue, ensuring that all timers are registered.
2. Advances time by the smallest timeout, so long it's less than the given duration.
3. Drives the timer in order to enqueue their associated futures to the task queue.
4. Calls yield_now() to drive these expired timeouts.
5. Repeats these steps until the provided timeout value is reached, and the clock has advanced by exactly that provided timeout.

This function requires time to be paused, and can only be used with a current-thread or local executor.

### DIFF
--- a/tokio/src/runtime/time/handle.rs
+++ b/tokio/src/runtime/time/handle.rs
@@ -3,8 +3,8 @@ use std::fmt;
 
 /// Handle to time driver instance.
 pub(crate) struct Handle {
-    pub(super) time_source: TimeSource,
-    pub(super) inner: super::Inner,
+    pub(crate) time_source: TimeSource,
+    pub(crate) inner: super::Inner,
 }
 
 impl Handle {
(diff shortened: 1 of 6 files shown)


ADD-SP:
I left my concern of this feature in https://github.com/tokio-rs/tokio/issues/7463#issuecomment-3086539989.

PR #7465, opened 2025-07-18
feat: introduce `NotifiedOwned`

ariaandika:
Allows returning an owned notification future that can be moved across threads. Inspired by #7231. Continuing from #7391

## Motivation

The `Notify` api returns `Notified<'_>` Future for the non-blocking wait of  But having a lifetime makes it hard to use in `Future`  since it requires self referencing lifetime.

In most if not all use cases, even in the example, the `Notify` will be wrapped in an `Arc`, so having an owned version will eliminate lifetime requirement.

Other use case mentioned in #7391 is to launch tasks and then notify them at the same time using `notify_waiters`.

```rust
let n = Arc::new(Notify::new()); // We can't use notified here because we need to send it.

tokio::spawn({
  let n = n.clone();
  async move {
    // if we need the code to be correct we need to register the waiter here
    tokio::sleep(10).await;
    n.notified().await;
    println!("event received");
}})

n.notify_waiters();
```

## Solution

Introduced new a type `NotifiedOwned`, a `Future` that holds `Arc` instead of shared reference to `Notify`, eliminating lifetime requirement. Add corresponding `Notify::notified_owned`. The new method can only be called if `self` is within an `Arc`.

### DIFF
--- a/tokio/src/sync/notify.rs
+++ b/tokio/src/sync/notify.rs
@@ -17,6 +17,7 @@ use std::panic::{RefUnwindSafe, UnwindSafe};
 use std::pin::Pin;
 use std::ptr::NonNull;
 use std::sync::atomic::Ordering::{self, Acquire, Relaxed, Release, SeqCst};
+use std::sync::Arc;
 use std::task::{Context, Poll, Waker};
 
 type WaitList = LinkedList<Waiter, <Waiter as linked_list::Link>::Target>;
@@ -397,6 +398,28 @@ pub struct Notified<'a> {
 unsafe impl<'a> Send for Notified<'a> {}
 unsafe impl<'a> Sync for Notified<'a> {}
 
+/// Future returned from [`Notify::notified_owned()`].
+///
+/// This future is fused, so once it has completed, any future calls to poll
+/// will immediately return `Poll::Ready`.
+#[derive(Debug)]
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub struct NotifiedOwned {
+    /// The `Notify` being received on.
+    notify: Arc<Notify>,
+
+    /// The current state of the receiving process.
+    state: State,
+
+    /// Number of calls to `notify_waiters` at the time of creation.
+    notify_waiters_calls: usize,
+
+    /// Entry in the waiter `LinkedList`.
+    waiter: Waiter,
+}
+
+unsafe impl Sync for NotifiedOwned {}
+
 #[derive(Debug)]
 enum State {
     Init,
@@ -541,6 +564,52 @@ impl Notify {
         }
     }
 
+    /// Wait for a notification with an owned `Future`.
+    ///
+    /// Unlike [`Self::notified`] which returns a future tied to the `Notify`'s
+    /// lifetime, `notified_owned` creates a self-contained future that owns its
+    /// notification state, making it safe to move between threads.
+    ///
+    /// See [`Self::notified`] for more details.
+    ///
+    /// # Cancel safety
+    ///
+    /// This method uses a queue to fairly distribute notifications in the order
+    /// they were requested. Cancelling a call to `notified` makes you lose your
+    /// place in the queue.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use std::sync::Arc;
+    /// use tokio::sync::Notify;
+    ///
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let notify = Arc::new(Notify::new());
+    ///     // register all waiters, in this case 1
+    ///     let notified = notify.notified_owned();
+    ///
+    ///     // spawn the taks waiting for the notification
+    ///     tokio::spawn(async move {
+    ///         notified.await; // Waits for the notification
+    ///         println!("Notified!");
+    ///     });
+    ///
+    ///     notify.notify_waiters(); // Sends a notification
+    /// }
+    /// ```
+    pub fn notified_owned(self: &Arc<Self>) -> NotifiedOwned {
+        // we load the number of times notify_waiters
+        // was called and store that in the future.
+        let state = self.state.load(SeqCst);
+        NotifiedOwned {
+            notify: Arc::clone(self),
+            state: State::Init,
+            notify_waiters_calls: get_num_notify_waiters_calls(state),
+            waiter: Waiter::new(),
+        }
+    }
     /// Notifies the first waiting task.
     ///
     /// If a task is currently waiting, that task is notified. Otherwise, a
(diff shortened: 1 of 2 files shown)


tokio/src/sync/notify.rs
```diff
@@ -397,6 +398,28 @@ pub struct Notified<'a> {
...
+    state: State,
+
+    /// Number of calls to `notify_waiters` at the time of creation.
+    notify_waiters_calls: usize,
+
+    /// Entry in the waiter `LinkedList`.
+    waiter: Waiter,
+}
+
+unsafe impl Sync for NotifiedOwned {}
```
ADD-SP: Why isn't this `Send`?

tokio/src/sync/notify.rs
```diff
@@ -541,6 +564,52 @@ impl Notify {
...
+    /// Unlike [`Self::notified`] which returns a future tied to the `Notify`'s
+    /// lifetime, `notified_owned` creates a self-contained future that owns its
+    /// notification state, making it safe to move between threads.
+    ///
+    /// See [`Self::notified`] for more details.
+    ///
+    /// # Cancel safety
+    ///
+    /// This method uses a queue to fairly distribute notifications in the order
+    /// they were requested. Cancelling a call to `notified` makes you lose your
```
ADD-SP:
```suggestion
    /// they were requested. Cancelling a call to `notified_owned` makes you lose your
```

tokio/src/sync/notify.rs
```diff
@@ -397,6 +398,28 @@ pub struct Notified<'a> {
 unsafe impl<'a> Send for Notified<'a> {}
 unsafe impl<'a> Sync for Notified<'a> {}
 
+/// Future returned from [`Notify::notified_owned()`].
+///
+/// This future is fused, so once it has completed, any future calls to poll
+/// will immediately return `Poll::Ready`.
+#[derive(Debug)]
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub struct NotifiedOwned {
```
ADD-SP: I suggest using `OwnedNotified` to align with existing pattern (*[`tokio::sync`]*).



[`tokio::sync`]: https://docs.rs/tokio/1.46.1/tokio/sync/index.html

tokio/src/sync/notify.rs
```diff
@@ -932,265 +1001,331 @@ impl Notified<'_> {
...
+    fn drop(&mut self) {
+        // Safety: The type only transitions to a "Waiting" state when pinned.
+        let project = unsafe { Pin::new_unchecked(self).project() };
+        drop_notified(project);
+    }
+}
+
+// ===== Shared logic for `Notified` and `NotifiedOwned` =====
+
+type NotifiedProject<'a> = (&'a Notify, &'a mut State, &'a usize, &'a Waiter);
```
ADD-SP: Is this intended to avoid too many parameters of function `drop_notified` / `poll_notified`?

ADD-SP (commented):
Could you write some tests for this new feature?

tokio/src/sync/notify.rs
ariaandika: No, just a pattern that i noticed where both function uses the custom projection type. Is it preferred to just have all be individual arguments ?

tokio/src/sync/notify.rs
ariaandika: Ok

tokio/src/sync/notify.rs
ariaandika: Because its already `Send` ?

tokio/src/sync/notify.rs
ariaandika: The `project` method already returns tuple, so if i made it to individual arguments, we have to spread it to call `drop_notified` and `poll_notified`

tokio/src/sync/notify.rs
ADD-SP: Ah, you are right.

tokio/src/sync/notify.rs
ADD-SP: This is new public type, we also need to add tests in `tokio/tests/async_send_sync.rs`.

tokio/src/sync/notify.rs
```diff
@@ -17,6 +17,7 @@ use std::panic::{RefUnwindSafe, UnwindSafe};
 use std::pin::Pin;
 use std::ptr::NonNull;
 use std::sync::atomic::Ordering::{self, Acquire, Relaxed, Release, SeqCst};
+use std::sync::Arc;
```
ADD-SP, line 20: We prefer to use `crate::loom::sync::Arc`, so that we can run loom tests.

tokio/src/sync/notify.rs
ADD-SP: > Is it preferred to just have all be individual arguments?

I don't remember tokio having this pattern, it is unusual for me. I guess this is just my personal flavor of having all individual parameters.

For now, let's keep this, unless another reviewer also ask this topic.

ariaandika:
I did not know about loom `arbitrary_self_types` feature is, and considering the error help message, changing `notified_owned` to take `self: Arc<Self>` means callers must have the ownership of `Arc<Notify>`, though callers can just `.clone()` before calling `notified_owned`, which in my opinion is slightly inconvenient.

Should i proceed ?

Darksonn:
Do not enable the `arbitrary_self_types` feature. We can skip using the loom Arc, but anything else in `sync` should come from Loom.

Darksonn:
And yes you should take `self: Arc<Self>` instead of forcing a clone of the `Arc` on the end-user even if one is not needed.

tokio/src/sync/notify.rs
Darksonn: You can do this if you make a real struct. Don't use a tuple.

ADD-SP: Ok, please make a real struct, or having all the individual arguments. both are fine.

ariaandika: Ok, i will make it into struct.

tokio/src/sync/notify.rs
```diff
@@ -1143,20 +1305,15 @@ impl Notified<'_> {
...
-    fn drop(&mut self) {
+    fn drop_notified(self) {
         // Safety: The type only transitions to a "Waiting" state when pinned.
-        let (notify, state, _, waiter) = unsafe { Pin::new_unchecked(self).project() };
+        let NotifiedProject {
+            notify,
+            state,
+            waiter,
+            ..
+        } = self;
```
ADD-SP: There is no `unsafe` block here, we don't need the safety comment.

tokio/src/sync/notify.rs
```diff
@@ -541,6 +574,52 @@ impl Notify {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use std::sync::Arc;
+    /// use tokio::sync::Notify;
+    ///
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let notify = Arc::new(Notify::new());
+    ///     // register all waiters, in this case 1
```
ADD-SP: What is 'case 1'?

ADD-SP (commented):
Overall looks good to me.

tokio/src/sync/notify.rs
ariaandika: Its from the previous PR, didn't think it needs to be there.

I copied the example since that use case is better for an example docs than my use case :sweat_smile:.

tokio/src/sync/notify.rs
ariaandika: Whoops, missed that one.

ADD-SP (commented):
Implementation looks good to me, could you add some tests in `tokio/tests/sync_notify.rs`?

ariaandika:
The underlying logic is exactly the same, is additional tests necessary ?

Also, is it okay if i just copied from existing tests of `Notfied<'_>` ? I don't know special test case for an owned version.

ADD-SP (commented):
> The underlying logic is exactly the same, is additional tests necessary ?

I think additional tests are necessary as we may (not likely) to change the implementation in the future, so we might have different implementation. 

In principle, tests under `tokio/tests` are black box tests, which should not care about the implementation details too much.

> I don't know special test case for an owned version.

I think the only special case is that make sure the `OwnedNotified` is `'static` because this is the whole point of this feature.

I guess this is just my personal flavor, it should be easy to eyeball if a type is '`static` or not, seems we don't need a test to cover this.

> is it okay if i just copied from existing tests of Notfied<'_> ?

Apparently, we don't want to copy such a large test code, Let me think a little bit. Feel free to share you through about this.

ariaandika:
Well, looking on the other API, `Semaphore`, the tests between `sync_semaphore.rs` and `sync_semaphore_owned.rs` is pretty similar, one just swap the `Semaphore::acquire` and `Semaphore::acquire_owned`.

ADD-SP:
> Well, looking on the other API, `Semaphore`, the tests between `sync_semaphore.rs` and `sync_semaphore_owned.rs` is pretty similar, one just swap the `Semaphore::acquire` and `Semaphore::acquire_owned`.

I didn't come up with a better idea, I think it is ok to follow the approach of `sync_semaphore_owned.rs`.

ariaandika:
You're welcome, glad the feature can be added.

PR #7467, opened 2025-07-19
[WIP] 

ADD-SP:
## Blocked by #7461 

*There are something need to be polished, before polishing something, I drafted this PR to get the CI error early.*

## High-level design

Please checkout #7384.

**TBD**


tokio-util/tests/time_delay_queue.rs
```diff
@@ -522,43 +514,6 @@ async fn insert_after_ready_poll() {
     assert_eq!("3", res[2]);
 }
 
-#[tokio::test]
-async fn reset_later_after_slot_starts() {
```
ADD-SP, line 526: ## Note

This test was removed because it was testing the behavior of `tokio::runtime::time::entry::reset`, which is removed in this PR.

tokio-util/tests/time_delay_queue.rs
```diff
@@ -584,43 +539,6 @@ async fn reset_inserted_expired() {
     assert_eq!(queue.len(), 0);
 }
 
-#[tokio::test]
-async fn reset_earlier_after_slot_starts() {
```
ADD-SP, line 588: ## Note

This test was removed because it was testing the behavior of `tokio::runtime::time::entry::reset`, which is removed in this PR.

tokio-util/tests/time_delay_queue.rs
```diff
@@ -221,7 +219,7 @@ async fn reset_much_later() {
 
     sleep(ms(20)).await;
 
-    assert!(queue.is_woken());
+    assert_ready_some!(poll!(queue));
```
ADD-SP: ## Note

`queue.reset_at` resets inner `Sleep`, however, the new implementation will drop the inner timer and create a new one. So the waker will not be called, we have to poll manually.

tokio-util/tests/time_delay_queue.rs
```diff
@@ -248,7 +246,7 @@ async fn reset_twice() {
...
```
ADD-SP: ## Note

`queue.reset_at` resets inner `Sleep`, however, the new implementation will drop the inner timer and create a new one. So the waker will not be called, we have to poll manually.

tokio/src/runtime/scheduler/current_thread/mod.rs
```diff
@@ -646,10 +765,18 @@ impl Schedule for Arc<Handle> {
...
+                } else {
+                    // runtime is shutting down
+                    // OR waking up expired timers
+
+                    // Track that a task was scheduled from **outside** of the runtime.
+                    self.shared.scheduler_metrics.inc_remote_schedule_count();
+
+                    // Schedule the task
+                    self.shared.inject.push(task);
+                    self.driver.unpark();
```
ADD-SP: ## Note

Since the wheel is stored in the `Core`, the `cx.core` is `None` while waking up timers.

While we could store Option<Wheel> in Core, this would double the number of possible states in the local time subsystem, which isn't worth it.

I know this creates some lock contention on the inject queue, but it doesn't shows the regression on the benchmark, so I kept it for now. Feel free to share your thought and possible optimization.

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -1160,6 +1274,7 @@ impl Handle {
         for remote in &self.shared.remotes[..] {
             remote.unpark.unpark(&self.driver);
         }
+        self.driver.unpark();
```
ADD-SP: ## Note

There might be a worker is parked on driver, so we also need to unpark the driver.

ADD-SP: Resolved, this is an incorrect fix as `remote.unpark.unpark` is designed to unpark the driver as well, the correct fix should be setting `PARKED_DREIVER` flag here.

https://github.com/tokio-rs/tokio/blob/6d868d96ce9df27c5ec1be3bba960ebe80ee4793/tokio/src/runtime/scheduler/multi_thread/park.rs#L73-L87

tokio/src/util/mod.rs
```diff
@@ -41,7 +39,6 @@ mod wake_list;
     feature = "fs",
     feature = "rt",
     feature = "signal",
-    feature = "time",
 ))]
 pub(crate) use wake_list::WakeList;
```
ADD-SP, line 43: ## Note

We no long use `WakeList` in the time subsystem.

tokio/src/runtime/time/mod.rs
```diff
@@ -160,250 +129,10 @@ impl Driver {
...
-impl fmt::Debug for Inner {
-    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
-        fmt.debug_struct("Inner").finish()
-    }
 }
 
-#[cfg(test)]
-mod tests;
+// #[cfg(test)]
+// mod tests;
```
ADD-SP: ## Note

Blocked by #7461

ADD-SP: Unblocked by https://github.com/tokio-rs/tokio/pull/7473#issuecomment-3138506618.

tokio/src/runtime/scheduler/current_thread/mod.rs
```diff
@@ -62,6 +70,18 @@ struct Core {
...
+    /// Worker local timer wheel
+    wheel: Wheel,
+
+    #[cfg(feature = "time")]
+    /// Channel for sending timers that need to be cancelled
+    timer_cancel_tx: mpsc::Sender<EntryHandle>,
+
+    #[cfg(feature = "time")]
+    /// Channel for receiving timers that need to be cancelled
+    timer_cancel_rx: mpsc::Receiver<EntryHandle>,
```
ADD-SP: ## Note

We may want to encapsulate these into a single structure, please leave comment here if you like this idea.

tokio/src/runtime/scheduler/multi_thread/park.rs
```diff
@@ -71,11 +71,8 @@ impl Parker {
     }
 
     pub(crate) fn park_timeout(&mut self, handle: &driver::Handle, duration: Duration) {
-        // Only parking with zero is supported...
-        assert_eq!(duration, Duration::from_millis(0));
-
         if let Some(mut driver) = self.inner.shared.driver.try_lock() {
-            driver.park_timeout(handle, duration);
```
ADD-SP, line 78: ## Note

This assertion was here because the following `driver.park_timeout` didn't set the parker state to `PARKED_DRIVER`, this makes sense because zero timeout doesn't need to unpark.

I removed this assertion and then invoked `self.inner.park_driver` to ensure the parker state is correct with non-zero timeout.

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -753,9 +797,41 @@ impl Context {
...
+                    (None, Some(park_duration)) => Some(park_duration),
+                    (None, None) => None,
+                };
+                if util::time::pre_auto_advance(&handle.driver, dur) {
+                    (Some(Duration::ZERO), dur)
+                } else {
+                    (dur, None)
+                }
+            }
+        };
```
ADD-SP: ## Note

I called several utilities to driver the local timers. You might ask me why we don't just merge it into single function since we always have to write it that way. This is a reasonable point, if you'd like me to do this, please left a comment here.

And I know the name of these function is bad, feel free to suggest a better one.

tokio/src/runtime/time/entry.rs
ADD-SP, line 1: ## Note

This file was replaced by `tokio/src/runtime/time/timer.rs`.

tokio/src/runtime/time/source.rs
```diff
@@ -22,11 +21,7 @@ impl TimeSource {
     pub(crate) fn instant_to_tick(&self, t: Instant) -> u64 {
         // round up
         let dur: Duration = t.saturating_duration_since(self.start_time);
-        let ms = dur
-            .as_millis()
-            .try_into()
-            .unwrap_or(MAX_SAFE_MILLIS_DURATION);
-        ms.min(MAX_SAFE_MILLIS_DURATION)
```
ADD-SP, line 29: ## Note

We don't have `MAX_SAFE_MILLIS_DURATION` anymore, so I removed it.

tokio/src/runtime/time/tests/mod.rs
```diff
@@ -121,70 +156,33 @@ fn change_waker() {
...
-            handle
-                .time_source()
-                .instant_to_tick(start + Duration::from_millis(2500)),
-        );
-
         jh.join().unwrap();
-
-        assert!(finished_early.load(Ordering::Relaxed));
     })
 }
```
ADD-SP, line 187: ## Note

This implementation always create a `Entry` while resetting a `Sleep` or `Timeout`, so I removed this test.

tokio/src/runtime/time/tests/mod.rs
```diff
@@ -205,31 +203,33 @@ fn poll_process_levels() {
 
     let mut entries = vec![];
 
-    for i in 0..normal_or_miri(1024, 64) {
-        let mut entry = Box::pin(TimerEntry::new(
-            handle.inner.clone(),
-            handle.inner.driver().clock().now() + Duration::from_millis(i),
-        ));
+    rt.block_on(async {
```
ADD-SP: ## Note

I put codes inside the `rt.block_on` to make sure we can get the current timer wheel by thread local variable.

tokio/src/runtime/time/tests/mod.rs
```diff
@@ -121,70 +156,33 @@ fn change_waker() {
...
-        // advance 2s
-        time.process_at_time(time.time_source().now(clock) + 2_000_000_000);
+        change_waker_rx.blocking_recv().unwrap();
 
-        jh.join().unwrap();
-    })
-}
-
-#[test]
-fn reset_future() {
```
ADD-SP, line 140: ## Note

This implementation always create a `Entry` while resetting a `Sleep` or `Timeout`, so I removed this test.

tokio/src/runtime/time/tests/mod.rs
```diff
@@ -240,30 +240,16 @@ fn poll_process_levels_targeted() {
...
-
-    handle.process_at_time(62);
-    assert!(e1.as_mut().poll_elapsed(&mut context).is_pending());
-    handle.process_at_time(192);
-    handle.process_at_time(192);
-}
-
-#[test]
-#[cfg(not(loom))]
-fn instant_to_tick_max() {
```
ADD-SP, line 259: ## Note

We don't have `MAX_SAFE_MILLIS_DURATION` anymore, so I removed it.

tokio/src/time/error.rs
```diff
@@ -46,11 +46,6 @@ impl From<Kind> for Error {
 #[derive(Debug, PartialEq, Eq)]
 pub struct Elapsed(());
 
-#[derive(Debug)]
-pub(crate) enum InsertError {
-    Elapsed,
-}
-
```
ADD-SP, line 53: ## Note

This error never happens in this implementation, and it is not a public type, so I remove it.

tokio/src/time/sleep.rs
```diff
@@ -352,22 +352,12 @@ impl Sleep {
...
-        me.entry.as_mut().reset(deadline, false);
-    }
-
     fn reset_inner(self: Pin<&mut Self>, deadline: Instant) {
+        use crate::runtime::scheduler;
         let mut me = self.project();
-        me.entry.as_mut().reset(deadline, true);
+        me.entry.as_mut().cancel();
+        me.entry
+            .set(Timer::new(scheduler::Handle::current(), deadline));
```
ADD-SP: ## Note

I know this introduces extra overhead while resetting the `Sleep`.

Since the current PR is pretty large, this is a trade-off. I can add the `reset` optimization in a follow-up PR if we really want it.

Please left a comment here if you are concerned about this change.

tokio/src/runtime/time/wheel/entry.rs
```diff
@@ -0,0 +1,242 @@
...
+    pub(crate) fn as_entry_ptr(&self) -> NonNull<Entry> {
+        self.ptr
+    }
+
+    /// # Safety
+    ///
+    /// [`Self::ptr`] must be a valid pointer to an [`Entry`].
+    pub(crate) unsafe fn drop_entry(&self) {
+        drop(Box::from_raw(self.ptr.as_ptr()));
+    }
```
ADD-SP: ## Note

Manually dropping creates extra mental burden, so I tried to make the number of callers as small as possible. Now, only `tokio/src/runtime/time/wheel/mod.rs` calls this method.

tokio/src/runtime/time/wheel/entry.rs
```diff
@@ -0,0 +1,242 @@
...
+    state: AtomicU8,
+}
+
+/// The entry in the timer wheel.
+pub(crate) struct Entry {
+    /// The pointers used by the intrusive linked list.
+    pointers: linked_list::Pointers<Entry>,
+
+    inner: Arc<Inner>,
+}
```
ADD-SP: ## Note

This `Entry` doesn't keep (almost) any data (like state, waker) directly.

Without the protection of global `Mutex<Wheel>`, it is complicated to know if the entry has already been drop from other worker threads, so I simply chose the `Arc` to hold (almost) all data.

Otherwise, almost all methods on the `Handle` will be unsafe methods, it is not worth doing it.

tokio/src/runtime/time/wheel/entry.rs
```diff
@@ -0,0 +1,242 @@
...
+        let old = self.inner.state.swap(STATE_PENDING, SeqCst);
+        assert_eq!(old, STATE_REGISTERED, "Entry not registered");
+        Ok(())
+    }
+
+    /// # Panic
+    ///
+    /// Panics if receiver side is closed, this is usually caused by
+    /// the shutdown logic dropping the receiver side too early.
+    pub(crate) fn cancel(&self) {
```
ADD-SP, line 241: ## Note

Apparently, we can simply remove the entry from wheel if it is in the current wheel. I prefer to do it in a follow-up PR.

Please leave comments here if you want it in this PR.

tokio/src/runtime/time/wheel/entry.rs
Darksonn: You could probably move the `Pointers` inside the `Arc` too, similar to how `JoinSet` works.

tokio/src/runtime/time/wheel/entry.rs
```diff
@@ -0,0 +1,242 @@
...
+    pub(crate) fn transition_to_registered(&self, cancel_tx: Sender<Handle>) {
+        {
+            let mut maybe_tx = self.inner.cancel_tx.lock();
+            assert!(maybe_tx.is_none(), "cancel sender already set");
+            *maybe_tx = Some(cancel_tx);
+            // lock is dropped here
+        }
+        let old = self.inner.state.swap(STATE_REGISTERED, SeqCst);
+        assert_eq!(old, STATE_UNREGISTERED, "Entry not unregistered");
+    }
```
ADD-SP: ## Note

In case of concurrent call to `transition_to_registered` and `cancel`, there might be a race condition, I will look at this.

tokio/src/runtime/time/wheel/entry.rs
ADD-SP: This is a better approach, I have finished your suggestion.

tokio/src/runtime/time/wheel/entry.rs
ADD-SP: Outdated

tokio/src/runtime/time/wheel/entry.rs
```diff
@@ -0,0 +1,317 @@
...
+            Err(STATE_CANCELLING) => TransitionToPending::Cancelling,
+            Err(actual) => panic!("state is corrupted ({actual})"),
+        }
+    }
+
+    /// # Panic
+    ///
+    /// Panics if receiver side is closed, this is usually caused by
+    /// the shutdown logic dropping the receiver side too early.
+    pub(crate) fn transition_to_cancelling(&self) {
```
ADD-SP: ## Note

Apparently, we can simply remove the entry from wheel if it is in the current local wheel. I prefer to do it in a follow-up PR.

Please leave comments here if you want it in this PR.

tokio/src/runtime/time/entry.rs
ADD-SP: Outdated

tokio/src/runtime/time/wheel/entry.rs
ADD-SP: Outdated

tokio/src/runtime/time/wheel/entry.rs
ADD-SP: Fixed

tokio/src/runtime/time/wheel/cancellation_queue.rs
```diff
@@ -0,0 +1,257 @@
...
+
+            fn next(&mut self) -> Option<Self::Item> {
+                match self.head {
+                    Some(head) => unsafe {
+                        let atomic_next = head.as_ref().cancel_pointer();
+                        let mut next = atomic_next.load(SeqCst);
+                        while head != self.tail.unwrap() && next.is_null() {
+                            spin_loop();
+                            next = atomic_next.load(SeqCst);
+                        }
```
Darksonn: Spinloops are generally pretty problematic performance-wise. Can we just protect the linked list by a mtuex instead of using atomics?

ADD-SP: I think the 'window of spin loop' is extremely small, I drew a diagram to explain it.

queue

tokio/src/runtime/time/wheel/cancellation_queue.rs
Darksonn, line 1: Since you now protect the list by a mutex, I would expect it to use the `LinkedList` utility that we use elsewhere in Tokio.

ADD-SP: Utilizing the `LinkedList` requires `unsafe impl linked_list::Link for Entry`, however, the existing `impl` has already been used by the wheel's linked list, we cannot impl it twice.

Of course, we can write something like `unsafe impl linked_list::Link for (Entry, ())`, but it is a bit weird pattern and not straightforward.

Do you have suggestions?

ADD-SP: I didn't come up with a better solution, `impl linked_list::Link for (Entry,)` is the best one for now. I have pushed a new commit for it.

Darksonn: Can we use a `repr(transparent)` wrapper struct instead of a tuple? Otherwise that part LGTM.

ADD-SP: We might just need a ZST to make the second `impl linked_list::Link` happen.

Initially, I thought the intrusive list relies on the layout of `L` internally, but I didn't find such usage. Finally I found #2790 and realized using two generic parameters here is just for `const fn`.

https://github.com/tokio-rs/tokio/blob/925c614c89d0a26777a334612e2ed6ad0e7935c3/tokio/src/util/linked_list.rs#L19-L28

tokio/src/runtime/time/wheel/cancellation_queue.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// Behavior is undefined if any of the following conditions are violated:
+    ///
+    /// - `hdl` must not in any cancellation queue.
+    unsafe fn push_front(&mut self, hdl: EntryHandle) {
+        self.list.push_front(hdl);
+    }
+
+    fn iter(&mut self) -> impl Iterator<Item = EntryHandle> {
+        let mut list = std::mem::take(&mut self.list);
+        std::iter::from_fn(move || list.pop_front())
```
Darksonn: What happens if the iterator is dropped before all items are consumed? Probably they are leaked since `LinkedList` has no destructor. Can we return a custom iterator with a safer drop?

tokio/src/runtime/scheduler/current_thread/mod.rs
```diff
@@ -62,6 +68,18 @@ struct Core {
...
+    /// Worker local timer wheel
+    wheel: Wheel,
+
+    #[cfg(feature = "time")]
+    /// Channel for sending timers that need to be cancelled
+    timer_cancel_tx: cancellation_queue::Sender,
+
+    #[cfg(feature = "time")]
+    /// Channel for receiving timers that need to be cancelled
+    timer_cancel_rx: cancellation_queue::Receiver,
```
Darksonn: If you're storing the sender and receiver in the same location, then I think we can just merge the two structs into one. I also think this lets us get rid of the `Arc` inside the queue.

tokio/src/runtime/time/wheel/entry.rs
```diff
@@ -0,0 +1,345 @@
...
+use std::task::Waker;
+
+pub(crate) type EntryList = linked_list::LinkedList<Entry, Entry>;
+
+/// A pure new entry, no any changes to the state.
+const STATE_UNREGISTERED: u8 = 0;
+
+/// The entry is being registered to the timer wheel,
+/// and also saving the `cancel_tx` to the entry.
+const STATE_BUSY_REGISTERING: u8 = 1;
```
Darksonn: What does this mean? I don't understand "saving the cancel_tx"

tokio/src/runtime/time/wheel/entry.rs
```diff
@@ -0,0 +1,345 @@
...
+    _pin: PhantomPinned,
+}
+
+/// Safety: There is a field is neither `Send` nor `Sync`.
+///
+/// - [`Self::cancel_tx`]: This is protected by [`Self::state`].
+///
+/// [`cancellation_queue`]: `super::cancellation_queue`
+unsafe impl Send for Entry {}
+unsafe impl Sync for Entry {}
```
Darksonn: Why are these required?

tokio/src/runtime/time/wheel/entry.rs
```diff
@@ -0,0 +1,345 @@
...
+    cancel_pointers: linked_list::Pointers<Entry>,
+
+    /// The tick when this entry is scheduled to expire.
+    deadline: u64,
+
+    /// The currently registered waker.
+    waker: AtomicWaker,
+
+    /// The mpsc channel used to cancel the entry.
+    cancel_tx: UnsafeCell<Option<Sender>>,
```
Darksonn: Can't this just be a `Sender`? Sending is `&self`, so I don't think you need mutability.

tokio/src/runtime/time/wheel/entry.rs
ADD-SP: I think we can remove `unsafe impl Send`, but we still need to keep the `Sync` as the `!Sync` will propagate to Scheduler handle.

tokio/src/runtime/time/wheel/cancellation_queue.rs
ADD-SP: Good catch! I have replaced it with a custom iterator with a `Drop` implementation, also added a test for this bug.

tokio/src/runtime/time/wheel/entry.rs
ADD-SP: The `Entry` needs to save a `CancellationQueueSender` while being registered into the local timer wheel, I may need to use a more reasonable name for `cancel_tx`. I have updated this comment.

Relevant context: <https://github.com/tokio-rs/tokio/pull/7467#discussion_r2307499297>.

tokio/src/runtime/time/wheel/entry.rs
ADD-SP: Since tokio timer can be polled in another runtime, it is possible that we cannot access any worker thread's local timer wheel at the first poll. In this case, we need to create the `Entry` first, and then send it to the timer inject queue, the `Sender` is not available at that time.

https://github.com/tokio-rs/tokio/blob/009d15870dd76a7504dba713d48e8ae1df7ea17d/tokio/src/runtime/time/timer.rs#L60-L82

tokio/src/runtime/scheduler/current_thread/mod.rs
ADD-SP: My motivation is to avoid `tokio::time::sleep` or `tokio::time::timeout` implementation to take out entries from the cancellation queue, the scheduler should be the only one can take out entries.

If we merge `Sender` and `Receiver` into a single struct, we might have to use method visibility control to make sure `tokio::time::sleep` or `tokio::time::timeout` cannot take out entries. It looks like a unusual approach.

ADD-SP: <del>It would be better adding an `Receiver::new_sender(&self) -> Sender`, so we don't need to save both. I will try it later.</del>

After careful consideration, I think it is unusual to construct a `Sender` using `Receiver`, I'm not going to implementation this kind of interface even this is just an internal API.

tokio/src/runtime/time/wheel/entry.rs
Darksonn: What I mean is ... one of the fields is making this necessary. Can we fix it on the field rather than the struct?

ADD-SP: I'm considering whether use `Mutex<State>` and `State::Registered(Sender)`, will let you know after a careful consideration.

ADD-SP: I have fixed the field, details are attached to <https://gist.github.com/ADD-SP/4037e25256b8dc8a4956962415de2356#file-y-faq-md>.

tokio/src/runtime/time/wheel/entry.rs
ADD-SP: Resolved this conversation, see <https://github.com/tokio-rs/tokio/pull/7467#discussion_r2316156818>.

tokio/src/runtime/time/mod.rs
```diff
@@ -87,33 +88,8 @@ use std::{num::NonZeroU64, ptr::NonNull};
...
-    is_shutdown: AtomicBool,
 
-    // When `true`, a call to `park_timeout` should immediately return and time
-    // should not advance. One reason for this to be `true` is if the task
-    // passed to `Runtime::block_on` called `task::yield_now()`.
-    //
-    // While it may look racy, it only has any effect when the clock is paused
-    // and pausing the clock is restricted to a single-threaded runtime.
-    #[cfg(feature = "test-util")]
-    did_wake: AtomicBool,
```
Darksonn, line 107: Moving these atomics into their own `Arc` is best avoided. That's an allocation per integer, and they may be stored far away from each other hurting cache locality.

tokio/src/runtime/time/wheel/cancellation_queue.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+#[derive(Debug)]
+pub(crate) struct Receiver {
+    inner: Arc<Mutex<Inner>>,
+
+    // Technically, receiver is `Sync`, however, we only
+    // need single receiver for cancellation purpose,
+    // so we make it `!Sync` to prevent abusing.
+    _not_sync: PhantomData<*const ()>,
```
Darksonn: The method is `&mut self`. That should be enough to prevent that.

Darksonn (commented):
I'm sorry for taking long between reviewes, but this is a really large PR and I find it difficult to review. Here are a few things I noticed while looking at it, but I still don't feel like I have an overview.

tokio/src/runtime/time/mod.rs
ADD-SP: Let me revisit the shutdown process and relevant dataflows, will come back later.

tokio/src/runtime/time/wheel/cancellation_queue.rs
ADD-SP: fixed

tokio/src/runtime/time/mod.rs
ADD-SP: Improved, details are attached to <https://gist.github.com/ADD-SP/4037e25256b8dc8a4956962415de2356#file-d-propagate-shutdown-notification-md>.

tokio/src/runtime/scheduler/current_thread/mod.rs
ADD-SP: done

tokio/src/runtime/scheduler/current_thread/mod.rs
ADD-SP: The lock contention of inject queue is still high if the timer is always fired, `tokio::time::Interval` is a typical use case.

Fixed by <https://github.com/tokio-rs/tokio/pull/7467/commits/4862ef7028f7bfd6085beb9b386001286707f8e8>.

tokio/src/runtime/time/mod.rs
```diff
@@ -128,281 +107,67 @@ impl Driver {
...
-
-        lock.next_wake = lock
-            .wheel
-            .poll_at()
-            .map(|t| NonZeroU64::new(t).unwrap_or_else(|| NonZeroU64::new(1).unwrap()));
-
-        drop(lock);
-
-        waker_list.wake_all();
+    pub(crate) struct Context2 {
```
ADD-SP: ## Note

Bad name, feel free to suggest a new one.

.github/workflows/ci.yml
```diff
@@ -1058,6 +1058,7 @@ jobs:
           CARGO_TARGET_WASM32_WASIP1_RUNNER: "wasmtime run --"
           CARGO_TARGET_WASM32_WASIP1_THREADS_RUNNER: "wasmtime run -W bulk-memory=y -W threads=y -S threads=y --"
           RUSTFLAGS: --cfg tokio_unstable -Dwarnings -C target-feature=+atomics,+bulk-memory -C link-args=--max-memory=67108864
+          RUSTDOCFLAGS: -C link-args=--max-memory=67108864
```
ADD-SP, line 1061: ## Note

https://github.com/tokio-rs/tokio/blob/d060401f6c7dca4a20674e3ad63ad7f1b228aa31/tokio-util/src/io/sync_bridge.rs#L91

tokio/src/runtime/time/wheel/mod.rs
```diff
@@ -70,80 +80,82 @@ impl Wheel {
...
     /// # Return
     ///
-    /// Returns `Ok` when the item is successfully inserted, `Err` otherwise.
-    ///
-    /// `Err(Elapsed)` indicates that `when` represents an instant that has
-    /// already passed. In this case, the caller should fire the timeout
-    /// immediately.
-    ///
-    /// `Err(Invalid)` indicates an invalid `when` argument as been supplied.
+    /// * `true`: The entry was successfully inserted.
```
martin-g: This should be updated to `Insert::Success`, `Insert::Elapsed` and `Insert::Cancelling`

tokio/src/runtime/context.rs
```diff
@@ -198,3 +183,24 @@ cfg_rt! {
         }
     }
 }
+
+cfg_rt_or_time! {
```
martin-g: Is this the correct macro ? The code below cares only about `rt` but not for `time`.

ADD-SP: This method is also called by the timer.

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -793,6 +849,143 @@ impl Context {
...
+        /// * Process expired timers.
+        fn maintain_local_timers_after_parking(
+            &self,
+            core: Box<Core>,
+            auto_advance_duration: Option<Duration>
+        ) -> Box<Core> {
+            let handle = &self.worker.handle;
+            let (core, ()) = self.enter_with_time_context(core, |time_cx| {
+                util::time::post_auto_advance(&handle.driver, auto_advance_duration);
+                util::time::process_expired_timers(&mut time_cx.wheel, &handle.driver);
```
Darksonn: Can we avoid the `RefCell` like this?

1. Obtain linked list of timers to wake up, while holding `RefMut` on core.
2. Release `RefMut` on core.
3. Iterate linked list to wake up tasks, without `RefMut`.

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -793,6 +849,143 @@ impl Context {
...
+        ) -> MaintainLocalTimer {
+            let handle = &self.worker.handle;
+            let (core, (should_yield, next_timer)) =
+                self.enter_with_time_context(core, |time_cx| {
+                    util::time::remove_cancelled_timers(&mut time_cx.wheel, &mut time_cx.canc_rx);
+                    let should_yield = util::time::insert_inject_timers(
+                        &mut time_cx.wheel,
+                        &time_cx.canc_tx,
+                        handle.take_remote_timers(),
+                    );
```
Darksonn: For cancellation, we probably also have to move timers to linked list, and cancel after releasing `RefMut`. (I think.)

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -1131,6 +1324,26 @@ impl Handle {
...
     }
 
+    cfg_time! {
+        /// Push a timer handle from the remote thread.
+        pub(crate) fn push_remote_timer(&self, hdl: EntryHandle) {
+            {
+                let mut synced = self.shared.synced.lock();
+                synced.inject_timers.push(hdl);
+            }
+            self.notify_parked_remote();
```
Darksonn: Maybe we should use a different mutex from task injectoin queue for contention?

tokio/src/runtime/time/wheel/entry.rs
```diff
@@ -0,0 +1,370 @@
...
+    Unregistered(Waker),
+
+    /// The entry is registered to the timer wheel,
+    /// but not in the pending queue of the timer wheel.
+    Registered(Sender, Waker, ThreadId),
+
+    /// The entry is in the pending queue of the timer wheel,
+    /// and not in any wheel level, which means that
+    /// the entry is reached its deadline and waiting to be woken up.
+    Pending(Sender, Waker, ThreadId),
```
Darksonn: Since `block_in_place` may move the `Core` from one thread to another, using the thread id may be a problem.

tokio/src/runtime/time/wheel/entry.rs
```diff
@@ -0,0 +1,370 @@
...
+                panic!("corrupted state: {state:#?}")
+            }
+            // don't unlock — poisoning the `Mutex` stops others from using the bad state.
+            PrivState::Pending(..) => panic!("corrupted state: State::Pending"),
+            // don't wake up cancelling entries
+            PrivState::Cancelling { .. } => (),
+        }
+    }
+
+    pub(crate) fn register_waker(&self, waker: &Waker) {
```
Darksonn: Consider dropping old waker after releasing lock.

tokio/src/runtime/time/timer.rs
```diff
@@ -0,0 +1,187 @@
...
+                };
+
+                if let Some(TimeContext::Running { wheel, canc_tx: _ }) = maybe_time_cx {
+                    if thread_id == cur_thread_id {
+                        // Safety:
+                        // 1. entry is either in slots or pending list
+                        // 2. entry is registered in this thread
+                        unsafe {
+                            wheel.remove(entry);
+                        }
```
Darksonn: Maybe we can just call `entry.transition_to_cancelling()` always here to avoid relying on thread id?

tokio/src/runtime/scheduler/multi_thread/worker.rs
ADD-SP: Great idea! It is implemented in <https://github.com/tokio-rs/tokio/pull/7467/commits/bc243b818e2aa42b8bafab2d8edbec918fc250fb>.

tokio/src/runtime/scheduler/multi_thread/worker.rs
ADD-SP: Move this discussion to <https://github.com/tokio-rs/tokio/pull/7467#discussion_r2532011757>.

tokio/src/runtime/time/wheel/entry.rs
ADD-SP: Good catch! The `ThreadId` here is not enough safe for canceling the timer locally. Fixed by <https://github.com/tokio-rs/tokio/pull/7467/commits/9a4d4390460776452ea69dd38b3a6abdd37d144e>, which send the timer to the cancellation queue unconditionally. It’s not worth adding this complexity for a performance optimization at this stage.

tokio/src/runtime/time/wheel/entry.rs
ADD-SP: Done by <https://github.com/tokio-rs/tokio/pull/7467/commits/cde17cdd03e3f7f79a1e0ada56a48dc10b9e9109>.

tokio/src/runtime/time/timer.rs
ADD-SP: Move this discussion to <https://github.com/tokio-rs/tokio/pull/7467#discussion_r2532013784>.

tokio/src/runtime/time_alt/entry.rs
```diff
@@ -0,0 +1,550 @@
...
+    /// - The [`Entry::wheel_pointers`] is NOT currently in use.
+    /// - The [`Entry::extra_pointers`] is NOT currently in use.
+    WokenUp,
+
+    /// The [`Entry`] is in the cancellation queue,
+    ///
+    /// - The [`Entry::wheel_pointers`] is MAYBE in use by [`super::Level::slot`] or [`super::Wheel::pending`].
+    /// - The [`Entry::extra_pointers`] is currently in use by [`super::cancellation_queue`].
+    Cancelling(Cancelling),
+}
```
Darksonn, line 17: I've been thinking about how to simplify the `PrivState` enum, and I have a proposal.

I think the only state we actually need is:
* a boolean for whether the timer has been cancelled.
* when in a timer wheel, some mechanism to insert it into the todo-list on the same worker thread as the timer wheel.

To see why, first let me summarize the different lists again:

1. The runtime has a global queue for timer events. (Uses `extra_pointers`.)
2. For each worker thread there is:
   * A timer wheel with a linked list for each bucket. (Uses `wheel_pointers`.)
   * A todo-list of timers that need processing. (Uses `extra_pointers`.)

The main idea here is to replace the cancellation queue with a todo-list that both handles insertion and cancellation.

So what the runtime then does is as follows:

When registering a new timer, it goes on the local todo-list if on a worker thread, otherwise the global queue.

When canceling a timer, set the cancelled boolean, then directly add it to the todo-list on the same worker thread as the timer wheel. (If already cancelled, do nothing.)

Before parking, we process (see below) the global queue and local todo-list before going to sleep.

After parking, we turn the timer wheel. All expired timers are added to the local todo-list. Then the local todo-list is processed using the same logic as before parking.

To process the todo list we do the following:
* If the item is marked cancelled, then we just remove it from the timer wheel and then free it.
* If the item is not marked cancelled but the deadline is expired, then wake the waker, mark it cancelled, and process it like a cancelled item.
* Otherwise, insert it into the timer wheel.

ADD-SP: Good idea! done

tokio/src/runtime/time/wheel/entry.rs
```diff
@@ -0,0 +1,262 @@
...
+            waker.wake();
+        }
+    }
+
+    pub(crate) fn register_cancel_tx(&self, cancel_tx: Sender) {
+        let mut lock = self.entry.state.lock();
+        if !lock.cancelled && !lock.woken_up {
+            // don't unlock — poisoning the `Mutex` stops others from using the bad state.
+            assert!(
+                lock.cancel_tx.replace(cancel_tx).is_none(),
```
Darksonn:
```suggestion
            let old_tx = lock.cancel_tx.replace(cancel_tx);
            assert!(
                old_tx.is_none(),
```

tokio/src/runtime/time_alt/entry.rs
```diff
@@ -0,0 +1,262 @@
...
+
+    pub(crate) fn cancel(&self) {
+        let mut lock = self.entry.state.lock();
+        if !lock.cancelled {
+            lock.cancelled = true;
+            if let Some(cancel_tx) = lock.cancel_tx.take() {
+                drop(lock);
+                unsafe {
+                    cancel_tx.send(self.clone());
+                }
```
Darksonn, line 252: Needs safety comment.

tokio/src/runtime/time/wheel/entry.rs
```diff
@@ -0,0 +1,262 @@
...
+pub(crate) struct Entry {
+    /// The intrusive pointer used by [`super::cancellation_queue`].
+    cancel_pointers: linked_list::Pointers<Entry>,
+
+    /// The intrusive pointer used by any of the following queues:
+    ///
+    /// - [`super::Level::slot`]
+    /// - [`super::RegistrationQueue`]
+    /// - [`super::WakeQueue`]
+    extra_pointers: linked_list::Pointers<Entry>,
```
Darksonn: I think I need to see a summary of why all push/remove operations on these lists are safe.

tokio/src/runtime/time_alt/entry.rs
```diff
@@ -0,0 +1,262 @@
...
+    cancelled: bool,
+    woken_up: bool,
+    waker: Option<Waker>,
+    cancel_tx: Option<Sender>,
+}
+
+#[derive(Debug)]
+pub(crate) struct Entry {
+    /// The intrusive pointer used by [`super::cancellation_queue`].
+    cancel_pointers: linked_list::Pointers<Entry>,
```
Darksonn, line 22: Needs summary of why we don't cancel twice (list safety requirements).

tokio/src/runtime/time/wheel/entry.rs
Darksonn: Think about what it would take to support re-registering the same timer without causing trouble here.

tokio/src/runtime/scheduler/util.rs
```diff
@@ -0,0 +1,206 @@
+cfg_rt_and_time! {
+    pub(crate) mod time {
```
Darksonn: Rustfmt does not work inside `cfg_*` macros, so please avoid wrapping entire files in them. Consider e.g. moving the macro to the parent file so it just wraps `mod filename;` instead of the actual contents.

tokio/src/runtime/time/tests/mod.rs
Darksonn, line 1: Could you check if various scenarios are covered by tests?

* Create timer on worker thread, on current-thread rt, or outside of runtime.
* Cancel on same worker thread, different worker thread, outside of runtime.

ADD-SP: It is hard to cover the multi-thread runtime even in unit test, I tried my best to cover the current runtime: <https://github.com/tokio-rs/tokio/pull/7467/commits/ef4c37f9ded2136666ff60182ec2c85777472f17>.

tokio/src/runtime/scheduler/util.rs
ADD-SP: done by <https://github.com/tokio-rs/tokio/pull/7467/commits/0215b54053808e579a0b687291a982621c1132d0>

tokio/src/runtime/time_alt/entry.rs
ADD-SP: done by <https://github.com/tokio-rs/tokio/pull/7467/commits/aebf2b3b447ebd26c873e7c5aa38c787a3e43853>

tokio/src/runtime/time/wheel/entry.rs
ADD-SP: done by <https://github.com/tokio-rs/tokio/pull/7467/commits/aebf2b3b447ebd26c873e7c5aa38c787a3e43853>

tokio/src/runtime/time/wheel/entry.rs
ADD-SP: > I think I need to see a summary of why all push/remove operations on these lists are safe.

done by <https://github.com/tokio-rs/tokio/pull/7467/commits/aebf2b3b447ebd26c873e7c5aa38c787a3e43853>

tokio/src/runtime/time_alt/entry.rs
ADD-SP: done by <https://github.com/tokio-rs/tokio/pull/7467/commits/aebf2b3b447ebd26c873e7c5aa38c787a3e43853>

Darksonn:
It would be good to check whether this has the same bug as https://github.com/tokio-rs/tokio/issues/6682, since this also introduces sharding of timers.

carllerche:
Thanks to both @ADD-SP and @Darksonn for sticking with this. I know it is a big complex one.

I commented in the issue that I am fine w/ the benchmark perf tradeoff.

This change is a really big one that is fundamentally changing a core part of Tokio. We can't just just drop it in all at once. What I suggest is we take a phased approach, similar to what I did with the alt multi-threaded runtime. The alt multi-threaded runtime ultimately didn't land, but that is actually a success IMO because a) bugs were uncovered after merging it that were not during testing / loom and b) I was able to get real services to test the runtime and was not able to demonstrate any real-world benefits (micro benches can be decoupled from reality).

So, what I would suggest is to start a new PR with the code from this one.
- Keep the old timer code as the default timer
- Add your timer code as a completely separate timer implementation.
- Use an enum for the runtime to pick between which one is enabled (similar to the alt MT runtime)
- Only enable the new timer strategy with `tokio_unstable`. 
- Use a runtime configuration setting (only available with `tokio_unstable`) to opt into the new timer implementation.

Then we can merge the PR and start the next phase of testing: getting real service teams to try it out. Having **both** available in a released version of Tokio makes this testing easier. You can use deployment configuration to only use the new timer code in a fraction of the production servers. This reduces risk and lets us collect data.

Assuming the first phase works, we can then consider swapping the default timer implementation to the new one when `tokio_unstable` is enabled, followed by making it available to all.

Also, I think being able to pick from timer implementations is a long term win. For example, we may want to have a "current-thread" optimized timer, and/or variants that are only very coarse. For example, maybe @Noah-Kennedy / and other pingora maintainers could bring in what they need to a timer variant.

ADD-SP:
> It would be good to check whether this has the same bug as #6682, since this also introduces sharding of timers.

Thanks for this context! @Darksonn 

The root cause of #6682 is that next_wake is not up to date when a timer is [`reregister`]ed. In this PR, we don’t have this issue because:

- When the local wheel is available, the existing worker thread is still running, so we don’t need next_wake to decide whether to unpark the resource driver.
- When the local wheel is not available, [`push_remote_timer`] always calls [`notify_parked_remote`].


[`reregister`]: https://github.com/tokio-rs/tokio/blob/9a1b076c0084e2bca8558818bffd2273ec5ae5bd/tokio/src/runtime/time/mod.rs#L355
[`push_remote_timer`]: https://github.com/ADD-SP/tokio/blob/c1a8676a315e063a742f88d05728ec35c7a82c40/tokio/src/runtime/scheduler/multi_thread/worker.rs#L1346
[`notify_parked_remote`]: https://github.com/ADD-SP/tokio/blob/c1a8676a315e063a742f88d05728ec35c7a82c40/tokio/src/runtime/scheduler/multi_thread/worker.rs#L1351

tokio/src/time/interval.rs
```diff
@@ -444,7 +444,8 @@ impl Interval {
         #[cfg(not(all(tokio_unstable, feature = "tracing")))]
         let instant = poll_fn(|cx| self.poll_tick(cx));
 
-        instant.await
+        let r = instant.await;
+        r
```
Darksonn: ?

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -1131,6 +1351,27 @@ impl Handle {
...
+        {
+            let mut synced = self.shared.synced.lock();
+            synced.inject_timers.push(hdl);
+        }
+        self.notify_parked_remote();
+    }
+
+    #[cfg(all(tokio_unstable, feature = "time"))]
+    pub(crate) fn take_remote_timers(&self) -> Vec<time_alt::EntryHandle> {
+        assert_eq!(self.timer_flavor, TimerFlavor::Alternative,);
```
Darksonn:
```suggestion
        assert_eq!(self.timer_flavor, TimerFlavor::Alternative);
```

tokio/src/runtime/scheduler/mod.rs
```diff
@@ -249,6 +293,28 @@ cfg_rt! {
...
+        // #[cfg(all(tokio_unstable, feature = "time", feature = "rt-multi-thread", not(target_os = "wasi"), test))]
+        // pub(crate) fn with_time_local_context<F, R>(&self, f: F) -> R
+        // where
+        //     F: FnOnce(Option<&mut crate::runtime::time_alt::LocalContext>) -> R,
+        // {
+        //     match self {
+        //         Context::CurrentThread(_) => panic!("the alternative timer implementation is not supported on CurrentThread runtime"),
+        //         Context::MultiThread(context) => context.with_time_local_context(f),
+        //     }
+        // }
```
Darksonn: ?

Darksonn:
@carllerche Good call to make it optional. I think it's a lot easier to merge this PR now. There are a few pieces of test code remaining and such, but once that's removed I think we can ship it as an optional configuration.

PR #7468, opened 2025-07-19
Fix io Mock deadlock with `wait` action.

### LINKED ISSUE #7445: Splitting a tokio-test Mock instance can cause deadlock/lockup
Using a `tokio_test::io::Mock` to read and write simultaneously via `tokio::io::split` causes lockup if the `Mock` is waiting, and there's no data to read.

This is a simple test, enough to trigger it for me:

```rust
#[tokio::test]
async fn check_io() {
    let socket = tokio_test::io::Builder::new()
        .wait(Duration::from_millis(10))
        .write([0].as_slice())
        .build();

    let (mut recv, mut send) = tokio::io::split(socket);
    tokio::spawn(async move { recv.read_u8().await.unwrap() });
    send.write_u8(0).await.unwrap();
}
```

I'm not super familiar with the Mock API, but I'd expect this to finish after 10ms, possibly with an error due to the extra read. Instead, it hangs forever.

As far as I'm aware, this should not be a problem unless the Mock is blocking in a poll call of some sort, which may be a deeper issue?

**Version**
```
├── tokio v1.45.0
│   └── tokio-macros v2.5.0 (proc-macro)
└── tokio-openssl v0.6.5
    └── tokio v1.45.0 (*)
└── tokio v1.45.0 (*)
├── tokio v1.45.0 (*)
└── tokio-util v0.7.15
    └── tokio v1.45.0 (*)
├── tokio v1.45.0 (*)
└── tokio-test v0.4.4
    ├── tokio v1.45.0 (*)
    └── tokio-stream v0.1.17
        └── tokio v1.45.0 (*)
```
**Platform**
`Linux [...] 5.15.0-139-generic #149~20.04.1-Ubuntu SMP [...] x86_64 x86_64 x86_64 GNU/Linux`

Cyberboss:
## Motivation

Addressing potential deadlock by Mock io system.

## Solution

A single `Sleep` future was being shared on two separate code_paths, despite being callable from different async contexts. As described in the issue, this can result in the waker being unintentionally overwritten.

Fixing this by creating a distinct future for both `poll_write` and `poll_read`.

First contribution to tokio. Mainly unsure if the test is in the right module if someone wants to point that out.

Fixes #7445

### DIFF
--- a/tokio-test/src/io.rs
+++ b/tokio-test/src/io.rs
@@ -69,7 +69,8 @@ enum Action {
 struct Inner {
     actions: VecDeque<Action>,
     waiting: Option<Instant>,
-    sleep: Option<Pin<Box<Sleep>>>,
+    read_sleep: Option<Pin<Box<Sleep>>>,
+    write_sleep: Option<Pin<Box<Sleep>>>,
     read_wait: Option<Waker>,
     rx: UnboundedReceiverStream<Action>,
     name: String,
@@ -199,7 +200,8 @@ impl Inner {
 
         let inner = Inner {
             actions,
-            sleep: None,
+            read_sleep: None,
+            write_sleep: None,
             read_wait: None,
             rx,
             waiting: None,
@@ -363,12 +365,12 @@ impl AsyncRead for Mock {
         buf: &mut ReadBuf<'_>,
     ) -> Poll<io::Result<()>> {
         loop {
-            if let Some(ref mut sleep) = self.inner.sleep {
+            if let Some(ref mut sleep) = self.inner.read_sleep {
                 ready!(Pin::new(sleep).poll(cx));
             }
 
             // If a sleep is set, it has already fired
-            self.inner.sleep = None;
+            self.inner.read_sleep = None;
 
             // Capture 'filled' to monitor if it changed
             let filled = buf.filled().len();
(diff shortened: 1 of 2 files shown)


ADD-SP (commented):
I'm re-thinking the behavior of `Builder::wait()` to ensure the behavior doesn't surprise the downstream,  For this example, what should `elapsed`'s value be?

```rust
use tokio_test::io::Builder;

let mock = Builder::Default()
    .wait(Duration::from_sec(30))
    .read([0].as_slice())
    .write([0].as_slice())
    .build();

let (mut recv, mut send) = tokio::io::split(mock);

tokio::spawn(async move {
    let _ = recv.read_u8().await;
})

// let's assume that the previous `recv.read_u8()` has been polled
// at this point.

    let now = Instant::now();
    let _ = send.write_u8(0).await;
    now.elapsed
})

let elapsed = jh.await.unwrap();
```

1. `elapsed` is 30 seconds, which means that the `wait()` blocks all the following I/O operations.
2. `elapsed` is almost zero, which means that the `wait()` only block the next operation, and other following operations are not affected.

For (1), read and write are very likely to be woken up at the same time, and write might be polled before the read, leading the test failure due to the write occurs before the read.

For (2), write never be blocked, but this behavior might surprise the downstream, my feeling when I first see this example is that the read always occurs before the write.

I think (1) is unlikely to surprise the downstream, however, we need to do something to avoid the write occurs before the read.

Could you share your through about this topic? Which behavior is less surprise-able?

ADD-SP (commented):
I renamed this PR to:

> io: fix the lost waker issue of `tokio_test::io::Mock`

ADD-SP:
Hi @Cyberboss, are you still working on this PR? I believe the actual fix is much complex than oneline change, feel free to leave comments in the original issue if you are thinking possible solutions.

Cyberboss:
Going to close this out. Think it's a bit beyond what I thought it was

PR #7470, opened 2025-07-20
ci: use ubuntu-24.04-arm instead of ubuntu-22.04-arm

taiki-e:
## Motivation

The upstream issue causing crashes (https://github.com/rust-lang/rust/issues/135867) has been fixed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -583,9 +583,9 @@ jobs:
           - target: armv5te-unknown-linux-gnueabi
             os: ubuntu-latest
           - target: armv7-unknown-linux-gnueabihf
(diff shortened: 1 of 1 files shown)


ADD-SP (approved):
Additional context

> Thanks for your patience with the random failures. We discovered it was an issue with the underlying virtual machine and have migrated all the runners to a different arm64 sku (dpdsv5) and this should solve the errors you were seeing. We reported the issues to the Microsoft Azure team, and they are working to debug them, but in the meantime we have migrated our runners so that you can use the runners without issue.
>
> [*actions/partner-runner-images#36*](https://github.com/actions/partner-runner-images/issues/36#issuecomment-2679504033)

PR #7475, opened 2025-07-22
future: add adapters of `CancellationToken` for `FutureExt`

### LINKED ISSUE #7466: Consider add `timeout_at` and `with_cancel` method in `tokio_util:: FutureExt` trait
Hello! In most cases where I'm currently using this, I need to simultaneously use `tokio::time::timeout/timeout_at` and `CancellationToken::run_until_cancelled` method to listen for both timeout and exit signals. The current implementation approach is `cancellationtoken.run_until_cancelled(myfut.timeout(duration))`, where `timeout` method comes from `tokio_util::FutureExt`. If a method like `with_cancel` could be added to `FutureExt`, it would enable method chaining and make the code much more concise.

```rust
use std::{future::ready, time::Duration};

use tokio_util::{sync::CancellationToken, time::FutureExt};

#[tokio::main]
async fn main() {
    let cancel = CancellationToken::new();

    match cancel
        .run_until_cancelled(ready(1).timeout(Duration::from_millis(100)))
        .await
    {
        Some(Ok(res)) => {
            println!("{res}")
        }
        Some(Err(_)) => {
            println!("elapsed")
        }
        None => {
            println!("cancelled")
        }
    }
}
```

after add `with_cancel` method

```rust
use std::{
    future::ready,
    pin::Pin,
    task::{Context, Poll},
    time::Duration,
};

use pin_project_lite::pin_project;
use tokio::time::{Instant, Timeout};
use tokio_util::sync::{CancellationToken, WaitForCancellationFuture};

trait FutureExt: Future {
    fn timeout(self, timeout: Duration) -> Timeout<Self>
    where
        Self: Sized,
    {
        tokio::time::timeout(timeout, self)
    }

    // ********** new method **********
    fn timeout_at(self, timeout_at: Instant) -> Timeout<Self>
    where
        Self: Sized,
    {
        tokio::time::timeout_at(timeout_at, self)
    }

    // ********** new method **********
    fn with_cancel(self, cancel: &CancellationToken) -> RunUntilCancelledFuture<'_, Self>
    where
        Self: Sized,
    {
        RunUntilCancelledFuture {
            cancellation: cancel.cancelled(),
            future: self,
        }
    }
}

impl<T: Future + ?Sized> FutureExt for T {}

// ******* This RunUntilCancelledFuture is copy from run_until_cancelled method *********
pin_project! {
    struct RunUntilCancelledFuture<'a, F: Future> {
        #[pin]
        cancellation: WaitForCancellationFuture<'a>,
        #[pin]
        future: F,
    }
}

impl<'a, F: Future> Future for RunUntilCancelledFuture<'a, F> {
    type Output = Option<F::Output>;

        let this = self.project();
        if let Poll::Ready(res) = this.future.poll(cx) {
            Poll::Ready(Some(res))
        } else if this.cancellation.poll(cx).is_ready() {
            Poll::Ready(None)
        } else {
            Poll::Pending
        }
    }
}

#[tokio::main]
async fn main() {
    let cancel = CancellationToken::new();
    match ready(1)
        .with_cancel(&cancel)
        .timeout(Duration::from_millis(100))
        .await        // <-------- here we use fluent call chains --------
    {
        Ok(Some(res)) => {
            println!("{res}")
        }
        Ok(None) => {
            println!("elapsed")
        }
        Err(_) => {
            println!("cancelled")
        }
    }
}
```

Moreover, the `timeout_at` method is also very useful in loops that require timeout limits.
```rust
#[tokio::main]
async fn main() {
    let cancel = CancellationToken::new();
    let timeout_at = Instant::now() + Duration::from_secs(10);

    let mut entries = Vec::new();
    loop {
        // ***** In scenarios where messages are continuously read from file/network ******
        match reader
            .read_entries()
            .timeout_at(timeout_at)
            .with_cancel(&cancel)
            .await
        {
            Some(Ok(res)) => {
                entries.extend(res);
            }
            Some(Err(_)) | None => break,
        }
    }
}
```

bitcapybara:
Closes #7466 

## Motivation

In my cases, I typically need to simultaneously listen for both `tokio::time::timeout` timeouts and `CancellationToken::cancelled` events to guide future exits. I noticed that `tokio_util::time::FutureExt enables method chaining, which is very helpful for improving code conciseness. Therefore, I hope to add a with_cancel method to FutureExt, allowing for calling patterns like future.timeout(Duration).with_cancel(CancellationToken).await.

## Solution

Move `FutureExt` to the `tokio-util::future` module and add methods like `timeout_at`, `with_cancel` and `with_cancel_owned`


Darksonn:
To avoid a breaking change, a re-export must be added to the time module so that existing users can continue using it via the time module.

bitcapybara:
I re-export the path of `FutureExt` in `time` module, please review it again, thanks!

bitcapybara:
Should `future` module be treated as a feature? Currently, the `future` module internally depends on `tokio::time`, which was previously enabled when the `time` feature was enabled.

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -439,3 +411,59 @@ impl Future for WaitForCancellationFutureOwned {
         }
     }
 }
+
+pin_project! {
+    /// A Future that is resolved once the corresponding [`CancellationToken`]
+    /// is cancelled or a given Future gets resolved. It is biased towards the
+    /// Future completion.
+    #[must_use = "futures do nothing unless polled"]
+    pub struct RunUntilCancelledFuture<'a, F: Future> {
```
ADD-SP: This creates a new public type, downstream imports this from the `tokio_util::sync::cancellation_token`. However, the `CancellationToken` itself never use this new type from the view of the downstream. This might confuse the downstream.

tokio-util/src/future.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// let token = CancellationToken::new();
+    /// let child_token = token.child_token();
+    /// tokio::spawn(async move {
+    ///     tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+    ///     token.cancel();
+    /// });
+    /// assert!(rx.with_cancel(&child_token).await.is_none())
+    /// }
+    /// ```
+    fn with_cancel(
```
ADD-SP: I personally prefer `with_cancellation_token`, because the developer can see at a glance what parameters are needed for this function, this reduces the mental burden.

I don't prefer `with_cancellation_token_ref` because this doesn't align to the naming pattern of `CancellationToken` itself.

tokio-util/src/future.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// let token = CancellationToken::new();
+    /// let child_token = token.child_token();
+    /// tokio::spawn(async move {
+    ///     tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+    ///     token.cancel();
+    /// });
+    /// assert!(rx.with_cancel_owned(child_token).await.is_none())
+    /// # }
+    /// ```
+    fn with_cancel_owned(
```
ADD-SP: I personally prefer `with_cancellation_token_owned`, because the developer can see at a glance what parameters are needed for this function, this reduces the mental burden.

Darksonn:
> Should `future` module be treated as a feature?

I would suggest keeping a `time` feature to match the Tokio features. The trait can provide the timeout methods only when the `time` feature is enabled.

tokio-util/src/sync/cancellation_token.rs
bitcapybara: Hello! Are you suggesting to change this type to ‎`pub(crate)`? This `RunUntilCancelledFuture` is returned by the `with_cancellation_token` method in `FutureExt` trait, which is public, and `RunUntilCancelledFuture` seems can not be more private. If I made it `pub(crate)`, there is an compile warning `type RunUntilCancelledFuture is more private than the item FutureExt::with_cancellation_token`.

tokio-util/src/future.rs
bitcapybara: Ok, I have renamed the two methods to ‎`with_cancellation_token` and ‎`with_cancellation_token_owned`.

bitcapybara:
> > Should `future` module be treated as a feature?
> 
> I would suggest keeping a `time` feature to match the Tokio features. The trait can provide the timeout methods only when the `time` feature is enabled.

Ok, I have removed the `future` feature and enabled `timeout`/`timeout_at` methods only when `time` feature enabled

tokio-util/src/sync/cancellation_token.rs
bitcapybara: I have changed these two new types in ‎`sync::cancellation_token` to ‎`pub(crate) use cancellation_token::{RunUntilCancelledFuture, RunUntilCancelledFutureOwned};`, so downstream users will not see these two types.

ADD-SP: I think there are two solutions.

One is that making both `CancellationToken::run_until_cancelled` and `CancellationToken::run_until_cancelled_owned` return a concrete type, which means that these functions are no longer async functions, so that we can make the `RunUntilCancelledFuture` public.

Another one is that write a new public type in the `tokio-util/src/futures.rs` and impl `Future` for it.

The first solution might be overkill for this feature, and `with_cancellation_token` returns `RunUntilCancelledFuture` is a little bit unusual for me, a type named `WithCancellationToken<'_>` is much more align to the tokio's pattern (see [`AsyncReadExt`]).

I prefer the second solution, and feel free to share your solution.

[`AsyncReadExt`]: https://docs.rs/tokio/1.46.1/tokio/io/trait.AsyncReadExt.html

bitcapybara: Thanks for your reply! Solution 2 is a very good for me, but since the logic of `with_cancellation_token` is the same as `run_until_cancelled`, the code for `WithCancellationToken` and `RunUntilCancelledFuture` would be the same. I'm concerned this might cause code maintenance issues. If such code duplication is acceptable, I'd be happy to make modifications according to this solution.

Darksonn: Please make sure that extension traits that return futures use a private type. This ensures that we can replace them with a real `async fn` once our MSRV is large enough to use async fn in traits. See `AsyncReadExt` for examples.

ADD-SP: > I'm concerned this might cause code maintenance issues. If such code duplication is acceptable, I'd be happy to make modifications according to this solution.

The duplication should be small enough and manageable, and we currently don't have a better solution. Let's try this, if we end up realizing this causes too much maintenance issue, then we can discuss later.

bitcapybara: Ok, I understand. I have added a new file called `with_cancellation_token` under the `future` module, and in this file, I have added these two types. Please review again. Thanks.

tokio-util/src/future/with_cancellation_token.rs
```diff
@@ -0,0 +1,83 @@
...
+use pin_project_lite::pin_project;
+
+use crate::sync::{CancellationToken, WaitForCancellationFuture, WaitForCancellationFutureOwned};
+
+pin_project! {
+    /// A Future that is resolved once the corresponding [`CancellationToken`]
+    /// is cancelled or a given Future gets resolved. It is biased towards the
+    /// Future completion.
+    #[must_use = "futures do nothing unless polled"]
+    pub struct WithCancellationTokenFuture<'a, F: Future> {
```
ADD-SP, line 19:
```suggestion
    pub(crate) struct WithCancellationTokenFuture<'a, F: Future> {
```

I know this is not exported, but we can make the visible scope more clear.

tokio-util/src/future/with_cancellation_token.rs
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+}
+
+pin_project! {
+    /// A Future that is resolved once the corresponding [`CancellationToken`]
+    /// is cancelled or a given Future gets resolved. It is biased towards the
+    /// Future completion.
+    #[must_use = "futures do nothing unless polled"]
+    pub struct WithCancellationTokenFutureOwned<F: Future> {
```
ADD-SP, line 54:
```suggestion
    pub(crate) struct WithCancellationTokenFutureOwned<F: Future> {
```

tokio-util/src/future.rs
```diff
@@ -0,0 +1,125 @@
+//! An extension trait for Futures that provides a variety of convenient adapters.
+
+mod with_cancellation_token;
+
+use std::future::Future;
+
+use with_cancellation_token::{WithCancellationTokenFuture, WithCancellationTokenFutureOwned};
```
ADD-SP:
```suggestion
mod with_cancellation_token;
use with_cancellation_token::{WithCancellationTokenFuture, WithCancellationTokenFutureOwned};

use std::future::Future;
```

nit: we prefer make the import statement more closer to the relevant `mod` statement.

tokio-util/src/future/with_cancellation_token.rs
bitcapybara: I tried this, but there will be a compiler warning: “type WithCancellationTokenFuture is more private than item FutureExt::with_cancellation_token.”

ADD-SP: Ah, I see, let's close this comment.

tokio-util/src/future/with_cancellation_token.rs
ADD-SP: Resolved, see https://github.com/tokio-rs/tokio/pull/7475#discussion_r2227159087.

tokio-util/src/future/with_cancellation_token.rs
bitcapybara: I looked at the code in `AsyncReadExt`, and the types it uses, such as `ReadU8` and `ReadU16`, are`pub`, while their `new` methods are `pub(crate)`. Maybe we can refer to this approach when implementing `WithCancellationTokenFuture`?

ADD-SP: > Maybe we can refer to this approach when implementing

That is also my thought, so I closed this comment.

tokio-util/src/future.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// async fn dox() {
+    /// let (tx, rx) = oneshot::channel::<()>();
+    /// let token = CancellationToken::new();
+    /// let child_token = token.child_token();
+    /// tokio::spawn(async move {
+    ///     tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+    ///     token.cancel();
+    /// });
+    /// assert!(rx.with_cancellation_token(&child_token).await.is_none())
+    /// }
```
ADD-SP:
```suggestion
    /// # async fn dox() {
    /// let (tx, rx) = oneshot::channel::<()>();
    /// let token = CancellationToken::new();
    /// let child_token = token.child_token();
    /// tokio::spawn(async move {
    ///     tokio::time::sleep(std::time::Duration::from_millis(10)).await;
    ///     token.cancel();
    /// });
    /// assert!(rx.with_cancellation_token(&child_token).await.is_none())
    /// # }
```

tokio-util/src/future.rs
```diff
@@ -0,0 +1,124 @@
...
+    ///
+    /// ```rust
+    /// use tokio::sync::oneshot;
+    /// use tokio_util::future::FutureExt;
+    /// use tokio_util::sync::CancellationToken;
+    ///
+    /// async fn dox() {
+    /// let (tx, rx) = oneshot::channel::<()>();
+    /// let token = CancellationToken::new();
+    /// let child_token = token.child_token();
```
ADD-SP: Why do we need a child_token here? I asked this because our example should be simple enough.

bitcapybara: Using `child_token` here doesn't have any special meaning. Using `token.clone()` here would also work fine - there's no difference.😊

ADD-SP: I prefer the `token.clone()`, the child-token made me start thinking about the purpose of child_token, and end up realizing there is nothing special.

Downstream developers just want to understand the API, we should care about things that might causes unnecessary mental burden. I guess downstream developers might think similarly to me while reading the example.

bitcapybara: Ok, I have changed it to `token.clone()`

tokio-util/src/future.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// ```
+    fn with_cancellation_token(
+        self,
+        cancellation_token: &CancellationToken,
+    ) -> WithCancellationTokenFuture<'_, Self>
+    where
+        Self: Sized,
+    {
+        WithCancellationTokenFuture::new(cancellation_token, self)
+    }
```
ADD-SP, line 97: We also need to check if the `cancellation_token` has already been cancelled before jumping into the behavior biased toward the future completion.

This is a recently fixed bug, I just realized this is also what we need to care in this feature, see #7462 for more details.

bitcapybara: I noticed that the implementation of `Timeout` Future also biased waiting for the inner Future to complete first. Here, the behavior of `WithCancellationToke`n is roughly the same as `Timeout`. If we want to ensure that `timeout` and `with_cancellation_token` are biased to complete. I have two solutions:

1. Perhaps we can redefine and return an Enum Future type, with one of its variants being a ready type, which is a breaking change for timeout method.
2. Alternatively, we could just modify the behavior of `with_cancellation_token`, which wouldn’t affect my case anyway

ADD-SP: > I noticed that the implementation of `Timeout` Future also biased waiting for the inner Future to complete first.

Yes, but this doesn't surprise the downstream. Consider the following three scenarios.

1. `socket.read_u8().timeout(Duration::from_secs(10).await`
2. The waker will be stored by both I/O driver and Time driver.
3. I/O is ready at the 3rd second.
4. I/O driver wake up the future, and biased behavior polls the `read_u8` at first.
5. Downstream get an `u8` from the socket.

***

1. `socket.read_u8().timeout(Duration::from_secs(10).await`
2. The waker will be stored by both I/O driver and Time driver.
3. At the 10th second, I/O event is still not ready, and the Time driver wakes up the future.
4. The biased behavior polls the `read_u8` at first, and then polls the timers.
5. Downstream gets the timeout error.

***

1. `socket.read_u8().timeout(Duration::from_secs(10).await`
2. The waker will be stored by both I/O driver and Time driver.
3. I/O event is ready at the 10th second, at the same time, the Time driver also fires the timer.
4. The biased behavior polls the `read_u8` at first.
5. Downstream get an `u8` from the socket.

***

The above three scenarios would not seem to surprise downstream.

bitcapybara: Oh, got it. I have another confusion. Regarding the behavior of the `timeout_at` method,, if the provided Instant is a time point in the past, should the timeout be polled first, or should the inner Future be polled first?

bitcapybara: I tried making `timeout_at` return an Enum Future type, but there’s a problem: when a past Instant is passed in, it needs to return an `Elapsed` error. However, I can’t construct it because its `new` method is `pub(crate)`.

bitcapybara: For the `WithCancellationToken` Future, should we check the `CancellationToken` only once at the beginning of the method call as in #7462 , or should we biased polling the `CancellationToken`’s Future every time `poll` method is called? Which do you prefer?

bitcapybara: It seems that implementing functionality similar to #7462 is not very straightforward. Since #7462 is an async function, it can directly return None. However, FutureExt returns a Future type, so to achieve similar functionality, it might require returning an enum Future type (similar to `tokio_util::either::Either`, but `Either` is a `pub` type and cannot be used here) with one variant being Ready. 
In my current implementation, `WithCancellationToken` is biased toward first checking whether the `CancellationToken` has already been cancelled when polling, in order to achieve this effect.

tokio-util/src/future.rs
Darksonn: The meaning of `child_token()` is not the same as `token.clone()`. With a cloned token, there is only a single token that is either cancelled or not. Whereas with a child token, you have two tokens A and B where cancelling A also cancels B, but cancelling B does not cancel A.

tokio-util/src/future.rs
Darksonn: We do not want duplicate implementations. If you need to create a future here, then you should also change `run_until_cancelled` to use that new future instead of the existing `RunUntilCancelledFuture`. One option could be to have `RunUntilCancelledFuture` store an `Option<WaitForCancellationFuture>` so that it can be created on first poll.

ADD-SP: Ah, yes, this is a better solution, which doesn't change the interface of `CancellationToken`, also doesn't duplicate the implementation.

bitcapybara: Hello, so according to my understanding, your meaning is that if the `CancellationToken` has already been cancelled, then return a `RunUntilCancelledFuture` that carries a `WaitForCancellationFuture` with the value  `None`?

ADD-SP: > if the provided Instant is a time point in the past, should the timeout be polled first, or should the inner Future be polled first?

The current implementation of `tokio::time::timeout` always polls the inner `Future` at first, I'd like to keep it as it is as no one complain about it.

ADD-SP: > Hello, so according to my understanding, your meaning is that if the CancellationToken has already been cancelled, then return a RunUntilCancelledFuture that carries a WaitForCancellationFuture with the value None?

I think this is the core idea, let's implement this at first, then we can discuss some details.

ADD-SP: > Hello, so according to my understanding, your meaning is that if the CancellationToken has already been cancelled, then return a RunUntilCancelledFuture that carries a WaitForCancellationFuture with the value None?

I think this is the core idea, let's implement this at first, then we can discuss some details.

tokio-util/src/sync/cancellation_token.rs
```diff
@@ -439,3 +411,77 @@ impl Future for WaitForCancellationFutureOwned {
...
+        }
+    }
+}
+
+pin_project! {
+    /// A Future that is resolved once the corresponding [`CancellationToken`]
+    /// is cancelled or a given Future gets resolved. It is biased towards the
+    /// Future completion.
+    #[must_use = "futures do nothing unless polled"]
+    pub struct RunUntilCancelledFutureOwned<F: Future> {
```
ADD-SP: Not sure if you have already tried to make it `pub(crate)`?

bitcapybara: Oh, that’s my mistake—I thought it would give the previous “more private” compile warning.

tokio-util/src/future/with_cancellation_token.rs
```diff
@@ -0,0 +1,73 @@
...
+pin_project! {
+    /// A Future that is resolved once the corresponding [`CancellationToken`]
+    /// is cancelled or a given Future gets resolved. It is biased towards the
+    /// [`CancellationToken`] cancelled.
+    #[must_use = "futures do nothing unless polled"]
+    pub struct WithCancellationTokenFuture<'a, F: Future> {
+        #[pin]
+        run_until_cancelled: Option<RunUntilCancelledFuture<'a, F>>
+    }
+}
```
ADD-SP:
```suggestion
pin_project! {
    /// A [`Future`] that is resolved once the corresponding [`CancellationToken`]
    /// is cancelled or a given [`Future`] gets resolved.
    ///
    /// This future is Immediately resolved if the corresponding [`CancellationToken`]
    /// is already cancelled, otherwise, in case of concurrent completion and
    /// cancellation, this is biased towards the future completion.
    #[must_use = "futures do nothing unless polled"]
    pub struct WithCancellationTokenFuture<'a, F: Future> {
        #[pin]
        run_until_cancelled: Option<RunUntilCancelledFuture<'a, F>>
    }
}
```

> It is biased towards the [`CancellationToken`] cancelled.

Emm, my understanding of this sentence is that this future is always biased towards the cancellation, but this is only true if the token is already cancelled before constructing this struct.

So I tried to write another version to avoid confusion.

tokio-util/src/future.rs
```diff
@@ -0,0 +1,124 @@
...
+    ///     tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+    ///     token.cancel();
+    /// });
+    /// assert!(rx.with_cancellation_token(&token_clone).await.is_none())
+    /// # }
+    /// ```
+    fn with_cancellation_token(
+        self,
+        cancellation_token: &CancellationToken,
+    ) -> WithCancellationTokenFuture<'_, Self>
```
ADD-SP: Please also highlight the fairness here, we could refer the below one.

https://github.com/tokio-rs/tokio/blob/911ab21d7012a50e53971ad1292a9f18de22d4c8/tokio-util/src/sync/cancellation_token.rs#L271-L275

tokio-util/src/lib.rs
```diff
@@ -59,3 +59,5 @@ pub mod either;
 pub use bytes;
 
 mod util;
+
+pub mod future;
```
ADD-SP, line 63: Since exposing a new module at the crate root is a significant change, I'd like to hear @Darksonn 's opinion. Does this change make sense for you?

ADD-SP (commented):
Please also add some tests against these two new APIs.

tokio-util/src/future/with_cancellation_token.rs
bitcapybara: Thanks! I forgot to update the comments.

tokio-util/src/future/with_cancellation_token.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use pin_project_lite::pin_project;
+
+use crate::sync::{CancellationToken, RunUntilCancelledFuture, RunUntilCancelledFutureOwned};
+
+pin_project! {
+    /// A [`Future`] that is resolved once the corresponding [`CancellationToken`]
+    /// is cancelled or a given [`Future`] gets resolved.
+    ///
+    /// This future is Immediately resolved if the corresponding [`CancellationToken`]
```
ADD-SP:
```suggestion
    /// This future is immediately resolved if the corresponding [`CancellationToken`]
```

I found a typo written by myself.

tokio-util/src/future/with_cancellation_token.rs
```diff
@@ -0,0 +1,79 @@
...
+            None => Poll::Ready(None),
+        }
+    }
+}
+
+pin_project! {
+    /// A [`Future`] that is resolved once the corresponding [`CancellationToken`]
+    /// is cancelled or a given [`Future`] gets resolved.
+    ///
+    /// This future is Immediately resolved if the corresponding [`CancellationToken`]
```
ADD-SP:
```suggestion
    /// This future is immediately resolved if the corresponding [`CancellationToken`]
```

tokio-util/src/time/mod.rs
```diff
@@ -8,45 +8,18 @@
...
-use std::future::Future;
 use std::time::Duration;
-use tokio::time::Timeout;
 
 mod wheel;
 
 pub mod delay_queue;
 
+/// re-export `FutureExt` to avoid breaking change
+pub use crate::future::FutureExt;
```
ADD-SP:
```suggestion
// re-export `FutureExt` to avoid breaking change
pub use crate::future::FutureExt;
```

docstring will be rendered in the `docs.rs`, which is not what we want to do.

tokio-util/src/future.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// fluent call chains.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use tokio::{sync::oneshot, time::Duration};
+    /// use tokio_util::future::FutureExt;
+    ///
+    /// # async fn dox() {
+    /// let (tx, rx) = oneshot::channel::<()>();
```
ADD-SP:
```suggestion
    /// let (_tx, rx) = oneshot::channel::<()>();
```

This makes me understand the example quicker than before.

tokio-util/src/future.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// fluent call chains.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use tokio::{sync::oneshot, time::{Duration, Instant}};
+    /// use tokio_util::future::FutureExt;
+    ///
+    /// # async fn dox() {
+    /// let (tx, rx) = oneshot::channel::<()>();
```
ADD-SP:
```suggestion
    /// let (_tx, rx) = oneshot::channel::<()>();
```

This makes me understand the example quicker than before.

tokio-util/src/future.rs
```diff
@@ -0,0 +1,136 @@
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use tokio::sync::oneshot;
+    /// use tokio_util::future::FutureExt;
+    /// use tokio_util::sync::CancellationToken;
+    ///
+    /// # async fn dox() {
+    /// let (tx, rx) = oneshot::channel::<()>();
```
ADD-SP:
```suggestion
    /// let (_tx, rx) = oneshot::channel::<()>();
```

This makes me understand the example quicker than before.

tokio-util/src/future.rs
```diff
@@ -0,0 +1,136 @@
...
```
ADD-SP:
```suggestion
    /// let (_tx, rx) = oneshot::channel::<()>();
```

This makes me understand the example quicker than before.

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    let ready_with_cancellation_token_fut = ready_fut.with_cancellation_token(&token);
+
+    pin!(ready_with_cancellation_token_fut);
+
+    let res = ready_with_cancellation_token_fut
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+
+    assert_eq!(res, Poll::Ready(Some(())));
```
ADD-SP: You may want [`tokio_test::assert_ready_eq`].

[`tokio_test::assert_ready_eq`]: https://docs.rs/tokio-test/0.4.4/tokio_test/macro.assert_ready_eq.html

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    let ready_with_cancellation_token_fut = pending_fut.with_cancellation_token(&token);
+
+    pin!(ready_with_cancellation_token_fut);
+
+    let res = ready_with_cancellation_token_fut
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+
+    assert_eq!(res, Poll::Pending);
```
ADD-SP: Please try the [`tokio_test::assert_pending`].

[`tokio_test::assert_pending`]: https://docs.rs/tokio-test/0.4.4/tokio_test/macro.assert_pending.html

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+    let (waker, _) = new_count_waker();
+    let token = CancellationToken::new();
+
+    let pending_fut = pending::<()>();
+
+    let ready_with_cancellation_token_fut = pending_fut.with_cancellation_token(&token);
+
+    pin!(ready_with_cancellation_token_fut);
+
+    let res = ready_with_cancellation_token_fut
```
ADD-SP:
```suggestion
    let pending_with_cancellation_token_fut = pending_fut.with_cancellation_token(&token);

    pin!(pending_with_cancellation_token_fut);

    let res = pending_with_cancellation_token_fut
```

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+    let token = CancellationToken::new();
+    token.cancel();
+
+    let pending_fut = pending::<()>();
+
+    let ready_with_cancellation_token_fut = pending_fut.with_cancellation_token(&token);
+
+    pin!(ready_with_cancellation_token_fut);
+
+    let res = ready_with_cancellation_token_fut
```
ADD-SP:
```suggestion
    let pending_with_cancellation_token_fut = pending_fut.with_cancellation_token(&token);

    pin!(pending_with_cancellation_token_fut);

    let res = pending_with_cancellation_token_fut
```

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+    assert_eq!(res, Poll::Pending);
+
+    token.cancel();
+
+    let res = ready_with_cancellation_token_fut
```
ADD-SP:
```suggestion
    let pending_with_cancellation_token_fut = pending_fut.with_cancellation_token(&token);

    pin!(pending_with_cancellation_token_fut);

    let res = pending_with_cancellation_token_fut
        .as_mut()
        .poll(&mut Context::from_waker(&waker));
    assert_eq!(res, Poll::Pending);

    token.cancel();

    let res = pending_with_cancellation_token_fut
```

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+use tokio_util::{future::FutureExt, sync::CancellationToken};
+
+#[test]
+fn ready_future_with_cancellation_token_test() {
+    let (waker, _) = new_count_waker();
+    let token = CancellationToken::new();
+
+    let ready_fut = ready(());
+
+    let ready_with_cancellation_token_fut = ready_fut.with_cancellation_token(&token);
```
ADD-SP: nit: I PERSONALLY prefer a shorter name like `ready_with_token_fut`, this is just my personal flavor, not a blocker.

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    let ready_with_cancellation_token_fut = pending_fut.with_cancellation_token(&token);
+
+    pin!(ready_with_cancellation_token_fut);
+
+    let res = ready_with_cancellation_token_fut
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+
+    assert_eq!(res, Poll::Ready(None));
```
ADD-SP: [`tokio_test::assert_ready_eq`]

[`tokio_test::assert_ready_eq`]: https://docs.rs/tokio-test/0.4.4/tokio_test/macro.assert_ready_eq.html

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+    assert_eq!(res, Poll::Pending);
+
+    token.cancel();
+
+    let res = ready_with_cancellation_token_fut
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+    assert_eq!(res, Poll::Ready(None));
```
ADD-SP: [`tokio_test::assert_ready_eq`]

[`tokio_test::assert_ready_eq`]: https://docs.rs/tokio-test/0.4.4/tokio_test/macro.assert_ready_eq.html

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+    let pending_fut = pending::<()>();
+
+    let ready_with_cancellation_token_fut = pending_fut.with_cancellation_token(&token);
+
+    pin!(ready_with_cancellation_token_fut);
+
+    let res = ready_with_cancellation_token_fut
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+    assert_eq!(res, Poll::Pending);
```
ADD-SP: [`tokio_test::assert_pending`]

[`tokio_test::assert_pending`]: https://docs.rs/tokio-test/0.4.4/tokio_test/macro.assert_pending.htm

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    let res = ready_with_cancellation_token_fut
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+
+    assert_eq!(res, Poll::Pending);
+}
+
+#[test]
+fn future_with_already_cancelled_token_test() {
```
ADD-SP:
```suggestion
fn pending_fut_with_already_cancelled_token_test() {
```

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+        }
+    }
+    let (waker, wake_count) = new_count_waker();
+    let token = CancellationToken::new();
+    let fut = TestFuture::default().with_cancellation_token(&token);
+    pin!(fut);
+
+    // first poll, TestFuture returned Pending
+    let res = fut.as_mut().poll(&mut Context::from_waker(&waker));
+    assert_eq!(res, Poll::Pending);
```
ADD-SP: [`tokio_test::assert_pending`]

[`tokio_test::assert_pending`]: https://docs.rs/tokio-test/0.4.4/tokio_test/macro.assert_pending.htm

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+    // due to the polling fairness (biased behavior) of WithCancellationToken Future,
+    // subsequent polls are biased toward polling TestFuture, which results in always returning Ready.
+    let res = fut.as_mut().poll(&mut Context::from_waker(&waker));
+    assert_eq!(res, Poll::Ready(Some(())));
+
+    let res = fut.as_mut().poll(&mut Context::from_waker(&waker));
+    assert_eq!(res, Poll::Ready(Some(())));
+
+    let res = fut.as_mut().poll(&mut Context::from_waker(&waker));
+    assert_eq!(res, Poll::Ready(Some(())));
```
ADD-SP: [`tokio_test::assert_ready_eq`]

[`tokio_test::assert_ready_eq`]: https://docs.rs/tokio-test/0.4.4/tokio_test/macro.assert_ready_eq.html

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    let ready_with_cancellation_token_fut = ready_fut.with_cancellation_token_owned(token);
+
+    pin!(ready_with_cancellation_token_fut);
+
+    let res = ready_with_cancellation_token_fut
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+
+    assert_eq!(res, Poll::Ready(Some(())));
```
ADD-SP: [`tokio_test::assert_ready_eq`]

[`tokio_test::assert_ready_eq`]: https://docs.rs/tokio-test/0.4.4/tokio_test/macro.assert_ready_eq.html

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    let ready_with_cancellation_token_fut = pending_fut.with_cancellation_token_owned(token);
+
+    pin!(ready_with_cancellation_token_fut);
+
+    let res = ready_with_cancellation_token_fut
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+
+    assert_eq!(res, Poll::Pending);
```
ADD-SP: [`tokio_test::assert_pending`]

[`tokio_test::assert_pending`]: https://docs.rs/tokio-test/0.4.4/tokio_test/macro.assert_pending.htm

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+    let (waker, _) = new_count_waker();
+    let token = CancellationToken::new();
+
+    let pending_fut = pending::<()>();
+
+    let ready_with_cancellation_token_fut = pending_fut.with_cancellation_token_owned(token);
+
+    pin!(ready_with_cancellation_token_fut);
+
+    let res = ready_with_cancellation_token_fut
```
ADD-SP:
```suggestion
    let pending_with_cancellation_token_fut = pending_fut.with_cancellation_token_owned(token);

    pin!(pending_with_cancellation_token_fut);

    let res = pending_with_cancellation_token_fut
```

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+
+    assert_eq!(res, Poll::Ready(None));
```
ADD-SP: [`tokio_test::assert_ready_eq`]

[`tokio_test::assert_ready_eq`]: https://docs.rs/tokio-test/0.4.4/tokio_test/macro.assert_ready_eq.html

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+    let token = CancellationToken::new();
+    token.cancel();
+
+    let pending_fut = pending::<()>();
+
+    let ready_with_cancellation_token_fut = pending_fut.with_cancellation_token_owned(token);
+
+    pin!(ready_with_cancellation_token_fut);
+
+    let res = ready_with_cancellation_token_fut
```
ADD-SP:
```suggestion
    let pending_with_cancellation_token_fut = pending_fut.with_cancellation_token_owned(token);

    pin!(pending_with_cancellation_token_fut);

    let res = pending_with_cancellation_token_fut
```

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+    assert_eq!(res, Poll::Pending);
+
+    token.cancel();
+
+    let res = ready_with_cancellation_token_fut
```
ADD-SP:
```suggestion
    let pending_with_cancellation_token_fut =
        pending_fut.with_cancellation_token_owned(token.clone());

    pin!(ready_with_cancellation_token_fut);

    let res = pending_with_cancellation_token_fut
        .as_mut()
        .poll(&mut Context::from_waker(&waker));
    assert_eq!(res, Poll::Pending);

    token.cancel();

    let res = pending_with_cancellation_token_fut
```

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+    assert_eq!(res, Poll::Ready(None));
```
ADD-SP: [`tokio_test::assert_ready_eq`]

[`tokio_test::assert_ready_eq`]: https://docs.rs/tokio-test/0.4.4/tokio_test/macro.assert_ready_eq.html

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    let ready_with_cancellation_token_fut =
+        pending_fut.with_cancellation_token_owned(token.clone());
+
+    pin!(ready_with_cancellation_token_fut);
+
+    let res = ready_with_cancellation_token_fut
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+    assert_eq!(res, Poll::Pending);
```
ADD-SP: [`tokio_test::assert_pending`]

[`tokio_test::assert_pending`]: https://docs.rs/tokio-test/0.4.4/tokio_test/macro.assert_pending.htm

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+        }
+    }
+    let (waker, wake_count) = new_count_waker();
+    let token = CancellationToken::new();
+    let fut = TestFuture::default().with_cancellation_token(&token);
+    pin!(fut);
+
+    // first poll, TestFuture returned Pending
+    let res = fut.as_mut().poll(&mut Context::from_waker(&waker));
+    assert_eq!(res, Poll::Pending);
```
ADD-SP: [`tokio_test::assert_pending`]

[`tokio_test::assert_pending`]: https://docs.rs/tokio-test/0.4.4/tokio_test/macro.assert_pending.htm

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+    // due to the polling fairness (biased behavior) of WithCancellationToken Future,
+    // subsequent polls are biased toward polling TestFuture, which results in always returning Ready.
+    let res = fut.as_mut().poll(&mut Context::from_waker(&waker));
+    assert_eq!(res, Poll::Ready(Some(())));
+
+    let res = fut.as_mut().poll(&mut Context::from_waker(&waker));
+    assert_eq!(res, Poll::Ready(Some(())));
+
+    let res = fut.as_mut().poll(&mut Context::from_waker(&waker));
+    assert_eq!(res, Poll::Ready(Some(())));
```
ADD-SP: [`tokio_test::assert_ready_eq`]

[`tokio_test::assert_ready_eq`]: https://docs.rs/tokio-test/0.4.4/tokio_test/macro.assert_ready_eq.html

bitcapybara: Thank you so much for such a meticulous and detailed review of my code!

tokio-util/src/future.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// ```
+    #[cfg(feature = "time")]
+    fn timeout_at(self, deadline: tokio::time::Instant) -> tokio::time::Timeout<Self>
+    where
+        Self: Sized,
+    {
+        tokio::time::timeout_at(deadline, self)
+    }
+
+    /// Similar to [`CancellationToken::run_until_cancelled`], but with the advantage that it is easier to write
```
ADD-SP:
```suggestion
    /// Similar to [`CancellationToken::run_until_cancelled`],
    /// but with the advantage that it is easier to write
```

This line it too long (115 columns).

tokio-util/src/future.rs
```diff
@@ -0,0 +1,136 @@
...
+        self,
+        cancellation_token: &CancellationToken,
+    ) -> WithCancellationTokenFuture<'_, Self>
+    where
+        Self: Sized,
+    {
+        WithCancellationTokenFuture::new(cancellation_token, self)
+    }
+
+    /// Similar to [`CancellationToken::run_until_cancelled_owned`], but with the advantage that it is easier to write
```
ADD-SP:
```suggestion
    /// Similar to [`CancellationToken::run_until_cancelled_owned`],
    /// but with the advantage that it is easier to write
```


This line it too long (115 columns).

tokio-util/src/future.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// use tokio_util::future::FutureExt;
+    ///
+    /// # async fn dox() {
+    /// let (_tx, rx) = oneshot::channel::<()>();
+    ///
+    /// let res = rx.timeout(Duration::from_millis(10)).await;
+    /// assert!(res.is_err());
+    /// # }
+    /// ```
+    #[cfg(feature = "time")]
```
ADD-SP: Please use `cfg_time!` to render the feature gate in the `docs.rs`.



We could refer the below example

https://github.com/tokio-rs/tokio/blob/ce41896f8dcbc6249df3279600f45f7a65915cf6/tokio-util/src/lib.rs#L51-L53

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+    // due to the polling fairness (biased behavior) of WithCancellationToken Future,
+    // subsequent polls are biased toward polling TestFuture, which results in always returning Ready.
+    let res = fut.as_mut().poll(&mut Context::from_waker(&waker));
+    assert_ready_eq!(res, Some(()));
+
+    let res = fut.as_mut().poll(&mut Context::from_waker(&waker));
+    assert_ready_eq!(res, Some(()));
+
+    let res = fut.as_mut().poll(&mut Context::from_waker(&waker));
+    assert_ready_eq!(res, Some(()));
```
ADD-SP: It is not encouraged to poll a completed future.

> Panics
> 
> Once a future has completed (returned Ready from poll), calling its poll method again may panic, block forever, or cause other kinds of problems; the Future trait places no requirements on the effects of such a call. However, as the poll method is not marked unsafe, Rust’s usual rules apply: calls must never cause undefined behavior (memory corruption, incorrect use of unsafe functions, or the like), regardless of the future’s state.
>
> [`std::future::poll`](https://doc.rust-lang.org/std/future/trait.Future.html#panics)

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
```
ADD-SP: It is not encouraged to poll a completed future.

> Panics
> 
> Once a future has completed (returned Ready from poll), calling its poll method again may panic, block forever, or cause other kinds of problems; the Future trait places no requirements on the effects of such a call. However, as the poll method is not marked unsafe, Rust’s usual rules apply: calls must never cause undefined behavior (memory corruption, incorrect use of unsafe functions, or the like), regardless of the future’s state.
>
> [`std::future::poll`](https://doc.rust-lang.org/std/future/trait.Future.html#panics)

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+        type Output = ();
+
+        fn poll(mut self: std::pin::Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Self::Output> {
+            if !self.polled {
+                self.polled = true;
+                return Poll::Pending;
+            }
+            Poll::Ready(())
+        }
+    }
```
ADD-SP: nit: We'd better avoiding the duplicated implementation. We also want a better name like `ReadyOnTheSecondPollFuture`.

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    let res = pending_with_token_fut
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+
+    assert_pending!(res);
+}
+
+#[test]
+fn pending_fut_with_already_cancelled_token_owned_test() {
```
ADD-SP, line 205: We also want a `ready_fut_with_already_cancelled_token_owned_test` to cover the https://github.com/tokio-rs/tokio/pull/7475#discussion_r2229891851.

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    let res = pending_with_token_fut
+        .as_mut()
+        .poll(&mut Context::from_waker(&waker));
+
+    assert_pending!(res);
+}
+
+#[test]
+fn pending_fut_with_already_cancelled_token_test() {
```
ADD-SP, line 84: We also want a `ready_fut_with_already_cancelled_token_test` to cover the https://github.com/tokio-rs/tokio/pull/7475#discussion_r2229891851.

tokio-util/src/future.rs
```diff
@@ -9,57 +9,61 @@ use crate::sync::CancellationToken;
...
+        /// # }
+        /// ```
+        #[cfg(feature = "time")]
+        fn timeout_at(self, deadline: tokio::time::Instant) -> tokio::time::Timeout<Self>
+        where
+            Self: Sized,
+        {
+            tokio::time::timeout_at(deadline, self)
+        }
     }
```
ADD-SP: Actually, we can put both two methods into a single `cfg_time!`.

bitcapybara: Oh, yes, I will fix it

tokio-util/src/time/mod.rs
```diff
@@ -8,45 +8,18 @@
...
-use std::future::Future;
 use std::time::Duration;
-use tokio::time::Timeout;
 
 mod wheel;
 
 pub mod delay_queue;
 
+// re-export `FutureExt` to avoid breaking change
+pub use crate::future::FutureExt;
```
ADD-SP:
```suggestion
// re-export `FutureExt` to avoid breaking change
#[doc(inline)]
pub use crate::future::FutureExt;
```

tokio-util/src/future.rs
```diff
@@ -0,0 +1,136 @@
...
+        fn timeout_at(self, deadline: tokio::time::Instant) -> tokio::time::Timeout<Self>
+        where
+            Self: Sized,
+        {
+            tokio::time::timeout_at(deadline, self)
+        }
+    }
+
+    /// Similar to [`CancellationToken::run_until_cancelled`],
+    /// but with the advantage that it is easier to write
```
ADD-SP:
```suggestion
    /// but with the advantage that it is easier to write.
```

tokio-util/src/future.rs
```diff
@@ -0,0 +1,136 @@
...
+        cancellation_token: &CancellationToken,
+    ) -> WithCancellationTokenFuture<'_, Self>
+    where
+        Self: Sized,
+    {
+        WithCancellationTokenFuture::new(cancellation_token, self)
+    }
+
+    /// Similar to [`CancellationToken::run_until_cancelled_owned`],
+    /// but with the advantage that it is easier to write
```
ADD-SP:
```suggestion
    /// but with the advantage that it is easier to write.
```

tokio-util/src/future.rs
```diff
@@ -0,0 +1,136 @@
...
+        fn timeout_at(self, deadline: tokio::time::Instant) -> tokio::time::Timeout<Self>
+        where
+            Self: Sized,
+        {
+            tokio::time::timeout_at(deadline, self)
+        }
+    }
+
+    /// Similar to [`CancellationToken::run_until_cancelled`],
+    /// but with the advantage that it is easier to write.
```
ADD-SP: Did we remove too many words?

> but with the advantage that it is easier to write fluent call chains

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,266 @@
...
+
+    // first poll, ReadyOnTheSecondPollFuture returned Pending
+    let res = fut.as_mut().poll(&mut Context::from_waker(&waker));
+    assert_pending!(res);
+
+    token.cancel();
+    assert_eq!(wake_count, 1);
+
+    // due to the polling fairness (biased behavior) of WithCancellationToken Future,
+    // subsequent polls are biased toward polling ReadyOnTheSecondPollFuture, which results in always returning Ready.
```
ADD-SP:
```suggestion
    // subsequent polls are biased toward polling ReadyOnTheSecondPollFuture,
    // which results in always returning Ready.
```

tokio-util/tests/future.rs
```diff
@@ -0,0 +1,266 @@
...
```
ADD-SP:
```suggestion
    // subsequent polls are biased toward polling ReadyOnTheSecondPollFuture,
    // which results in always returning Ready.
```

ADD-SP (commented):
We are approaching 100 comments, overall is looks good to me, I think we just need to polish something like docs and tests, thanks you very much for your patience.

By the way, since this PR publish a new module `tokio_util::future`, this is a significant change, 
after approving this PR, I'd like to keep hold on merge for a few days to see if other maintainers concerned about this.

tokio-util/src/future.rs
bitcapybara: Oh, sorry, I didn't notice that, I will fix it

ADD-SP (approved):
Thanks! Let's hold on the merge button for a few days to see if other maintainers concern about publishing a new module `tokio_util::future`.

bitcapybara:
You're welcome, and thank you for spending so much time reviewing my code. 😊

PR #7479, opened 2025-07-23
Implement `skip_until`

WilliamVenner:
## Motivation

[`std::io::BufRead::skip_until`](https://doc.rust-lang.org/std/io/trait.BufRead.html#method.skip_until) was stabilized in Rust 1.83.0. Let's implement it into Tokio, too.

## Solution

I have simply adapted existing `read_until` code to discard the read bytes rather than writing them in to a buffer.


ADD-SP:
This breaks the MSRV, current MSRV is Rust 1.70, this requires 1.83.

PR #7495, opened 2025-07-30
ci: fix clippy warnings triggered under specific cfg

mox692:
Context: https://github.com/tokio-rs/tokio/pull/7321#issuecomment-3132627052.

Added tokio's specific cfgs to the clippy ci settings, and fixed some warnings generated by that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -843,6 +843,8 @@ jobs:
       # Run clippy
       - name: "clippy --all"
         run: cargo clippy --all --tests --all-features --no-deps
+        env:
+          RUSTFLAGS: --cfg tokio_unstable --cfg tokio_taskdump --cfg tokio_uring -Dwarnings
 
   docs:
     name: docs
--- a/tokio/tests/macros_select.rs
+++ b/tokio/tests/macros_select.rs
@@ -661,7 +661,9 @@ mod unstable {
             .build()
             .unwrap();
         let rt1_values = rt1.block_on(async {
-            let _ = tokio::spawn(async { (select_0_to_9().await, select_0_to_9().await) }).await;
+            tokio::spawn(async { (select_0_to_9().await, select_0_to_9().await) })
+                .await
+                .unwrap()
         });
 
         let rt2 = tokio::runtime::Builder::new_multi_thread()
@@ -670,7 +672,9 @@ mod unstable {
             .build()
             .unwrap();
         let rt2_values = rt2.block_on(async {
-            let _ = tokio::spawn(async { (select_0_to_9().await, select_0_to_9().await) }).await;
+            tokio::spawn(async { (select_0_to_9().await, select_0_to_9().await) })
+                .await
+                .unwrap()
         });
 
         assert_eq!(rt1_values, rt2_values);
--- a/tokio-util/src/task/join_map.rs
+++ b/tokio-util/src/task/join_map.rs
@@ -541,9 +541,9 @@ where
     /// assert!(!map.abort("goodbye universe"));
     /// # }
     /// ```
-    pub fn abort<Q: ?Sized>(&mut self, key: &Q) -> bool
+    pub fn abort<Q>(&mut self, key: &Q) -> bool
     where
-        Q: Hash + Eq,
+        Q: ?Sized + Hash + Eq,
         K: Borrow<Q>,
     {
         match self.get_by_key(key) {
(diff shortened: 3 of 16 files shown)


tokio/tests/macros_select.rs
```diff
@@ -661,7 +661,9 @@ mod unstable {
             .build()
             .unwrap();
         let rt1_values = rt1.block_on(async {
-            let _ = tokio::spawn(async { (select_0_to_9().await, select_0_to_9().await) }).await;
+            tokio::spawn(async { (select_0_to_9().await, select_0_to_9().await) })
+                .await
+                .unwrap()
         });
```
mox692, line 674: Here, I'm assuming that the original test code wanted to compare these two tupples (`rt1_values`, `rt2_values`), but actually this was comparing two unit values.

.github/workflows/ci.yml
```diff
@@ -843,6 +843,8 @@ jobs:
       # Run clippy
       - name: "clippy --all"
         run: cargo clippy --all --tests --all-features --no-deps
+        env:
+          RUSTFLAGS: --cfg tokio_unstable --cfg tokio_taskdump --cfg tokio_uring -Dwarnings
```
ADD-SP: I think we also need to cover `--all-features` but no any `cfg`s.

mox692: Fixed: 355f013.

tokio/tests/macros_select.rs
mox692: Hmm this test is flaky? ... I'll dig in a bit

https://github.com/tokio-rs/tokio/actions/runs/16648337632/job/47114406819?pr=7495

mox692: It turns out that just using `Builder::rng_seed()` doesn't garantee a deterministic branch selection of the `select!`.
Added a commit [92a0e8b](https://github.com/tokio-rs/tokio/pull/7495/commits/92a0e8b4818dee1323e69dd7180724453139f704) to fix that.

<details>

<summary>Details</summary>

The function used to decide the `select!` branch is `thread_rng_n()`, and the values generated by this function depend on the seed created by `RngSeedGenerator`. This `RngSeedGenerator` is used globally inside the runtime, and its seed is updated every time `next_seed()` is called.

In this test, `next_seed()` is called 5 times in total:

a. The `block_on` thread calls `build_threaded_runtime`’s [`next_generator()`](https://github.com/tokio-rs/tokio/blob/4e3f17bce3addf0bbf626e4847bab73e28328181/tokio/src/runtime/builder.rs#L1623-L1624) (**twice**)  
b. The `block_on` thread calls [`worker::create()`](https://github.com/tokio-rs/tokio/blob/4e3f17bce3addf0bbf626e4847bab73e28328181/tokio/src/runtime/scheduler/multi_thread/worker.rs#L269)  
c. The `block_on` thread calls [`enter_runtime()`](https://github.com/tokio-rs/tokio/blob/4e3f17bce3addf0bbf626e4847bab73e28328181/tokio/src/runtime/context/runtime.rs#L49)  
d. The `worker` thread calls [`enter_runtime()`](https://github.com/tokio-rs/tokio/blob/4e3f17bce3addf0bbf626e4847bab73e28328181/tokio/src/runtime/context/runtime.rs#L49) (the `select!` macro uses the seed generated here)

I checked the actual execution order of these 5 operations locally, and found that there are two cases:

- order1 (armv5te often follows this order)
    - a → b → **d** → c
- order2 (other platforms often follow this order)
    - a → b → **c** → d

This shows that there is a race condition about which thread calls `enter_runtime()` and following `next_seed()` first. (from the perspective of the worker thread running `select!`, it could use either the 3rd seed or 4th seed)

In the fix commit, we changed the behavior so that the `block_on` thread always calls `enter_runtime()` _after_ the `worker` thread does, in a deterministic way.

</details>

tokio/tests/macros_select.rs
```diff
@@ -655,22 +655,40 @@ mod unstable {
...
+        let (tx, rx) = std::sync::mpsc::channel();
         let rt1 = tokio::runtime::Builder::new_multi_thread()
             .worker_threads(1)
+            .on_thread_park(move || tx.send(()).unwrap())
             .rng_seed(RngSeed::from_bytes(seed))
             .build()
             .unwrap();
+
+        // This makes sure that `enter_runtime()` from worker thread is called before the one from main thread,
+        // ensuring that the RNG state is consistent.
```
ADD-SP: There are two places to consume the `rng_seed` for multi thread runtime.

1. `block_on`
2. Launching worker threads.

Once the `on_thread_park` called, it is very likely that all worker_threads have started running, so this change reduces flakiness.

I leave more context here to help future developers debugging this test, it would be better adding this into the code, but this is not a blocker.

mox692: I added a PR link to the comment to make it easier for developers to follow the context.

[6e41ee5](https://github.com/tokio-rs/tokio/pull/7495/commits/6e41ee5e1f4dd34a0032be50b807c3bd012e908d)

PR #7512, opened 2025-08-04
util/future: fix `FutureExt` bias and fairness docs

lucab:
This fixes the docstrings on `FutureExt` so that the bias and fairness notes are correct and consistent in all cases.
All cancellation-related wrappers are biased towards the completion of the inner future, but they do initially check if the token is already cancelled at construction time.

Followup for https://github.com/tokio-rs/tokio/pull/7475


ADD-SP (commented):
I renamed the title to ' future: fix confusing docs of FutureExt for cancellation adapters'

PR #7530, opened 2025-08-13
Implement `Ord` for `task::Id`

tyilo:
## Motivation

I want to use the id from a `JoinSet` as a key in a `BTreeMap`.

## Solution

Implement `PartialOrd` and `Ord` for `Id`.

### DIFF
--- a/tokio/src/runtime/task/id.rs
+++ b/tokio/src/runtime/task/id.rs
@@ -16,7 +16,7 @@ use std::{fmt, num::NonZeroU64};
 ///   [`task::id()`](crate::task::id()) functions and from outside the task via
 ///   the [`JoinHandle::id()`](crate::task::JoinHandle::id()) function.
 #[cfg_attr(docsrs, doc(cfg(all(feature = "rt"))))]
-#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq)]
+#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
 pub struct Id(pub(crate) NonZeroU64);
 
 /// Returns the [`Id`] of the currently running task.


tyilo:
I don't really care about the ordered property of a `BTreeMap`, but I think it would be faster than a `HashMap`.

For instance both [`std::io::ErrorKind`](https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html) and [`std::any::TypeId`](https://doc.rust-lang.org/stable/std/any/struct.TypeId.html) implements `Ord` even though they have no "natural" ordering.

See also [C-COMMON-TRAITS](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits)

<hr>

The concrete code I wanted it for is this (with `HashMap` replaced by `BTreeMap`):
```rust
use std::collections::HashMap;
use std::hash::Hash;

use tokio::task::JoinSet;

trait InsertOnce<K>
where
    K: ?Sized,
{
    type V: ?Sized;

    fn insert_once(&mut self, key: K, value: Self::V);
}

impl<K, V> InsertOnce<K> for HashMap<K, V>
where
    K: Eq + Hash,
{
    type V = V;

    fn insert_once(&mut self, key: K, value: Self::V) {
        let previous = self.insert(key, value);
        assert!(previous.is_none());
    }
}

impl<V> InsertOnce<usize> for Vec<Option<V>> {
    type V = V;

    fn insert_once(&mut self, key: usize, value: Self::V) {
        let entry = &mut self[key];
        assert!(entry.is_none());
        *entry = Some(value);
    }
}

pub async fn spawn_and_join_all<I, F, T>(futures: I, max_concurrent: usize) -> Vec<T>
where
    I: IntoIterator<Item = F>,
    F: Future<Output = T> + Send + 'static,
    T: Send + 'static,
{
    let mut futures = futures.into_iter().fuse();

    let mut set = JoinSet::new();
    let mut ids_to_index = HashMap::new();

    for f in futures.by_ref().take(max_concurrent) {
        let id = set.spawn(f).id();
        ids_to_index.insert_once(id, ids_to_index.len());
    }

    let mut total_started = set.len();
    let lower_bound = total_started + futures.size_hint().0;
    let mut result: Vec<Option<T>> = Vec::with_capacity(lower_bound);
    result.resize_with(total_started, Default::default);

    while let Some(r) = set.join_next_with_id().await {
        let (id, v) = r.unwrap();
        let index = ids_to_index.remove(&id).unwrap();
        result.insert_once(index, v);

        if let Some(f) = futures.next() {
            let id = set.spawn(f).id();
            ids_to_index.insert_once(id, total_started);
            total_started += 1;
            result.push(None);
        }
    }

    result.into_iter().map(|v| v.unwrap()).collect()
}
```

ADD-SP:
I believe that now we have two topics.

## <del>Task ID may be reused</del>

> ```rust
>     fn insert_once(&mut self, key: K, value: Self::V) {
>         let previous = self.insert(key, value);
>         assert!(previous.is_none());
>     }
>
> // ----- omitted -----
>
>         if let Some(f) = futures.next() {
>             let id = set.spawn(f).id();
>             ids_to_index.insert_once(id, total_started);
>             total_started += 1;
>             result.push(None);
>         }
> ```

<del>There is a potential panic case.</del>

> Task IDs are unique relative to other currently running tasks. When a task completes, the same ID may be used for another task.
>
> [*tokio::task::Id*](https://docs.rs/tokio/1.47.1/tokio/task/struct.Id.html)

## Add `PartialOrd ` and `Ord` for convenience

Thanks for your mentioning [`std::io::ErrorKind`] and [`std::any::TypeId`], I searched their PRs and read the discussion.

* Implement more traits for `std::io::ErrorKind` (https://github.com/rust-lang/rust/pull/35911)
* Add PartialOrd, Ord derivations to TypeId (https://github.com/rust-lang/rust/pull/38981)

I now think it makes sense to add these two features to make things easier (even they don't have natural ordering) for downstream users so I am not opposed to this change at this point.

<del>However, firstly, it would be better taking care of the potential panic I mentioned above, then we can re-evaluate do we actually need a `PartialOrd ` in your case.</del>

<del>Again, I am not opposed to this change after reading the above two PRs, but I still want to keep it as it is until there is a use case that absolutely requires this change.</del>

[`tokio::task::Id`]: https://docs.rs/tokio/1.47.1/tokio/task/struct.Id.html
[`std::io::ErrorKind`]: https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html
[`std::any::TypeId`]: https://doc.rust-lang.org/stable/std/any/struct.TypeId.html

tyilo:
> There is a potential panic case.

Not really, because of the call to `ids_to_index.remove(&id)`.

ADD-SP (commented):
Looks good to me, I also learned some interesting stories today.

Since merging this PR will make `PartialOrd` and `Ord` be the part of semantic version guarantee, I would like to keep this PR open for a few days so that the community and other maintainers can participate in our discussion during this time window.

carllerche:
Given the TypeId precedence, I am fine adding Ord / PartialOrd implementation. I think it is fairly safe.

Darksonn (approved):
Thanks.

Issue #7519, opened 2025-08-08
Use latest rust for wasm tests


mox692:
Currently we're seeing wasm test failure since rust 1.89.0 release, so we should fix this.
We should revert https://github.com/tokio-rs/tokio/pull/7518 once we can confirm that our wasm tests work fine on the latest Rust version.

taiki-e:
We probably need to add `ignore-wasm` annotations to the failing doctests.  (Example: https://github.com/tokio-rs/bytes/pull/790. See https://github.com/drager/wasm-pack/issues/1518#issuecomment-3172337213 for more.)

mox692:
Looks like we need to add a bunch of annotations ;)

```
$ wasm-pack test --node -- --features "macros sync"
...

test result: FAILED. 93 passed; 283 failed; 2 ignored; 0 measured; 0 filtered out; finished in 1.50s
```

ADD-SP:
Resolved by <https://github.com/tokio-rs/tokio/pull/7537>.

PR #7537, opened 2025-08-16
time: add `ignore-wasm` to failing wasm tests

lblack00:
Addresses failing time tests: #7519 

## Motivation

Currently, the MSRV is pinned temporarily and there is a large amount of failures for wasm doctests with the Rust 1.89.0 release.

## Solution

Add `ignore-wasm` to failing doctests in `tokio/src/time` until these tests pass with the new release.

### DIFF
--- a/tokio/src/time/instant.rs
+++ b/tokio/src/time/instant.rs
@@ -78,7 +78,7 @@ impl Instant {
     ///
     /// # Examples
     ///
-    /// ```
+    /// ```ignore-wasm
     /// use tokio::time::{Duration, Instant, sleep};
     ///
     /// #[tokio::main]
@@ -99,7 +99,7 @@ impl Instant {
     ///
     /// # Examples
     ///
-    /// ```
+    /// ```ignore-wasm
     /// use tokio::time::{Duration, Instant, sleep};
     ///
     /// #[tokio::main]
@@ -120,7 +120,7 @@ impl Instant {
     ///
     /// # Examples
     ///
-    /// ```
+    /// ```ignore-wasm
     /// use tokio::time::{Duration, Instant, sleep};
     ///
     /// #[tokio::main]
--- a/tokio/src/time/interval.rs
+++ b/tokio/src/time/interval.rs
@@ -23,7 +23,7 @@ use std::task::{ready, Context, Poll};
 ///
 /// # Examples
 ///
-/// ```
+/// ```ignore-wasm
 /// use tokio::time::{self, Duration};
 ///
 /// #[tokio::main]
@@ -49,7 +49,7 @@ use std::task::{ready, Context, Poll};
 /// would only be executed once every three seconds, and not every two
 /// seconds.
 ///
-/// ```
+/// ```ignore-wasm
 /// use tokio::time;
 ///
 /// async fn task_that_takes_a_second() {
@@ -89,7 +89,7 @@ pub fn interval(period: Duration) -> Interval {
 ///
 /// # Examples
 ///
-/// ```
+/// ```ignore-wasm
 /// use tokio::time::{interval_at, Duration, Instant};
 ///
 /// #[tokio::main]
@@ -151,7 +151,7 @@ fn internal_interval_at(
 /// Sometimes, an [`Interval`]'s tick is missed. For example, consider the
 /// following:
 ///
-/// ```
+/// ```ignore-wasm
 /// use tokio::time::{self, Duration};
 /// # async fn task_that_takes_one_to_three_millis() {}
 ///
@@ -413,7 +413,7 @@ impl Interval {
     ///
     /// # Examples
     ///
-    /// ```
+    /// ```ignore-wasm
     /// use tokio::time;
     ///
     /// use std::time::Duration;
@@ -501,7 +501,7 @@ impl Interval {
     ///
     /// # Examples
     ///
-    /// ```
+    /// ```ignore-wasm
     /// use tokio::time;
     ///
     /// use std::time::Duration;
@@ -533,7 +533,7 @@ impl Interval {
     ///
     /// # Examples
     ///
-    /// ```
+    /// ```ignore-wasm
     /// use tokio::time;
     ///
     /// use std::time::Duration;
@@ -565,7 +565,7 @@ impl Interval {
     ///
     /// # Examples
     ///
-    /// ```
+    /// ```ignore-wasm
     /// use tokio::time;
     ///
     /// use std::time::Duration;
@@ -601,7 +601,7 @@ impl Interval {
     ///
     /// # Examples
     ///
-    /// ```
+    /// ```ignore-wasm
     /// use tokio::time::{self, Instant};
     ///
     /// use std::time::Duration;
--- a/tokio/src/time/mod.rs
+++ b/tokio/src/time/mod.rs
@@ -23,7 +23,7 @@
 //!
 //! Wait 100ms and print "100 ms have elapsed"
 //!
-//! ```
+//! ```ignore-wasm
 //! use std::time::Duration;
 //! use tokio::time::sleep;
 //!
@@ -63,7 +63,7 @@
 //! would only be executed once every three seconds, and not every two
 //! seconds.
 //!
-//! ```
+//! ```ignore-wasm
 //! use tokio::time;
 //!
 //! async fn task_that_takes_a_second() {
--- a/tokio/src/time/sleep.rs
+++ b/tokio/src/time/sleep.rs
@@ -25,7 +25,7 @@ use std::task::{self, ready, Poll};
 ///
 /// Wait 100ms and print "100 ms have elapsed".
 ///
-/// ```
+/// ```ignore-wasm
 /// use tokio::time::{sleep_until, Instant, Duration};
 ///
 /// #[tokio::main]
@@ -85,7 +85,7 @@ pub fn sleep_until(deadline: Instant) -> Sleep {
 ///
 /// Wait 100ms and print "100 ms have elapsed".
 ///
-/// ```
+/// ```ignore-wasm
 /// use tokio::time::{sleep, Duration};
 ///
 /// #[tokio::main]
@@ -140,7 +140,7 @@ pin_project! {
     ///
     /// Wait 100ms and print "100 ms have elapsed".
     ///
-    /// ```
+    /// ```ignore-wasm
     /// use tokio::time::{sleep, Duration};
     ///
     /// #[tokio::main]
@@ -152,7 +152,7 @@ pin_project! {
     ///
     /// Use with [`select!`]. Pinning the `Sleep` with [`tokio::pin!`] is
     /// necessary when the same `Sleep` is selected on multiple times.
-    /// ```no_run
+    /// ```no_run,ignore-wasm
     /// use tokio::time::{self, Duration, Instant};
     ///
     /// #[tokio::main]


ADD-SP:
Please also revert #7518 .

lblack00:
These annotations are only for a small section of the failing wasm tests, which yields:
`test result: FAILED. 93 passed; 265 failed; 20 ignored; 0 measured; 0 filtered out; finished in 3.91s`

I plan on opening PRs for the remaining sections to it split up based on module, unless it is preferred to aggregate all annotations here

mox692:
@lblack00

Thank you for the PR!

> I plan on opening PRs for the remaining sections to it split up based on module, unless it is preferred to aggregate all annotations here


I think we would want to have 1 PR (1 commit), instead of having multiple commits of this. Having some large file changes for this is fine with me.

Also, as @ADD-SP mention, we can revert #7518 in this PR.

lblack00:
Test results yield:
`test result: ok. 93 passed; 0 failed; 285 ignored; 0 measured; 0 filtered out; finished in 1.59s`

mox692:
Now the doc test is hitting `compile_error!` here:
https://github.com/tokio-rs/tokio/blob/925c614c89d0a26777a334612e2ed6ad0e7935c3/tokio/src/lib.rs#L463-L475

(it looks like `cargo test --lib` wouldn't hit this cfg, but `cargo test --doc` does.)

You could update the cfgs in the ci [settings](https://github.com/tokio-rs/tokio/blob/925c614c89d0a26777a334612e2ed6ad0e7935c3/.github/workflows/ci.yml#L1095) like this:
```diff
- cargo test -p tokio --target ${{ matrix.target }} --features full
+ cargo test -p tokio --target ${{ matrix.target }} --features "sync,macros,io-util,rt,time"
```

.github/workflows/ci.yml
```diff
@@ -1092,7 +1092,7 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
       - name: WASI test tokio full
-        run: cargo test -p tokio --target ${{ matrix.target }} --features full
+        run: cargo test -p tokio --target ${{ matrix.target }} --features "sync,macros,io-util,rt,time"
```
ADD-SP, line 1028: Why didn't 1.88 trigger this line? It looks unrelated to the docs.

https://github.com/tokio-rs/tokio/blob/925c614c89d0a26777a334612e2ed6ad0e7935c3/tokio/src/lib.rs#L463-L475

lblack00: My best guess is that it was [this change added in 1.89](https://github.com/rust-lang/cargo/pull/15462/), which would previously skip doctests silently when `--target` flag is set for a target that is not the host.

Running these tests locally on 1.88 with `--verbose` reveals this note:
```
note: skipping doctests for tokio v1.47.1 (/.../tokio/tokio) (lib), cross-compilation doctests are not yet supported
See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#doctest-xcompile for more information.
```

So I suppose that the doctests were not running at all on 1.88, and with this change on 1.89, it got snagged when doctests were unconditionally enabled during cross-compilation

ADD-SP: > So I suppose that the doctests were not running at all on 1.88, and with this change on 1.89, it got snagged when doctests were unconditionally enabled during cross-compilation

I don't think this code is written just for docstest? Or did I miss something? Even the docstest was disabled in 1.88, the `cargo test --lib` was still able to run.

https://github.com/tokio-rs/tokio/blob/925c614c89d0a26777a334612e2ed6ad0e7935c3/tokio/src/lib.rs#L463-L475

mox692: Yes, it looks like `cargo test --lib --target wasm32-wasip1` didn't trigger `compiler_error!` even before this PR.

```bash
# lib test doesn't trigger `compiler_error!` in both 1.88, 1.89.
RUSTFLAGS="--cfg tokio_unstable -Dwarnings -C link-args=--max-memory=67108864" \
  cargo test --lib -p tokio --target wasm32-wasip1 --features full
  
# doc test triggers `compiler_error!` in 1.89, and doesn't in 1.88.
RUSTFLAGS="--cfg tokio_unstable -Dwarnings -C link-args=--max-memory=67108864" \
  cargo test --doc -p tokio --target wasm32-wasip1 --features full
```


I haven't looked into details, but there might be a cargo issue related to this.

mox692 (approved):
Looks good to me.

.github/workflows/ci.yml
lblack00: > I don't think this code is written just for docstest?

On second thought, you're 100% correct. To be honest I don't know, and this particular error confused me between the version changes.

ADD-SP (commented):
See <https://github.com/tokio-rs/tokio/pull/7537#discussion_r2297104898>, It might be a cargo issue, we'd better having a minimal reproducible example, otherwise, it is hard for me to say I know what I am doing.

ADD-SP:
I made an minimal reproducible example, it seems unrelated to neither Rust version nor wasm. If I understand correctly, since there is a `not(my_unstable)`, the `compile_error!` should never fire.

@mox692 Do you have any idea? It look like a Cargo issue.

```toml
[package]
name = "feature-flag"
version = "0.1.0"
edition = "2024"

[dependencies]

[features]
epoll = []

[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = ['cfg(my_unstable)'] }
```

```rust
// src/lib.rs
#[cfg(all(
    not(my_unstable),
    any(
        feature = "epoll",
    )
))]
compile_error!("blablabla");
```

```bash
# no error
RUSTFLAGS="--cfg my_unstable" cargo +1.88 test --lib --features epoll
RUSTFLAGS="--cfg my_unstable" cargo +1.89 test --lib --features epoll

# compilation error
RUSTFLAGS="--cfg my_unstable" cargo +1.88 test --doc --features epoll
RUSTFLAGS="--cfg my_unstable" cargo +1.89 test --doc --features epoll
```

.github/workflows/ci.yml
ADD-SP: See <https://github.com/tokio-rs/tokio/pull/7537#issuecomment-3228612557> for a minimal reproducible example.

mox692:
Not sure. It looks like cfg isn't passed to cargo properly. It's probably worth filing an issue on cargo side?

lblack00:
@ADD-SP This does not trigger `compile_fail!` on minimal example:
```bash
# no error on --doc
RUSTFLAGS="--cfg my_unstable" RUSTDOCFLAGS="--cfg my_unstable" cargo +1.88 test --doc --features epoll
RUSTFLAGS="--cfg my_unstable" RUSTDOCFLAGS="--cfg my_unstable" cargo +1.89 test --doc --features epoll

# no error on --lib
RUSTFLAGS="--cfg my_unstable" cargo +1.88 test --lib --features epoll
RUSTFLAGS="--cfg my_unstable" cargo +1.89 test --lib --features epoll
```

If we run `cargo +1.88 test --doc --features epoll` with *only* `RUSTFLAGS` or `RUSTDOCFLAGS`, then `compile_fail!` is triggered:
```bash
RUSTDOCFLAGS="--cfg my_unstable" cargo +1.88 test --doc --features epoll

   Compiling foo v0.1.0 (/.../foo)
error: blablabla
 --> src/lib.rs:8:1
  |
8 | compile_error!("blablabla");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `foo` (lib) due to 1 previous error
```

```bash
RUSTFLAGS="--cfg my_unstable" cargo +1.88 test --doc --features epoll

    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.19s
   Doc-tests foo
error: blablabla
 --> src/lib.rs:8:1
  |
8 | compile_error!("blablabla");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to 1 previous error

error: doctest failed, to rerun pass `--doc`
```

I believe this boils down to how running `cargo test --doc` compiles a crate using `rustc` *and* `rustdoc`, where `RUSTFLAGS` are passed to `rustc` and `RUSTDOCFLAGS` are passed to `rustdoc`. Even if there are no doctests, running the minimal example without `RUSTDOCFLAGS` will fail as `rustdoc` compiles in a separate context than `rustc` and `--cfg=my_stable` isn't passed to `rustdoc`.

.github/workflows/ci.yml
ADD-SP: Follow up discussion of <https://github.com/tokio-rs/tokio/pull/7537#issuecomment-3229432077>.

Does this mean we need to pass the same `cfg`s to `RUSTDOCFLAGS` while running the `cargo test` and `cargo nextest`?

I think the good news is that we now understand why `cargo test --lib` and `cargo test --doc` yield different results.

lblack00: If I understand it correctly, we only need to pass `--cfg tokio_unstable` or any other `cfg` to `RUSTDOCFLAGS` and the codegen flags will be propagated to `rustdoc` from the `RUSTFLAGS`.

If we want to revert the WASI test to use `full` then we can do something like this:
```bash
CARGO_TARGET_WASM32_WASIP1_RUNNER="wasmtime run --" \
RUSTFLAGS="--cfg tokio_unstable -Dwarnings -C target-feature=+atomics,+bulk-memory -C link-args=--max-memory=67108864" \
RUSTDOCFLAGS="--cfg tokio_unstable" \
cargo test -p tokio --target wasm32-wasip1 --features full
```

Which shows more annotations need to be added.

I'm not sure about `cargo nextest`, but it looks like something similar is already being done
https://github.com/tokio-rs/tokio/blob/925c614c89d0a26777a334612e2ed6ad0e7935c3/.github/workflows/ci.yml#L306-L317

mox692: Can you add the `RUSTDOCFLAGS="--cfg tokio_unstable"` and fix some tests? Then I think we're good.

tokio-stream/src/empty.rs
```diff
@@ -23,7 +23,7 @@ unsafe impl<T> Sync for Empty<T> {}
 ///
 /// Basic usage:
 ///
-/// ```
+/// ```ignore-wasm
```
ADD-SP: I would be better keeping `Stream` tests running on wasm, we could try this approach to enable such tests that can be executed on wasm.

```rust
/// # use futures::executor::block_on;
/// use tokio_stream::{self as stream, StreamExt};
///
/// # block_on(async {
/// let mut none = stream::empty::<i32>();
///
/// assert_eq!(None, none.next().await);
/// # })
```

lblack00: I went with this approach except for a couple of tests where I changed the tokio flavor to `current_thread`, which seemed to work. I'm just not sure if this is the most idiomatic way of going about this.

https://github.com/tokio-rs/tokio/blob/5bf9656351c458a246849f73b75f538d3b62f7ab/tokio-stream/src/stream_map.rs#L68-L71

I also could not get the wrapper which implements `Stream` for `TcpListener` to work with either approach (`tokio-stream/src/wrappers/tcp_listener.rs`) as the `TcpListener::bind` function is not cfg'd for WASI so that was the only test I left unchanged with respect to the annotations.

ADD-SP: > I went with this approach except for a couple of tests where I changed the tokio flavor to current_thread, which seemed to work. I'm just not sure if this is the most idiomatic way of going about this.

I think `#[tokio::main(current_thread)]` is better, as it is simpler. Would you mind trying this?

lblack00: I tried replacing `#[tokio::main(flavor = "current_thread")` with `#[tokio::main(current_thread)` and received this error on `Stream` tests:

```
---- tokio-stream/src/stream_ext.rs - stream_ext::StreamExt::fuse (line 490) stdout ----
error: Set the runtime flavor with #[tokio::main(flavor = "current_thread")].
  --> tokio-stream/src/stream_ext.rs:520:15
   |
32 | #[tokio::main(current_thread)]
   |               ^^^^^^^^^^^^^^
```

Is this syntax supported?

ADD-SP: Ah, sorry, could you try `#[tokio::main(flavor = "current_thread")` to replace the `futures::executor::block_on`?

lblack00: Done :-)

tokio-util/src/codec/length_delimited.rs
```diff
@@ -78,7 +78,7 @@
 //! The following will parse a `u16` length field at offset 0, omitting the
 //! frame head in the yielded `BytesMut`.
 //!
-//! ```
+//! ```ignore-wasm
```
ADD-SP: Could we enable the tests like this that can be executed on wasm?

tokio-util/src/codec/mod.rs
```diff
@@ -14,7 +14,7 @@
 //! [`FramedWrite`] are first framed according to a specific codec, and then sent to
 //! an implementor of [`AsyncWrite`].
 //!
-//! ```
+//! ```ignore-wasm
```
ADD-SP: Could we also enable these kind of tests?

tokio-util/src/either.rs
```diff
@@ -40,7 +40,7 @@ use tokio::io::{AsyncBufRead, AsyncRead, AsyncSeek, AsyncWrite, ReadBuf, Result}
 /// When the output type is the same, we can wrap each future in `Either` to avoid the
 /// issue:
 ///
-/// ```
+/// ```ignore-wasm
```
ADD-SP: Could we enable this test?

tokio-util/src/io/read_arc.rs
```diff
@@ -9,7 +9,7 @@ use tokio::io::{AsyncRead, AsyncReadExt};
 ///
 /// # Example
 ///
-/// ```
+/// ```ignore-wasm
```
ADD-SP: Could we enable this test?

tokio-util/src/io/read_buf.rs
```diff
@@ -11,7 +11,7 @@ use tokio::io::AsyncRead;
 ///
 /// # Example
 ///
-/// ```
+/// ```ignore-wasm
```
ADD-SP: Could we enable this test?

tokio-util/src/io/reader_stream.rs
```diff
@@ -15,7 +15,7 @@ pin_project! {
     ///
     /// # Example
     ///
-    /// ```
+    /// ```ignore-wasm
```
ADD-SP: Could we enable this test?

tokio-util/src/io/sync_bridge.rs
```diff
@@ -77,7 +77,7 @@ use tokio::io::{
 /// incrementally. This avoids blocking and improves performance over using
 /// `SyncIoBridge`.
 ///
-/// ```rust
+/// ```rust,ignore-wasm
```
ADD-SP: Could we enable this test?

lblack00: I initially couldn't get this test to be enabled until I lowered the memory allocated in the test from 64KiB to 16KiB, specifically for the threads runner.

I tried using `futures::executor::block_on` and adjusting the max memory limit of the test in `RUSTFLAGS` but neither seemed to work. Neither did simply adjusting the flavor to `current_thread`.

Not sure what the deeper problem is there. Except maybe it's just due to the single-threaded execution?

lblack00:
I removed `async fn main` from rendered docs except for two instances that I thought came down to either sacrificing some syntactic confusion or readability for docs. Let me know if these aren't appropriate.
https://github.com/tokio-rs/tokio/blob/58f199e42c520aa15b5a53838b4fc3093b4319ba/tokio-stream/src/stream_ext.rs#L515-L518
https://github.com/tokio-rs/tokio/blob/58f199e42c520aa15b5a53838b4fc3093b4319ba/tokio-util/src/io/sync_bridge.rs#L107-L110

tokio-util/src/io/sync_bridge.rs
```diff
@@ -104,7 +104,10 @@ use tokio::io::{
 ///    Ok(hash)
 ///}
 ///
+/// # /*
 /// #[tokio::main]
+/// # */
```
ADD-SP: left-over code?

tokio-util/src/task/join_map.rs
ADD-SP, line 1: JoinMap tests should not be disabled in wasm.

tokio-util/src/task/task_tracker.rs
ADD-SP, line 1: Could we enable doc tests?

tokio-util/src/util/poll_buf.rs
```diff
@@ -11,7 +11,7 @@ use std::task::{ready, Context, Poll};
 ///
 /// # Example
 ///
-/// ```
+/// ```ignore-wasm
```
ADD-SP: Could we enable this test?

tokio/src/io/util/empty.rs
ADD-SP, line 1: Could we enable doc tests?

tokio/src/io/util/repeat.rs
ADD-SP, line 1: Could we enable doc tests?

tokio/src/io/util/sink.rs
ADD-SP, line 1: Could we enable doc tests?

tokio/src/macros/join.rs
ADD-SP, line 1: Could we enable doc tests?

tokio/src/macros/pin.rs
ADD-SP, line 1: Could we enable doc tests?

tokio/src/macros/select.rs
ADD-SP, line 1: Could we enable doc tests?

tokio/src/macros/try_join.rs
ADD-SP, line 1: Could we enable doc tests?

tokio-util/src/io/sync_bridge.rs
lblack00: ~~No, I left a~~ [#7537 (comment)](https://github.com/tokio-rs/tokio/pull/7537#issuecomment-3259292086). ~~If you look at this particular test, I don't think it's as clear cut as rendering out `async fn main` as you pointed out here~~ [#7537 (comment)](https://github.com/tokio-rs/tokio/pull/7537#pullrequestreview-3185382477).

~~I'm open to reverting this, I've just found there's similar approaches in existing doctests even before this PR such as this written by Darksonn which I believe I mistakenly changed~~
https://github.com/tokio-rs/tokio/blob/master/tokio-stream/src/stream_ext.rs#L340-L344

I went with your approach, I think my point was negligible all things considered. Also too time consuming versus a relatively simple `sed` script.

lblack00:
There's **108** tests remaining that still have `ignore-wasm` annotations, which are primarily due to features unavailable on WASM targets. Such as: `fs`, `net`, `rt-multi-threaded`, etc...

Although, a common exception to this are these tests, for example, with blocking mechanisms (marked by `*`):
```
* tokio/src/runtime/builder.rs - runtime::builder::UnhandledPanic::Ignore (line 154)
tokio/src/runtime/task/error.rs - runtime::task::error::JoinError::is_panic (line 48)
tokio/src/runtime/task/join.rs - runtime::task::join::JoinHandle (line 108)
* tokio/src/sync/broadcast.rs - sync::broadcast::Receiver<T>::blocking_recv (line 1524)
* tokio/src/sync/mpsc/unbounded.rs - sync::mpsc::unbounded::UnboundedReceiver<T>::blocking_recv (line 299)
* tokio/src/sync/oneshot.rs - sync::oneshot::Receiver<T>::blocking_recv (line 1191)
* tokio/src/sync/rwlock.rs - sync::rwlock::RwLock<T>::blocking_read (line 491)
* tokio/src/sync/rwlock.rs - sync::rwlock::RwLock<T>::blocking_write (line 833)
* tokio/src/task/join_set.rs - task::join_set::JoinSet<T>::spawn_blocking (line 211)
tokio/src/task/mod.rs - task (line 93)
```

tokio/src/runtime/runtime.rs
```diff
@@ -183,7 +183,7 @@ impl Runtime {
     ///
     /// # Examples
     ///
-    /// ```
+    /// ```ignore-wasm
```
ADD-SP: We can partially enable this kind of test by cfg(target_os), see https://github.com/tokio-rs/tokio/blob/5f3f5b0be4ec933aed2290ea9d929b5daadae321/tokio/src/net/udp.rs#L2137

tokio/src/sync/broadcast.rs
```diff
@@ -1521,7 +1521,7 @@ impl<T: Clone> Receiver<T> {
     /// context.
     ///
     /// # Examples
-    /// ```
+    /// ```ignore-wasm
```
ADD-SP: These kind of test can also be enabled partially, see https://github.com/tokio-rs/tokio/blob/5f3f5b0be4ec933aed2290ea9d929b5daadae321/tokio/src/net/udp.rs#L2137

tokio/src/runtime/task/join.rs
```diff
@@ -91,21 +91,21 @@ cfg_rt! {
...
+    /// let result = join_handle.await??;
+    /// assert_eq!(result, 8);
+    /// Ok(())
+    /// # }
     /// ```
     ///
     /// If the task panics, the error is a [`JoinError`] that contains the panic:
     ///
-    /// ```
+    /// ```ignore-wasm
```
ADD-SP: Not sure if this doctest can also be enabled.

tokio/src/runtime/task/error.rs
```diff
@@ -45,7 +45,7 @@ impl JoinError {
     ///
     /// # Examples
     ///
-    /// ```
+    /// ```ignore-wasm
```
ADD-SP: Emm, I'm not sure why can't we enable this test. Is panic unavailable in WASM?

lblack00: https://github.com/rust-lang/rust/blob/a9d0a6f15533a364816c4d81e2192009ef601d33/compiler/rustc_target/src/spec/base/wasm.rs#L73-L78

Not in a traditional sense apparently. It appears there is no official documentation regarding this behavior, but there are some comments in `rustc` about how the default behavior for panics in WASM is currently `panic=abort` for all WASM targets instead of unwinding.

tokio/src/runtime/task/join.rs
lblack00: Unfortunately looks similar to [#7537 (comment)](https://github.com/tokio-rs/tokio/pull/7537#discussion_r2353829242)

tokio/src/runtime/runtime.rs
lblack00: @ADD-SP I'm curious what you would prefer here. I tried doing something like this:
```
    /// ...
    ///
    /// fn main() {
    ///     # if cfg!(target_family = "wasm") { return; }
    ///     let rt = Runtime::new().unwrap();
    ///
    ///     ...
    /// }
```

Which fails as there are macros that disable `Runtime::new()` on WASM targets. While `cfg!` is evaluated at compilation, the compiler still parses `Runtime::new()` (which it throws an error for).

Would something like this, which conditionally compiles the whole thing, be fine?

```
    /// # #[cfg(not(target_family = "wasm"))]
    /// # {
    /// use tokio::runtime::Runtime;
    ///
    /// ...
    ///
    /// fn main() {
    ///     ...
    /// }
    /// # }
```

ADD-SP: I'm ok for conditional complilation as long as it can get rid of the exclamation marks and also doesn't make the docstring too complicated.

If the docstring is getting crazy, please let me know, we can accept some exclamation marks.

lblack00: Two things I noticed:

1) Cannot partially enable tests this way when it has a `should_panic` annotation, as it doesn't panic

2) There are couple instances where the docstring is getting a little confusing from a glance, e.g.
```
//! # #[cfg(not(target_family = "wasm"))]
//! # {
//! ...
//!
//! # async fn docs() {
//! ...
//! # }
//! # }
```
And perhaps one with the most potential confusion,
```
    /// ```
    /// # #[cfg(not(target_family = "wasm"))]
    /// # {
    /// ...
    /// # fn dox() {
    /// ...
    ///
    /// # let handle = ...
    /// ...
    /// # });
    /// # }
    /// # }
    /// ```
```

tokio/src/sync/oneshot.rs
```diff
@@ -1188,7 +1188,7 @@ impl<T> Receiver<T> {
     ///
     /// # Examples
     ///
-    /// ```
+    /// ```ignore-wasm
```
lblack00: Single annotation I missed in partially enabling, would prefer to wait for feedback before applying changes

tokio/src/runtime/handle.rs
```diff
@@ -134,6 +139,7 @@ impl Handle {
     /// # handle.join().unwrap();
     /// # });
```
lblack00, line 140: This is the specific docstring where we may want an exclamation mark for overall clarity

tokio/src/runtime/handle.rs
```diff
@@ -134,6 +139,7 @@ impl Handle {
...
     /// # }
+    /// # }
```
lblack00, line 142: This is the specific docstring where we may want an exclamation mark for overall clarity

ADD-SP: > This is the specific docstring where we may want an exclamation mark for overall clarity

Emm, sorry, I don't understand, could you explain more?

tokio/src/sync/oneshot.rs
ADD-SP: > Single annotation I missed in partially enabling, would prefer to wait for feedback before applying changes

I think the current approach (`# #[tokio::main(flavor = "current_thread")]`) is fine.

tokio/src/runtime/runtime.rs
ADD-SP: Ah, this DOES look a bit messy. Personally, I think it's acceptable as is, but it shouldn't get any more complicated—otherwise it'll become hard to understand.

If fixing an exclamation mark requires a much more complicated approach than this, then we should just accept the exclamation mark.

lblack00: Understood, and yeah I agree, it is a tad on the messy side.

As this is the only docstring where it gets as complex as it does with nonrenderable code, and you find it acceptable, I'll leave it as is.

tokio/src/sync/oneshot.rs
lblack00: Unfortunately, this is one of those tests mentioned here [#7537 (comment)](https://github.com/tokio-rs/tokio/pull/7537#issuecomment-3277339246) where we attempted
`# #[tokio::main(flavor = "current_thread")]`.

I partially enabled it using conditional compilation, as long as that is okay.

tokio/src/runtime/handle.rs
lblack00: For other readers, see [#7537 comment](https://github.com/tokio-rs/tokio/pull/7537#discussion_r2365861155)

Issue #7281, opened 2025-04-22
`tokio::sync::watch::Receiver`  does not adhere to the specification


beroal:
**Version**
```
└── tokio v1.44.2
    └── tokio-macros v2.5.0 (proc-macro)
```

**Platform**
`Linux beroal 6.14.3-arch1-1 #1 SMP PREEMPT_DYNAMIC Sun, 20 Apr 2025 12:38:52 +0000 x86_64 GNU/Linux`

**Description**

Hi. When exploring whether `Receiver` can detect that the corresponding `Sender`s are dropped, I found the following behavior demonstrated by the program below.

```rust
use std::error::Error;
use tokio::sync::watch;

fn new_watch<T: Clone>(initial: T)
-> Result<(watch::Sender<T>, watch::Receiver<T>), watch::error::SendError<T>> {
    let (tx, rx) = watch::channel(initial.clone());
    let () = tx.send(initial)?;
    Ok((tx, rx))
}

async fn has_changed() -> Result<(), Box<dyn Error>> {
    let (_tx, rx) = new_watch(()).map_err(Box::new)?;
    rx.has_changed().map_err(Box::new)?;
    Ok(())
}

async fn drop_sender_then_has_changed() -> Result<(), Box<dyn Error>> {
    let rx = new_watch(()).map_err(Box::new)?.1;
    rx.has_changed().map_err(Box::new)?;
    Ok(())
}

async fn drop_sender_then_changed() -> Result<(), Box<dyn Error>> {
    let mut rx = new_watch(()).map_err(Box::new)?.1;
    rx.changed().await.map_err(Box::new)?;
    Ok(())
}

#[tokio::main]
async fn main() {
    eprintln!("has_changed: {:?}", has_changed().await);
    eprintln!(
        "drop_sender_then_has_changed: {:?}",
        drop_sender_then_has_changed().await,
    );
    eprintln!(
        "drop_sender_then_changed: {:?}",
        drop_sender_then_changed().await,
    );
}
```

`drop_sender_then_has_changed` returns an error. The specification of [`watch::Receiver::has_changed`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.has_changed) says,

> Returns an error if the channel has been closed.

However, not all receivers have been dropped. It's likely that the fact that all senders have been dropped is responsible for the error because my `has_changed` does not return an error.

[...]

PR #7556, opened 2025-08-28
fix: return `RecvError` if channel is closed when calling `watch::Receiver::changed`

DSharifi:
## Motivation

Solves #7281 . The specification of [watch::Receiver::changed](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.changed) says,
>"This method returns an error if and only if the Sender is dropped."

A regression test has been added, `changed_errors_on_closed_channel`, to verify that the proposed changes fixes the implementation of `changed` to conform to he specification.

## Solution

The PR fixes the bug in `changed` by adding an early check in `changed_impl` on whether the channel is closed before checking if the value is changed.

### DIFF
--- a/tokio/src/sync/watch.rs
+++ b/tokio/src/sync/watch.rs
@@ -825,8 +825,13 @@ impl<T> Receiver<T> {
         &mut self,
         mut f: impl FnMut(&T) -> bool,
     ) -> Result<Ref<'_, T>, error::RecvError> {
-        let mut closed = false;
+        let state = self.shared.state.load();
+        let mut closed = state.is_closed();
         loop {
+            if closed {
+                return Err(error::RecvError(()));
+            }
+
             {
                 let inner = self.shared.value.read();
 
@@ -852,10 +857,6 @@ impl<T> Receiver<T> {
                         }
                     };
                 }
-            }
-
-            if closed {
-                return Err(error::RecvError(()));
             }
 
             // Wait for the value to change.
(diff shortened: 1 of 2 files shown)


DSharifi:
Thanks @ADD-SP  for posting your thoughts!

Indeed, we should definitely align the behavior of both methods `changed` and `has_changed`. I don't have a strong opinion in what direction we go with to achieve that behavior, but here are my thoughts:

## 1. Don't return `RecvError` until the latest value is marked as seen
The use cases I've usually encountered where a watcher is used, the actor that receives the `receiver` handle is only alive as long as the channel and producer half is alive, meaning it's ideal to receive an `Err(RecvError)` IFF the channel is closed. For that reason I'd prefer this option.

Given that this change is a bug fix to match documented behavior I believe option 1 is a safe path.

## 2. Return `RecvError` regardless of whether a new value exists
For use cases where it is critical that the last value is preserved before shutting down, I'd expect such implementations to call [`borrow`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.borrow) and process the last value before shutting down. Thus I think it's fine if we do ~~don't~~ return `RecvError` whether a new value exists or not.

However, like you mention option 2 is less likely to break any dependents if they somehow depend on the current behavior.

 ------

### Sidenote
While we are on the topic of unifying the behavior, I find [`wait_for`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.wait_for)'s behavior also a bit odd, and its documentation makes me no wiser. 

From the docs:
>**If the channel is closed, then wait_for will return a [RecvError](https://docs.rs/tokio/latest/tokio/sync/watch/error/struct.RecvError.html)**. Once this happens, no more messages can ever be sent on the channel. When an error is returned, it is guaranteed that the closure has been called on the last value, and that it returned false for that value. (If the closure returned true, then the last value would have been returned instead of the error.)

The first sentence and the last sentence in parentheses seem to conflict each other. When testing the behavior I found that `wait_for` does indeed never return a `RecvError` if the predicate returns true, regardless of whether the value is seen or not.

DSharifi:
Closing this PR.

Thinking more over it, your suggestion of instead changing the behavior of `changed` to not return `RecvError` for unseen values makes more sense.

Will reopen a new PR for it shortly.

PR #7564, opened 2025-08-30
docs: Add cancellation safety note for `DelayQueue`'s `StreamExt::next` implementation

### LINKED ISSUE #7374: Document cancellation safety for DelayQueue

When using tokio::select macro, knowing the cancellation safety is useful information to prevent against any side effects.

DelayQueue seems to have cancellation safety but it would be nice for a guarantee to be provided.

None

**Additional context**

None

DSharifi:
## Motivation

Resolves #7374 

The usage of a `DelayQueue` often involves using its implementation of `StreamExt::next` in `select!` statements, where users need to know whether the method is cancellation safe or not.

`DelayQueue`'s `Stream::poll_next` implementation which is backed by `poll_expired` does not maintain any intermediary state with side effects on its internal state when returning `Pending`, thus `StreamExt::next` should be cancellation safe.

## Solution

Added a note that its implementation is cancellation safe.

### DIFF
--- a/tokio-util/src/time/delay_queue.rs
+++ b/tokio-util/src/time/delay_queue.rs
@@ -40,7 +40,7 @@ use std::task::{self, ready, Poll, Waker};
 /// # `Stream` implementation
 ///
 /// Items are retrieved from the queue via [`DelayQueue::poll_expired`]. If no delays have
-/// expired, no items are returned. In this case, `Poll::Pending` is returned and the
+/// expired, no items are returned. In this case, [`Poll::Pending`] is returned and the
 /// current task is registered to be notified once the next item's delay has
 /// expired.
 ///
@@ -66,9 +66,13 @@ use std::task::{self, ready, Poll, Waker};
 /// Capacity can be checked using [`capacity`] and allocated preemptively by using
 /// the [`reserve`] method.
 ///
+/// # Cancel safety
+///
+/// [`DelayQueue`]'s implementation of [`futures::StreamExt::next`] is cancellation safe.
+///
 /// # Usage
 ///
-/// Using `DelayQueue` to manage cache entries.
+/// Using [`DelayQueue`] to manage cache entries.
 ///
 /// ```rust,no_run
 /// use tokio_util::time::{DelayQueue, delay_queue};


tokio-util/src/time/delay_queue.rs
```diff
@@ -118,7 +122,8 @@ use std::task::{self, ready, Poll, Waker};
 /// [`insert`]: method@Self::insert
 /// [`insert_at`]: method@Self::insert_at
 /// [`Key`]: struct@Key
-/// [`Stream`]: https://docs.rs/futures/0.1/futures/stream/trait.Stream.html
+/// [`Stream`]: https://docs.rs/futures/0.3.31/futures/stream/trait.Stream.html
+/// [`StreamExt::next`]: https://docs.rs/futures/0.3.31/futures/stream/trait.StreamExt.html#method.next
```
ADD-SP: Since [`tokio_stream::StreamExt::next`] documents its own cancellation safety, and `tokio::select!` [explicitly references](https://github.com/tokio-rs/tokio/blob/0284d1b5c8ea5aff5b30c254200fb0a46c21d67c/tokio/src/macros/select.rs#L111) it, it’s more appropriate to link directly to [tokio_stream::StreamExt::next].

[`tokio_stream::StreamExt::next`]: https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.next

tokio-util/src/time/delay_queue.rs
```diff
@@ -66,9 +66,13 @@ use std::task::{self, ready, Poll, Waker};
 /// Capacity can be checked using [`capacity`] and allocated preemptively by using
 /// the [`reserve`] method.
 ///
+/// # Cancel safety
```
ADD-SP:
```suggestion
/// # Cancellation safety
```

I personally prefer the "Cancellation safety", not a blocker.

ADD-SP (commented):
Overall looks good to me.

tokio-util/src/time/delay_queue.rs
DSharifi: Good point, makes sense to point to `tokio_stream::StreamExt::next` instead.

~~Question regarding the wording in https://github.com/tokio-rs/tokio/blob/0284d1b5c8ea5aff5b30c254200fb0a46c21d67c/tokio/src/macros/select.rs#L111~~

~~What does **any** mean here? Does it mean any type defined in the `tokio`  ( also `tokio_util`then?) crate or **any** possible type? Is it not possible to create a custom type that implements Stream that doesn't provide cancellation safety?~~

tokio-util/src/time/delay_queue.rs
DSharifi: Same, I prefer "Cancellation" as well over "Cancel" but went with "Cancel" as I saw it's more used in the existing documentation and wanted to keep it consistent:

Grepping for terms in the repo:
- "/// # Cancel safety" -> 123 results
- "/// # Cancellation safety" -> 8 results


Feel free to apply the change if you think we should use "Cancellation"

tokio-util/src/time/delay_queue.rs
ADD-SP: > Is it not possible to create a custom type that implements Stream that doesn't provide cancellation safety?

It is possible for a malicious implementation of `Stream::poll_next` to break the cancellation safety of `StreamExt::next`.

---

I believe the more accurate description is

"on stream whose `poll_next` is cancellation safe"

ADD-SP: I think it might not be a big problem, a reasonable `Stream` implementation should be cancellation safe.

ADD-SP (approved):
Thanks!

DSharifi:
Thanks for the review! @ADD-SP

PR #7569, opened 2025-09-02
sync: use `UnsafeCell::get_mut` in `Mutex::get_mut` and `RwLock::get_mut`

kpp:
Remove two `unsafe {}` in the code. Blocked until rustc 1.83 MSRV.


ADD-SP (approved):
The `UnsafeCell::get_mut` is available since 1.50.0, and become const since 1.83.

PR #7574, opened 2025-09-04
ci: fill in the missing dependencies of the `test-pass` job

ADD-SP:
The job `test-pass` is intended to include all jobs to prevent merge PR without passing the CI. see also #2690.


Darksonn:
We probably just want to remove it. It doesn't really work for blocking github because if a dependency job fails, then `test-pass` falls into a cancelled state rather than a failed state, and the cancelled state does *not* block merging even if the job is listed as required.

ADD-SP:
You are right, I found an example.

PR #7578, opened 2025-09-05
stream: improve the the docs of `TcpListenerStream`

martin-g:
The example of TcpListenerStream uses listener1.chain(listener2). This way it will always poll the first listener until it is fully consumed, i.e. closed. Then it will start polling the second listener. It would be better to use `listener1.merge(listener2)` instead. This way both listeners will be polled one after the other.

While here also fix a typo in the constants for LOCALHOST - ipv4_listener actually used Ipv6Addr::LOCALHOST and ipv6_listener - Ipv4Addr::LOCALHOST

## Motivation

The documentation of TcpListenerStream was using the wrong constants. Also the `chain()` method will try to consumer the first Stream fully before polling the second. Using `merge()` would be better because it will poll both listeners one after the other.

## Solution

Improve the sample in the documentation.


ADD-SP (approved):
Thanks!

PR #7580, opened 2025-09-05
io: clarify the zero capacity case of `AsyncRead::poll_read`

martin-g:
## Motivation

5b9a290acd25d88416e70c73c49a54106fcf02b5 improved the trait documentation for the cases when the buf length didn't change but forgot to do the same for the method documentation.

## Solution

Apply the improvement also for the method documentation.


ADD-SP (approved):
Thanks!

PR #7581, opened 2025-09-05
examples: fix the write length in the `connect-tcp` example

martin-g:
## Motivation

The connect-udp example reads `n` bytes from the reader but writes the full 1024 bytes buffer to stdout. It should write only the `n` bytes it just read.

## Solution

Write only `n` bytes to stdout: 
```rust
Bytes::copy_from_slice(&buf[..n])
```


ADD-SP (approved):
Thanks!

PR #7586, opened 2025-09-07
task: improve the example of `poll_proceed`

ADD-SP:
## Motivation

I'm thinking about how to do <https://github.com/tokio-rs/tokio/pull/7565#discussion_r2315140342> properly, but I found the example of [`poll_proceed`] is a bit long. I always prefer short example because it allows people like me grasp the basic usage in 10 seconds.

[`poll_proceed`]: https://docs.rs/tokio/1.47.1/tokio/task/coop/fn.poll_proceed.html


tokio/src/task/coop/mod.rs
```diff
@@ -306,75 +306,44 @@ cfg_coop! {
...
-    ///             Poll::Pending
-    ///         }
+    ///        let coop = ready!(coop::poll_proceed(cx));
+    ///        if buf.remaining() == 0 {
+    ///            Poll::Ready(Ok(()))
+    ///        } else {
+    ///            while buf.remaining() > 0 {
+    ///                let len = ZEROES.len().min(buf.remaining());
+    ///                buf.put_slice(&ZEROES[..len]);
+    ///            }
```
Darksonn: You can improve the implementation a bit using [the `BufMut` impl](https://docs.rs/tokio/latest/tokio/io/struct.ReadBuf.html#impl-BufMut-for-ReadBuf%3C'a%3E).
```suggestion
    ///            buf.put_bytes(0, buf.remaining());
```

Darksonn:
Actually, I'm not so sure about the example. The branch returns `Poll::Ready` and we generally always `.made_progress()` when we return `Ready`. The other branch writes to the buffer but returns `Pending`, but if you wrote anything you *should* return `Ready`.

ADD-SP:
> Actually, I'm not so sure about the example. The branch returns `Poll::Ready` and we generally always `.made_progress()` when we return `Ready`. The other branch writes to the buffer but returns `Pending`, but if you wrote anything you _should_ return `Ready`.

Ah, bad mistake, I have fixed it.

ADD-SP:
I considered adding a reasonable `Pending` case, but I didn't find a reasonable one, so I chose to return `Poll::Ready` unconditionally.

Darksonn:
What about having the example be "add coop to the futures mpsc channel"?

ADD-SP:
> What about having the example be "add coop to the futures mpsc channel"?

Looks still straightforward. I chose to not include the `UnboundedSender`, as the `Sink` requires too many methods to be implemented, which makes the example too long.

image

tokio/src/task/coop/mod.rs
```diff
@@ -306,75 +306,41 @@ cfg_coop! {
...
+    ///            Poll::Pending => Poll::Pending,
+    ///        }
     ///     }
     /// }
-    ///
-    /// impl<T> Unpin for CountdownLatch<T> {}
     /// ```
+    ///
+    /// [`futures::channel::mpsc::UnboundedReceiver`]: https://docs.rs/futures/0.3.31/futures/channel/mpsc/struct.UnboundedReceiver.html
```
Darksonn: Can we avoid a http link here? At the very least, I want to avoid the version number.

ADD-SP: link was removed

tokio/src/task/coop/mod.rs
```diff
@@ -306,74 +306,38 @@ cfg_coop! {
...
-    ///             // other work.
-    ///             this.waker = Some(cx.waker().clone());
-    ///             Poll::Pending
-    ///         }
+    ///         match self.receiver.poll_next_unpin(cx) {
+    ///            Poll::Ready(v) => {
+    ///               // We received a value, so consume budget.
+    ///               coop.made_progress();
+    ///               Poll::Ready(v)
+    ///            }
```
Darksonn: We use four spaces for indentation.

Darksonn (approved):
One last nit. Feel free to merge after this.

PR #7590, opened 2025-09-08
task: add `tokio_util::task::JoinQueue`

nickkuk:
New `tokio_util::task::JoinQueue` structure as FIFO-ordered analog of `tokio::task::JoinSet`.

Issue: #7576

_Edit:_ renamed from `JoinDeque` to `JoinQueue`.

### DIFF
--- a/tokio-util/src/task/join_queue.rs
+++ b/tokio-util/src/task/join_queue.rs
@@ -1,0 +1,342 @@
+use super::AbortOnDropHandle;
+use std::{
+    collections::VecDeque,
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tokio::{
+    runtime::Handle,
+    task::{AbortHandle, Id, JoinError, JoinHandle},
+};
+
+/// A collection of tasks spawned on a Tokio runtime.
+///
+/// A `JoinQueue` can be used to await the completion of the tasks in FIFO
+/// order. That is, if tasks are spawned in the order A, B, C, then
+/// awaiting the next completed task will always return A first, then B,
+/// then C, regardless of the order in which the tasks actually complete.
+///
+/// All of the tasks must have the same return type `T`.
+///
+/// When the `JoinQueue` is dropped, all tasks in the `JoinQueue` are
+/// immediately aborted.
+#[derive(Debug)]
+pub struct JoinQueue<T>(VecDeque<AbortOnDropHandle<T>>);
+
+impl<T> JoinQueue<T> {
+    /// Create a new empty `JoinQueue`.
+    pub const fn new() -> Self {
+        Self(VecDeque::new())
+    }
+
+    /// Creates an empty `JoinQueue` with space for at least `capacity` tasks.
+    pub fn with_capacity(capacity: usize) -> Self {
+        Self(VecDeque::with_capacity(capacity))
+    }
+
+    /// Returns the number of tasks currently in the `JoinQueue`.
+    ///
+    /// This includes both tasks that are currently running and tasks that have
+    /// completed but not yet been removed from the queue because outputting of
+    /// them waits for FIFO order.
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    /// Returns whether the `JoinQueue` is empty.
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    /// Spawn the provided task on the `JoinQueue`, returning an [`AbortHandle`]
+    /// that can be used to remotely cancel the task.
+    ///
+    /// The provided future will start running in the background immediately
+    /// when this method is called, even if you don't await anything on this
+    /// `JoinQueue`.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if called outside of a Tokio runtime.
+    ///
+    /// [`AbortHandle`]: tokio::task::AbortHandle
+    #[track_caller]
+    pub fn spawn<F>(&mut self, task: F) -> AbortHandle
+    where
+        F: Future<Output = T> + Send + 'static,
+        T: Send + 'static,
+    {
+        self.push_back(tokio::spawn(task))
+    }
+
+    /// Spawn the provided task on the provided runtime and store it in this
+    /// `JoinQueue` returning an [`AbortHandle`] that can be used to remotely
+    /// cancel the task.
+    ///
+    /// The provided future will start running in the background immediately
+    /// when this method is called, even if you don't await anything on this
+    /// `JoinQueue`.
+    ///
+    /// [`AbortHandle`]: tokio::task::AbortHandle
+    #[track_caller]
+    pub fn spawn_on<F>(&mut self, task: F, handle: &Handle) -> AbortHandle
+    where
+        F: Future<Output = T> + Send + 'static,
+        T: Send + 'static,
+    {
+        self.push_back(handle.spawn(task))
+    }
+
+    /// Spawn the provided task on the current [`LocalSet`] and store it in this
+    /// `JoinQueue`, returning an [`AbortHandle`] that can be used to remotely
+    /// cancel the task.
+    ///
+    /// The provided future will start running in the background immediately
+    /// when this method is called, even if you don't await anything on this
+    /// `JoinQueue`.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if it is called outside of a `LocalSet`.
+    ///
+    /// [`LocalSet`]: tokio::task::LocalSet
+    /// [`AbortHandle`]: tokio::task::AbortHandle
+    #[track_caller]
+    pub fn spawn_local<F>(&mut self, task: F) -> AbortHandle
+    where
+        F: Future<Output = T> + 'static,
+        T: 'static,
+    {
+        self.push_back(tokio::task::spawn_local(task))
+    }
+
+    /// Spawn the blocking code on the blocking threadpool and store
+    /// it in this `JoinQueue`, returning an [`AbortHandle`] that can be
+    /// used to remotely cancel the task.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if called outside of a Tokio runtime.
+    ///
+    /// [`AbortHandle`]: tokio::task::AbortHandle
+    #[track_caller]
+    pub fn spawn_blocking<F>(&mut self, f: F) -> AbortHandle
+    where
+        F: FnOnce() -> T + Send + 'static,
+        T: Send + 'static,
+    {
+        self.push_back(tokio::task::spawn_blocking(f))
+    }
+
+    /// Spawn the blocking code on the blocking threadpool of the
+    /// provided runtime and store it in this `JoinQueue`, returning an
+    /// [`AbortHandle`] that can be used to remotely cancel the task.
+    ///
+    /// [`AbortHandle`]: tokio::task::AbortHandle
+    #[track_caller]
+    pub fn spawn_blocking_on<F>(&mut self, f: F, handle: &Handle) -> AbortHandle
+    where
+        F: FnOnce() -> T + Send + 'static,
+        T: Send + 'static,
+    {
+        self.push_back(handle.spawn_blocking(f))
+    }
+
+    fn push_back(&mut self, jh: JoinHandle<T>) -> AbortHandle {
+        let join_handle = AbortOnDropHandle::new(jh);
+        let abort_handle = join_handle.abort_handle();
+        self.0.push_back(join_handle);
+        abort_handle
+    }
+
+    /// Waits until the next task in FIFO order completes and returns its output.
+    ///
+    /// Returns `None` if the queue is empty.
+    ///
+    /// # Cancel Safety
+    ///
+    /// This method is cancel safe. If `join_next` is used as the event in a `tokio::select!`
+    /// statement and some other branch completes first, it is guaranteed that no tasks were
+    /// removed from this `JoinQueue`.
+    pub async fn join_next(&mut self) -> Option<Result<T, JoinError>> {
+        std::future::poll_fn(|cx| self.poll_join_next(cx)).await
+    }
+
+    /// Waits until the next task in FIFO order completes and returns its output,
+    /// along with the [task ID] of the completed task.
+    ///
+    /// Returns `None` if the queue is empty.
+    ///
+    /// When this method returns an error, then the id of the task that failed can be accessed
+    /// using the [`JoinError::id`] method.
+    ///
+    /// # Cancel Safety
+    ///
+    /// This method is cancel safe. If `join_next_with_id` is used as the event in a `tokio::select!`
+    /// statement and some other branch completes first, it is guaranteed that no tasks were
+    /// removed from this `JoinQueue`.
+    ///
+    /// [task ID]: tokio::task::Id
+    /// [`JoinError::id`]: fn@tokio::task::JoinError::id
+    pub async fn join_next_with_id(&mut self) -> Option<Result<(Id, T), JoinError>> {
+        std::future::poll_fn(|cx| self.poll_join_next_with_id(cx)).await
+    }
+
+    /// Aborts all tasks and waits for them to finish shutting down.
+    ///
+    /// Calling this method is equivalent to calling [`abort_all`] and then calling [`join_next`] in
+    /// a loop until it returns `None`.
+    ///
+    /// This method ignores any panics in the tasks shutting down. When this call returns, the
+    /// `JoinQueue` will be empty.
+    ///
+    /// [`abort_all`]: fn@Self::abort_all
+    /// [`join_next`]: fn@Self::join_next
+    pub async fn shutdown(&mut self) {
+        self.abort_all();
+        while self.join_next().await.is_some() {}
+    }
+
+    /// Awaits the completion of all tasks in this `JoinQueue`, returning a vector of their results.
+    ///
+    /// The results will be stored in the order they were spawned, not the order they completed.
+    /// This is a convenience method that is equivalent to calling [`join_next`] in
+    /// a loop. If any tasks on the `JoinQueue` fail with an [`JoinError`], then this call
+    /// to `join_all` will panic and all remaining tasks on the `JoinQueue` are
+    /// cancelled. To handle errors in any other way, manually call [`join_next`]
+    /// in a loop.
+    ///
+    /// # Cancel Safety
+    /// This method is not cancel safe as it calls `join_next` in a loop. If you need
+    /// cancel safety, manually call `join_next` in a loop with `Vec` accumulator.
+    ///
+    /// [`join_next`]: fn@Self::join_next
+    /// [`JoinError::id`]: fn@tokio::task::JoinError::id
+    pub async fn join_all(mut self) -> Vec<T> {
+        let mut output = Vec::with_capacity(self.len());
+
+        while let Some(res) = self.join_next().await {
+            match res {
+                Ok(t) => output.push(t),
+                Err(err) if err.is_panic() => std::panic::resume_unwind(err.into_panic()),
+                Err(err) => panic!("{err}"),
+            }
+        }
+        output
+    }
+
+    /// Aborts all tasks on this `JoinQueue`.
+    ///
+    /// This does not remove the tasks from the `JoinQueue`. To wait for the tasks to complete
+    /// cancellation, you should call `join_next` in a loop until the `JoinQueue` is empty.
+    pub fn abort_all(&mut self) {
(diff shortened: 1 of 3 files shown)


tokio-util/src/task/join_deque.rs
```diff
@@ -0,0 +1,87 @@
...
+/// order. That is, if tasks are spawned in the order A, B, C, then
+/// awaiting the next completed task will always return A first, then B,
+/// then C, regardless of the order in which the tasks actually complete.
+///
+/// All of the tasks must have the same return type `T`.
+///
+/// When the `JoinDeque` is dropped, all tasks in the `JoinDeque` are
+/// immediately aborted.
+#[derive(Debug)]
+pub struct JoinDeque<T>(VecDeque<JoinHandle<T>>);
```
ADD-SP: [`tokio_util::task::AbortOnDropHandle`] would be better.

[`tokio_util::task::AbortOnDropHandle`]: https://docs.rs/tokio-util/0.7.16/tokio_util/task/struct.AbortOnDropHandle.html

tokio-util/src/task/join_deque.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+
+impl<T> Drop for JoinDeque<T> {
+    fn drop(&mut self) {
+        for join_handle in self.0.drain(..) {
+            join_handle.abort();
+        }
+    }
+}
```
ADD-SP:
```suggestion
```

See <https://github.com/tokio-rs/tokio/pull/7590#discussion_r2331762837>

nickkuk:
> I think we should replicate most of the methods from `JoinSet`.

Ok, I pushed implementation of all methods except `try_join_next` and `try_join_next_with_id`. Do you know how to implement them?

tokio-util/src/task/join_deque.rs
```diff
@@ -0,0 +1,366 @@
...
+    /// [`AbortHandle`]: tokio::task::AbortHandle
+    /// [`spawn_local`]: Self::spawn_local
+    #[track_caller]
+    pub fn spawn_local_on<F>(&mut self, task: F, local_set: &LocalSet) -> AbortHandle
+    where
+        F: Future<Output = T> + 'static,
+        T: 'static,
+    {
+        self.insert(local_set.spawn_local(task))
+    }
```
ADD-SP: Since `LocalSet` might be deprecated in the future, we can exclude this method for now, see #6741.

tokio-util/src/task/join_deque.rs
```diff
@@ -0,0 +1,366 @@
...
+    #[track_caller]
+    pub fn spawn_blocking_on<F>(&mut self, f: F, handle: &Handle) -> AbortHandle
+    where
+        F: FnOnce() -> T + Send + 'static,
+        T: Send + 'static,
+    {
+        self.insert(handle.spawn_blocking(f))
+    }
+
+    fn insert(&mut self, jh: JoinHandle<T>) -> AbortHandle {
```
ADD-SP: Even this is an internal interface, I still prefer the `push_back` or similar name.

tokio-util/src/task/join_queue.rs
```diff
@@ -0,0 +1,346 @@
...
+    /// The results will be stored in the order they were spawned, not the order they completed.
+    /// This is a convenience method that is equivalent to calling [`join_next`] in
+    /// a loop. If any tasks on the `JoinQueue` fail with an [`JoinError`], then this call
+    /// to `join_all` will panic and all remaining tasks on the `JoinQueue` are
+    /// cancelled. To handle errors in any other way, manually call [`join_next`]
+    /// in a loop.
+    ///
+    /// [`join_next`]: fn@Self::join_next
+    /// [`JoinError::id`]: fn@tokio::task::JoinError::id
+    pub async fn join_all(mut self) -> Vec<T> {
```
Darksonn, line 217: All async functions must have a cancel safety section in the documentation.

tokio-util/src/task/join_queue.rs
```diff
@@ -0,0 +1,346 @@
...
+            Some(jh) => jh,
+        };
+        if let Poll::Ready(res) = Pin::new(jh).poll(cx) {
+            drop(self.0.pop_front().unwrap().detach());
+            Poll::Ready(Some(res))
+        } else {
+            // A JoinHandle generally won't emit a wakeup without being ready unless
+            // the coop limit has been reached. We yield to the executor in this
+            // case.
+            cx.waker().wake_by_ref();
```
Darksonn: This is wrong because `self.0.front_mut()` is *not* the same as the `pop_notified()` used by `JoinSet`. Since you always poll a `JoinHandle` no matter what, you should not call `wake_by_ref()` here. This will cause needless spinning.

Please add a test that would have caught this bug. You can look at tests using `tokio_test::spawn` for examples of how you can detect whether wakeups are sent.

nickkuk: I've removed these, but still need to understand more Tokio internals for implementing correct test.

Also `try_join_next` and `try_join_next_with_id` methods are missing and I don't see that it is possible to implement them currently without some access to Tokio internals.

Darksonn: You can try these steps:

1. Create a `JoinQueue` with a task on it. The task can wait on an oneshot channel so you can control it.
2. Call `let join_next = tokio_test::task::spawn(join_queue.join_next())` to create a tracked future.
3. Run `assert_pending!(join_next.poll())`
4. Run `assert!(!join_next.is_woken())` to assert not woken.
5. Make the task exit using the oneshot channel. Wait for it to exit with `yield_now()`.
6. Run `assert!(join_next.is_woken())` to assert woken by task exit.
7. Run `assert_ready!(join_next.poll())` to get output

ADD-SP (commented):
* Please avoid force-pushing during the review process.
* Could you add tests for this feature?

nickkuk:
@ADD-SP yes, added.

tokio-util/src/task/join_queue.rs
```diff
@@ -0,0 +1,342 @@
...
+    ///  * `Poll::Ready(Some(Err(err)))` if the next task in this `JoinQueue` has panicked or been
+    ///    aborted. The `err` is the `JoinError` from the panicked/aborted task.
+    ///  * `Poll::Ready(None)` if the `JoinQueue` is empty.
+    pub fn poll_join_next(&mut self, cx: &mut Context<'_>) -> Poll<Option<Result<T, JoinError>>> {
+        let jh = match self.0.front_mut() {
+            None => return Poll::Ready(None),
+            Some(jh) => jh,
+        };
+        if let Poll::Ready(res) = Pin::new(jh).poll(cx) {
+            drop(self.0.pop_front().unwrap().detach());
```
ADD-SP, line 275: Why do we need the `detach()` here?

tokio-util/src/task/join_queue.rs
```diff
@@ -0,0 +1,342 @@
...
+    pub fn poll_join_next_with_id(
+        &mut self,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<(Id, T), JoinError>>> {
+        let jh = match self.0.front_mut() {
+            None => return Poll::Ready(None),
+            Some(jh) => jh,
+        };
+        if let Poll::Ready(res) = Pin::new(jh).poll(cx) {
+            let jh = self.0.pop_front().unwrap().detach();
```
ADD-SP, line 316: Why do we need the `detach()` here?

tokio-util/tests/task_join_queue.rs
ADD-SP, line 1: We still need more tests to cover more cases. Only one test is not enough.

tokio-util/src/task/join_queue.rs
```diff
@@ -0,0 +1,342 @@
...
+    task::{Context, Poll},
+};
+use tokio::{
+    runtime::Handle,
+    task::{AbortHandle, Id, JoinError, JoinHandle},
+};
+
+/// A collection of tasks spawned on a Tokio runtime.
+///
+/// A `JoinQueue` can be used to await the completion of the tasks in FIFO
```
ADD-SP:
```suggestion
/// A [`JoinQueue`] can be used to await the completion of the tasks in FIFO
```

nit

tokio-util/src/task/join_queue.rs
```diff
@@ -0,0 +1,342 @@
...
+/// order. That is, if tasks are spawned in the order A, B, C, then
+/// awaiting the next completed task will always return A first, then B,
+/// then C, regardless of the order in which the tasks actually complete.
+///
+/// All of the tasks must have the same return type `T`.
+///
+/// When the `JoinQueue` is dropped, all tasks in the `JoinQueue` are
```
ADD-SP:
```suggestion
/// When the [`JoinQueue`] is dropped, all tasks in the `JoinQueue` are
```
nit

tokio-util/src/task/join_queue.rs
```diff
@@ -0,0 +1,342 @@
...
+    pub const fn new() -> Self {
+        Self(VecDeque::new())
+    }
+
+    /// Creates an empty `JoinQueue` with space for at least `capacity` tasks.
+    pub fn with_capacity(capacity: usize) -> Self {
+        Self(VecDeque::with_capacity(capacity))
+    }
+
+    /// Returns the number of tasks currently in the `JoinQueue`.
```
ADD-SP:
```suggestion
    /// Returns the number of tasks currently in the [`JoinQueue`].
```
nit

tokio-util/src/task/join_queue.rs
```diff
@@ -0,0 +1,342 @@
...
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    /// Returns whether the `JoinQueue` is empty.
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    /// Spawn the provided task on the `JoinQueue`, returning an [`AbortHandle`]
```
ADD-SP:
```suggestion
    /// Spawn the provided task on the [`JoinQueue`], returning an [`AbortHandle`]
```
nit

tokio-util/src/task/join_queue.rs
nickkuk: I just thought that it could be a bit faster, i.e. abort then drop vs just drop.

tokio-util/src/task/join_queue.rs
nickkuk: I just thought that it could be a bit faster, i.e. abort then drop vs just drop.

tokio-util/tests/task_join_queue.rs
nickkuk: I've just added more tests, all are adapted from `JoinSet` ones.

tokio-util/src/task/join_queue.rs
```diff
@@ -0,0 +1,342 @@
...
+    /// Awaits the completion of all tasks in this [`JoinQueue`], returning a vector of their results.
+    ///
+    /// The results will be stored in the order they were spawned, not the order they completed.
+    /// This is a convenience method that is equivalent to calling [`join_next`] in
+    /// a loop. If any tasks on the [`JoinQueue`] fail with an [`JoinError`], then this call
+    /// to `join_all` will panic and all remaining tasks on the [`JoinQueue`] are
+    /// cancelled. To handle errors in any other way, manually call [`join_next`]
+    /// in a loop.
+    ///
+    /// # Cancel Safety
```
ADD-SP, line 210:
```suggestion
    /// # Cancel Safety

```

tokio-util/tests/task_join_queue.rs
```diff
@@ -0,0 +1,208 @@
...
+    assert_eq!(queue.len(), 0);
+}
+
+#[tokio::test(start_paused = true)]
+async fn test_join_queue_join_all() {
+    let mut queue = JoinQueue::new();
+
+    for i in 0..5 {
+        queue.spawn(async move {
+            tokio::time::sleep(Duration::from_secs(5 - i as u64)).await;
```
ADD-SP: It looks like we want to simulate the reverse completion order of futures, however, due to the [Auto-advance], this might not work as our expected.

I suggest achieving this by several oneshot channels.

[Auto-advance]: https://docs.rs/tokio/1.47.1/tokio/time/fn.advance.html#auto-advance

tokio-util/tests/task_join_queue.rs
```diff
@@ -0,0 +1,208 @@
...
+async fn test_join_queue_shutdown() {
+    let mut queue = JoinQueue::new();
+
+    for _ in 0..5 {
+        queue.spawn(futures::future::pending::<()>());
+    }
+
+    queue.shutdown().await;
+    assert_eq!(queue.len(), 0);
+}
```
ADD-SP, line 147: We also want to check if the underlying futures is actually aborted.

tokio-util/tests/task_join_queue.rs
```diff
@@ -0,0 +1,208 @@
...
+
+#[tokio::test(start_paused = true)]
+async fn test_join_queue_with_manual_abort() {
+    let mut queue = JoinQueue::new();
+    let mut num_canceled = 0;
+    let mut num_completed = 0;
+    for i in 0..16 {
+        let abort = queue.spawn(async move {
+            tokio::time::sleep(Duration::from_secs(20 - i as u64)).await;
```
ADD-SP: It looks like we want to simulate the reverse completion order of futures, however, due to the [Auto-advance], this might not work as our expected.

I suggest achieving this by several oneshot channels.

[Auto-advance]: https://docs.rs/tokio/1.47.1/tokio/time/fn.advance.html#auto-advance

tokio-util/src/task/join_queue.rs
ADD-SP: It might not make many differences, but it confused developers, I prefer removing it.

tokio-util/tests/task_join_queue.rs
ADD-SP: Discussion moved to <https://github.com/tokio-rs/tokio/pull/7590#discussion_r2355602217>.

tokio-util/src/task/join_queue.rs
nickkuk: As it is one of two methods that can be hot (i.e. as a part of `queue.join_next().await`; another one is `queue.spawn`) I wouldn't do things that are unnecessary. Can I just add an explanatory comment to not confuse developers?

tokio-util/src/task/join_queue.rs
ADD-SP: Discussion moved to <https://github.com/tokio-rs/tokio/pull/7590#discussion_r2355602217>.

tokio-util/src/task/join_queue.rs
ADD-SP: > I just thought that it could be a bit faster, i.e. abort then drop vs just drop.

Why is manually detaching faster?

nickkuk: I added the comment that without `detach` additional unnecessary `abort` will be called for already finished future.

ADD-SP: I think this is reasonable.

tokio-util/src/task/join_queue.rs
```diff
@@ -0,0 +1,349 @@
...
+    pub fn poll_join_next(&mut self, cx: &mut Context<'_>) -> Poll<Option<Result<T, JoinError>>> {
+        let jh = match self.0.front_mut() {
+            None => return Poll::Ready(None),
+            Some(jh) => jh,
+        };
+        if let Poll::Ready(res) = Pin::new(jh).poll(cx) {
+            // Use `detach` to avoid calling `abort` on a task that has already completed.
+            // Dropping `AbortOnDropHandle` would abort the task, but since it is finished,
+            // we only need to drop the `JoinHandle` for cleanup.
+            drop(self.0.pop_front().unwrap().detach());
```
ADD-SP, line 275: Another question, why did we use explicit drop?

nickkuk: No real reason for this, I just like to do something with function result when it is not `()` to show that I'm aware of it.

ADD-SP: Not a blocker, but I personally prefer `let _ = self.0.pop_front().unwrap().detach()`, as `_` will also be dropped immediately, and it looks better to me.

nickkuk: Ok, I changed it. Actually I also use `let _ = ...` more often than explicit `drop`.

nickkuk: But CI is against:
```
error: non-binding `let` on a future
   --> tokio-util/src/task/join_queue.rs:275:13
    |
275 |             let _ = self.0.pop_front().unwrap().detach();
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: consider awaiting the future or dropping explicitly with `std::mem::drop`
```

tokio-util/src/task/join_queue.rs
```diff
@@ -0,0 +1,349 @@
...
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tokio::{
+    runtime::Handle,
+    task::{AbortHandle, Id, JoinError, JoinHandle},
+};
+
+/// A collection of tasks spawned on a Tokio runtime.
```
ADD-SP:
```suggestion
/// A FIFO queue for tasks spawned on a Tokio runtime.
```

tokio-util/tests/task_join_queue.rs
```diff
@@ -0,0 +1,228 @@
...
+    tokio::time::sleep(Duration::from_secs(2)).await;
+
+    queue.abort_all();
+    assert_eq!(queue.len(), 10);
+
+    let mut count = 0;
+    while let Some(res) = queue.join_next().await {
+        if count < 5 {
+            assert!(res.is_err());
+            assert!(res.unwrap_err().is_cancelled());
```
ADD-SP:
```suggestion
            assert!(res.unwrap_err().is_cancelled());
```

The `unwrap_err` can cover the `assert!(res.is_err())`.

tokio-util/tests/task_join_queue.rs
```diff
@@ -0,0 +1,228 @@
...
+            }
+            None => break,
+        }
+    }
+
+    assert_eq!(num_canceled, 8);
+    assert_eq!(num_completed, 8);
+}
+
+#[tokio::test(flavor = "current_thread")]
```
ADD-SP: Emm, I think the default flavor is the `current_thread`. Is there any specific reason we set it explicitly?

tokio-util/tests/task_join_queue.rs
```diff
@@ -0,0 +1,228 @@
...
+                assert_eq!(res, num_completed * 2);
+                num_completed += 1;
+            }
+            Some(Err(e)) => {
+                assert!(e.is_cancelled());
+                num_canceled += 1;
+            }
+            None => break,
+        }
+    }
```
ADD-SP: Is `while let Some(v) = queue.jon_next().await {}` more straightforward?

tokio-util/src/task/join_queue.rs
ADD-SP: Ah, let's use `drop` here.

nickkuk:
It seems that now CI failed with errors unrelated to this PR's changes.

@ADD-SP should we restart CI, or should I update branch, or something else?

ADD-SP:
> It seems that now CI failed with errors unrelated to this PR's changes.
> 
> @ADD-SP should we restart CI, or should I update branch, or something else?

These failures are caused by the recent Rust release (Rust 1.90.0), you could update the branch after <https://github.com/tokio-rs/tokio/pull/7630> is merged.

ADD-SP (commented):
Overall looks good to me, @nickkuk Could you sync the changes from the base branch?

tokio-util/src/task/join_queue.rs
```diff
@@ -0,0 +1,349 @@
...
+            Poll::Pending
+        }
+    }
+}
+
+impl<T> Default for JoinQueue<T> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
ADD-SP: Can this be derived by `#[derive(Default)]`?

ADD-SP (approved):
Overall looks good to me.

tokio-util/src/task/join_queue.rs
nickkuk: Hmm, on the other hand, `#[derive(Default)]` adds implicit `T: Default` constraint.

ADD-SP: Ah, I see, let's revert it.

PR #7592, opened 2025-09-09
doc: Fix the version used as an example how to use the latest minor of LTS

martin-g:
## Motivation

The version in the test (1.32.x) was lagging behind the one in the test snippet for Cargo.toml

## Solution

Sync the versions in the text and in the snippet


Darksonn (approved):
Thanks.

PR #7599, opened 2025-09-09
io: export Chain for users of AsyncReadExt::chain

aatifsyed:
## Motivation

I want to downcast a `hyper::upgrade::Upgraded`.
This also makes the code consistent with `Take`, and other combinators like those found in `std::iter`


ADD-SP:
> I want to downcast a `hyper::upgrade::Upgraded`.

Sorry, I don't understand your usecase, could you please elaborate your usecase?

aatifsyed:
> Sorry, I don't understand your usecase, could you please elaborate your usecase?

I'll rephrase :)

I want to downcast a trait object. 
To downcast, I'd like to name the trait object. 
I can't name a `Chain`. This PR makes that possible. 

Also makes `Chain` more consistent with `Take`, which is already nameable.

PR #7604, opened 2025-09-10
net: Fix copy/paste errors in udp peek methods

martin-g:
## Motivation

* On Windows the self.io.peek() method may return an error, not `peek_from`
* poll_peek() does not return the sender address on success. It returns Poll::Ready(Ok(())) indicating no problems during the peek
* A comment mentions `recv` that read the `n` bytes. It is actually self.io.peek()

## Solution

Fix the typos / copy-paste errors in docs and comments


Darksonn (approved):
Thanks.

PR #7607, opened 2025-09-11
net: Update the docs of net::unix::pipe::OpenOptions that read_write is supported on Android too

martin-g:
## Motivation

https://github.com/tokio-rs/tokio/pull/7426 made `tokio::net::unix::pipe::OpenOptions::read_write` available on Android too

## Solution

Update the docs to mention `Android` next to `Linux`

### DIFF
--- a/tokio/src/net/unix/pipe.rs
+++ b/tokio/src/net/unix/pipe.rs
@@ -99,7 +99,7 @@ pub fn pipe() -> io::Result<(Sender, Receiver)> {
 /// # }
 /// ```
 ///
-/// Opening a [`Sender`] on Linux when you are sure the file is a FIFO:
+/// Opening a [`Sender`] on Linux/Android when you are sure the file is a FIFO:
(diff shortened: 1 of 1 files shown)


tokio/src/net/unix/pipe.rs
```diff
@@ -99,7 +99,7 @@ pub fn pipe() -> io::Result<(Sender, Receiver)> {
 /// # }
 /// ```
 ///
-/// Opening a [`Sender`] on Linux when you are sure the file is a FIFO:
+/// Opening a [`Sender`] on Linux/Android when you are sure the file is a FIFO:
```
Darksonn, line 102: The doc cfg stuff already makes it clear that it works on Android:

image

IMO Android is Linux and we don't need to *also* mention this in the text.

martin-g: Fair enough!

PR #7614, opened 2025-09-12
metrics: clarify that `num_alive_tasks` is not strongly consistent

joe-amzn:
## Motivation

I was confused by this to I opened  https://github.com/tokio-rs/tokio/issues/7591

## Solution

Update the documentation so future people know not to trust this metric when using the multithreaded runtime.

### DIFF
--- a/tokio/src/runtime/metrics/runtime.rs
+++ b/tokio/src/runtime/metrics/runtime.rs
@@ -54,6 +54,10 @@ impl RuntimeMetrics {
     /// This counter increases when a task is spawned and decreases when a
     /// task exits.
     ///
+    /// Note: When using the multi-threaded runtime this number may not
+    /// not have strong consistency i.e. no tasks may be running but the metric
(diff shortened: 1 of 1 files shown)


tokio/src/runtime/metrics/runtime.rs
```diff
@@ -54,6 +54,10 @@ impl RuntimeMetrics {
     /// This counter increases when a task is spawned and decreases when a
     /// task exits.
     ///
+    /// Note: When using the multi-threaded runtime this number may not
+    /// not have strong consistency i.e. no tasks may be running but the metric
+    /// reports otherwise.
+    ///
```
ADD-SP, line 60: I would say `RuntimeMetrics` is not designed to be accurate, I think it would better move this "Note" to the struct level.

ADD-SP (approved):
Thanks! I don't want to block this PR too long for nits.

PR #7617, opened 2025-09-13
fs: use FileOptions inside fs::File to support uring

WowVeryLogin:
Very straightforward patch: use FileOptions inside fs::File implementation the same way as std does. By this we automatically support io_uring path for File::open and File::create.


mox692 (approved):
Thanks!

PR #7618, opened 2025-09-14
examples: update to http 1

tottoto:
## Motivation

`http` version 1 could be used in the examples.

## Solution

Updates to `http` 1.


ADD-SP (approved):
Thanks!

PR #7620, opened 2025-09-14
Use std::os::fd instead of std::os::unix::io

martin-g:
## Motivation

std::os::fd has been stabilised since 1.66: https://releases.rs/docs/1.66.0/

std::os::fd are usable for unix and WASI

## Solution

Replace all occurrences of `std::os::unix::io` with `std::os::fd`


ADD-SP:
Thanks for your PR! Since the `std::os::unix::io` is not deprecated, and this doesn't significantly simplify the codebase, we don't have a strong motivation for this change.

PR #7621, opened 2025-09-15
fs: use feature for io-uring instead of cfg

mox692:
Derived from #7547, with a few updates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,8 +120,8 @@ jobs:
       - name: test tokio full
         run: |
           set -euxo pipefail
-          cargo nextest run --features full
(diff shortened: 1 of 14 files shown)


ADD-SP (approved):
Looks good to me.

ADD-SP:
Emm, Why did this happen? It looks strange.

mox692:
I think github actions is waiting for jobs that have already been deleted.
@Darksonn Are you able to fix this? IIRC, this is related to the repository settings.

Darksonn:
Done.

ADD-SP:
Thanks for @Sytten 's initial work!

Sytten:
Thanks for pushing it through! 🚀

PR #7628, opened 2025-09-17
net: clarify the supported platform of `set_reuseport()` and `reuseport()`

VbhvGupta:
This PR fixes documentation mismatches introduced by cfg changes that exclude Cygwin from certain unix-only APIs.

- Update docstrings in tokio/src/net/tcp/socket.rs so `set_reuseport` and `reuseport` explicitly mention Cygwin alongside Solaris & Illumos.

### Rationale:
The implementation cfgs exclude Cygwin `(not(target_os = "cygwin"))`, but the prose still listed only Solaris & Illumos. 
That mismatch confuses users who attempt to compile for `*-pc-cygwin`. 
This PR makes docstring consistent with the code.

No runtime or API changes are included here.


ADD-SP (approved):
Thanks!

PR #7629, opened 2025-09-18
sync: close the `broadcast::Sender` in `broadcast::Sender::new()`

martin-g:
## Motivation

tokio::sync::broadcast::Sender::closed() should return Poll::Ready if there are no receivers for the channel.
A newly created Sender with `Sender::new()` currently returns Poll::Pending although there are no receivers for it.

## Solution

Initialize Tail::closed with `receivers_count == 0`


ADD-SP (commented):
@martin-g Could you sync the changes from the base branch to fix the CI failures?

PR #7634, opened 2025-09-19
fs: emit compilation error without `tokio_unstable` for `io-uring`

ADD-SP:
The fixup of <https://github.com/tokio-rs/tokio/pull/7621> for better error message. Otherwise, the error message is a bit encrypted.

```bash
$ cargo run
error: failed to select a version for `tokio`.
    ... required by package `uring-test v0.1.0 (xxxx)`
versions that meet the requirements `*` (locked to 1.47.1) are: 1.47.1

package `uring-test` depends on `tokio` with feature `io-uring` but `tokio` does not have that feature.
 A required dependency with that name exists, but only optional dependencies can be used as features.


failed to select a version for `tokio` which could resolve this conflict
```

This is a bit complicated. It seems like it's not worth the complexity, but we still have to do it because we also want to convert `--cfg tokio_taskdump` into a Cargo feature, and taskdump already has similar user-friendly error messages.

https://github.com/tokio-rs/tokio/blob/3b5a15dfdfc5dc789dcb80cef0986fb2fa5cac0a/tokio/src/lib.rs#L477-L478

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -210,7 +210,7 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
 
-      - name: Check tests --unstable --features ${{ env.TOKIO_STABLE_FEATURES }}
+      - name: Check tests --features ${{ env.TOKIO_STABLE_FEATURES }}
         run: cargo check --workspace --tests --features $TOKIO_STABLE_FEATURES
 
   valgrind:
@@ -340,15 +340,13 @@ jobs:
         with:
           tool: cargo-nextest
       - uses: Swatinem/rust-cache@v2
-      # Run `tokio` with "unstable" and "taskdump" cfg flags.
+      # Since the internal-mt-counters feature is only for debugging purposes,
+      # we can enable all features including unstable.
       - name: check tokio full --cfg unstable --cfg internal-mt-counters
         run: |
           set -euxo pipefail
-          # taskdump is an unstable feature, but it can only be enabled
-          # by --cfg tokio_taskdump, not by a feature flag, so we can
-          # use $TOKIO_STABLE_FEATURES here.
-          cargo nextest run --features $TOKIO_STABLE_FEATURES
-          cargo test --doc --features $TOKIO_STABLE_FEATURES
+          cargo nextest run --all-features
+          cargo test --doc --all-features
         working-directory: tokio
         env:
           RUSTFLAGS: --cfg tokio_unstable --cfg tokio_internal_mt_counters -Dwarnings
@@ -747,9 +745,7 @@ jobs:
           cargo hack --remove-dev-deps --workspace
           # Update Cargo.lock to minimal version dependencies.
           cargo update -Z minimal-versions
-          # We use `--features "full,test-util"` instead of `--all-features` since
-          # `--all-features` includes `io_uring`, which requires '--cfg tokio_unstable'.
-          cargo hack check -p tokio --features full,test-util --ignore-private
+          cargo hack check -p tokio --features $TOKIO_STABLE_FEATURES --ignore-private
           cargo hack check -p tokio-macros -p tokio-stream -p tokio-util -p tokio-test --all-features --ignore-private
       - name: "check --all-features --unstable -Z minimal-versions"
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,9 @@ env:
   # - tokio-test/Cargo.toml
   # - tokio-stream/Cargo.toml
   rust_min: '1.70'
+  # This excludes unstable features like io_uring,
+  # which require '--cfg tokio_unstable'.
+  TOKIO_STABLE_FEATURES: "full,test-util"
 
 defaults:
   run:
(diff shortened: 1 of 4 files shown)


tokio-macros/src/lib.rs
```diff
@@ -539,7 +539,7 @@ pub fn main_rt(args: TokenStream, item: TokenStream) -> TokenStream {
 ///         panic!("This panic will shutdown the runtime.");
 ///     }).await;
 /// }
-/// # #[cfg(not(tokio_unstable))]
```
ADD-SP, line 542: I removed it due to this

<details>
<summary>Click to see the image</summary>



</details>

tokio-macros/src/lib.rs
```diff
@@ -560,7 +560,7 @@ pub fn main_rt(args: TokenStream, item: TokenStream) -> TokenStream {
 ///             }).await;
 ///         })
 /// }
-/// # #[cfg(not(tokio_unstable))]
```
ADD-SP, line 563: See <https://github.com/tokio-rs/tokio/pull/7634/files#r2365346442>.

mox692:
Sorry for the delay, I'll check this tomorrow.

.github/workflows/ci.yml
```diff
@@ -204,8 +209,9 @@ jobs:
         run: sed -i '/\[features\]/a plsend = ["parking_lot/send_guard"]' tokio/Cargo.toml
 
       - uses: Swatinem/rust-cache@v2
-      - name: Check tests with all features enabled
-        run: cargo check --workspace --all-features --tests
+
+      - name: Check tests --unstable --features ${{ env.TOKIO_STABLE_FEATURES }}
```
mox692: I think this test will not use unstable features.

```suggestion
      - name: Check tests --features ${{ env.TOKIO_STABLE_FEATURES }}
```

.github/workflows/ci.yml
```diff
@@ -336,8 +344,11 @@ jobs:
       - name: check tokio full --cfg unstable --cfg internal-mt-counters
         run: |
           set -euxo pipefail
-          cargo nextest run --all-features
-          cargo test --doc --all-features
+          # taskdump is an unstable feature, but it can only be enabled
+          # by --cfg tokio_taskdump, not by a feature flag, so we can
+          # use $TOKIO_STABLE_FEATURES here.
+          cargo nextest run --features $TOKIO_STABLE_FEATURES
+          cargo test --doc --features $TOKIO_STABLE_FEATURES
```
mox692: Does this part really use the taskdump feature?

.github/workflows/ci.yml
```diff
@@ -724,14 +740,17 @@ jobs:
...
         run: |
           # Remove dev-dependencies from Cargo.toml to prevent the next `cargo update`
           # from determining minimal versions based on dev-dependencies.
           cargo hack --remove-dev-deps --workspace
           # Update Cargo.lock to minimal version dependencies.
           cargo update -Z minimal-versions
-          cargo hack check --all-features --ignore-private
+          # We use `--features "full,test-util"` instead of `--all-features` since
+          # `--all-features` includes `io_uring`, which requires '--cfg tokio_unstable'.
+          cargo hack check -p tokio --features full,test-util --ignore-private
```
mox692: Can't we use `TOKIO_STABLE_FEATURES` here?

.github/workflows/ci.yml
```diff
@@ -780,10 +794,12 @@ jobs:
...
       # Run clippy
-      - name: "clippy --all ${{ matrix.rustflags }}"
-        run: cargo clippy --all --tests --all-features --no-deps
+      - name: "clippy --all --features ${{ env.TOKIO_STABLE_FEATURES }}"
+        run: cargo clippy --all --tests --no-deps --features $TOKIO_STABLE_FEATURES
+      - name: "clippy --all --all-features --unstable"
+        run: cargo clippy --all --tests --no-deps --all-features
         env:
-          RUSTFLAGS: ${{ matrix.rustflags }}
+          RUSTFLAGS: --cfg tokio_unstable --cfg tokio_taskdump -Dwarnings
```
mox692, line 798: If I understand correctly, this change removed the case where we don't use `tokio_unstable`?

ADD-SP: There is a `cargo cliipy` in the near above, did you mean this?

```yaml
      - name: "clippy --all --features ${{ env.TOKIO_STABLE_FEATURES }}"
        run: cargo clippy --all --tests --no-deps --features $TOKIO_STABLE_FEATURES
```

.github/workflows/ci.yml
ADD-SP: Emm, it looks doesn't use the taskdump feature. Initially, this job was introduced in <https://github.com/tokio-rs/tokio/commit/9eb3f5b556c1484f17368b193001ce972219d980>, but is looks not related to the taskdump.

Fixed in <https://github.com/tokio-rs/tokio/pull/7634/commits/3cee7fd0441b9bc966874176c8fd62035627ddab>.

.github/workflows/ci.yml
mox692: Never mind, I missed that one.

mox692 (approved):

PR #7636, opened 2025-09-20
task: add `JoinQueue::try_join_next[_with_id]` methods

nickkuk:
I want to add missing `try_join_next` and `try_join_next_with_id` methods to `tokio_util::task::JoinQueue`, but I'm not sure that it is correct usage of `futures_util::task::noop_waker()`.

Can anyone give me a hint how to implement them correctly? Or maybe Tokio need to expose any new methods, e.g., something like this?
```rust
impl JoinHandle<T> {
    pub fn try_into_result(self) -> Result<Result<T, JoinError>, Self> { ... }
}
```

### DIFF
--- a/tokio-util/src/task/join_queue.rs
+++ b/tokio-util/src/task/join_queue.rs
@@ -183,6 +183,53 @@ impl<T> JoinQueue<T> {
         std::future::poll_fn(|cx| self.poll_join_next_with_id(cx)).await
     }
 
+    /// Tries to join the next task in FIFO order if it has completed.
+    ///
+    /// Returns `None` if the queue is empty or if the next task is not yet ready.
+    pub fn try_join_next(&mut self) -> Option<Result<T, JoinError>> {
+        let waker = futures_util::task::noop_waker();
+        let mut cx = Context::from_waker(&waker);
+
+        let jh = self.0.front_mut()?;
+        if let Poll::Ready(res) = Pin::new(jh).poll(&mut cx) {
+            // Use `detach` to avoid calling `abort` on a task that has already completed.
+            // Dropping `AbortOnDropHandle` would abort the task, but since it is finished,
+            // we only need to drop the `JoinHandle` for cleanup.
+            drop(self.0.pop_front().unwrap().detach());
+            Some(res)
+        } else {
+            None
+        }
+    }
+
+    /// Tries to join the next task in FIFO order if it has completed and return its output,
+    /// along with its [task ID].
+    ///
+    /// Returns `None` if the queue is empty or if the next task is not yet ready.
+    ///
+    /// When this method returns an error, then the id of the task that failed can be accessed
+    /// using the [`JoinError::id`] method.
+    ///
+    /// [task ID]: crate::task::Id
+    /// [`JoinError::id`]: fn@crate::task::JoinError::id
+    pub fn try_join_next_with_id(&mut self) -> Option<Result<(Id, T), JoinError>> {
+        let waker = futures_util::task::noop_waker();
+        let mut cx = Context::from_waker(&waker);
+
+        let jh = self.0.front_mut()?;
+        if let Poll::Ready(res) = Pin::new(jh).poll(&mut cx) {
+            // Use `detach` to avoid calling `abort` on a task that has already completed.
+            // Dropping `AbortOnDropHandle` would abort the task, but since it is finished,
+            // we only need to drop the `JoinHandle` for cleanup.
+            let jh = self.0.pop_front().unwrap().detach();
+            let id = jh.id();
+            drop(jh);
+            Some(res.map(|output| (id, output)))
+        } else {
+            None
+        }
+    }
+
     /// Aborts all tasks and waits for them to finish shutting down.
     ///
     /// Calling this method is equivalent to calling [`abort_all`] and then calling [`join_next`] in


ADD-SP (commented):
You may want [`tokio_util::task::AbortOnDropHandle::is_finished`].

[`tokio_util::task::AbortOnDropHandle::is_finished`]: https://docs.rs/tokio-util/0.7.16/tokio_util/task/struct.AbortOnDropHandle.html#method.is_finished

nickkuk:
> You may want [`tokio_util::task::AbortOnDropHandle::is_finished`](https://docs.rs/tokio-util/0.7.16/tokio_util/task/struct.AbortOnDropHandle.html#method.is_finished).

Yeah, it calls and checks [`self.raw.header().state.load()`](https://github.com/tokio-rs/tokio/blob/1b98d5ad85f44a2d2df91b520990ded04db81371/tokio/src/runtime/task/join.rs#L252C21-L252C51), but literally the same [is done in `JoinHandle::poll`](https://github.com/tokio-rs/tokio/blob/1b98d5ad85f44a2d2df91b520990ded04db81371/tokio/src/runtime/task/harness.rs#L422) before taking result.

My real question is whether passing `futures_util::task::noop_waker()` breaks some Tokio internal invariant.

ADD-SP:
> My real question is whether passing futures_util::task::noop_waker() breaks some Tokio internal invariant.

According to the contract of the `Future` trait, an `noop_waker` should not causes language-level UB, but might cause the  application-level bugs. However, for the usecase of `JoinQueue::try_join_next`, it should be ok.

Personally, I feel the `noop_waker` here is a bit hacky. Would `jh.is_finished` then [`now_or_never`] be better?


[`now_or_never`]: https://docs.rs/futures-util/0.3.31/futures_util/future/trait.FutureExt.html#method.now_or_never

nickkuk:
Looks like `now_or_never` does almost the same:
```rust
fn now_or_never(self) -> Option<Self::Output>
where
    Self: Sized,
{
    let noop_waker = crate::task::noop_waker();
    let mut cx = Context::from_waker(&noop_waker);

    let this = self;
    pin_mut!(this);
    match this.poll(&mut cx) {
        Poll::Ready(x) => Some(x),
        _ => None,
    }
}
```

After look at it more, I think that additional `is_finished` should be used anyway because for not completed future polling [tries to set a waker](https://github.com/tokio-rs/tokio/blob/1b98d5ad85f44a2d2df91b520990ded04db81371/tokio/src/runtime/task/harness.rs#L427), which can break further internal Tokio logic.

I'm pushing your suggestion, will add tests a bit later.

tokio-util/tests/task_join_queue.rs
ADD-SP, line 1: Could you also add a test for `try_join_next_with_id`?

tokio-util/tests/task_join_queue.rs
```diff
@@ -221,3 +221,61 @@ async fn test_join_queue_join_next_with_id() {
...
+        let _ = rx3.await;
+    });
+
+    assert_eq!(queue.len(), 3);
+    assert!(queue.try_join_next().is_none());
+    assert_eq!(queue.len(), 3);
+    assert!(queue.try_join_next().is_none());
+    assert_eq!(queue.len(), 3);
+    assert!(queue.try_join_next().is_none());
+    assert_eq!(queue.len(), 3);
```
ADD-SP: Emm, I guess we repeated it three times because of the length of `JoinQueue` is 3. I think only one time is enough, we don't want to introduce too much complexity into the test set.

tokio-util/tests/task_join_queue.rs
```diff
@@ -221,3 +221,61 @@ async fn test_join_queue_join_next_with_id() {
...
+    tx1.send(()).unwrap();
+    tokio::task::yield_now().await;
+
+    assert_eq!(queue.len(), 3);
+    assert!(queue.try_join_next().is_some());
+    assert_eq!(queue.len(), 2);
+    assert!(queue.try_join_next().is_none());
+    assert_eq!(queue.len(), 2);
+    assert!(queue.try_join_next().is_none());
+    assert_eq!(queue.len(), 2);
```
ADD-SP: Emm, why did we need this?

tokio-util/tests/task_join_queue.rs
```diff
@@ -221,3 +221,61 @@ async fn test_join_queue_join_next_with_id() {
...
+
+    tx3.send(()).unwrap();
+    tokio::task::yield_now().await;
+
+    assert_eq!(queue.len(), 2);
+    assert!(queue.try_join_next().is_none());
+    assert_eq!(queue.len(), 2);
+    assert!(queue.try_join_next().is_none());
+    assert_eq!(queue.len(), 2);
```
ADD-SP: Why did we need this?

tokio-util/tests/task_join_queue.rs
```diff
@@ -221,3 +221,61 @@ async fn test_join_queue_join_next_with_id() {
...
+
+    assert_eq!(queue.len(), 2);
+    assert!(queue.try_join_next().is_some());
+    assert_eq!(queue.len(), 1);
+    assert!(queue.try_join_next().is_some());
+    assert!(queue.is_empty());
+    assert!(queue.try_join_next().is_none());
+    assert!(queue.is_empty());
+    assert!(queue.try_join_next().is_none());
+    assert!(queue.is_empty());
```
ADD-SP: Why did we need this?

tokio-util/src/task/join_queue.rs
```diff
@@ -183,6 +184,46 @@ impl<T> JoinQueue<T> {
...
+    /// Tries to join the next task in FIFO order if it has completed.
+    ///
+    /// Returns `None` if the queue is empty or if the next task is not yet ready.
+    pub fn try_join_next(&mut self) -> Option<Result<T, JoinError>> {
+        if self.0.front()?.is_finished() {
+            // Use `detach` to avoid calling `abort` on a task that has already completed.
+            // Dropping `AbortOnDropHandle` would abort the task, but since it is finished,
+            // we only need to drop the `JoinHandle` for cleanup.
+            let jh = self.0.pop_front().unwrap().detach();
+            let res = jh.now_or_never().unwrap();
```
ADD-SP: A short comment here for `unwrap` would be better.

tokio-util/src/task/join_queue.rs
```diff
@@ -183,6 +184,46 @@ impl<T> JoinQueue<T> {
...
+    /// [task ID]: tokio::task::Id
+    /// [`JoinError::id`]: fn@tokio::task::JoinError::id
+    pub fn try_join_next_with_id(&mut self) -> Option<Result<(Id, T), JoinError>> {
+        if self.0.front()?.is_finished() {
+            // Use `detach` to avoid calling `abort` on a task that has already completed.
+            // Dropping `AbortOnDropHandle` would abort the task, but since it is finished,
+            // we only need to drop the `JoinHandle` for cleanup.
+            let jh = self.0.pop_front().unwrap().detach();
+            let id = jh.id();
+            let res = jh.now_or_never().unwrap();
```
ADD-SP: A short comment here for `unwrap` would be better.

tokio-util/tests/task_join_queue.rs
nickkuk: Of course, from the outside we know that `try_join_next` is idempotent when nothing changes in the test circumstances. But it’s exactly this property that is being tested by repeating the same checks several times.

tokio-util/tests/task_join_queue.rs
nickkuk: https://github.com/tokio-rs/tokio/pull/7636#discussion_r2366149838

tokio-util/tests/task_join_queue.rs
nickkuk: https://github.com/tokio-rs/tokio/pull/7636#discussion_r2366149838

tokio-util/tests/task_join_queue.rs
nickkuk: https://github.com/tokio-rs/tokio/pull/7636#discussion_r2366149838

tokio-util/tests/task_join_queue.rs
ADD-SP: This looks reasonable, could you add a short comment on repeated code or even use `for` loop to make the purpose more clear?

nickkuk:
@ADD-SP I added test for `try_join_next_with_id`, but something wrong happening now: for pushed 130 tasks last three report that they are finished, but `jh.now_or_never()` returns `None`:
```rust
/// Temporary function to debug the state of all tasks in the queue.
pub fn dbg_state(&mut self)
where
    T: std::fmt::Debug,
{
    let mut counter = 0;
    for jh in &mut self.0 {
        let finished = jh.is_finished();
        let res = jh.now_or_never();
        println!("task {counter}, finished: {finished}, result: {res:?}");
        counter += 1;
    }
}
```

Output:
```sh
...
task 124, finished: true, result: Some(Ok(()))
task 125, finished: true, result: Some(Ok(()))
task 126, finished: true, result: Some(Ok(()))
task 127, finished: true, result: None
task 128, finished: true, result: None
task 129, finished: true, result: None
```

ADD-SP:
> @ADD-SP I added test for `try_join_next_with_id`, but something wrong happening now: for pushed 130 tasks last three report that they are finished, but `jh.now_or_never()` returns `None`:
> 
> ```rust
> /// Temporary function to debug the state of all tasks in the queue.
> pub fn dbg_state(&mut self)
> where
>     T: std::fmt::Debug,
> {
>     let mut counter = 0;
>     for jh in &mut self.0 {
>         let finished = jh.is_finished();
>         let res = jh.now_or_never();
>         println!("task {counter}, finished: {finished}, result: {res:?}");
>         counter += 1;
>     }
> }
> ```
> 
> Output:
> 
> ```shell
> ...
> task 124, finished: true, result: Some(Ok(()))
> task 125, finished: true, result: Some(Ok(()))
> task 126, finished: true, result: Some(Ok(()))
> task 127, finished: true, result: None
> task 128, finished: true, result: None
> task 129, finished: true, result: None
> ```

It appears to be a race condition. You can try calling `tokio::task::yield_now()` multiple times.

nickkuk:
> It appears to be a race condition. You can try calling tokio::task::yield_now() multiple times.

```rust
for _ in 0..1000000 {
}
```
changes it to
```sh
...
task 124, finished: true, result: Some(Ok(()))
task 125, finished: true, result: Some(Ok(()))
task 126, finished: true, result: Some(Ok(()))
task 127, finished: true, result: Some(Ok(()))
task 128, finished: true, result: None
task 129, finished: true, result: None
```

But the real problem is that for `try_join_next[_with_id]` methods we can no longer guarantee that combination of checking `jh.is_finished()` and calling `jh.now_or_never().unwrap()` works.

ADD-SP:
> > It appears to be a race condition. You can try calling tokio::task::yield_now() multiple times.
> 
> ```rust
> for _ in 0..1000000 {
>     tokio::task::yield_now().await;
> }
> ```
> 
> changes it to
> 
> ```shell
> ...
> task 124, finished: true, result: Some(Ok(()))
> task 125, finished: true, result: Some(Ok(()))
> task 126, finished: true, result: Some(Ok(()))
> task 127, finished: true, result: Some(Ok(()))
> task 128, finished: true, result: None
> task 129, finished: true, result: None
> ```
> 
> But the real problem is that for `try_join_next[_with_id]` methods we can no longer guarantee that combination of checking `jh.is_finished()` and calling `jh.now_or_never().unwrap()` works.

Emm, maybe caused by the cooperative scheduling. Could you try <https://docs.rs/tokio/1.47.1/tokio/task/coop/fn.unconstrained.html>?

If this fix the problem, then we need to fix the `try_join_next` and `try_join_next_with_id` method (try to avoid cooperative scheduling internally?).

nickkuk:
@ADD-SP Thank you, everything works with `unconstrained`! Also for just in case I decided to remove unwrapping of `now_or_never()` and return `None` in this case too.

tokio-util/src/task/join_queue.rs
```diff
@@ -183,6 +184,55 @@ impl<T> JoinQueue<T> {
...
+    /// Tries to join the next task in FIFO order if it has completed.
+    ///
+    /// Returns `None` if the queue is empty or if the next task is not yet ready.
+    pub fn try_join_next(&mut self) -> Option<Result<T, JoinError>> {
+        let jh = self.0.front_mut()?;
+        if jh.is_finished() {
+            // Since this function is not async and cannot be forced to yield, we should
+            // disable budgeting when we want to check for the `JoinHandle` readiness.
+            let jh = tokio::task::coop::unconstrained(jh);
+            let res = jh.now_or_never()?;
```
ADD-SP: Emm, why should we return `None` here?

tokio-util/tests/task_join_queue.rs
```diff
@@ -192,23 +192,119 @@ async fn test_join_queue_join_next_with_id() {
...
+            Err(err) => panic!("failed: {err}"),
+        }
+    }
+
+    assert_eq!(count, TASK_NUM);
+    assert_eq!(joined, spawned);
+}
+
+#[tokio::test]
+async fn test_join_queue_try_join_next() {
```
ADD-SP, line 226: Could you also test the to ensure the `try_join_next` has disabled the cooperative scheduling?

tokio-util/tests/task_join_queue.rs
nickkuk: Sure! I've just added `check_try_join_next_is_noop` sub-function to check exactly this.

tokio-util/tests/task_join_queue.rs
```diff
@@ -208,7 +208,97 @@ async fn test_join_queue_join_next_with_id() {
...
+    assert!(queue.try_join_next().is_some());
+    assert_eq!(queue.len(), 1);
+    assert!(queue.try_join_next().is_some());
+    assert!(queue.is_empty());
+    check_try_join_next_is_noop(&mut queue);
+}
+
+#[tokio::test]
+async fn test_join_queue_try_join_next_with_id() {
+    const TASK_NUM: u32 = 1000;
```
ADD-SP, line 321: It would be better adding an short comment here to explain how does cooperative scheduling affect this test.

tokio-util/src/task/join_queue.rs
nickkuk: In current version `jh` is not owned join handle, namely `jh: Unconstrained<&mut AbortOnDropHandle<T>>`. The method `now_or_never` tries to [take](https://github.com/tokio-rs/tokio/blob/1b98d5ad85f44a2d2df91b520990ded04db81371/tokio/src/runtime/task/harness.rs#L283) result from it once and either success or fail. In either case it is not dropped on this line 196 and can be polled once more (with possible [panic](https://github.com/tokio-rs/tokio/blob/1b98d5ad85f44a2d2df91b520990ded04db81371/tokio/src/runtime/task/core.rs#L412) if we try to take result more than once).

In successfull case I'm removing from queue and dropping this `JoinHandle` on the next line. As for fail case, I thought it is impossible (after `jh.is_finished() == true`) and used `unwrap()` before, but it appears to be not the case without using of `unconstrained`. So, nothing wrong here if `now_or_never()` fails, it is just additional protection.

nickkuk:
@ADD-SP I added another test for disabled coop scheduling, similar that I found for `JoinSet`.

tokio-util/src/task/join_queue.rs
nickkuk: Do you suggest me to return back `jh.now_or_never().unwrap()` here with a comment about jh completeness?

ADD-SP: I'm not entirely sure if this scenario exists: when `jh` has finished and cooperative scheduling is disabled, `jh.poll` still returns a `Pending`.

If this should not happen, we should panic here. Otherwise, we might need to tolerate it.

nickkuk: > I'm not entirely sure if this scenario exists: when jh has finished and cooperative scheduling is disabled, jh.poll still returns a Pending.

Could we consider such scenario as bug in `tokio` + `tokio-util` combo? In this case panic is more suitable, I can even make something like `.expect("Tokio bug happened, please open an issue")`.

ADD-SP: > Could we consider such scenario as bug in `tokio` + `tokio-util` combo?

For the existing implementation of tokio, we can `.unwrap()`, but I personally don't want to make this guarantee to `JoinHandle` and `AbortOnDropHandle`, so I'm re-thinking the no-op waker solution.


> After look at it more, I think that additional is_finished should be used anyway because for not completed future polling [tries to set a waker](https://github.com/tokio-rs/tokio/blob/1b98d5ad85f44a2d2df91b520990ded04db81371/tokio/src/runtime/task/harness.rs#L427), which can break further internal Tokio logic.

Emm, why does this break Tokio's internal logic? The waker here is not for the future tracked by the `JoinHandle`, but rather the waker of the `JoinHandle` itself. Would this actually cause issues?

From a semantic standpoint, the `JoinHandle` behaves like [`std::thread::JoinHandle`]: it never drives the task it is tracking. Therefore, polling this kind of task tracker with a no-op waker should not affect the tracked task.


[`std::thread::JoinHandle`]: https://doc.rust-lang.org/1.90.0/std/thread/struct.JoinHandle.html

tokio-util/src/task/join_queue.rs
```diff
@@ -183,6 +184,47 @@ impl<T> JoinQueue<T> {
...
 
+    /// Tries to join the next task in FIFO order if it has completed.
+    ///
+    /// Returns `None` if the queue is empty or if the next task is not yet ready.
+    pub fn try_join_next(&mut self) -> Option<Result<T, JoinError>> {
+        let jh = self.0.front_mut()?;
+        // Since this function is not async and cannot be forced to yield, we should
+        // disable budgeting when we want to check for the `JoinHandle` readiness.
+        let jh = tokio::task::coop::unconstrained(jh);
+        let res = jh.now_or_never()?;
```
ADD-SP: nit, the semantic of `now_or_never` is a bit confusing here, could we write an helper function and it internally uses the noop_waker?

tokio-util/src/task/join_queue.rs
nickkuk: I pushed another commit where I removed `jh.is_finished()` check and leave only `jh.now_or_never()?` (aka no-op waker).

Tests seems to be fine. Do you know if I can test something more?

tokio-util/tests/task_join_queue.rs
```diff
@@ -208,7 +208,137 @@ async fn test_join_queue_join_next_with_id() {
...
+    check_try_join_next_is_noop(&mut queue);
+}
+
+#[tokio::test]
+async fn test_join_queue_try_join_next_disabled_coop() {
+    // This number is large enough to trigger coop. Without using `tokio::task::coop::unconstrained`
+    // inside `try_join_next` this test fails on `assert!(coop_count == 0)`.
+    const TASK_NUM: u32 = 1000;
+
+    static SEM: tokio::sync::Semaphore = tokio::sync::Semaphore::const_new(0);
```
ADD-SP: Will `Arc` work? I'd like to avoid `static` in tests because a `static` var is a global state.

ADD-SP: Follow up question, I saw the `test_join_queue_try_join_next_with_id` ues the `watch` channel to achieve the same work, I'm ok for both, but we'd better using the same approach to reduce the mental burden.

tokio-util/tests/task_join_queue.rs
```diff
@@ -208,7 +208,137 @@ async fn test_join_queue_join_next_with_id() {
...
+            SEM.add_permits(1);
+        });
+    }
+
+    let _ = SEM.acquire_many(TASK_NUM).await.unwrap();
+
+    let mut count = 0;
+    let mut coop_count = 0;
+    while !queue.is_empty() {
+        match queue.try_join_next() {
```
ADD-SP, line 302: Emm, why not `while let Some`? I guess this is because you want to collect the `coop_count`, if so, shall we do the same thing for `test_join_queue_try_join_next_with_id`?

tokio-util/tests/task_join_queue.rs
```diff
@@ -208,7 +208,137 @@ async fn test_join_queue_join_next_with_id() {
...
+            }
+        }
+        count += 1;
+    }
+    assert!(coop_count == 0);
+    assert_eq!(count, TASK_NUM);
+}
+
+#[tokio::test]
+async fn test_join_queue_try_join_next_with_id() {
```
ADD-SP:
```suggestion
async fn test_join_queue_try_join_next_with_id_disabled_coop() {
```

tokio-util/src/task/join_queue.rs
nickkuk: I've added `try_poll_handle` helper. Looks like we need something like this in `JoinHandle` itself. But I don't like that while it uses handle by `&mut`, it makes handle unpollable in the future.

On the other hand, such function can consume handle and return itself if it is not ready:
```rust
impl JoinHandle<T> {
    pub fn try_into_result(self) -> Result<Result<T, JoinError>, Self> { ... }
}
```
but it is not very good for `JoinQueue` because we have to `pop_front` then `push_front` in pending case.

ADD-SP: I personally think the `try_poll_handle` is enough for both correctness and semantic. Since we have solved this problem using `try_poll_handle`, we don't need the `JoinHandle::try_into_result` for now, unless someone asks this with another reasonable usecase.

tokio-util/src/task/join_queue.rs
ADD-SP: It looks good.

tokio-util/tests/task_join_queue.rs
```diff
@@ -208,7 +208,137 @@ async fn test_join_queue_join_next_with_id() {
...
+    let (tx2, rx2) = oneshot::channel::<()>();
+    queue.spawn(async {
+        let _ = rx2.await;
+    });
+    let (tx3, rx3) = oneshot::channel::<()>();
+    queue.spawn(async {
+        let _ = rx3.await;
+    });
+
+    fn check_try_join_next_is_noop(queue: &mut JoinQueue<()>) {
```
ADD-SP, line 243: Could you add a short comment on this function to explain that this is to test the idempotent of `try_join_next`.

ADD-SP (commented):
I think the implementation part looks good, I left some comments on tests.

tokio-util/tests/task_join_queue.rs
nickkuk: > I'm ok for both, but we'd better using the same approach to reduce the mental burden

You are right that these two approaches finally solves the same problem, but on another hand, I was happy to have two a bit different tests when found them  side by side among `JoinSet` ones: [first](https://github.com/tokio-rs/tokio/blob/b48586f560062e5597f8ef7e9003a3b6cf3b2af0/tokio/tests/task_join_set.rs#L235) and [second](https://github.com/tokio-rs/tokio/blob/b48586f560062e5597f8ef7e9003a3b6cf3b2af0/tokio/tests/task_join_set.rs#L275).

If you really think that it is mental burden to have both approaches, I can definitely unify it.

ADD-SP: > If you really think that it is mental burden to have both approaches, I can definitely unify it.

I don't have strong feeling to unify it, feel free to keep it as is.

By the way, could we get rid of the static variable? We want to avoid introducing global state in test.

nickkuk: > By the way, could we get rid of the static variable? We want to avoid introducing global state in test.

Sure, I removed it in the latest commit.

ADD-SP (approved):
Thanks!

Issue #7637, opened 2025-09-20
`COUNT` constant in `join*!` macros can collide with user's constant used in futures


robinhundt:
**Version**

1.46

**Platform**
Linux nixos 6.12.45 #1-NixOS SMP PREEMPT_DYNAMIC Thu Sep  4 13:31:56 UTC 2025 x86_64 GNU/Linux

**Description**

The commit https://github.com/tokio-rs/tokio/commit/8e999e380643160f0e944fd5708de2bd3d1b0d51#diff-71d8f976620bd493969f9c0ecf5db22825ba06ca05acdcce17f400c5c6ae2c11 changed the scope of the `COUNT` constant of the `join!` and `try_join!` macros so that it is in the same scope as the expansion of the futures expressions `$e`. If a user of these macros has their own `COUNT` constant in scope, this leads to a compile error due to mismatching types (`usize` and `u32` in my case) or in the worst case to a change in behavior if the types are compatible but the values are different ([repro](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=337c32f36b04ebc70370b1cf4ee09515)).

Apparently macros are not hygienic for items, which includes consts... :(

**I tried this code**:

In one of my projects I have a few test cases which look like this:
```rust
    #[tokio::test]
    async fn test_extension() {
        let _g = init_tracing();
        const COUNT: usize = 2 * DEFAULT_OT_BATCH_SIZE;
        let (c1, c2) = local_conn().await.unwrap();
        let rng1 = StdRng::seed_from_u64(42);
        let mut rng2 = StdRng::seed_from_u64(24);
        let choices = random_choices(COUNT, &mut rng2);
        let mut sender = SemiHonestOtExtensionSender::new_with_rng(c1, rng1);
        let mut receiver = SemiHonestOtExtensionReceiver::new_with_rng(c2, rng2);
        let (send_ots, recv_ots) =
            tokio::try_join!(sender.send(COUNT), receiver.receive(&choices)).unwrap();
        for ((r, s), c) in recv_ots.into_iter().zip(send_ots).zip(choices) {
            assert_eq!(r, s[c.unwrap_u8() as usize]);
        }
    }
```
[source](https://github.com/robinhundt/CryProt/blob/cde3e70e58445370e4da941d4b9fa3c7aaff3f5d/cryprot-ot/src/extension.rs#L798-L813)

**I expected to see this happen**:

The code should still compile after a minor version upgrade from 1.45 to 1.46.

[...]

PR #7638, opened 2025-09-20
 fix(macros/(try_)join): don't leak const COUNT in macro expansion

jlizen:
Closes: #7637

## Motivation

#7307 added a `biased` mode to `join!` and `try_join!`. However, it did so in a way that leaked the `const COUNT: u32` into the outer scope of the macro expansion. This resulted in it shadowing any caller-specified `const COUNT`s in the futures passed into the macro call - which either caused a compiler failure in case of type mismatch, or the variable constants being overwritten.

This change fixes this by refactoring to keep the macro expansion's declaration of `const COUNT` within its `poll_fn` body. This restores the original scoping prior to the `biased` feature.

## Solution
This was a bit fiddly since we want to use a const generic with our `Rotator` helper to avoid increasing the memory footprint of the `*join` future, but the rotator is instantiated in the outer scope. At the same time, we don't want to just use an obscure variable name to reduce the chance of collision.

@robinhundt [had a nice suggestion](https://github.com/tokio-rs/tokio/issues/7637#issuecomment-3315239456) to instead use a helper trait with an associated type, along with some marker structs per rotator type. This allows us to signal whether we are in biased mode or not in our tt muncher expansions, but defer directly naming the `const COUNT` generic due to the additional indirection layer. We thus pass in the literal `$($total:tt)` expression as our generic in our ultimate macro expansion and don't need the `const COUNT` declaration in outer scope.

This preserves the reduced memory footprint while also sidestepping hygiene issues. It technically does result in us calculating the `0 + 1 + 2`... expression an additional time (at the rotator declaration and also at the `const COUNT` declaration inside `poll_fn` scope, though I would hope that the compiler will optimize that away. Anyway it's a cheap, one-time cost on initialization and less important than memory footprint.

The main tradeoff is that this is a bit more complicated, but not actually any less readable macro impl code. And then, this is not an implementation detail that concerns the end user.

## Testing

I structured this PR as two atomic commits.

The first commit demonstrates the regression with tests that pass by asserting that the macro hygiene issue indeed is shadowing the user-provided generic.

The second commit implements the solution described above, and flips the tests over to asserting that the user-provided value is respected.


jlizen:
cc @ADD-SP whenever you have a moment, thanks!

jlizen:
Appreciate the eyes!

PR #7639, opened 2025-09-21
io: fix the hang-forever issue of `Mock` when the write buf is empty

ADD-SP:
The `tokio_test::io::Mock` is buggy, I try to improve it without breaking change.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -6,6 +6,8 @@ env:
   RUST_STABLE: stable
   RUST_NIGHTLY: nightly-2025-01-25
   RUSTFLAGS: -D warnings
+  # This excludes unstable features like io_uring, which require '--cfg tokio_unstable'.
+  TOKIO_STABLE_FEATURES: full,test-util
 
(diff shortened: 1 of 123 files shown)


tokio-test/tests/io.rs
```diff
@@ -170,3 +176,84 @@ async fn multiple_wait() {
         start.elapsed().as_millis()
     );
 }
+
+// No matter which usecase, it doesn't make sense for a read
```
martin-g:
```suggestion
// No matter which usecase, it doesn't make sense for a read to
```

tokio-test/tests/io.rs
```diff
@@ -170,3 +176,84 @@ async fn multiple_wait() {
...
+}
+
+// The `Mock` must never panic, even if cloned multiple times.
+// However, at present, cloning the builder under certain
+// conditions causes a panic.
+//
+// Fixing this would require making `Mock` non-`Clone`,
+// which is a breaking change.
+//
+// Since we want be aware of the fixing of this bug,
```
martin-g:
```suggestion
// Since we want to be aware of the fixing of this bug,
```

PR #7640, opened 2025-09-21
[WIP] io: add `shutdown()` and `shutdown_error()` on `tokio_test::io::Builder`

ADD-SP:
blocked by <https://github.com/tokio-rs/tokio/pull/7639>.


ADD-SP:
I closed this PR. As you can see, this feature even breaks tests of the tokio itself, so it is very like a breaking change for downstream.

PR #7661, opened 2025-10-04
sync: improve unbounded send method docs

mattiapitossi:
## Motivation

While trying to understand how the send for unbounded channel works, I wasn't sure why an unbounded channel never requires any form of waiting. I found a similar explanation in the sync's mod.rs that explains it well, so I was wondering if it's worth adding it also within the send docs.

## Solution

Adds explanation on why an unbounded channel never requires any form of waiting.


ADD-SP (approved):

ADD-SP:
Thanks!

PR #7666, opened 2025-10-09
task: Add LocalKey::try_get

tisonkun:
TRIVIAL AS IS. Align with `with` and `try_with`.

### DIFF
--- a/tokio/tests/task_local.rs
+++ b/tokio/tests/task_local.rs
@@ -145,3 +145,27 @@ async fn poll_after_take_value_should_fail() {
     // Poll the future after `take_value` has been called
     fut.await;
 }
+
+#[tokio::test]
+async fn get_value() {
+    tokio::task_local! {
+        static KEY: u32
+    }
+
+    KEY.scope(1, async {
+        assert_eq!(KEY.get(), 1);
+        assert_eq!(KEY.try_get().unwrap(), 1);
+    })
+    .await;
+
+    let fut = KEY.scope(1, async {
+        let result = KEY.try_get();
+        // The task local value no longer exists.
+        assert!(result.is_err());
+    });
+    let mut fut = Box::pin(fut);
+    fut.as_mut().take_value();
+
+    // Poll the future after `take_value` has been called
+    fut.await;
+}


tokio/src/task/task_local.rs
```diff
@@ -275,6 +275,17 @@ impl<T: Clone + 'static> LocalKey<T> {
...
         self.with(|v| v.clone())
     }
+
+    /// Returns a copy of the task-local value
+    /// if the task-local value implements `Clone`.
+    ///
+    /// If the task-local with the associated key is not present, this
+    /// method will return an `AccessError`. For a panicking variant,
+    /// see `get`.
+    #[track_caller]
```
ADD-SP: Why `#[track_caller]` here?

tisonkun: @ADD-SP This follows how `get` is implemented and provide a `try_` variant.

If this is undesired, the reason to have a `get` that forwards to `self.with(|v| v.clone())` doesn't stand also.

ADD-SP: > If this is undesired, the reason to have a get that forwards to self.with(|v| v.clone()) doesn't stand also.

Sorry, I don't understand. I might need to rephrase my question: What issues would we encounter without `#[track_caller]`?

tisonkun: This follows what `LocalKey::get` does.

With a `git blame` it points out to https://github.com/tokio-rs/tokio/pull/4848 that could give more background.

tisonkun:
> A follow-up question, what's the usecase of this feature?

This is answered in https://github.com/tokio-rs/tokio/pull/7666#discussion_r2415372173.

tokio/src/task/task_local.rs
ADD-SP: For `LocalKey::get` the `#[track_caller]` is used for reporting the location of caller. For example.

```rust
fn foo() {
    panic!("foo");
}

fn main() {
    foo();
}
```

Without the `#[track_caller]`, the error message points to the `fn foo` instead of `fn main`. Apparently, pointing to `fn main` is more friendly.

```
thread 'main' panicked at src/main.rs:2:5:
foo
```

However, if we add the `#[track_caller]` to the `fn foo`.

```rust
#[track_caller]
fn foo() {
    panic!("foo");
}

fn main() {
    foo();
}
```

```
thread 'main' panicked at src/main.rs:7:5:
foo
```

The error message points to the `fn main`, which helps downstream to troubleshoot their code.

ADD-SP: This is why I asked the reason why we need the `#[track_caller]` here. The `try_get` doesn't panic, why do we still need the `#[track_caller]`? Did I miss another reason?

tisonkun: I get the point now. Yes we can remove `#[track_caller]` here.

tisonkun:
```suggestion
```

tisonkun:
> Could you add tests for this new method? I think `tokio/tests/task_local.rs` is a good place.

Added at 5c7114e90e4333ff3c2f5778432ab4f75d6c7d5b

PR #7694, opened 2025-10-18
task: add tests for `spawn_local` in panic scenarios

ADD-SP:
Follow up of #7562

### DIFF
--- a/tokio/tests/task_builder.rs
+++ b/tokio/tests/task_builder.rs
@@ -1,12 +1,9 @@
 #![cfg(all(tokio_unstable, feature = "tracing"))]
 
 use std::rc::Rc;
-use tokio::{
-    task::{Builder, LocalSet},
-    test,
-};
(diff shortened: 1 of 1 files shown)


tokio/tests/task_join_set.rs
```diff
@@ -407,6 +407,24 @@ async fn try_join_next_with_id() {
 mod spawn_local {
     use super::*;
 
+    #[test]
+    #[should_panic(
+        expected = "`spawn_local` called from outside of a `task::LocalSet` or `runtime::LocalRuntime`"
+    )]
+    fn panic_outside_any_runtime() {
+        let mut set = JoinSet::new();
```
mattiapitossi, line 415: I think these tests are also missing for `JoinQueue`

ADD-SP: thanks! added

PR #7696, opened 2025-10-19
fs: support io_uring with `tokio::fs::read`

Daksh14:
## Motivation

We slowly use io uring everywhere here I made the simple change of supporting `fs::read` however it might not be as simple. Let me know if the unsafe I used is correct or not. 

We currently use the blocking `std::fs::MetaData` to obtain file size for buffer capacity and extend the length of the vector according to the bytes read in the CQE. This implementation sounds good on paper to me.

Later we should implement an internal statx helper, in this PR or a seperate PR to make our uring implementation less painful to use. As this pr failed https://github.com/tokio-rs/tokio/pull/7616 

Lets put statx helper in different PR to avoid merging an inefficient read implementation given io uring is about being more efficient in file IO

## Solution

Continue adopting io uring 

strace on a `tokio::fs::read` after this change
```
io_uring_setup(256, {flags=0, sq_thread_cpu=0, sq_thread_idle=0, sq_entries=256, cq_entries=512, features=IORING_FEAT_SINGLE_MMAP|IORING_FEAT_NODROP|IORING_FEAT_SUBMIT_STABL
E|IORING_FEAT_RW_CUR_POS|IORING_FEAT_CUR_PERSONALITY|IORING_FEAT_FAST_POLL|IORING_FEAT_POLL_32BITS|IORING_FEAT_SQPOLL_NONFIXED|IORING_FEAT_EXT_ARG|IORING_FEAT_NATIVE_WORKERS
|IORING_FEAT_RSRC_TAGS|IORING_FEAT_CQE_SKIP|IORING_FEAT_LINKED_FILE|IORING_FEAT_REG_REG_RING|IORING_FEAT_RECVSEND_BUNDLE|IORING_FEAT_MIN_TIMEOUT|IORING_FEAT_RW_ATTR, sq_off=
{head=0, tail=4, ring_mask=16, ring_entries=24, flags=36, dropped=32, array=8256, user_addr=0}, cq_off={head=8, tail=12, ring_mask=20, ring_entries=28, overflow=44, cqes=64,
 flags=40, user_addr=0}}) = 9
mmap(NULL, 16384, PROT_READ|PROT_WRITE, MAP_SHARED|MAP_POPULATE, 9, 0x10000000) = 0xfaf0bf1e2000
mmap(NULL, 9280, PROT_READ|PROT_WRITE, MAP_SHARED|MAP_POPULATE, 9, 0) = 0xfaf0be71d000
epoll_ctl(5, EPOLL_CTL_ADD, 9, {events=EPOLLIN|EPOLLRDHUP|EPOLLET, data=0}) = 0
io_uring_enter(9, 1, 0, 0, NULL, 128)   = 1
futex(0xfaf0bf2557f0, FUTEX_WAIT_PRIVATE, 1, NULL) = 0
mmap(NULL, 2162688, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_STACK, -1, 0) = 0xfaf0be50d000
mprotect(0xfaf0be51d000, 2097152, PROT_READ|PROT_WRITE) = 0
rt_sigprocmask(SIG_BLOCK, ~[], [], 8)   = 0
clone3({flags=CLONE_VM|CLONE_FS|CLONE_FILES|CLONE_SIGHAND|CLONE_THREAD|CLONE_SYSVSEM|CLONE_SETTLS|CLONE_PARENT_SETTID|CLONE_CHILD_CLEARTID, child_tid=0xfaf0be71c150, parent_
tid=0xfaf0be71c150, exit_signal=0, stack=0xfaf0be50d000, stack_size=0x20e960, tls=0xfaf0be71c7a0} => {parent_tid=[746758]}, 88) = 746758
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
futex(0xfaf0be71c850, FUTEX_WAKE_PRIVATE, 1) = 1
io_uring_enter(9, 1, 0, 0, NULL, 128)   = 1
futex(0xfaf0bf2557f0, FUTEX_WAIT_PRIVATE, 1, NULL) = -1 EAGAIN (Resource temporarily unavailable)
close(10)                               = 0
```

### DIFF
--- a/tokio/src/fs/read_uring.rs
+++ b/tokio/src/fs/read_uring.rs
@@ -106,10 +106,9 @@ async fn small_probe_read(
     Ok((r_fd, r_buf, is_eof))
 }
 
-// Takes a amount of length to read and reads until we exhaust the given length
-// to read or EOF reached.
+// Takes a amount of length to read and returns a singluar read in the buffer
 //
-// Returns size_read, the result and EOF reached or not
+// Returns the file descriptor, buffer and EOF reached or not
 async fn op_read(
     mut fd: OwnedFd,
     mut buf: Vec<u8>,
--- a/tokio/src/io/uring/read.rs
+++ b/tokio/src/io/uring/read.rs
@@ -38,7 +38,7 @@ impl Cancellable for Read {
 
 impl Op<Read> {
     // Submit a request to read a FD at given length and offset into a
-    // dynamic buffer with uninitialized memory. The read happens on unitialized
+    // dynamic buffer with uninitialized memory. The read happens on uninitialized
     // buffer and no overwriting happens.
 
     // SAFETY: The `len` of the amount to be read and the buffer that is passed


tokio/src/fs/read.rs
```diff
@@ -45,5 +45,58 @@ use std::{io, path::Path};
...
+        .unwrap_or(0);
+
+    let buf = Vec::with_capacity(size);
+
+    let fd: OwnedFd = file
+        .try_into_std()
+        .expect("unexpected in-flight operation detected")
+        .into();
+
+    let (read_size, mut buf) = Op::read(fd, buf)?.await?;
```
Darksonn: Like `fs::write`, you have to call it in a loop. It's not guaranteed to give you all of the data in a single read.

tokio/src/fs/read.rs
```diff
@@ -45,5 +45,58 @@ use std::{io, path::Path};
...
+
+    // SAFETY:
+    // 1. The buffer is initialized with `size` capacity
```
ADD-SP: Just a reminder: it’s possible for another process to append data to the same file concurrently, so we can’t assume that the final `read_size` equals the `size` reported by `fs.metadata()` earlier.

tokio/src/io/uring/read.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+impl Op<Read> {
+    /// Submit a request to open a file.
+    pub(crate) fn read(fd: OwnedFd, mut buf: Vec<u8>) -> io::Result<Self> {
+        let buf_mut_ptr = buf.as_mut_ptr();
+        let cap = buf.capacity() as _;
+
+        let read_op = opcode::Read::new(types::Fd(fd.as_raw_fd()), buf_mut_ptr, cap).build();
```
ADD-SP:
```rust
// tokio/src/fs/read.rs | async fn read_uring(..)
    let size = file
        .metadata()
        .await
        .map(|m| m.len() as usize)
        .ok()
        .unwrap_or(0);
```

What will happen if the `cap` is zero?

tokio/src/fs/read.rs
```diff
@@ -45,5 +45,75 @@ use std::{io, path::Path};
...
+
+    let mut fd: OwnedFd = file
+        .try_into_std()
+        .expect("unexpected in-flight operation detected")
+        .into();
+
+    let mut size_read = 0;
+    let mut offset = 0;
+
+    while size_read < size {
```
ADD-SP: This `while` loop is problematic if the `size` is zero due to `unwrap_or(0)`.

tokio/src/io/uring/read.rs
```diff
@@ -0,0 +1,42 @@
...
+use io_uring::{opcode, types};
+use std::{
+    io,
+    os::fd::{AsRawFd, OwnedFd},
+};
+
+#[derive(Debug)]
+pub(crate) struct Read {
+    fd: OwnedFd,
+}
```
Darksonn, line 11: You must transfer ownership of the buffer to the `Read` operation. Passing a `&mut Vec<u8>` is not enough.

tokio/src/fs/read.rs
```diff
@@ -45,5 +45,75 @@ use std::{io, path::Path};
...
+    use std::os::fd::OwnedFd;
+
+    let file = OpenOptions::new().read(true).open(path).await?;
+
+    let size = file
+        .metadata()
+        .await
+        .map(|m| m.len() as usize)
+        .ok()
+        .unwrap_or(0);
```
Darksonn: What does the stdlib `std::read` do in this scenario?

tokio/src/fs/read.rs
Darksonn: Yeah, you have to keep reading until you get a length of size zero from the read operation.

tokio/src/fs/read.rs
```diff
@@ -45,5 +45,75 @@ use std::{io, path::Path};
...
+
+        size_read += _size_read as usize;
+        offset += _size_read as u64;
+
+        fd = _fd;
+    }
+
+    // SAFETY:
+    // 1. The buffer is initialized with `size` capacity
+    // 2. the read_size is the number of bytes read from the file
```
martin-augment:
```suggestion
    // 2. the size_read is the number of bytes read from the file
```

tokio/src/fs/read.rs
```diff
@@ -45,5 +45,75 @@ use std::{io, path::Path};
...
+    let mut fd: OwnedFd = file
+        .try_into_std()
+        .expect("unexpected in-flight operation detected")
+        .into();
+
+    let mut size_read = 0;
+    let mut offset = 0;
+
+    while size_read < size {
+        let left_to_read = (size - size_read) as u32;
```
martin-augment:
```suggestion
        let left_to_read = u32::try_from(size - size_read).unwrap_or(u32::MAX);
```
To properly support files bigger than 4GB.

tokio/src/fs/read.rs
```diff
@@ -45,5 +45,75 @@ use std::{io, path::Path};
...
+        .await
+        .map(|m| m.len() as usize)
+        .ok()
+        .unwrap_or(0);
+
+    let mut buf = Vec::with_capacity(size);
+
+    let mut fd: OwnedFd = file
+        .try_into_std()
+        .expect("unexpected in-flight operation detected")
```
martin-augment: The method returns a Result, so let's return an Err instead of panic-ing.

tokio/src/io/uring/read.rs
```diff
@@ -0,0 +1,42 @@
...
+
+impl Cancellable for Read {
+    fn cancel(self) -> CancelData {
+        CancelData::Read(self)
+    }
+}
+
+impl Op<Read> {
+    /// Submit a request to open a file
```
martin-augment:
```suggestion
    /// Submit a request to read from a file
```

tokio/src/fs/read.rs
```diff
@@ -45,5 +45,75 @@ use std::{io, path::Path};
...
+        size_read += _size_read as usize;
+        offset += _size_read as u64;
+
+        fd = _fd;
+    }
+
+    // SAFETY:
+    // 1. The buffer is initialized with `size` capacity
+    // 2. the read_size is the number of bytes read from the file
+    unsafe { buf.set_len(size_read as _) };
```
martin-augment: Is `as _` needed here ?
`size_read` is `usize` and `set_len()` expects a `usize`

tokio/src/fs/read.rs
```diff
@@ -45,5 +45,75 @@ use std::{io, path::Path};
...
+        .expect("unexpected in-flight operation detected")
+        .into();
+
+    let mut size_read = 0;
+    let mut offset = 0;
+
+    while size_read < size {
+        let left_to_read = (size - size_read) as u32;
+
+        let (_size_read, _fd) = Op::read(fd, &mut buf, left_to_read, offset)?.await?;
```
martin-augment: The leading underscores are usually used for `unused` variables.
Maybe rename those to `bytes_read` and `returned_fd` ?!

tokio/src/fs/read.rs
martin-augment: I agree that falling back to 0 seems wrong here - ` while size_read < size` won't enter at all.
IIRC special files like `/dev/stdin`, `/proc/**`, pipes, sockets will always return 0.
IMO the `size` should default to something bigger than 0 but small enough to not allocate too much for the unknown cases. In addition the `size` should be used only for the Vec capacity but the logic for the reading should read until EOF.

tokio/src/fs/read.rs
ADD-SP: I think we should panic here because this error should never happen unless there is a bug inside the tokio.

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,195 @@
...
+                    buf = r_buf;
+                    offset += size_read as u64;
+                    read_len -= size_read;
+
+                    // 1. In case of no size_hint and a large file, if we keep reading
+                    // PROBE_SIZE, we want to increment number of short reads in order to gradually
+                    // increase read size per Op submission
+                    // 2. In case of small reads by the kernel, also gradually increase
+                    // read size per Op submission to read files in lesser cycles
+                    if size_read_usize <= buf.len() {
```
martin-g: `buf.len` (`r_buf`) is updated at line 119 and it accumulates the old length + `size_read_usize`.
I think this `if` check will always pass.

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,195 @@
...
+        .await
+        .map(|m| usize::try_from(m.len()).unwrap_or(usize::MAX))
+        .ok();
+
+    let fd: OwnedFd = file
+        .try_into_std()
+        .expect("unexpected in-flight operation detected")
+        .into();
+
+    let buf = Vec::with_capacity(size_hint.unwrap_or(0) + 1);
```
martin-g: `size_hint` could be `usize::MAX` here (see line 25). Adding `1` more will lead to overflow.

Daksh14: @martin-g noted I'll use a `saturating_add` here

tokio/src/fs/read_uring.rs
Daksh14: I see, I'll change the `buf.len()` to the `read_len` to detect shorter reads, how else do you think I should detect shorter reads

martin-g: You could add something like `let requested = read_len;` anywhere before `read_len -= size_read;` and use `if size_read_usize < requested {` here.

tokio/tests/fs_uring_read.rs
```diff
@@ -0,0 +1,212 @@
...
+    tmp.write_all(&buf).unwrap();
+    let path = tmp.path().to_path_buf();
+
+    (tmp, path)
+}
+
+fn create_buf(length: usize) -> Vec<u8> {
+    let mut buf = Vec::with_capacity(length);
+
+    for i in 0..buf.capacity() {
```
martin-g:
```suggestion
    for i in 0..length {
```
See https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.with_capacity
```rust
let mut vec = Vec::with_capacity(10);

// The vector contains no items, even though it has capacity for more
assert_eq!(vec.len(), 0);
assert!(vec.capacity() >= 10);
```
`.capacity()` may return more than the requested

martin-g:
```rust
fn create_buf(length: usize) -> Vec<u8> {
    (0..length).map(|i| i as u8).collect()
}
```

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,195 @@
...
+            Err(e) => {
+                if e.kind() == ErrorKind::Interrupted {
+                    buf = r_buf;
+                    fd = r_fd;
+
+                    continue;
+                } else {
+                    return Err(e);
+                }
+            }
```
ADD-SP:
```suggestion
            Err(e) if e.kind() == ErrorKind::Interrupted {
                buf = r_buf;
                fd = r_fd;
                continue;
            }
            Err(e) => return Err(e),
```

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,195 @@
...
+
+        loop {
+            // read into spare capacity
+            let (res, r_fd, mut r_buf) = Op::read(fd, buf, read_len, offset).await;
+
+            match res {
+                Ok(size_read) => {
+                    if size_read == 0 {
+                        return Ok(r_buf);
+                    }
```
ADD-SP:
```suggestion
                Ok(0) => return Ok(r_buf),
                Ok(size_read) => {
```

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,195 @@
...
+        // prepare the spare capacity to be read into
+        let spare = buf.capacity() - buf.len();
+        let buf_len = cmp::min(spare, MAX_READ_SIZE);
+
+        let mut read_len = {
+            // SAFETY:
+            // 1. buf_len cannot be greater than u32::MAX because max_read_size
+            // is u32::MAX
+            // 2. This is faster than a cast
+            unsafe { TryInto::<u32>::try_into(buf_len).unwrap_unchecked() }
```
ADD-SP: It’s not worth introducing an unsafe operation here; the major bottleneck of this function is the memory allocator and system calls, not the computation.

tokio/src/fs/read_uring.rs
ADD-SP: We need a short comment to explain the reason of "plus one".

tokio/src/fs/read_uring.rs
Daksh14: switched to `as` cast with the comment included

tokio/src/fs/read_uring.rs
Daksh14: let me know if its descriptive enough @ADD-SP

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,183 @@
...
+        .map(|m| usize::try_from(m.len()).unwrap_or(usize::MAX))
+        .ok();
+
+    let fd: OwnedFd = file
+        .try_into_std()
+        .expect("unexpected in-flight operation detected")
+        .into();
+
+    // extra single capacity for the whole size to fit without any reallocation
+    let buf = Vec::with_capacity(size_hint.unwrap_or(0).saturating_add(1));
```
ADD-SP: > extra single capacity for the whole size to fit without any reallocation

Why doesn't the standard library do it?

```rust
// https://doc.rust-lang.org/1.90.0/src/std/fs.rs.html#308
pub fn read<P: AsRef<Path>>(path: P) -> io::Result<Vec<u8>> {
    fn inner(path: &Path) -> io::Result<Vec<u8>> {
        let mut file = File::open(path)?;
        let size = file.metadata().map(|m| m.len() as usize).ok();
        let mut bytes = Vec::try_with_capacity(size.unwrap_or(0))?;
        io::default_read_to_end(&mut file, &mut bytes, size)?;
        Ok(bytes)
    }
    inner(path.as_ref())
}
```

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,183 @@
...
+
+pub(crate) async fn read_uring(path: &Path) -> io::Result<Vec<u8>> {
+    let file = OpenOptions::new().read(true).open(path).await?;
+
+    // TODO: use io uring in the future to obtain metadata
+    let size_hint: Option<usize> = file
+        .metadata()
+        .await
+        .map(|m| usize::try_from(m.len()).unwrap_or(usize::MAX))
+        .ok();
```
ADD-SP: The purpose of `size_hint` is to improve performance by avoiding memory allocation. However, it involves thread pool mechanisms—which may not be more efficient than multiple memory allocations due to the overhead of creating/terminating threads and inter-thread synchronization operations.

We need performance data to demonstrate that this approach is faster than the method without using `size_hint`.

I asked this because I want to eliminate unnecessary logic as much as possible.

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,183 @@
...
+const MAX_READ_SIZE: usize = u32::MAX as usize;
+
+pub(crate) async fn read_uring(path: &Path) -> io::Result<Vec<u8>> {
+    let file = OpenOptions::new().read(true).open(path).await?;
+
+    // TODO: use io uring in the future to obtain metadata
+    let size_hint: Option<usize> = file
+        .metadata()
+        .await
+        .map(|m| usize::try_from(m.len()).unwrap_or(usize::MAX))
```
ADD-SP: The `Vec::with_capacity` panics if the capacity exceeds `isize::MAX`.

tokio/src/fs/read_uring.rs
Daksh14: Because they react on the bytes read rather than how many bytes to read ahead of time. In case of a short read during EOF and `small_probe_read` is called, it allocates `PROBE_SIZE` extra space which can be an expensive operation depending on the current size of the buffer

tokio/src/fs/read_uring.rs
Daksh14: my question is Why would you think that? Would you think a statx helper will eliminate this problem or it will still involve thread/pool mechanisms? I guess it will also be a future SQE. 

I will include some benchmarks and upload them in a separate branch to test this @ADD-SP

tokio/src/fs/read_uring.rs
Daksh14: Fixed with a `as` cast

tokio/src/fs/read_uring.rs
ADD-SP: Just two simple benchmarks are enough. One is for small file, another one is for large file.

tokio/src/io/uring/read.rs
```diff
@@ -0,0 +1,54 @@
...
+    // Submit a request to read a FD at given length and offset into a
+    // dynamic buffer with uinitialized memory. The read happens on unitialized
+    // buffer and no overwiting happens.
+
+    // SAFETY: The `len` of the amount to be read and the buffer that is passed
+    // should have capacity > len.
+    //
+    // If `len` read is higher than vector capacity then setting its length by
+    // the caller in terms of size_read can be unsound.
+    pub(crate) fn read(fd: OwnedFd, mut buf: Vec<u8>, len: u32, offset: u64) -> Self {
```
mox692, line 49: This function should be marked as `unsafe` if the caller needs to follow this.

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,179 @@
...
+        offset += size_read as u64;
+    }
+
+    loop {
+        if buf.len() == buf.capacity() && buf.capacity() == start_cap {
+            // The buffer might be an exact fit. Let's read into a probe buffer
+            // and see if it returns `Ok(0)`. If so, we've avoided an
+            // unnecessary increasing of the capacity. But if not, append the
+            // probe buffer to the primary buffer and let its capacity grow.
+            let (size_read, r_fd, r_buf) = small_probe_read(fd, buf, offset).await?;
```
mox692: > If so, we've avoided an unnecessary increasing of the capacity.

This doesn't sound correct, since current `small_probe_read` would allocate by calling `buf.reserve_exact(PROBE_SIZE);`.

Daksh14: yea and it would exactly increase the capacity by `PROBE_SIZE`, if not then `try_reserve` gets called which doubles the capacity

mox692: ~~I suppose we can avoid extra allocation at all if `read` returns `0`.~~

~~https://github.com/rust-lang/rust/blob/17e7324d44c7b310b776302880fdd08d9252118a/library/std/src/io/mod.rs#L426-L441~~


EDIT: Never mind, we can't do that for the same reason as https://github.com/tokio-rs/tokio/pull/7567#discussion_r2329613543.

tokio/src/io/uring/read.rs
Daksh14: I would agree but there's nothing inherently unsafe in calling this function nor it is unsound.
The safety comment is there since this function writes in uninitialized capacity the caller will try to `set_len` and if they do that on the basis of size_read and do not have sufficient capacity/len then it will be UB

tokio/src/fs/read.rs
Daksh14: max read size at a time is u32::MAX, we read the rest in other next iterations

tokio/src/io/uring/read.rs
Daksh14: Changed, let me know what you think now

tokio/src/fs/read.rs
Daksh14: Fixed those

tokio/src/fs/read.rs
Daksh14: In the future, if we know we're reading more than u32::MAX then we can batch 2 read requests to avoid extra syscalls

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,179 @@
...
+    mut fd: OwnedFd,
+    mut buf: Vec<u8>,
+) -> io::Result<Vec<u8>> {
+    let mut offset = 0;
+    let mut consecutive_short_reads = 0;
+
+    let start_cap = buf.capacity();
+
+    // if buffer has no room and no size_hint, start with a small probe_read from 0 offset
+    if (size_hint.is_none() || size_hint == Some(0)) && buf.capacity() - buf.len() < (PROBE_SIZE) {
```
ADD-SP:
```suggestion
    if (size_hint.is_none() || size_hint == Some(0)) && buf.capacity() - buf.len() < PROBE_SIZE {
```

spurious brackets?

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    if has_enough {
+        // if we have more than PROBE_SIZE bytes in the buffer already then
+        // don't call reserve as we might potentially read 0 bytes
+        let last_bytes = &buf[buf.len() - PROBE_SIZE..];
+        temp_arr.copy_from_slice(last_bytes);
+
+        // SAFETY: We're decreasing the length of the buffer and len is greater
+        // than PROBE_SIZE. So we can read into the discarded length
+        unsafe { buf.set_len(buf.len() - PROBE_SIZE) };
```
Darksonn: This can just be truncate.

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,214 @@
...
+            return Ok(r_buf);
+        }
+
+        buf = r_buf;
+        fd = r_fd;
+        offset += size_read as u64;
+    }
+
+    loop {
+        if buf.len() == buf.capacity() && buf.capacity() == start_cap {
```
Darksonn: I think we should just not call `small_probe_read` if the length is less than `SMALL_PROBE_READ`. In that case, we can just call `buf.reserve(SMALL_PROBE_READ)` and continue to the normal code below. This will simplify `small_probe_read` because `has_enough` will always be true so you can remove those codepaths.

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,214 @@
...
+            offset += size_read as u64;
+        }
+
+        // buf is full, need more capacity
+        if buf.len() == buf.capacity() {
+            if consecutive_short_reads > 1 {
+                buf.try_reserve(PROBE_SIZE.saturating_mul(consecutive_short_reads))?;
+            } else {
+                buf.try_reserve(PROBE_SIZE)?;
+            }
```
Darksonn: I don't understand why there is this `consecutive_short_reads` logic. When you reserve, it will double the capacity internally so we do not need this logic.

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        // doesn't matter if we have a valid size_hint or not, if we do more
+        // than 2 consecutive_short_reads, gradually increase the buffer
+        // capacity to read more data at a time
+
+        // prepare the spare capacity to be read into
+        let spare = buf.capacity() - buf.len();
+        let buf_len = cmp::min(spare, MAX_READ_SIZE);
```
Darksonn:
```suggestion
        let buf_len = usize::min(buf.spare_capacity_mut().len(), MAX_READ_SIZE);
```

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,214 @@
...
+
+        // SAFETY: buf_len cannot be greater than u32::MAX because max_read_size
+        // is u32::MAX
+        let mut read_len = buf_len as u32;
+
+        loop {
+            // read into spare capacity
+            let (res, r_fd, mut r_buf) = Op::read(fd, buf, read_len, offset).await;
```
Darksonn: I don't think we need a loop inside a loop. In case of interrupted, just go around the outer loop again.

tokio/src/io/uring/read.rs
```diff
@@ -0,0 +1,54 @@
...
+        // don't overwrite on already written part
+        let written = buf.len();
+        let slice: &mut [u8] = &mut buf[written..];
+        let buf_mut_ptr = slice.as_mut_ptr();
```
Darksonn:
```suggestion
        assert!(buf.spare_capacity_mut().len() <= len as usize);
        let buf_mut_ptr = buf.spare_capacity_mut().as_mut_ptr();
```

tokio/src/io/uring/read.rs
```diff
@@ -0,0 +1,54 @@
...
+    fd: OwnedFd,
+    buf: Vec<u8>,
+}
+
+impl Completable for Read {
+    type Output = (io::Result<u32>, OwnedFd, Vec<u8>);
+
+    fn complete(self, cqe: CqeResult) -> Self::Output {
+        (cqe.result, self.fd, self.buf)
+    }
```
Darksonn, line 26: Moving the `set_len` here seems to simplify the other file quite a bit.
```suggestion
    type Output = (io::Result<u32>, OwnedFd, Vec<u8>);

    fn complete(mut self, cqe: CqeResult) -> Self::Output {
        if let Ok(len) = cqe.result {
            let new_len = self.buf.len() + len as usize;
            unsafe { self.buf.set_len(new_len) };
        }

        (cqe.result, self.fd, self.buf)
    }
```

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,214 @@
...
+                    // read size per Op submission to read files in lesser cycles
+                    if size_read <= requested {
+                        consecutive_short_reads += 1;
+                    } else {
+                        consecutive_short_reads = 0;
+                    }
+
+                    // keep reading if there's something left to be read
+                    if read_len > 0 {
+                        continue;
```
Darksonn: I think we can remove this consecutive short reads logic.
```suggestion
                    // keep reading if there's something left to be read
                    if read_len == size_read {
                        continue;
```

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,214 @@
...
+                // copy the bytes we just read into a temp buffer
+                temp_stack[..size_read_usize].copy_from_slice(&r_buf[old_len..]);
+
+                // ignore the bytes we just read so we can first extend the old
+                // copied stack array
+                unsafe { r_buf.set_len(old_len) }
+
+                // extend old copied PROBE_SIZE bytes
+                r_buf.extend_from_slice(&temp_arr);
+                r_buf.extend_from_slice(&temp_stack[..size_read_usize]);
```
Darksonn:
```suggestion
                r_buf.splice(old_len..old_len, temp_arr);
```

tokio/src/io/uring/read.rs
Daksh14: Can't do this because this gives a `MaybeUnint<T>`

Darksonn: `let buf_mut_ptr = buf.spare_capacity_mut().as_mut_ptr().cast();`

tokio/src/fs/read_uring.rs
Daksh14: Depends on the allocator impl, some allocators do not double the capacity, and in case of short reads by the kernel, we should ask for more data to read

tokio/src/fs/read_uring.rs
Daksh14: small_probe_read is still called at the start if no size_hint is provided

tokio/src/fs/read_uring.rs
Darksonn: It may not guarantee that it specifically doubles it, but it says this:

> Vec does not guarantee any particular growth strategy when reallocating when full, nor when reserve is called. The current strategy is basic and it may prove desirable to use a non-constant growth factor. Whatever strategy is used will of course guarantee O(1) amortized push.

Amortized constant time push is a sufficient guarantee here. We do not need to do anything special when calling reserve.

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,167 @@
...
+        // prepare the spare capacity to be read into
+        let buf_len = usize::min(buf.spare_capacity_mut().len(), MAX_READ_SIZE);
+
+        // SAFETY: buf_len cannot be greater than u32::MAX because max_read_size
+        // is u32::MAX
+        let mut read_len = buf_len as u32;
+
+        loop {
+            // read into spare capacity
+            let (res, r_fd, r_buf) = Op::read(fd, buf, read_len, offset).await;
```
Darksonn: This file has two different loops around `Op::read` to handle the interrupted error. Perhaps we should instead extract that logic to a separate function call, which behaves just like `Op::read` except that it retries in case of interrupted, then use that function from both places?

Daksh14: @Darksonn The newly changed code has custom handling with the match around the result of `Op::read` so it will be a little hard to put it in a function, I think what we have is fine for now

tokio/src/io/uring/read.rs
Daksh14: Added an assert in the function

tokio/src/fs/read_uring.rs
Daksh14: this is now removed as per https://github.com/tokio-rs/tokio/pull/7696#discussion_r2498854193

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,166 @@
...
+            Err(e) if e.kind() == ErrorKind::Interrupted => {
+                buf = r_buf;
+                fd = r_fd;
+
+                continue;
+            }
+            Err(e) => return Err(e),
+        }
+    }
+}
```
ADD-SP, line 134: This is a bit hard to follow, we may need a dedicated `Read` type just `[u8; N]`, for example:

Daksh14: This doesn't work because you cannot read into an Array, only a vec. The returned operation struct needs to have owned version of the Vector

ADD-SP: > This doesn't work because you cannot read into an Array, only a vec. The returned operation struct needs to have owned version of the Vector

Sorry, I don't understand, could you explain it in a bit details? Why cannot we read data into an Array?

tokio/src/fs/read_uring.rs
Darksonn: I gave it a try, and it seems to work well:
```diff
diff --git a/tokio/src/fs/read_uring.rs b/tokio/src/fs/read_uring.rs
index 037ca3ed..07843580 100644
--- a/tokio/src/fs/read_uring.rs
+++ b/tokio/src/fs/read_uring.rs
@@ -83,37 +83,17 @@ async fn read_to_end_uring(
         // prepare the spare capacity to be read into
         let buf_len = usize::min(buf.spare_capacity_mut().len(), MAX_READ_SIZE);
 
-        // SAFETY: buf_len cannot be greater than u32::MAX because max_read_size
-        // is u32::MAX
-        let mut read_len = buf_len as u32;
-
-        loop {
-            // read into spare capacity
-            let (res, r_fd, r_buf) = Op::read(fd, buf, read_len, offset).await;
-
-            match res {
-                Ok(0) => return Ok(r_buf),
-                Ok(size_read) => {
-                    fd = r_fd;
-                    buf = r_buf;
-                    offset += size_read as u64;
-                    read_len -= size_read;
-
-                    // keep reading if there's something left to be read
-                    if read_len > 0 {
-                        continue;
-                    } else {
-                        break;
-                    }
-                }
-                Err(e) if e.kind() == ErrorKind::Interrupted => {
-                    buf = r_buf;
-                    fd = r_fd;
-
-                    continue;
-                }
-                Err(e) => return Err(e),
+        // read into spare capacity
+        let (res, r_fd, r_buf) = op_read(fd, buf, buf_len as u32, offset).await;
+
+        match res {
+            Ok(0) => return Ok(r_buf),
+            Ok(size_read) => {
+                fd = r_fd;
+                buf = r_buf;
+                offset += size_read as u64;
             }
+            Err(e) => return Err(e),
         }
     }
 }
@@ -140,27 +120,38 @@ async fn small_probe_read(
         buf.reserve_exact(PROBE_SIZE);
     }
 
-    loop {
-        let (res, r_fd, mut r_buf) = Op::read(fd, buf, PROBE_SIZE_U32, offset).await;
+    let (res, r_fd, mut r_buf) = op_read(fd, buf, PROBE_SIZE_U32, offset).await;
 
-        match res {
-            // return early if we inserted into reserved PROBE_SIZE
-            // bytes
-            Ok(size_read) if !has_enough => return Ok((size_read, r_fd, r_buf)),
-            Ok(size_read) => {
-                let old_len = r_buf.len() - (size_read as usize);
+    match res {
+        // return early if we inserted into reserved PROBE_SIZE
+        // bytes
+        Ok(size_read) if !has_enough => Ok((size_read, r_fd, r_buf)),
+        Ok(size_read) => {
+            let old_len = r_buf.len() - (size_read as usize);
 
-                r_buf.splice(old_len..old_len, temp_arr);
+            r_buf.splice(old_len..old_len, temp_arr);
 
-                return Ok((size_read, r_fd, r_buf));
-            }
+            Ok((size_read, r_fd, r_buf));
+        }
+        Err(e) => Err(e),
+    }
+}
+
+async fn op_read(
+    mut fd: OwnedFd,
+    mut buf: Vec<u8>,
+    len: u32,
+    offset: u64,
+) -> io::Result<(u32, OwnedFd, Vec<u8>)> {
+    loop {
+        let (res, r_fd, mut r_buf) = Op::read(fd, buf, len, offset).await;
+
+        match res {
             Err(e) if e.kind() == ErrorKind::Interrupted => {
                 buf = r_buf;
                 fd = r_fd;
-
-                continue;
             }
-            Err(e) => return Err(e),
+            _ => return (res, r_fd, r_buf),
         }
     }
 }
```

Daksh14: @Darksonn I see what you mean now, I'll use this

Daksh14: After trying to implement this, it becomes quite ugly as we'll be returning `io::Result<(io::Result<u32>, OwnedFd, Vec<u8>)>` and we can't use `read_len` anymore but its useful. The match clauses that call `op_read` also become quite ugly like
```rust
match res {
            Ok((Ok(0), _, r_buf)) => return Ok(r_buf),
            Ok((Ok(size_read), r_fd, r_buf)) => {
                fd = r_fd;
                buf = r_buf;
                offset += size_read as u64;
                read_len -= size_read;
            }
            Ok((Err(e), _, _)) | Err(e) => return Err(e),
        }

```

ADD-SP: This looks still better than the nested loop.

Darksonn: The return type of `op_read` should probably be `(io::Result<u32>, OwnedFd, Vec<u8>)`, which fixes this.

Daksh14: in case of Err What do I return as tuple not wrapped in a result?

Darksonn: What do you mean?

ADD-SP: Just want to share my thoughts — this `match` statement is easy to understand and already an improvement over the nested loop. I suggest keeping this `match` statement unless there is a better approach.

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        // buf is full, need more capacity
+        if buf.len() == buf.capacity() {
+            buf.try_reserve(PROBE_SIZE)?;
+        }
+
+        // doesn't matter if we have a valid size_hint or not, if we do more
+        // than 2 consecutive_short_reads, gradually increase the buffer
+        // capacity to read more data at a time
```
Darksonn: This comment is outdated.

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,167 @@
...
+
+        // prepare the spare capacity to be read into
+        let buf_len = usize::min(buf.spare_capacity_mut().len(), MAX_READ_SIZE);
+
+        // SAFETY: buf_len cannot be greater than u32::MAX because max_read_size
+        // is u32::MAX
```
Darksonn: This is a cast, not an unsafe operation, so please remove SAFETY:.

Also `max_read_size` should be capital letters.

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,167 @@
...
+    if !has_enough {
+        Ok((size_read, r_fd, r_buf))
+    } else {
+        let old_len = r_buf.len() - (size_read as usize);
+
+        r_buf.splice(old_len..old_len, temp_arr);
+
+        Ok((size_read, r_fd, r_buf))
+    }
+}
```
Darksonn, line 107:
```suggestion
async fn small_probe_read(
    fd: OwnedFd,
    mut buf: Vec<u8>,
    offset: &mut u64,
) -> io::Result<(u32, OwnedFd, Vec<u8>)> {
    if buf.len() < PROBE_SIZE {
        buf.try_reserve(PROBE_SIZE)?;
        return op_read(fd, buf, offset, PROBE_SIZE_U32).await;
    }

    let mut temp_arr = [0; PROBE_SIZE];

    let back_bytes_len = buf.len() - PROBE_SIZE;
    temp_arr.copy_from_slice(&buf[back_bytes_len..]);

    // We're decreasing the length of the buffer and len is greater
    // than PROBE_SIZE. So we can read into the discarded length
    buf.truncate(back_bytes_len);

    let (size_read, r_fd, mut r_buf) = op_read(fd, buf, offset, PROBE_SIZE_U32).await?;

    r_buf.splice(back_bytes_len..back_bytes_len, temp_arr);

    Ok((size_read, r_fd, r_buf))
}
```

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,154 @@
...
+    r_buf.splice(back_bytes_len..back_bytes_len, temp_arr);
+
+    Ok((size_read, r_fd, r_buf))
+}
+
+async fn op_read(
+    mut fd: OwnedFd,
+    mut buf: Vec<u8>,
+    offset: &mut u64,
+    read_len: &mut u32,
```
Darksonn: There are no cases where the caller reads the value of `read_len` after the call, so a ref is not needed.
```suggestion
    read_len: u32,
```

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    temp_arr.copy_from_slice(&buf[back_bytes_len..]);
+
+    // We're decreasing the length of the buffer and len is greater
+    // than PROBE_SIZE. So we can read into the discarded length
+    buf.truncate(back_bytes_len);
+
+    let (size_read, r_fd, mut r_buf) = op_read(fd, buf, offset, &mut read_len).await?;
+
+    r_buf.splice(back_bytes_len..back_bytes_len, temp_arr);
```
Darksonn, line 104: We can handle allocation failure by error rather than panic by allocating before calling `splice`. Note that this doesn't allocate if `size_read == 0` since the capacity is guaranteed enough in that case.
```suggestion
    r_buf.try_reserve(PROBE_SIZE)?;
    r_buf.splice(back_bytes_len..back_bytes_len, temp_arr);
```

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,154 @@
...
+    // TODO: use io uring in the future to obtain metadata
+    let size_hint: Option<usize> = file.metadata().await.map(|m| m.len() as usize).ok();
+
+    let fd: OwnedFd = file
+        .try_into_std()
+        .expect("unexpected in-flight operation detected")
+        .into();
+
+    // extra single capacity for the whole size to fit without any reallocation
+    let buf = Vec::with_capacity(size_hint.unwrap_or(0));
```
Darksonn: We handle other allocation failures with an error, so let's do this one too. (This one is the most likely to fail.)
```suggestion
    let buf = Vec::new();
    if let Some(size_hint) = size_hint {
        buf.try_reserve(size_hint)?;
    }
```

tokio/tests/fs_uring_read.rs
```diff
@@ -0,0 +1,194 @@
...
+            }
+        });
+
+        std::thread::spawn(move || {
+            let rt: Runtime = rx.recv().unwrap();
+            rt.shutdown_timeout(Duration::from_millis(300));
+            done_tx.send(()).unwrap();
+        });
+
+        tx.send(rt).unwrap();
```
Darksonn: This is equivalent to just moving the runtime.
```suggestion
        std::thread::spawn(move || {
            rt.shutdown_timeout(Duration::from_millis(300));
            done_tx.send(()).unwrap();
        });

```
If you wanted an actual sleep for the loop to make some progress, then what you have now does not achieve that.

tokio/tests/fs_uring_read.rs
```diff
@@ -0,0 +1,194 @@
...
+    let bytes = read(path).await.unwrap();
+
+    assert_eq!(bytes, create_buf(20));
+}
+
+#[tokio::test]
+async fn cancel_op_future() {
+    let (_tmp_file, path): (Vec<NamedTempFile>, Vec<PathBuf>) = create_tmp_files(1);
+
+    let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
```
Darksonn, line 139: You only send one message, so use oneshot.

tokio/tests/fs_uring_read.rs
```diff
@@ -0,0 +1,194 @@
...
+    let handle = tokio::spawn(async move {
+        poll_fn(|cx| {
+            let fut = read(&path[0]);
+
+            // If io_uring is enabled (and not falling back to the thread pool),
+            // the first poll should return Pending.
+            let _pending = Box::pin(fut).poll_unpin(cx);
```
Darksonn: This code currently drops the read future right away. If you want the future to stay around until the `handle.abort()` call, then you need to modify this code.
```suggestion
        let fut = read(&path[0]);
        tokio::pin!(fut);
        poll_fn(|cx| {
            // If io_uring is enabled (and not falling back to the thread pool),
            // the first poll should return Pending.
            let _pending = fut.as_mut().poll(cx);
```

tokio/src/fs/read_uring.rs
Daksh14: @Darksonn Will this splice write in the reserved area? Because we can only index in the length and not capacity

Darksonn: No. The length of `r_buf` before the `op_read` is exactly `back_bytes_len`, so after the `op_read` we have `back_bytes_len <= r_buf.len()` which is the requirement for using `splice`.

tokio/tests/fs_uring_read.rs
Daksh14: I have pushed a slightly different version of this

tokio/tests/fs_uring_read.rs
Daksh14: Cannot use oneshot as `send` requires ownership

tokio/tests/fs_uring_read.rs
Daksh14: @Darksonn Should I sleep in the std::thread so the loop can make progress? Is that allowed or I shouldn't do it

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,159 @@
...
+    temp_arr.copy_from_slice(&buf[back_bytes_len..]);
+
+    // We're decreasing the length of the buffer and len is greater
+    // than PROBE_SIZE. So we can read into the discarded length
+    buf.truncate(back_bytes_len);
+
+    let (size_read, r_fd, mut r_buf) = op_read(fd, buf, offset, read_len).await?;
+
+    r_buf.try_reserve(PROBE_SIZE)?;
+    r_buf.splice(back_bytes_len..back_bytes_len, temp_arr);
```
ADD-SP, line 104: Emm, maybe a dumb question, why use `splice` instead of `extend_from_slice`?

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,159 @@
...
+        // buf is full, need more capacity
+        if buf.len() == buf.capacity() {
+            buf.try_reserve(PROBE_SIZE)?;
+        }
+
+        // prepare the spare capacity to be read into
+        let buf_len = usize::min(buf.spare_capacity_mut().len(), MAX_READ_SIZE);
+
+        // buf_len cannot be greater than u32::MAX because MAX_READ_SIZE
+        // is u32::MAX
```
martin-g: MAX_READ_SIZE is 64MiB (https://github.com/tokio-rs/tokio/pull/7696/files#diff-4faa26928d46778fd791e1e631b65dcc100bfe5c88eece9817f0256fd329ed16R17)

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,159 @@
...
+            // probe buffer to the primary buffer and let its capacity grow.
+            let (size_read, r_fd, r_buf) = small_probe_read(fd, buf, &mut offset).await?;
+
+            if size_read == 0 {
+                return Ok(r_buf);
+            }
+
+            buf = r_buf;
+            fd = r_fd;
+            offset += size_read as u64;
```
martin-g: Is this needed ?
The offset is incremented at https://github.com/tokio-rs/tokio/pull/7696/files#diff-4faa26928d46778fd791e1e631b65dcc100bfe5c88eece9817f0256fd329ed16R147

tokio/src/fs/read.rs
martin-g, line 15: This docu has to be updated for the uring fast path.

Daksh14: @martin-g Good catch, I added a io-uring section similar to the `Open` doc as its unstable

tokio/src/fs/read_uring.rs
Daksh14: It should be `is less than u32::MAX`

tokio/src/fs/read_uring.rs
Daksh14: Yes this is a double subtraction of the offset, more tests that detect these kinds of faults are welcome.

I have fixed it.

tokio/src/fs/read_uring.rs
Daksh14: @ADD-SP how will `extend_from_slice` let us put the temp_arr at the range `back_bytes_len..back_bytes_len`?

ADD-SP: Ah, I see, because the `op_read` might append data to the buffer.

tokio/src/fs/read_uring.rs
Daksh14: `op_read` is introduced

tokio/tests/fs_uring_read.rs
ADD-SP: 1. Poll the `read(path).await` manually to submit it to the kernel.
2. Then increase the semaphore.
3. Keep the spawn task pending for ever.

So that we can wait on the semaphore and then shutdown the runtime.

ADD-SP: Is this work for multi-thread runtime? For current thread runtime, this won't work. But I believe we can apply the similar pattern to the current thread runtime.

```rust
        rt.spawn(async move {
            let path = path[0].clone();
            let mut futs = vec![];

            // spawning a bunch of uring operations.
            for _ in 0..N {
                let path = path.clone();
                let cl = Arc::clone(&cl);
                let fut = Box::pin(read(path));
                assert_pending!(fut.poll(cx));
                futs.push(fut);
            }

            pending_forever().await;
        });

        std::thread::spawn(move || {
            rt.shutdown_timeout(Duration::from_millis(300));
            done_tx.send(()).unwrap();
        });

        done_rx.recv().unwrap();
```

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,157 @@
...
+async fn read_to_end_uring(
+    size_hint: Option<usize>,
+    mut fd: OwnedFd,
+    mut buf: Vec<u8>,
+) -> io::Result<Vec<u8>> {
+    let mut offset = 0;
+    let start_cap = buf.capacity();
+
+    // if buffer has no room and no size_hint, start with a small probe_read from 0 offset
+    if (size_hint.is_none() || size_hint == Some(0)) && buf.capacity() - buf.len() < PROBE_SIZE {
```
Darksonn: I think we can just delete this if. If the size hint is zero/missing, then we always have `buf.capacity() == 0`. If we delete it, then it will fall through to the loop, which will do the same thing as what this block does.

tokio/src/fs/read.rs
```diff
@@ -45,5 +55,23 @@ use std::{io, path::Path};
...
 pub async fn read(path: impl AsRef<Path>) -> io::Result<Vec<u8>> {
     let path = path.as_ref().to_owned();
+
+    #[cfg(all(
+        tokio_unstable,
+        feature = "io-uring",
+        feature = "rt",
+        feature = "fs",
+        target_os = "linux"
+    ))]
```
martin-g, line 65: Any reason why `cfg_io_uring! {` is not used here ?

tokio/src/io/uring/read.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl Cancellable for Read {
+    fn cancel(self) -> CancelData {
+        CancelData::Read(self)
+    }
+}
+
+impl Op<Read> {
+    // Submit a request to read a FD at given length and offset into a
+    // dynamic buffer with uinitialized memory. The read happens on unitialized
```
martin-g:
```suggestion
    // dynamic buffer with uninitialized memory. The read happens on uninitialized
```

tokio/src/io/uring/read.rs
```diff
@@ -0,0 +1,61 @@
...
+    // buffer and no overwiting happens.
```
martin-g:
```suggestion
    // buffer and no overwriting happens.
```

tokio/tests/fs_uring_read.rs
```diff
@@ -0,0 +1,202 @@
...
+    let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
+
+    let handle = tokio::spawn(async move {
+        let fut = read(path.clone());
+        tokio::pin!(fut);
+
+        poll_fn(move |cx| {
+            // If io_uring is enabled (and not falling back to the thread pool),
+            // the first poll should return Pending.
+            let _pending = fut.as_mut().poll(cx);
```
martin-g:
```suggestion
            assert_pending!(fut.as_mut().poll(cx));
```

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,141 @@
...
+use std::path::Path;
+
+// this algorithm is inspired from rust std lib version 1.90.0
+// https://doc.rust-lang.org/1.90.0/src/std/io/mod.rs.html#409
+const PROBE_SIZE: usize = 32;
+const PROBE_SIZE_U32: u32 = PROBE_SIZE as u32;
+
+// Max bytes we can read using io uring submission at a time
+// SAFETY: cannot be higher than u32::MAX for safe cast
+// Set to read max 64 blocks at time
```
martin-g: What kind of block is meant here ?
File system blocks are usually around 4096 bytes. Here we are talking about a block of 1MiB.

```suggestion
// Set to read max 64 MiB at time
```
?

tokio/tests/fs_uring_read.rs
```diff
@@ -0,0 +1,202 @@
...
+            tx.send(()).unwrap();
+
+            Poll::<()>::Pending
+        })
+        .await;
+    });
+
+    // Wait for the first poll
+
+    let _ = rx.try_recv();
```
martin-g: Why `try_recv()` ?
This may return Empty before the task has a chance to be executed.

tokio/src/fs/read.rs
Daksh14: @martin-augment its following the currently used code block macro  here in `tokio/src/fs/open_options.rs`, I can change those in a different "refactor" PR https://github.com/tokio-rs/tokio/blob/master/tokio/src/fs/open_options.rs#L592

tokio/src/io/uring/read.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl Cancellable for Read {
+    fn cancel(self) -> CancelData {
+        CancelData::Read(self)
+    }
+}
+
+impl Op<Read> {
+    // Submit a request to read a FD at given length and offset into a
+    // dynamic buffer with uninitialized memory. The read happens on unitialized
```
martin-g:
```suggestion
    // dynamic buffer with uninitialized memory. The read happens on uninitialized
```
you forgot this one

tokio/tests/fs_uring_read.rs
Daksh14: I added an assert with recv await

tokio/src/fs/read_uring.rs
Daksh14: @martin-augment this number came after benchmarking with io-uring against `std::fs::read` as its a reasonable amount of data for kernel to copy

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    temp_arr.copy_from_slice(&buf[back_bytes_len..]);
+
+    // We're decreasing the length of the buffer and len is greater
+    // than PROBE_SIZE. So we can read into the discarded length
+    buf.truncate(back_bytes_len);
+
+    let (size_read, r_fd, mut r_buf) = op_read(fd, buf, offset, read_len).await?;
+
+    r_buf.try_reserve(PROBE_SIZE)?;
```
mox692, line 103: It would be nice to describe the intent of this optimization.

```suggestion
    // If `size_read` returns zero due to reasons such as buffer's exact fit,
    // then this `try_reserve` does not perform allocation.
    r_buf.try_reserve(PROBE_SIZE)?;
```

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,135 @@
...
+        if buf.len() == buf.capacity() {
+            buf.try_reserve(PROBE_SIZE)?;
+        }
+
+        // prepare the spare capacity to be read into
+        let buf_len = usize::min(buf.spare_capacity_mut().len(), MAX_READ_SIZE);
+
+        // buf_len cannot be greater than u32::MAX because MAX_READ_SIZE
+        // is less than u32::MAX
+        let read_len = buf_len as u32;
```
mox692: nit: Although there is no risk to overflow in the current code, I would prefer an explicit check.

```suggestion
        let read_len = u32::try_from(buf_len).expect("buf_len must always fit in u32");
```

Daksh14: Do you think its ideal to panic when we can choose not to? @Darksonn

Darksonn: I think it's ok to panic here.

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    read_to_end_uring(fd, buf).await
+}
+
+async fn read_to_end_uring(mut fd: OwnedFd, mut buf: Vec<u8>) -> io::Result<Vec<u8>> {
+    let mut offset = 0;
+    let start_cap = buf.capacity();
+
+    loop {
+        if buf.len() == buf.capacity() && buf.capacity() == start_cap && buf.len() > PROBE_SIZE {
```
Darksonn:
```suggestion
        if buf.len() == buf.capacity() && buf.capacity() == start_cap && buf.len() >= PROBE_SIZE {
```

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,137 @@
...
+                *offset += size_read as u64;
+                read_len -= size_read;
+
+                if read_len == 0 || size_read == 0 {
+                    return Ok((size_read, r_fd, r_buf));
+                }
+
+                buf = r_buf;
+                fd = r_fd;
+            }
```
Darksonn, line 131: I do not think this makes sense.
```suggestion
            Ok(size_read) => return Ok((size_read, r_fd, r_buf)),
```
The current logic means that unless the buffer fits the file exactly, we perform the final read of length zero twice:

1. We call `op_read`, which performs a partial read and then a length-zero EOF read. It returns a non-zero length to the caller.
2. Since the caller got a non-zero length, it now needs to call `op_read` again to check for EOF.

Checking for EOF twice does not make sense.

tokio/src/fs/read_uring.rs
Daksh14: Updated the comment

tokio/tests/fs_uring_read.rs
```diff
@@ -0,0 +1,202 @@
...
+    let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
+
+    let handle = tokio::spawn(async move {
+        let fut = read(path.clone());
+        tokio::pin!(fut);
+
+        poll_fn(move |cx| {
+            // If io_uring is enabled (and not falling back to the thread pool),
+            // the first poll should return Pending.
+            assert_pending!(fut.as_mut().poll(cx));
```
ADD-SP: Since the `fut.as_mut().poll()` registers the waker, this assertion will be polled twice. Consider introduce noop waker here.

Daksh14: Replaced the contex of poll_fn with a noop waker @ADD-SP

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,143 @@
...
+
+                let is_eof = size_read == 0;
+
+                if read_len == 0 || is_eof {
+                    return Ok((r_fd, r_buf, is_eof));
+                }
+
+                buf = r_buf;
+                fd = r_fd;
+            }
```
Darksonn, line 131: Is there a reason to not simplify the code like this?
```suggestion
            Ok(size_read) => return Ok((r_fd, r_buf, size_read == 0)),
```

Darksonn (approved):

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,135 @@
...
+    let mut buf = Vec::new();
+
+    if let Some(size_hint) = size_hint {
+        buf.try_reserve(size_hint)?;
+    }
+
+    read_to_end_uring(fd, buf).await
+}
+
+async fn read_to_end_uring(mut fd: OwnedFd, mut buf: Vec<u8>) -> io::Result<Vec<u8>> {
```
ADD-SP, line 39: I'm not sure if I missed something, but wouldn't there be a bug if `buf.capacity()` is zero?

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,135 @@
...
+    r_buf.try_reserve(PROBE_SIZE)?;
+    r_buf.splice(back_bytes_len..back_bytes_len, temp_arr);
+
+    Ok((r_fd, r_buf, is_eof))
+}
+
+// Takes a amount of length to read and reads until we exhaust the given length
+// to read or EOF reached.
+//
+// Returns size_read, the result and EOF reached or not
```
martin-g:
```suggestion
// Returns the file descriptor, the result and EOF reached or not
```

tokio/src/io/uring/read.rs
martin-g:
```suggestion
    // dynamic buffer with uninitialized memory. The read happens on uninitialized
```

tokio/src/fs/read_uring.rs
paolobarbolini: Should this be re-visited now that we're using `try_reserve`?

I guess this must have been ispired by the std implementation https://doc.rust-lang.org/1.91.1/src/std/fs.rs.html#307

tokio/src/fs/read_uring.rs
```diff
@@ -0,0 +1,135 @@
...
+    let (r_fd, mut r_buf, is_eof) = op_read(fd, buf, offset, read_len).await?;
+    // If `size_read` returns zero due to reasons such as buffer's exact fit,
+    // then this `try_reserve` does not perform allocation.
+    r_buf.try_reserve(PROBE_SIZE)?;
+    r_buf.splice(back_bytes_len..back_bytes_len, temp_arr);
+
+    Ok((r_fd, r_buf, is_eof))
+}
+
+// Takes a amount of length to read and reads until we exhaust the given length
```
martin-g: `... until we exhaust the given length to read ...` is not correct.
There is no logic that `size_read == read_len` below. The caller has to keep calling `op_read()` in a loop.

tokio/src/fs/read_uring.rs
Daksh14: @ADD-SP why would there be a bug if cap is zero? We read till EOF

tokio/src/fs/read_uring.rs
Daksh14: This was changed recently, I have updated the comment

tokio/src/fs/read_uring.rs
ADD-SP: Ah, my fault.

ADD-SP (approved):
Thanks!

PR #7715, opened 2025-10-28
stream: add `ChunksTimeout::into_remainder`

muhamadazmy:
stream: add ChunksTimeout::into_remainder

Summary:
When the underlying stream is an exclusive reference (&mut stream), and we need to drop the `ChunksTimout` stream without losing the buffered items.

### DIFF
--- a/tokio-stream/src/stream_ext/chunks_timeout.rs
+++ b/tokio-stream/src/stream_ext/chunks_timeout.rs
@@ -33,6 +33,12 @@ impl<S: Stream> ChunksTimeout<S> {
             cap: max_size,
         }
     }
+
+    /// Consumes the [`ChunksTimeout`] and then returns all buffered items.
+    pub fn into_remainder(mut self: Pin<&mut Self>) -> Vec<S::Item> {
+        let me = self.as_mut().project();
+        std::mem::take(me.items)
+    }
 }
 
 impl<S: Stream> Stream for ChunksTimeout<S> {
--- a/tokio-stream/tests/stream_chunks_timeout.rs
+++ b/tokio-stream/tests/stream_chunks_timeout.rs
@@ -1,0 +1,30 @@
+#![warn(rust_2018_idioms)]
+
+use futures::FutureExt;
+use std::error::Error;
+use tokio::time;
+use tokio::time::Duration;
+use tokio_stream::{self as stream, StreamExt};
+use tokio_test::assert_pending;
+use tokio_test::task;
+
+#[tokio::test(start_paused = true)]
+async fn stream_chunks_remainder() -> Result<(), Box<dyn Error>> {
+    let stream1 =
+        stream::iter([5]).then(move |n| time::sleep(Duration::from_secs(1)).map(move |_| n));
+
+    let inner = stream::iter([1, 2, 3, 4]).chain(stream1);
+    tokio::pin!(inner);
+
+    let chunked = (&mut inner).chunks_timeout(10, Duration::from_millis(20));
+
+    let mut chunked = task::spawn(chunked);
+    assert_pending!(chunked.poll_next());
+
+    let remainder = chunked.enter(|_, stream| stream.into_remainder());
+
+    assert_eq!(remainder, vec![1, 2, 3, 4]);
+    time::advance(Duration::from_secs(2)).await;
+    assert_eq!(inner.next().await, Some(5));
+    Ok(())
+}


muhamadazmy:
## Use case
```rust
use std::time::Duration;

use tokio::sync::mpsc;
use tokio_stream::{StreamExt, wrappers::ReceiverStream};

#[tokio::main]
async fn main() {
    let (tx, rx) = mpsc::channel::<u64>(20);

    let mut receiver = ReceiverStream::new(rx);

    let chunked = (&mut receiver).chunks_timeout(3, Duration::from_secs(3));
    tokio::pin!(chunked);

    tx.send(10).await.unwrap();
    tx.send(20).await.unwrap();

    tokio::select! {
        Some(batch) = chunked.next() => {
            println!("Got: {:?}", batch);
        }
        _ = tokio::time::sleep(Duration::from_secs(1)) => {
            // another condition
        }
    }

    let remainder = chunked.into_remainder();

    // use `receiver` here
}
```

ADD-SP (commented):
There is a similar interface in the standard library: [`std::slice::ChunksExactMut:into_remainder`]. This might be a better interface for your requirements.

[`std::slice::ChunksExactMut:into_remainder`]: https://doc.rust-lang.org/1.91.0/std/slice/struct.ChunksExactMut.html#method.into_remainder

muhamadazmy:
Thank you @ADD-SP for your review. The original plan was to completely consume the stream indeed and just return the reminder. This however won't work with the Pinned ChunksTimeout stream which is required when polling next() in a select statement.

This means an `into_reminder` won't actually be useful because the only way for this to be useful is if the ChunksTimeout stream is not pinned. And in that case, the reminder will always guaranteed to be empty since there is no way to interrupt a `next().await` call (unless you are select statement which requires the stream to be pinned)

I will still rename the method to `reminder` instead of `remaining` to match the std interface.

tokio-stream/src/stream_ext/chunks_timeout.rs
```diff
@@ -33,6 +33,12 @@ impl<S: Stream> ChunksTimeout<S> {
             cap: max_size,
         }
     }
+
+    /// Drains the buffered items, returning them without waiting for the timeout or capacity limit.
+    pub fn reminder(mut self: Pin<&mut Self>) -> Vec<S::Item> {
```
ADD-SP:
```suggestion
    pub fn into_remainder(mut self: Pin<&mut Self>) -> Vec<S::Item> {
```

Does this make more sense?

tokio-stream/src/stream_ext/chunks_timeout.rs
```diff
@@ -33,6 +33,12 @@ impl<S: Stream> ChunksTimeout<S> {
             cap: max_size,
         }
     }
+
+    /// Drains the buffered items, returning them without waiting for the timeout or capacity limit.
```
ADD-SP:
```suggestion
    /// Consumes the [`ChunksTimeout`] and then returns all buffered items.
```

muhamadazmy:
Thank you so much @ADD-SP for your review. I applied all your comments. Let me know if you have more comments.

martin-g:
I think you could add the code example from https://github.com/tokio-rs/tokio/pull/7715#issuecomment-3456368612 as a snippet in the rustdoc for the new method.

Some new tests would be nice too!

ADD-SP:
> I think you could add the code example from [#7715 (comment)](https://github.com/tokio-rs/tokio/pull/7715#issuecomment-3456368612) as a snippet in the rustdoc for the new method.
> 
> Some new tests would be nice too!

@martin-g That example seems too long, and it’s not easy for downstream readers to grasp it in a few seconds. I’m okay with adding an example, but if we do, it should be simpler than the existing one.

tokio-stream/src/stream_ext/chunks_timeout.rs
```diff
@@ -33,6 +33,12 @@ impl<S: Stream> ChunksTimeout<S> {
             cap: max_size,
         }
     }
+
+    /// Consumes the [`ChunksTimeout`] and then returns all buffered items.
+    pub fn into_reminder(mut self: Pin<&mut Self>) -> Vec<S::Item> {
```
ADD-SP:
```suggestion
    pub fn into_remainder(mut self: Pin<&mut Self>) -> Vec<S::Item> {
```

typo?

muhamadazmy: Ops! sorry :(

muhamadazmy:
Dear @ADD-SP, thank you for your time! I’ve added a test case demonstrating how I would use it mainly to recover any buffered remainder, then restore the inner stream and continue consuming it without losing any items.

The example is a simplified version of the test, though I’m not sure if it’s simple enough for readers. Let me know what you think or if you have any suggestions to improve it

tokio-stream/src/stream_ext/chunks_timeout.rs
```diff
@@ -33,6 +33,34 @@ impl<S: Stream> ChunksTimeout<S> {
...
+    ///
+    ///     tokio::pin!(chunked);
+    ///
+    ///     // race with another future
+    ///     tokio::select! {
+    ///         Some(_chunk) = chunked.next() => {}
+    ///         _ = tokio::time::sleep(Duration::from_millis(3)) => {}
+    ///     }
+    ///
+    ///     assert_eq!(chunked.into_remainder(), vec![1, 2]);
```
ADD-SP: Please try to avoid introducing timers into tests, this can help us to make the test less flaky.  I suggest either improving this test or remove this test.

tokio-stream/tests/stream_chunks_remainder.rs
```diff
@@ -0,0 +1,26 @@
+#![warn(rust_2018_idioms)]
+#![cfg(all(feature = "time", not(target_os = "wasi")))] // Wasi does not support panic recovery
```
ADD-SP: Emm, are we testing panic recovery in this file?

tokio-stream/tests/stream_chunks_remainder.rs
ADD-SP, line 1: Please rename this file to `stream_chunks_timeout.rs`.

tokio-stream/tests/stream_chunks_remainder.rs
```diff
@@ -0,0 +1,26 @@
...
+    let inner = stream::iter([1, 2, 3, 4]).chain(stream::pending());
+    tokio::pin!(inner);
+
+    let chunked = (&mut inner).chunks_timeout(10, Duration::from_millis(20));
+    tokio::pin!(chunked);
+
+    tokio::select! {
+        Some(_chunk) = chunked.next() => {}
+        _ = tokio::time::sleep(Duration::from_millis(3)) => {}
+    }
```
ADD-SP: Could we poll the `chunked` manually? So that we can simplify the test by removing the timer and `tokio::select`.

ADD-SP (commented):
I saw you changed the title, is this intended?

tokio-stream/tests/stream_chunks_remainder.rs
muhamadazmy: Sounds good.

tokio-stream/tests/stream_chunks_remainder.rs
muhamadazmy: No, not really. I started with a copy of another file and did not notice. Sorry about that.

tokio-stream/src/stream_ext/chunks_timeout.rs
muhamadazmy: Sounds good to me. I agree that it we should remove it.

muhamadazmy:
@ADD-SP sorry about changing the title, that was indeed an accident. I use sapling which pushes title and description when submitting the PR. I will use the title you suggested.

muhamadazmy:
Dear @ADD-SP thank you so much for your tips and pointers. I like the tests much more now with the paused-time. Great feature that I have overlooked 😄 

Let me know if you still have any more comments.

ADD-SP (approved):
I have considered two different interfaces for this usecase.

# `into_remainder`

```rust
pub fn into_remainder(self: Pin<&mut Self>) -> Vec<T>;
```

## Pros

* This aligns to [`std::slice::ChunksExactMut:into_remainder`].

## Cons

* The receiver is `self: Pin<&mut Self>` instead of `self`, which is a bit different from the [`std::slice::ChunksExactMut:into_remainder`].
* The downstream can keep using it after calling into_remainder(). The semantics look a bit strange.
  ```rust
   pinned.as_mut().into_remainder();
   pinned.as_mut().into_remainder();
  ```

# `take_remainder`

```rust
pub fn take_remainder(self: &mut Pin<&mut Self>) -> Vec<T>;
```

## Pros

* The semantic looks more straightforward for downstream to keep using it after calling the `take_remainder`.
  ```rust
  let elems = pinned.take_remainder();
  let _ = pinned.poll_next(&mut cx);
  ```

## Cons

* The `&mut Pin<&mut Self>` is not a common pattern in Rust.

# Why choose `into_remainder`?

Because of the use case. The use case of this interface is to retrieve buffered items after `.chunked_timeout().await` is canceled, and the downstream is no longer going to keep polling it. For example:

```rust
tokio::select! {
    Some(batch) = chunked.next() => todo!(),
    _ = tokio::signal::ctrl_c() => (),
}

let elems = chunked.into_remainder();
```

For this use case, `into_remainder()` is better than `take_remainder() `semantically, because the downstream is not going to poll the chunked again, so consuming it makes more sense.


[`std::slice::ChunksExactMut:into_remainder`]: https://doc.rust-lang.org/1.91.0/std/slice/struct.ChunksExactMut.html#method.into_remainder

ADD-SP:
Thanks!

PR #7716, opened 2025-10-29
codec: remove unnecessary trait bounds on all `Framed` constructors

conradludgate:
## Motivation

Minor ergonomics, mostly. To avoid `Framed::from_parts(parts)` which always reserves 8KiB in the read and write buffers I have been using

```rust
fn my_from_parts<T, U>(parts: FramedParts<T, U>) -> Framed<T, U> {
    let mut framed = Framed::with_capacity(parts.io, parts.codec, 0);
    *framed.read_buffer_mut() = parts.read_buf;
    *framed.write_buffer_mut() = parts.write_buf;
    framed
}
```

It's a minor annoyance that the above function actually needs a `T: AsyncRead + AsyncWrite` bounds, whereas `Framed::from_parts(parts)` doesn't.

## Solution

Remove the bounds on all Framed constructors.

### DIFF
--- a/tokio-util/src/codec/framed.rs
+++ b/tokio-util/src/codec/framed.rs
@@ -124,9 +124,7 @@ impl<T, U> Framed<T, U> {
             },
         }
     }
-}
 
-impl<T, U> Framed<T, U> {
     /// Provides a [`Stream`] and [`Sink`] interface for reading and writing to this
     /// I/O object, using [`Decoder`] and [`Encoder`] to read and write the raw data.
     ///
--- a/tokio-util/src/codec/framed_read.rs
+++ b/tokio-util/src/codec/framed_read.rs
@@ -64,9 +64,7 @@ impl<T, D> FramedRead<T, D> {
             },
         }
     }
-}
 
-impl<T, D> FramedRead<T, D> {
     /// Returns a reference to the underlying I/O stream wrapped by
     /// `FramedRead`.
     ///
--- a/tokio-util/src/codec/framed_write.rs
+++ b/tokio-util/src/codec/framed_write.rs
@@ -61,9 +61,7 @@ impl<T, E> FramedWrite<T, E> {
             },
         }
     }
-}
 
-impl<T, E> FramedWrite<T, E> {
     /// Returns a reference to the underlying I/O stream wrapped by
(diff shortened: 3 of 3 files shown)


tokio-util/src/codec/framed.rs
```diff
@@ -41,10 +41,7 @@ pin_project! {
     }
 }
 
-impl<T, U> Framed<T, U>
-where
-    T: AsyncRead + AsyncWrite,
-{
+impl<T, U> Framed<T, U> {
```
ADD-SP, line 44: Could we merge this `impl` block into the Line#129 in `tokio-util/src/codec/framed.rs`?

https://github.com/tokio-rs/tokio/blob/4b516e332e07acaa5e572e0d74cbc97468df6191/tokio-util/src/codec/framed.rs#L129

tokio-util/src/codec/framed_read.rs
```diff
@@ -36,11 +36,7 @@ pin_project! {
 
 // ===== impl FramedRead =====
 
-impl<T, D> FramedRead<T, D>
-where
-    T: AsyncRead,
-    D: Decoder,
-{
+impl<T, D> FramedRead<T, D> {
```
ADD-SP, line 39: Ditto <https://github.com/tokio-rs/tokio/pull/7716/files#r2472323792>.

tokio-util/src/codec/framed_write.rs
```diff
@@ -35,10 +35,7 @@ pin_project! {
     }
 }
 
-impl<T, E> FramedWrite<T, E>
-where
-    T: AsyncWrite,
-{
+impl<T, E> FramedWrite<T, E> {
```
ADD-SP, line 38: Ditto <https://github.com/tokio-rs/tokio/pull/7716/files#r2472323792>.

tokio-util/src/codec/framed.rs
conradludgate: Yeah. I considered it. I wasn't sure if it would impact docs in a way that mattered but I will happily do that

ADD-SP: I think it should be safe as it just changes the layout of the document instead of the runtime behavior.

PR #7655, opened 2025-09-30
runtime: use cargo feature for taskdump instead of cfg

ADD-SP:
*close #5875 *

PR #7717, opened 2025-10-31
feat: USDT probes for tokio task events

conradludgate:
## Motivation

As discussed on [discord](https://discord.com/channels/500028886025895936/500336346770964480/1432426246700273766):

> My team isn't comfortable with the performance concerns of enabling tokio's tracing feature since we do actually use tracing in our application and such there is potentially a measurable cost of evaluating our EnvFilter every time in our performance critical services.

## Solution

Using Userspace Statically Defined Tracing (USDT) we expose lightweight probes that can be attached to at runtime with tools like bpftrace or dtrace. This is inspired by https://github.com/oxidecomputer/usdt.

The new functionality is behind a new unstable feature flag. Currently it only exposes some basic task events and not yet any resource events.

See USDT in the wild:
* https://github.com/oxidecomputer/tokio-dtrace
* https://github.com/aws/s2n-quic/blob/main/quic/s2n-quic-core/src/event/generated/metrics/probe.rs

### DIFF
--- a/tokio/src/util/usdt/macos.rs
+++ b/tokio/src/util/usdt/macos.rs
@@ -36,7 +36,7 @@ use std::arch::global_asm;
 
 cfg_rt!(
     #[inline(always)]
-    pub(super) fn task_details(task_id: u64, name: &str, file: &str, line: u32, col: u32) {
+    pub(super) fn task_details(task_id: u64, name: &str, file: &std::ffi::CStr, line: u32, col: u32) {
         unsafe extern "C" {
             #[link_name = "__dtrace_isenabled$tokio$task__details$v1"]
             fn task_details_enabled() -> i32;
@@ -55,13 +55,12 @@ cfg_rt!(
         if unsafe { task_details_enabled() } != 0 {
             // add nul bytes
             let name0 = [name.as_bytes(), b"\0"].concat();
-            let file0 = [file.as_bytes(), b"\0"].concat();
 
             unsafe {
                 __task_details(
                     task_id,
                     name0.as_ptr() as *const std::ffi::c_char,
-                    file0.as_ptr() as *const std::ffi::c_char,
+                    file.as_ptr(),
                     line,
                     col,
                 );
--- a/tokio/src/util/usdt/mod.rs
+++ b/tokio/src/util/usdt/mod.rs
@@ -48,12 +48,13 @@ cfg_usdt! {
         pub(crate) use usdt_impl::{waker_clone, waker_wake, waker_drop};
 
         #[inline(never)]
+        #[allow(clippy::incompatible_msrv)]
         pub(crate) fn start_task(kind: TaskKind, meta: SpawnMeta<'_>, id: Id, size: usize) {
             usdt_impl::task_start(id.as_u64(), kind as u8, size, meta.original_size);
             usdt_impl::task_details(
                 id.as_u64(),
                 meta.name.unwrap_or_default(),
-                meta.spawned_at.0.file(),
+                meta.spawned_at.0.file_as_c_str(),
                 meta.spawned_at.0.line(),
                 meta.spawned_at.0.column(),
             );
--- a/tokio/src/util/usdt/stapsdt.rs
+++ b/tokio/src/util/usdt/stapsdt.rs
@@ -136,12 +136,11 @@ cfg_rt!(
 
     // `inline(always)` is ok here since we only inline into `super::start_task` which is `inline(never)`
     #[inline(always)]
-    pub(super) fn task_details(task_id: u64, name: &str, file: &str, line: u32, col: u32) {
+    pub(super) fn task_details(task_id: u64, name: &str, file: &std::ffi::CStr, line: u32, col: u32) {
         #[cold]
-        fn task_details_inner(task_id: u64, name: &str, file: &str, line: u32, col: u32) {
+        fn task_details_inner(task_id: u64, name: &str, file: &std::ffi::CStr, line: u32, col: u32) {
             // add nul bytes
             let name0 = [name.as_bytes(), b"\0"].concat();
-            let file0 = [file.as_bytes(), b"\0"].concat();
 
             unsafe {
                 call_probe!(
@@ -151,7 +150,7 @@ cfg_rt!(
                     semaphore = sym __usdt_sema_tokio_task__details,
                     task_id = in(reg) task_id,
                     name = in(reg) name0.as_ptr(),
-                    file = in(reg) file0.as_ptr(),
+                    file = in(reg) file.as_ptr(),
                     line = in(reg) line,
                     col = in(reg) col,
                 );


tokio/Cargo.toml
```diff
@@ -106,6 +106,7 @@ socket2 = { version = "0.6.0", optional = true, features = ["all"] }
 # Requires `--cfg tokio_unstable` to enable.
 [target.'cfg(tokio_unstable)'.dependencies]
 tracing = { version = "0.1.29", default-features = false, features = ["std"], optional = true } # Not in full
+usdt = { version = "0.6.0", default-features = false, optional = true } # Not in full
```
hds: This shouldn't be a feature flag (because once they're in it's a breaking change for them to get removed, even if `tokio_unstable` isn't enabled.

Instead we should require 2 config flags: `tokio_unstable` and `tokio_usdt`.

This is similar to `tokio_taskdump`.

tokio/src/runtime/builder.rs
```diff
@@ -940,6 +940,9 @@ impl Builder {
     /// # }
     /// ```
     pub fn build(&mut self) -> io::Result<Runtime> {
+        #[cfg(all(tokio_unstable, feature = "usdt"))]
+        usdt::register_probes().map_err(|error| io::Error::new(io::ErrorKind::Other, error))?;
```
hds: The `usdt` documentation suggests that this should be exposed by libraries, but should only be invoked by the final binary.

tokio/src/util/usdt.rs
```diff
@@ -0,0 +1,88 @@
...
+        #[allow(non_snake_case)]
+        pub(crate) mod probes {
+            fn task__spawn(task_id: u64, name: &str, size: usize, file: &str, line: u32, col: u32) {}
+            fn task__block_on(task_id: u64, name: &str, size: usize, file: &str, line: u32, col: u32) {}
+            fn task__poll__start(task_id: u64) {}
+            fn task__poll__end(task_id: u64) {}
+            fn task__terminate(task_id: u64) {}
+
+            fn task__waker__clone(task_id: u64) {}
+            fn task__waker__wake(task_id: u64) {}
```
hds: Nice idea to represent waker as wake + drop and remove the need for wake_by_ref.

tokio/src/util/usdt.rs
```diff
@@ -0,0 +1,88 @@
...
+                probes::task__poll__end!(|| *this.task_id);
+                res
+            }
+        }
+
+        #[usdt::provider(provider = "tokio")]
+        #[allow(non_snake_case)]
+        pub(crate) mod probes {
+            fn task__spawn(task_id: u64, name: &str, size: usize, file: &str, line: u32, col: u32) {}
+            fn task__block_on(task_id: u64, name: &str, size: usize, file: &str, line: u32, col: u32) {}
```
hds: Maybe we need to benchmark, but I would consider having separate probes for spawning a task (or `block_on`) and the details associated with that task. This would remove the requirement that strings are copied through the kernel if all you want to do is get task spawning and poll times.

```suggestion
            fn task__spawn(task_id: u64) {}
            fn task__block_on(task_id: u64) {}
            fn task__details(task_id: u64, name: &str, size: usize, file: &str, line: u32, col: u32) {}
```

tokio/src/util/usdt.rs
conradludgate: I actually wasn't sure if I liked that 😅

tokio/src/util/usdt.rs
conradludgate: Hmm. I did have similar ideas but I'm not 100% sure what value there is in knowing that a task started without knowing the details of the task. You could already analyse busy/idle/schedule time with the poll/wake events alone.

tokio/src/runtime/builder.rs
conradludgate: In the crate docs

https://github.com/oxidecomputer/usdt/blob/9a58fa8a730daabda2633382cb96eaa4f9d9a531/usdt/src/lib.rs#L253-L260

> Alternatively, library developers may call this function during some initialization routines required by their library. There is no harm in calling this method multiple times, even in concurrent situations.

tokio/Cargo.toml
conradludgate: Fair 👍

conradludgate: Am I being silly or is there no `tokio_taskdump` cfg - instead it's `feature = "taskdump"`?

conradludgate: #7655

tokio/src/util/usdt.rs
hds: You miss the initial delay before the first poll if you don't know when the task was spawned.

tokio/src/runtime/builder.rs
hds: OK, fair enough.

tokio/Cargo.toml
hds: You're right. Things have changed and my memory appears to be wrong as to the initial reason. So this can stay as it was. Sorry.

tokio/src/util/usdt.rs
conradludgate: Right. Fair enough.

conradludgate:
Musings regarding usdt performance: https://github.com/oxidecomputer/usdt/issues/490

conradludgate:
I think I'm going to vendor the minimal subset of the code needed to support USDT on the platforms supported by the usdt crate. Removing any extra dependencies in the process.

I'd like to work with the oxide devs to improve this but currently usdt brings in a lot of unnecessary dependencies, and it also doesn't allow cross compilation. Lastly I want to rework the generated code to reduce any performance impact.

Because I will remove all dependencies, I'll move it back to a cfg tokio_usdt and remove the feature

hds:
Are we using checked in assembly and `usdt-impl` instead of the macros created by the `usdt` to avoid proc-macros? Or is there some other reason this is desireable?

Also, are the files in `utils/usdt` generated? Could we get some instructions on how included?

tokio/src/util/usdt.rs
```diff
@@ -0,0 +1,142 @@
+cfg_rt! {
```
hds, line 1: I think we probably want to rename this to `usdt/mod.rs` to keep in the style of the rest of the project.

conradludgate:
> Are we using checked in assembly and `usdt-impl` instead of the macros created by the `usdt` to avoid proc-macros? Or is there some other reason this is desireable?
> 
> Also, are the files in `utils/usdt` generated? Could we get some instructions on how included?

I had a lot of issues with using usdt directly.
1. It pulls in a lot of dependencies
2. On macos it runs dtrace at build time
    i. It is currently impossible to cross compile with it as a dependency due to the current code architecture
3. On linux I found that overly monomorphised probes would cause issues with linking, and usdt crate didn't give me the tools to work around that
4. I wanted to squeeze some more performance out of the asm.

The asm was originally generated using the proc macro, but has since been rewritten. Individual probes shouldn't need to be regenerated, but adding new probes will need some manual effort. I've tried to abstract the asm to some shared macros, I have some more work to do there. The macos code would require the most effort if new probes are needed but I will also document it.

Some notable changes I've made to the asm are
1. Not branching for trivial probes - eg `wake(task_id)` can just emit the NOP and skip checking if a probe is attached.
2. Outlining the probe if otherwise monomorphised (like the poll-start probe). In this case we do keep the branch, but the branch then just calls a function to keep the instruction count small and to keep the probe confined to 1 address.

What we lose by not using the usdt crate is the freebsd/illumos support. I don't have any freebsd/illumos setups available to test, and the code to support usdt on those platforms is the most complicated.

I don't consider any of these changes impossible for the usdt crate to support, but it's going to take effort and considerable rewrites of large amounts of their code. Improving the asm is relatively easy and I'd be comfortable sacrificing the build constraints when that's done. However I do consider the concise asm to be a blocker since we really intend for this feature to be as close to 0 overhead as possible

hds:
Thanks for the details! As you've said, I think it would be important to document these steps (maybe a README.md in the usdt directory).

For Illumos support maybe we can convince someone at Oxide to add it, since it would benefit them perhaps the most. (-:

.github/workflows/ci.yml
```diff
@@ -284,6 +284,41 @@ jobs:
...
+      - uses: Swatinem/rust-cache@v2
+      # Run `tokio` with "unstable" cfg flag.
+      - name: test tokio full --cfg unstable --usdt
+        run: |
+          set -euxo pipefail
+          cargo nextest run --features $TOKIO_STABLE_FEATURES,${{ matrix.extra_features }}
+          cargo test --doc --features $TOKIO_STABLE_FEATURES,${{ matrix.extra_features }}
+        working-directory: tokio
+        env:
+          RUSTFLAGS: --cfg tokio_unstable --cfg tokio_usdt -Dwarnings
```
Darksonn: We have moved away from adding new `--cfg` flags in cases like this. These days, we add a cargo feature, and fail the build if the cargo feature is enabled without also enabling `--cfg tokio_unstable`.

tokio/src/util/usdt/stapsdt.rs
```diff
@@ -0,0 +1,291 @@
...
+    };
+}
+
+#[inline(always)]
+pub(super) fn task_details(task_id: u64, name: &str, file: &str, line: u32, col: u32) {
+    #[cold]
+    fn task_details_inner(task_id: u64, name: &str, file: &str, line: u32, col: u32) {
+        // add nul bytes
+        let name0 = [name.as_bytes(), b"\0"].concat();
+        let file0 = [file.as_bytes(), b"\0"].concat();
```
Darksonn:
```suggestion
    fn task_details_inner(task_id: u64, name: &CStr, file: &CStr, line: u32, col: u32) {
```

Darksonn:
> Not branching for trivial probes - eg `wake(task_id)` can just emit the NOP and skip checking if a probe is attached.

Can you share more details about this? Do I understand correctly that this works by replacing nop instructions with a different instruction (call or jmp or similar)?

tokio/src/util/usdt/stapsdt.rs
Darksonn: Getting the filename of a `Location` as a `CStr` will be stable in 1.92.

conradludgate:
> > Not branching for trivial probes - eg `wake(task_id)` can just emit the NOP and skip checking if a probe is attached.
> 
> Can you share more details about this? Do I understand correctly that this works by replacing nop instructions with a different instruction (call or jmp or similar)?

That is correct. I'll try and verify the exact instructions that get replaced (iirc linux will use an interrupt and macos uses a function call), but I've observed in both macos and linux when using lldb that there is a nop instruction and some simple register ops for the generated code of wake_by_ref.

---

I'm unable to test it right now, but the eBPF docs claim that linux uses an interrupt: https://docs.ebpf.io/linux/concepts/usdt/#attaching-with-ebpf.

When testing on aarch64-apple-darwin, I see a NOP being replaced with a [FASTTRAP](https://github.com/apple/darwin-xnu/blob/2ff845c2e033bd0ff64b5b6aa6063a1f8f65aa32/bsd/arm/fasttrap_isa.h#L120) instruction.

tokio/src/util/usdt/stapsdt.rs
conradludgate: Some concerns:
1. This is the slow path - any conversions of &str -> &CStr should take place within this function.
2. I would rather not require nightly for the file location. We can wait 6 weeks.
3. Because the task name might contain a `\0` somewhere in the middle, it ends up being a fair bit of code and error handling to construct such a string.

conradludgate:
Here's some assembly differences, ignoring any label changes

<details>
<summary>wake_by_ref</summary>

```diff
.section .text.tokio::runtime::task::waker::wake_by_ref,"ax",@progbits
        .globl  tokio::runtime::task::waker::wake_by_ref
        .p2align        4
.type   tokio::runtime::task::waker::wake_by_ref,@function
tokio::runtime::task::waker::wake_by_ref:
        .cfi_startproc
        sub rsp, 24
        .cfi_def_cfa_offset 32
+       mov rax, qword ptr [rdi + 16]
+       mov rax, qword ptr [rax + 72]
+       mov rax, qword ptr [rdi + rax]
+       nop
        mov rax, qword ptr [rdi]
        lea rcx, [rsp + 8]
        lea rdx, [rsp + 16]
        .p2align        4
.LBB193_1:
        test al, 2
        jne .LBB193_2
        test al, 4
        jne .LBB193_4
        test al, 1
        jne .LBB193_8
        test rax, rax
        js .LBB193_14
        lea r8, [rax + 68]
        mov sil, 1
        jmp .LBB193_9
        .p2align        4
.LBB193_2:
        xor esi, esi
        mov r9, rcx
        xor r8d, r8d
        mov qword ptr [r9], r8
        cmp dword ptr [rsp + 8], 1
        je .LBB193_11
        jmp .LBB193_12
        .p2align        4
.LBB193_4:
        xor esi, esi
        mov r8, rax
        jmp .LBB193_9
.LBB193_8:
        mov r8, rax
        or r8, 4
        xor esi, esi
.LBB193_9:
        mov qword ptr [rsp + 8], 1
        mov r9, rdx
        mov qword ptr [r9], r8
        cmp dword ptr [rsp + 8], 1
        jne .LBB193_12
.LBB193_11:
        mov r8, qword ptr [rsp + 16]
        lock cmpxchg    qword ptr [rdi], r8
        jne .LBB193_1
.LBB193_12:
        test sil, sil
        je .LBB193_13
        mov rax, qword ptr [rdi + 16]
        add rsp, 24
        .cfi_def_cfa_offset 8
        jmp qword ptr [rax + 8]
.LBB193_13:
        .cfi_def_cfa_offset 32
        add rsp, 24
        .cfi_def_cfa_offset 8
        ret
.LBB193_14:
        .cfi_def_cfa_offset 32
        lea rdi, [rip + .Lanon.cd1a78ec98df3f76d56fd1466ac0099f.165]
        lea rdx, [rip + .Lanon.cd1a78ec98df3f76d56fd1466ac0099f.166]
        mov esi, 47
        call qword ptr [rip + core::panicking::panic@GOTPCREL]
```

</details>

<details>
<summary>poll</summary>

```diff
.section .text.tokio::runtime::task::raw::poll::hd96dfd798919c755,"ax",@progbits
	.p2align	4
.type	tokio::runtime::task::raw::poll::hd96dfd798919c755,@function
tokio::runtime::task::raw::poll::hd96dfd798919c755:
	.cfi_startproc
	.cfi_personality 155, DW.ref.rust_eh_personality
-	.cfi_lsda 27, .Lexception49
+	.cfi_lsda 27, .Lexception51
	push rbp
	.cfi_def_cfa_offset 16
	push r15
	.cfi_def_cfa_offset 24
	push r14
	.cfi_def_cfa_offset 32
	push r12
	.cfi_def_cfa_offset 40
	push rbx
	.cfi_def_cfa_offset 48
	sub rsp, 512
	.cfi_def_cfa_offset 560
	.cfi_offset rbx, -48
	.cfi_offset r12, -40
	.cfi_offset r14, -32
	.cfi_offset r15, -24
	.cfi_offset rbp, -16
	mov rbx, rdi
	call qword ptr [rip + tokio::runtime::task::state::State::transition_to_running::h51d78340084c6090@GOTPCREL]
	movzx eax, al
	lea rcx, [rip + .LJTI77_0]
	movsxd rax, dword ptr [rcx + 4*rax]
	add rax, rcx
	jmp rax
	mov rax, qword ptr [rip + tokio::runtime::task::waker::WAKER_VTABLE::h0490a160f2a7ec56@GOTPCREL]
	mov qword ptr [rsp], rax
	mov qword ptr [rsp + 8], rbx
	lea r14, [rbx + 32]
	mov rax, rsp
	mov qword ptr [rsp + 24], rax
	mov qword ptr [rsp + 32], 0
	mov qword ptr [rsp + 16], rax
	cmp dword ptr [rbx + 56], 0
	jne .LBB77_14
	mov rdi, qword ptr [rbx + 40]
	call qword ptr [rip + tokio::runtime::task::core::TaskIdGuard::enter::h31387a37d88abbe0@GOTPCREL]
-	lea rdi, [rbx + 64]
	mov qword ptr [rsp + 48], rax
+	mov r12, qword ptr [rbx + 40]
+	mov rax, qword ptr [rip + __usdt_sema_tokio_task__poll__start@GOTPCREL]
+	cmp word ptr [rax], 0
+	je .LBB77_5
+	mov rdi, r12
+	call qword ptr [rip + tokio::util::usdt::usdt_impl::task_poll_start::probe_inner::hda61800a23fcc40d@GOTPCREL]
+.LBB77_5:
+	lea rdi, [rbx + 64]
	lea rsi, [rsp + 16]
	call simple_echo_tcp::main::{{closure}}::h5671ec2a34b607f8
	mov ebp, eax
+	mov rax, qword ptr [rip + __usdt_sema_tokio_task__poll__end@GOTPCREL]
+	cmp word ptr [rax], 0
+	je .LBB77_8
+	mov rdi, r12
+	call qword ptr [rip + tokio::util::usdt::usdt_impl::task_poll_end::probe_inner::h476c2bd4bfb2eb31@GOTPCREL]
+.LBB77_8:
	lea rdi, [rsp + 48]
	call qword ptr [rip + <tokio::runtime::task::core::TaskIdGuard as core::ops::drop::Drop>::drop::hab02cca3e87fef69@GOTPCREL]
	test bpl, bpl
	je .LBB77_10
	mov rdi, rbx
	call qword ptr [rip + tokio::runtime::task::state::State::transition_to_idle::ha74e5e89ce88109b@GOTPCREL]
	movzx eax, al
	lea rcx, [rip + .LJTI77_1]
	movsxd rax, dword ptr [rcx + 4*rax]
	add rax, rcx
	jmp rax
	mov rdi, r14
	mov rsi, rbx
	call qword ptr [rip + tokio::runtime::scheduler::multi_thread::handle::<impl tokio::runtime::task::Schedule for alloc::sync::Arc<tokio::runtime::scheduler::multi_thread::handle::Handle>>::yield_now::hf966c4e666337dba@GOTPCREL]
	mov rdi, rbx
	call qword ptr [rip + tokio::runtime::task::state::State::ref_dec::h3033e08956b2d202@GOTPCREL]
	test al, al
	je .LBB77_55
	mov rdi, rbx
	call core::ptr::drop_in_place<tokio::runtime::task::core::Cell<simple_echo_tcp::main::{{closure}},alloc::sync::Arc<tokio::runtime::scheduler::multi_thread::handle::Handle>>>::h8c2b259dda47c6ee
	jmp .LBB77_54
	mov rdi, rbx
	call core::ptr::drop_in_place<tokio::runtime::task::core::Cell<simple_echo_tcp::main::{{closure}},alloc::sync::Arc<tokio::runtime::scheduler::multi_thread::handle::Handle>>>::h8c2b259dda47c6ee
.LBB77_54:
	mov esi, 384
	mov edx, 128
	mov rdi, rbx
	call qword ptr [rip + __rustc[de0091b922c53d7e]::__rust_dealloc@GOTPCREL]
	jmp .LBB77_55
	lea r14, [rbx + 32]
+	mov rax, qword ptr [rip + __usdt_sema_tokio_task__terminate@GOTPCREL]
+	cmp word ptr [rax], 0
+	je .LBB77_38
+	mov rdi, qword ptr [rbx + 40]
+	mov esi, 1
+	call qword ptr [rip + tokio::util::usdt::usdt_impl::task_terminate::probe_inner::h7db442807ccea7e2@GOTPCREL]
+.LBB77_38:
	mov dword ptr [rsp + 280], 2
	lea rsi, [rsp + 280]
	mov rdi, r14
	call tokio::runtime::task::core::Core<T,S>::set_stage::h6c06c0277547f365
.LBB77_39:
	xor eax, eax
.LBB77_41:
	mov rcx, qword ptr [rbx + 40]
	mov qword ptr [rsp + 56], rcx
	mov qword ptr [rsp + 64], rax
	mov qword ptr [rsp + 72], rdx
	mov dword ptr [rsp + 48], 1
	lea rsi, [rsp + 48]
	mov rdi, r14
	call tokio::runtime::task::core::Core<T,S>::set_stage::h6c06c0277547f365
	jmp .LBB77_42
.LBB77_10:
	mov rax, qword ptr [rip + __usdt_sema_tokio_task__terminate@GOTPCREL]
	cmp word ptr [rax], 0
	je .LBB77_12
	mov rdi, qword ptr [rbx + 40]
	xor esi, esi
	call qword ptr [rip + tokio::util::usdt::usdt_impl::task_terminate::probe_inner::h7db442807ccea7e2@GOTPCREL]
.LBB77_12:
	mov dword ptr [rsp + 48], 2
	lea rsi, [rsp + 48]
	mov rdi, r14
	call tokio::runtime::task::core::Core<T,S>::set_stage::h6c06c0277547f365
	xor ecx, ecx
.LBB77_25:
	mov qword ptr [rsp + 288], rcx
	mov qword ptr [rsp + 296], rax
	mov qword ptr [rsp + 304], rdx
	mov dword ptr [rsp + 280], 1
	lea rsi, [rsp + 280]
	mov rdi, r14
	call tokio::runtime::task::core::Core<T,S>::set_stage::h6c06c0277547f365
.LBB77_42:
	mov rdi, rbx
	call tokio::runtime::task::harness::Harness<T,S>::complete::hbafcc89c840d433c
.LBB77_55:
	add rsp, 512
	.cfi_def_cfa_offset 48
	pop rbx
	.cfi_def_cfa_offset 40
	pop r12
	.cfi_def_cfa_offset 32
	pop r14
	.cfi_def_cfa_offset 24
	pop r15
	.cfi_def_cfa_offset 16
	pop rbp
	.cfi_def_cfa_offset 8
	ret
	.cfi_def_cfa_offset 560
+	mov rax, qword ptr [rip + __usdt_sema_tokio_task__terminate@GOTPCREL]
+	cmp word ptr [rax], 0
+	je .LBB77_46
+	mov rdi, qword ptr [rbx + 40]
+	mov esi, 1
+	call qword ptr [rip + tokio::util::usdt::usdt_impl::task_terminate::probe_inner::h7db442807ccea7e2@GOTPCREL]
+.LBB77_46:
	mov dword ptr [rsp + 280], 2
	lea rsi, [rsp + 280]
	mov rdi, r14
	call tokio::runtime::task::core::Core<T,S>::set_stage::h6c06c0277547f365
	jmp .LBB77_39
.LBB77_14:
	lea rax, [rip + .Lanon.5efc90b22d074bac3f60c9ed09935ae4.75]
	mov qword ptr [rsp + 48], rax
	mov qword ptr [rsp + 56], 1
	mov qword ptr [rsp + 64], 8
	xorps xmm0, xmm0
	movups xmmword ptr [rsp + 72], xmm0
	lea rsi, [rip + .Lanon.5efc90b22d074bac3f60c9ed09935ae4.76]
	lea rdi, [rsp + 48]
	call qword ptr [rip + core::panicking::panic_fmt::h5138da2ef87be35b@GOTPCREL]
	ud2
	jmp .LBB77_51
	mov rdi, rax
	call qword ptr [rip + std::panicking::catch_unwind::cleanup::h80150e146981252a@GOTPCREL]
	jmp .LBB77_41
	call qword ptr [rip + core::panicking::panic_cannot_unwind::h2df093f6b1708ee2@GOTPCREL]
	mov rdi, rax
	call qword ptr [rip + std::panicking::catch_unwind::cleanup::h80150e146981252a@GOTPCREL]
	mov r15, rax
	test rax, rax
	je .LBB77_42
	mov r12, rdx
	mov rax, qword ptr [rdx]
	test rax, rax
	je .LBB77_30
	mov rdi, r15
	call rax
.LBB77_30:
	mov rsi, qword ptr [r12 + 8]
	test rsi, rsi
	je .LBB77_42
	mov rdx, qword ptr [r12 + 16]
	mov rdi, r15
	call qword ptr [rip + __rustc[de0091b922c53d7e]::__rust_dealloc@GOTPCREL]
	jmp .LBB77_42
	mov r14, rax
	mov rsi, qword ptr [r12 + 8]
	test rsi, rsi
	je .LBB77_35
	mov rdx, qword ptr [r12 + 16]
	mov rdi, r15
	jmp .LBB77_34
	call qword ptr [rip + core::panicking::panic_cannot_unwind::h2df093f6b1708ee2@GOTPCREL]
	mov r15, rax
-	lea rdi, [rsp + 48]
-	call qword ptr [rip + <tokio::runtime::task::core::TaskIdGuard as core::ops::drop::Drop>::drop::hab02cca3e87fef69@GOTPCREL]
+	mov rax, qword ptr [rip + __usdt_sema_tokio_task__poll__end@GOTPCREL]
+	cmp word ptr [rax], 0
+	je .LBB77_19
+	mov rdi, r12
+	call qword ptr [rip + tokio::util::usdt::usdt_impl::task_poll_end::probe_inner::h476c2bd4bfb2eb31@GOTPCREL]
	jmp .LBB77_19
-       call qword ptr [rip + core::panicking::panic_in_cleanup::h8f68387bb6cbbf54@GOTPCREL]
	mov rdi, rax
	call qword ptr [rip + std::panicking::catch_unwind::cleanup::h80150e146981252a@GOTPCREL]
	jmp .LBB77_41
	call qword ptr [rip + core::panicking::panic_cannot_unwind::h2df093f6b1708ee2@GOTPCREL]
.LBB77_51:
	mov r14, rax
	mov esi, 384
	mov edx, 128
	mov rdi, rbx
.LBB77_34:
	call qword ptr [rip + __rustc[de0091b922c53d7e]::__rust_dealloc@GOTPCREL]
.LBB77_35:
	mov rdi, r14
	call _Unwind_Resume@PLT
	mov r15, rax
.LBB77_19:
-	mov dword ptr [rsp + 280], 2
-	lea rsi, [rsp + 280]
+	lea rdi, [rsp + 48]
+	call qword ptr [rip + <tokio::runtime::task::core::TaskIdGuard as core::ops::drop::Drop>::drop::hab02cca3e87fef69@GOTPCREL]
+	jmp .LBB77_22
+	call qword ptr [rip + core::panicking::panic_in_cleanup::h8f68387bb6cbbf54@GOTPCREL]
+	mov r15, rax
+.LBB77_22:
	mov rdi, r14
-	call tokio::runtime::task::core::Core<T,S>::set_stage::h6c06c0277547f365
+	call core::ptr::drop_in_place<tokio::runtime::task::harness::poll_future::{{closure}}::Guard<simple_echo_tcp::main::{{closure}},alloc::sync::Arc<tokio::runtime::scheduler::current_thread::Handle>>>::he33e022ff410fd70
	mov rdi, r15
	call qword ptr [rip + std::panicking::catch_unwind::cleanup::h80150e146981252a@GOTPCREL]
	mov rcx, qword ptr [rbx + 40]
	jmp .LBB77_25
	call qword ptr [rip + core::panicking::panic_cannot_unwind::h2df093f6b1708ee2@GOTPCREL]
	call qword ptr [rip + core::panicking::panic_in_cleanup::h8f68387bb6cbbf54@GOTPCREL]
```

</details>

Darksonn:
The [docs found here](https://sourceware.org/systemtap/wiki/UserSpaceProbeImplementation) say that:

> SDT probes are designed to have a tiny runtime code and data footprint and no dynamic relocations.

But the assembly you shared contains relocations such as `__usdt_sema_tokio_task__poll__end@GOTPCREL`. Is the assembly correct?

conradludgate:
> The [docs found here](https://sourceware.org/systemtap/wiki/UserSpaceProbeImplementation) say that:
> 
> > SDT probes are designed to have a tiny runtime code and data footprint and no dynamic relocations.
> 
> But the assembly you shared contains relocations such as `__usdt_sema_tokio_task__poll__end@GOTPCREL`. Is the assembly correct?

I think this is a necessary part of how I'm currently avoiding over-monomorphisation. Each probe can only have one semaphore, so each probe callsite that wants to check the semaphore must necessarily need to relocate for the global.

---

I'd like to figure out why the monomorphisation is causing issues with the linker because then we could eliminate most of the semaphores entirely. 

I did also play around with trying to move the poll probes higher up the stack where the runtime is still polymorphic, but I wasn't too happy with it. Maybe I can rework/reconsider it.

conradludgate:
I've found the issue with monomorphisation. When multiple codegen units monomorphise the same function, they duplicate the notes. The linker later de-duplicates those functions but it unfortunately doesn't de-duplicate the notes. This leads to dangling notes.

The fix for now is to mark the notes as retained, which prevents the linker from garbage collecting these functions. Not ideal but it works.

With this change, the assembly now looks like

<details>
<summary>poll</summary>

```diff
.section .text.tokio::runtime::task::raw::poll::hd96dfd798919c755,"ax",@progbits
	.p2align	4
.type	tokio::runtime::task::raw::poll::hd96dfd798919c755,@function
tokio::runtime::task::raw::poll::hd96dfd798919c755:
	.cfi_startproc
	.cfi_personality 155, DW.ref.rust_eh_personality
-	.cfi_lsda 27, .Lexception49
+	.cfi_lsda 27, .Lexception48
	push rbp
	.cfi_def_cfa_offset 16
	push r15
	.cfi_def_cfa_offset 24
	push r14
	.cfi_def_cfa_offset 32
	push r12
	.cfi_def_cfa_offset 40
	push rbx
	.cfi_def_cfa_offset 48
	sub rsp, 512
	.cfi_def_cfa_offset 560
	.cfi_offset rbx, -48
	.cfi_offset r12, -40
	.cfi_offset r14, -32
	.cfi_offset r15, -24
	.cfi_offset rbp, -16
	mov rbx, rdi
	call qword ptr [rip + tokio::runtime::task::state::State::transition_to_running::h51d78340084c6090@GOTPCREL]
	movzx eax, al
	lea rcx, [rip + .LJTI77_0]
	movsxd rax, dword ptr [rcx + 4*rax]
	add rax, rcx
	jmp rax
	mov rax, qword ptr [rip + tokio::runtime::task::waker::WAKER_VTABLE::h0490a160f2a7ec56@GOTPCREL]
	mov qword ptr [rsp], rax
	mov qword ptr [rsp + 8], rbx
	lea r14, [rbx + 32]
	mov rax, rsp
	mov qword ptr [rsp + 24], rax
	mov qword ptr [rsp + 32], 0
	mov qword ptr [rsp + 16], rax
	cmp dword ptr [rbx + 56], 0
	jne .LBB77_14
	mov rdi, qword ptr [rbx + 40]
	call qword ptr [rip + tokio::runtime::task::core::TaskIdGuard::enter::h31387a37d88abbe0@GOTPCREL]
 	lea rdi, [rbx + 64]
	mov qword ptr [rsp + 48], rax
+	mov r12, qword ptr [rbx + 40]
+	nop
	lea rsi, [rsp + 16]
	call simple_echo_tcp::main::{{closure}}::h5671ec2a34b607f8
	mov ebp, eax
+	nop
	lea rdi, [rsp + 48]
	call qword ptr [rip + <tokio::runtime::task::core::TaskIdGuard as core::ops::drop::Drop>::drop::hab02cca3e87fef69@GOTPCREL]
	test bpl, bpl
	je .LBB77_10
	mov rdi, rbx
	call qword ptr [rip + tokio::runtime::task::state::State::transition_to_idle::ha74e5e89ce88109b@GOTPCREL]
	movzx eax, al
	lea rcx, [rip + .LJTI77_1]
	movsxd rax, dword ptr [rcx + 4*rax]
	add rax, rcx
	jmp rax
	mov rdi, r14
	mov rsi, rbx
	call qword ptr [rip + tokio::runtime::scheduler::multi_thread::handle::<impl tokio::runtime::task::Schedule for alloc::sync::Arc<tokio::runtime::scheduler::multi_thread::handle::Handle>>::yield_now::hf966c4e666337dba@GOTPCREL]
	mov rdi, rbx
	call qword ptr [rip + tokio::runtime::task::state::State::ref_dec::h3033e08956b2d202@GOTPCREL]
	test al, al
	je .LBB77_55
	mov rdi, rbx
	call core::ptr::drop_in_place<tokio::runtime::task::core::Cell<simple_echo_tcp::main::{{closure}},alloc::sync::Arc<tokio::runtime::scheduler::multi_thread::handle::Handle>>>::h8c2b259dda47c6ee
	jmp .LBB77_54
	mov rdi, rbx
	call core::ptr::drop_in_place<tokio::runtime::task::core::Cell<simple_echo_tcp::main::{{closure}},alloc::sync::Arc<tokio::runtime::scheduler::multi_thread::handle::Handle>>>::h8c2b259dda47c6ee
.LBB77_54:
	mov esi, 384
	mov edx, 128
	mov rdi, rbx
	call qword ptr [rip + __rustc[de0091b922c53d7e]::__rust_dealloc@GOTPCREL]
	jmp .LBB77_55
	lea r14, [rbx + 32]
+	mov rax, qword ptr [rbx + 40]
+	mov esx, 1
+	nop
	mov dword ptr [rsp + 280], 2
	lea rsi, [rsp + 280]
	mov rdi, r14
	call tokio::runtime::task::core::Core<T,S>::set_stage::h6c06c0277547f365
.LBB77_39:
	xor eax, eax
.LBB77_41:
	mov rcx, qword ptr [rbx + 40]
	mov qword ptr [rsp + 56], rcx
	mov qword ptr [rsp + 64], rax
	mov qword ptr [rsp + 72], rdx
	mov dword ptr [rsp + 48], 1
	lea rsi, [rsp + 48]
	mov rdi, r14
	call tokio::runtime::task::core::Core<T,S>::set_stage::h6c06c0277547f365
	jmp .LBB77_42
.LBB77_10:
+	mov rax, qword ptr [rbx + 40]
+	xor ecx, ecx
+	nop
	mov dword ptr [rsp + 48], 2
	lea rsi, [rsp + 48]
	mov rdi, r14
	call tokio::runtime::task::core::Core<T,S>::set_stage::h6c06c0277547f365
	xor ecx, ecx
.LBB77_25:
	mov qword ptr [rsp + 288], rcx
	mov qword ptr [rsp + 296], rax
	mov qword ptr [rsp + 304], rdx
	mov dword ptr [rsp + 280], 1
	lea rsi, [rsp + 280]
	mov rdi, r14
	call tokio::runtime::task::core::Core<T,S>::set_stage::h6c06c0277547f365
.LBB77_42:
	mov rdi, rbx
	call tokio::runtime::task::harness::Harness<T,S>::complete::hbafcc89c840d433c
.LBB77_55:
	add rsp, 512
	.cfi_def_cfa_offset 48
	pop rbx
	.cfi_def_cfa_offset 40
	pop r12
	.cfi_def_cfa_offset 32
	pop r14
	.cfi_def_cfa_offset 24
	pop r15
	.cfi_def_cfa_offset 16
	pop rbp
	.cfi_def_cfa_offset 8
	ret
	.cfi_def_cfa_offset 560
+	mov rax, qword ptr [rbx + 40]
+	mov ecx, 1
+	nop
	mov dword ptr [rsp + 280], 2
	lea rsi, [rsp + 280]
	mov rdi, r14
	call tokio::runtime::task::core::Core<T,S>::set_stage::h6c06c0277547f365
	jmp .LBB77_39
.LBB77_14:
	lea rax, [rip + .Lanon.5efc90b22d074bac3f60c9ed09935ae4.75]
	mov qword ptr [rsp + 48], rax
	mov qword ptr [rsp + 56], 1
	mov qword ptr [rsp + 64], 8
	xorps xmm0, xmm0
	movups xmmword ptr [rsp + 72], xmm0
	lea rsi, [rip + .Lanon.5efc90b22d074bac3f60c9ed09935ae4.76]
	lea rdi, [rsp + 48]
	call qword ptr [rip + core::panicking::panic_fmt::h5138da2ef87be35b@GOTPCREL]
	ud2
	jmp .LBB77_51
	mov rdi, rax
	call qword ptr [rip + std::panicking::catch_unwind::cleanup::h80150e146981252a@GOTPCREL]
	jmp .LBB77_41
	call qword ptr [rip + core::panicking::panic_cannot_unwind::h2df093f6b1708ee2@GOTPCREL]
	mov rdi, rax
	call qword ptr [rip + std::panicking::catch_unwind::cleanup::h80150e146981252a@GOTPCREL]
	mov r15, rax
	test rax, rax
	je .LBB77_42
	mov r12, rdx
	mov rax, qword ptr [rdx]
	test rax, rax
	je .LBB77_30
	mov rdi, r15
	call rax
.LBB77_30:
	mov rsi, qword ptr [r12 + 8]
	test rsi, rsi
	je .LBB77_42
	mov rdx, qword ptr [r12 + 16]
	mov rdi, r15
	call qword ptr [rip + __rustc[de0091b922c53d7e]::__rust_dealloc@GOTPCREL]
	jmp .LBB77_42
	mov r14, rax
	mov rsi, qword ptr [r12 + 8]
	test rsi, rsi
	je .LBB77_35
	mov rdx, qword ptr [r12 + 16]
	mov rdi, r15
	jmp .LBB77_34
	call qword ptr [rip + core::panicking::panic_cannot_unwind::h2df093f6b1708ee2@GOTPCREL]
	mov r15, rax
+	nop
	lea rdi, [rsp + 48]
	call qword ptr [rip + <tokio::runtime::task::core::TaskIdGuard as core::ops::drop::Drop>::drop::hab02cca3e87fef69@GOTPCREL]
	jmp .LBB77_19
	call qword ptr [rip + core::panicking::panic_in_cleanup::h8f68387bb6cbbf54@GOTPCREL]
	mov rdi, rax
	call qword ptr [rip + std::panicking::catch_unwind::cleanup::h80150e146981252a@GOTPCREL]
	jmp .LBB77_41
	call qword ptr [rip + core::panicking::panic_cannot_unwind::h2df093f6b1708ee2@GOTPCREL]
.LBB77_51:
	mov r14, rax
	mov esi, 384
	mov edx, 128
	mov rdi, rbx
.LBB77_34:
	call qword ptr [rip + __rustc[de0091b922c53d7e]::__rust_dealloc@GOTPCREL]
.LBB77_35:
	mov rdi, r14
	call _Unwind_Resume@PLT
	mov r15, rax
.LBB77_19:
+	mov rax, qword ptr [rbx + 40]
+	mov ecx, 2
+	nop
	mov dword ptr [rsp + 280], 2
	lea rsi, [rsp + 280]
	mov rdi, r14
	call tokio::runtime::task::core::Core<T,S>::set_stage::h6c06c0277547f365
	mov rdi, r15
	call qword ptr [rip + std::panicking::catch_unwind::cleanup::h80150e146981252a@GOTPCREL]
	mov rcx, qword ptr [rbx + 40]
	jmp .LBB77_25
	call qword ptr [rip + core::panicking::panic_cannot_unwind::h2df093f6b1708ee2@GOTPCREL]
	call qword ptr [rip + core::panicking::panic_in_cleanup::h8f68387bb6cbbf54@GOTPCREL]
```

</details>

conradludgate:
~~After tinkering some more, I found a proper solution that doesn't require using retained. Same assembly, but we still get to keep linker GC.~~

tokio/src/util/usdt/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+                        finish_task(self.0, TerminateKind::Panicked);
+                    }
+                }
+
+                let this = self.project();
+
+                usdt_impl::task_poll_start(this.task_id.as_u64());
+                let guard = Guard(*this.task_id);
+                let res = this.inner.poll(cx);
+                drop(guard);
```
martin-g: This calls `finish_task(..., Panicked)` (line 109) and below at line 122 if the result is Ready it calls `finish_task(..., Success)`. 
Same for `task_poll_end()` but unconditionally.
Is this intentional ?

tokio/src/util/usdt/macos.rs
```diff
@@ -0,0 +1,168 @@
...
+        extern "C" {
+            #[link_name = "__dtrace_probe$tokio$task__poll__end$v1$75696e7436345f74"]
+            fn __task_poll_end(task_id: core::ffi::c_ulonglong);
+        }
+
+        unsafe { __task_poll_end(task_id) }
+    }
+
+    #[inline(always)]
+    pub(crate) fn task_terminate(task_id: u64, reason: u8) {
```
martin-g:
```suggestion
    pub(super) fn task_terminate(task_id: u64, reason: u8) {
```
Does it need to be crate-private ? AFAIS it is used only locally (util/usdt/mod.rs)

tokio/src/util/usdt/macos.rs
```diff
@@ -0,0 +1,168 @@
...
+                    file0.as_ptr() as *const std::ffi::c_char,
+                    line,
+                    col,
+                );
+            }
+        }
+    }
+
+    #[inline(always)]
+    pub(super) fn task_start(task_id: u64, spawned: u8, size: usize, original_size: usize) {
```
martin-g:
```suggestion
    pub(super) fn task_start(task_id: u64, kind: u8, size: usize, original_size: usize) {
```

tokio/src/util/usdt/macos.rs
```diff
@@ -0,0 +1,168 @@
...
+        extern "C" {
+            #[link_name = "__dtrace_probe$tokio$task__start$v1$75696e7436345f74$75696e74385f74$75696e747074725f74$75696e747074725f74"]
+            fn __task_start(task_id: u64, kind: u8, size: usize, original_size: usize);
+        }
+
+        unsafe {
+            __task_start(task_id, spawned, size, original_size);
```
martin-g:
```suggestion
            __task_start(task_id, kind, size, original_size);
```

tokio/src/task/local.rs
```diff
@@ -441,6 +441,7 @@ cfg_rt! {
                 ))]
                 let future = task::trace::Trace::root(future);
                 let id = task::Id::next();
+                usdt::start_task(usdt::TaskKind::Spawn, meta, id, std::mem::size_of::<F>());
```
martin-g:
```suggestion
                usdt::start_task(usdt::TaskKind::SpawnLocal, meta, id, std::mem::size_of::<F>());
```

.github/workflows/ci.yml
```diff
@@ -1112,7 +1116,7 @@ jobs:
       - name: check-external-types
         env:
           RUSTFLAGS: --cfg tokio_unstable -Dwarnings
-          RUSTDOCFLAGS: --cfg tokio_unstable
+          RUSTDOCFLAGS: --cfg tokio_unstable ${{ matrix.extra_flags }}
```
martin-g: There is no `extra_flags` matrix key in this workflow

tokio/src/util/usdt/mod.rs
conradludgate: Good catch - it was meant to be a forget. Thanks

PR #7724, opened 2025-11-06
runtime: disable io-uring on `EPERM`

mox692:
## Motivation
Closes https://github.com/tokio-rs/tokio/issues/7691.

## Solution
If we get `EPERM` at the initial io-uring syscall (io_uring_setup), then we set the `Unsupported` flag internally, which disables subsequent io-uring operations.

### DIFF
--- a/tokio/src/runtime/io/driver/uring.rs
+++ b/tokio/src/runtime/io/driver/uring.rs
@@ -194,6 +194,13 @@ impl Handle {
                     self.set_uring_state(State::Unsupported);
                     Ok(false)
                 }
+                // If we get EPERM, io-uring syscalls may be blocked (for example, by seccomp).
+                // In this case, we try to fall back to spawn_blocking for this and future operations.
+                // See also: https://github.com/tokio-rs/tokio/issues/7691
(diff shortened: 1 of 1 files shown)


Darksonn (approved):
Testing this in CI sounds difficult, but can we perform a manual test?

ADD-SP (approved):
We can open another PR for testing if possible.

ADD-SP (approved):
Tested locally — this fix works as expected.

```bash
docker run --rm rust:latest bash -c "$(cat <<"EOF"
mkdir -p /tmp/test/src
cd /tmp/test
ln -s Cargo.toml a.toml

cat <<"EOF2" > Cargo.toml
[package]
edition = "2024"
name = "tokio-io-uring-symlink"

[dependencies]
tokio = { git = "https://github.com/tokio-rs/tokio.git", rev = "7f959e1049c8d252f4a1e18bbbabc94cfd91120d", features = ["full", "io-uring"] }
EOF2

cat <<"EOF2" > src/main.rs
#[tokio::main]
async fn main() {
    tokio::fs::File::open("a.toml").await.unwrap();
}
EOF2

RUSTFLAGS='--cfg tokio_unstable' cargo run
EOF
)"
```

PR #7727, opened 2025-11-09
util: fix assertion

YichiZhang0613:
## Motivation

This function's assertion is overly restrictive. It should be failed when MAX_ENTRIES is exceeded as message says. But it use  < MAX_ENTRIES instead of <= MAX_ENTRIES. According to MAX_ENTRIES's definition and usage, it should be <=.
```rust
pub fn insert_at(&mut self, value: T, when: Instant) -> Key {
        assert!(self.slab.len() < MAX_ENTRIES, "max entries exceeded");
...
```

## Solution

The solution is to change `assert!(self.slab.len() < MAX_ENTRIES, "max entries exceeded");` into `assert!(self.slab.len() <= MAX_ENTRIES, "max entries exceeded");`

### DIFF
--- a/tokio-util/src/time/delay_queue.rs
+++ b/tokio-util/src/time/delay_queue.rs
@@ -536,7 +536,7 @@ impl<T> DelayQueue<T> {
     /// [type]: #
     #[track_caller]
     pub fn insert_at(&mut self, value: T, when: Instant) -> Key {
-        assert!(self.slab.len() < MAX_ENTRIES, "max entries exceeded");
+        assert!(self.slab.len() <= MAX_ENTRIES, "max entries exceeded");
 
         // Normalize the deadline. Values cannot be set to expire in the past.
         let when = self.normalize_deadline(when);


ADD-SP (commented):
There is a `self.slab.insert` following this assertion. Let's say the `self.slab.len()` equals to `MAX_ENTRIES`, the following `self.slab.insert` will exceed the the `MAX_ENTRIES`. So I don't think we need to fix this assertion.

https://github.com/tokio-rs/tokio/blob/d3072b70576fd3fdfda5208c8d1d40aae80c8aec/tokio-util/src/time/delay_queue.rs#L539-L551

YichiZhang0613:
> There is a `self.slab.insert` following this assertion. Let's say the `self.slab.len()` equals to `MAX_ENTRIES`, the following `self.slab.insert` will exceed the the `MAX_ENTRIES`. So I don't think we need to fix this assertion.
> 
> https://github.com/tokio-rs/tokio/blob/d3072b70576fd3fdfda5208c8d1d40aae80c8aec/tokio-util/src/time/delay_queue.rs#L539-L551

OK.

PR #7731, opened 2025-11-12
docs: minor improvements for the README

mattiapitossi:
## Motivation

I would like to propose few minor corrections to the README (I think this can close also #6798, which has been there from some time)


Darksonn (approved):
Thanks.

PR #7734, opened 2025-11-13
tokio: enable more tests in Miri

RalfJung:
`poll_fns` has been fixed.

And as for that `runtime.rs` test: when the main thread is done, Miri gives the other threads a chance to finish before reporting a leak. At least in local testing, that seems to be enough for this test.

We can also remove the `-Zmiri-retag-fields` flag; it has been the default for more than 2 years already.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1058,6 +1058,7 @@ jobs:
           CARGO_TARGET_WASM32_WASIP1_RUNNER: "wasmtime run --"
           CARGO_TARGET_WASM32_WASIP1_THREADS_RUNNER: "wasmtime run -W bulk-memory=y -W threads=y -S threads=y --"
           RUSTFLAGS: --cfg tokio_unstable -Dwarnings -C target-feature=+atomics,+bulk-memory -C link-args=--max-memory=67108864
+          RUSTDOCFLAGS: -C link-args=--max-memory=67108864
 
       - name: WASI test tokio-stream
         run: cargo test -p tokio-stream --target ${{ matrix.target }} --features time,net,io-util,sync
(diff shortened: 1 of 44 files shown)


RalfJung:
poll_fns failed... does this not use the latest nightly?

EDIT: Ah, I found where I have to bump it.

.github/workflows/ci.yml
```diff
@@ -372,7 +372,7 @@ jobs:
           cargo miri nextest run --features full --lib --no-fail-fast
         working-directory: tokio
         env:
-          MIRIFLAGS: -Zmiri-disable-isolation -Zmiri-strict-provenance -Zmiri-retag-fields
+          MIRIFLAGS: -Zmiri-disable-isolation -Zmiri-strict-provenance
```
Darksonn, line 375: You are removing the retag fields argument?

RalfJung: As explained in the commit message, `-Zmiri-retag-fields` is the default (and has been the default for a long time already): https://github.com/rust-lang/miri/pull/2985.

tokio/src/runtime/tests/mod.rs
```diff
@@ -77,7 +77,7 @@ cfg_not_loom! {
     mod inject;
     mod queue;
 
-    #[cfg(not(miri))]
+    #[cfg(not(miri))] // takes a really long time with miri
     mod task_combinations;
```
RalfJung, line 81: Interestingly, there's `cfg!(miri)` logic inside this module, even though the module is disabled on Miri anyway...

Darksonn: I probably ran it under miri when I wrote the test.

ADD-SP (approved):

PR #7735, opened 2025-11-13
use parking_lot primitives also with Miri

RalfJung:
These got disabled in https://github.com/tokio-rs/tokio/pull/5919, but without any comment as to what was going wrong on Miri. Let's see if whatever problem occurred back then still exists.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -372,7 +372,7 @@ jobs:
           cargo miri nextest run --features full --lib --no-fail-fast
         working-directory: tokio
         env:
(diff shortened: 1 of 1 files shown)


RalfJung:
Not sure if that's what triggered the original PR, but we do hit https://github.com/Amanieu/parking_lot/issues/477 now since Tokio sets `-Zmiri-strict-provenance`.

Darksonn:
It sounds like we should just continue to use the stdlib mutex then?

RalfJung:
Yeah, probably best to wait until https://github.com/Amanieu/parking_lot/issues/477 is resolved.

Darksonn:
I will close. We can reopen if that is ever resolved.

PR #7736, opened 2025-11-15
experiment: use miri affected by data race UB bug

RalfJung:
See https://github.com/tokio-rs/tokio/issues/7712#issuecomment-3527731809 for context.


RalfJung:
So this would indeed have been caught, nice.

Issue #7729, opened 2025-11-10
`TcpListener::bind` example in `TcpSocket` docs is no longer correct


opara-tor:
**Version**
1.48.0

**Platform**
N/A

**Description**
The [docs for `TcpSocket` say](https://github.com/tokio-rs/tokio/blob/d709df25717e292609d93d4002a98232e63eccd3/tokio/src/net/tcp/socket.rs#L49):

> Calling `TcpListener::bind("127.0.0.1:8080")` is equivalent to:
>
> ```rust
> use tokio::net::TcpSocket;
>
> use std::io;
>
> #[tokio::main]
> async fn main() -> io::Result<()> {
>     let addr = "127.0.0.1:8080".parse().unwrap();
>
>     let socket = TcpSocket::new_v4()?;
>     // On platforms with Berkeley-derived sockets, this allows to quickly
>     // rebind a socket, without needing to wait for the OS to clean up the
>     // previous one.
>     //
>     // On Windows, this allows rebinding sockets which are actively in use,
>     // which allows "socket hijacking", so we explicitly don't set it here.
>     // https://docs.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse
>     socket.set_reuseaddr(true)?;
>     socket.bind(addr)?;
>
>     let listener = socket.listen(1024)?;
>
>     Ok(())
> }
> ```

But the backlog parameter was changed in mio in https://github.com/tokio-rs/mio/pull/1896. So 1024 is no longer used by `TcpListener::bind`.

I'm not sure how you'd like to fix this, since copying the logic from mio probably wouldn't be useful. Maybe a comment should be added here saying that the backlog size is actually platform-dependent?

ADD-SP:
> Maybe a comment should be added here saying that the backlog size is actually platform-dependent?

This looks reasonable to me. PR is welcome.

PR #7738, opened 2025-11-16
net: clarify platform-dependent backlog in TcpSocket docs

seakerOner:
Fixes: #7729

## Motivation

The documentation example for reproducing the behavior of  `TcpListener::bind` uses a hardcoded backlog value of 1024. However, Tokio now relies on Mio's platform-dependent default backlog configuration, so the value in the example was misleading.

## Solution

This commit updates the documentation by adding a note explaining that the
actual backlog used by `TcpListener::bind` varies by platform and that the
example value is only illustrative.


ADD-SP (approved):

ADD-SP:
Thanks!

PR #7742, opened 2025-11-22
net: clarify the drop behavior of `unix::OwnedWriteHalf`

darth-raijin:
## Motivation
Provide clarity in documentation, based on feedback from #7723. The goal is to provide context on dropping the write half of the socket connection.

## Solution
Inserted single word. 
What could be added, to provide even more context could be, the [reason](https://github.com/tokio-rs/tokio/issues/7723#issuecomment-3491309933) for this behavior but that could become too verbose.


ADD-SP (approved):
Thanks!

Issue #7751, opened 2025-12-01
TcpStream::set_linger can lead to blocking


dankmeme01:
**Version**
tokio v1.48.0

**Platform**
Linux dankpc 6.17.8-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Fri, 14 Nov 2025 06:54:07 +0000 x86_64 GNU/Linux

**Description**
`tokio::net::TcpStream` has a `set_linger` function, which sets the `SO_LINGER` option on the underlying socket, causing the `close()` and `shutdown()` syscalls to block on Linux until all pending data is sent, for a specified period of time. This means that dropping or shutting down such a stream will block the current worker thread until the timeout expires or all data is sent.

Simple example using https://tcpbin.com as a test endpoint

```rust
use std::{io, time::Duration};

use tokio::{io::AsyncWriteExt, net::TcpStream};
use tracing::info;

#[tokio::main]
async fn main() -> io::Result<()> {
    tracing_subscriber::fmt::init();

    let mut stream = TcpStream::connect("45.79.112.203:4242").await?;
    stream.set_linger(Some(Duration::from_secs(10)))?;
    info!("Socket connected");
    stream.write_all(&vec![0u8; 1024 * 1024]).await?;
    info!("Data sent");
    drop(stream);
    info!("Socket closed");

    Ok(())
}
```

```
2025-12-01T18:47:39.843644Z  INFO testing: Socket connected
2025-12-01T18:47:40.740568Z  INFO testing: Data sent
2025-12-01T18:47:41.132441Z  INFO testing: Socket closed
```

You can notice how the time difference between the last 2 logs is relatively large,  but becomes nearly zero if you comment the `set_linger` call.

Darksonn:
Perhaps we should mark our `set_linger` method deprecated with a reference to this problem?

ADD-SP:
Another option is to warn users about this blocking behavior in the documentation, while also informing them that they can mitigate this issue by disabling `SO_LINGER` before calling `close()`.

Darksonn:
I think the only thing that `SO_LINGER` changes is the behavior on `close()`, so telling users to disable it before calling `close()` doesn't make much sense to me.

ADD-SP:
> I think the only thing that `SO_LINGER` changes is the behavior on `close()`, so telling users to disable it before calling `close()` doesn't make much sense to me.

Make sense to me. The entire point of `SO_LINGER` is for `close()`.

[...]

PR #7752, opened 2025-12-02
net: deprecate `TcpStream::set_linger`

Darksonn:
I did not deprecate `TcpStream::linger()` because just reading the option seems harmless.

Closes: #7751

### DIFF
--- a/tokio/src/net/tcp/socket.rs
+++ b/tokio/src/net/tcp/socket.rs
@@ -424,7 +424,7 @@ impl TcpSocket {
     /// incorrect as it leads to blocking the thread when the socket is closed. For more details,
     /// please see:
     ///
-    /// > Volumes of communications have been devoted the the intricacies of `SO_LINGER` versus
+    /// > Volumes of communications have been devoted to the intricacies of `SO_LINGER` versus
     /// > non-blocking (`O_NONBLOCK`) sockets. From what I can tell, the final word is: don't do
     /// > it. Rely on the `shutdown()`-followed-by-`read()`-eof technique instead.
     /// >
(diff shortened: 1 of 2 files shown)


tokio/src/net/tcp/socket.rs
```diff
@@ -419,6 +419,17 @@ impl TcpSocket {
     ///
     /// If `SO_LINGER` is not specified, and the socket is closed, the system handles the call in a
     /// way that allows the process to continue as quickly as possible.
+    ///
+    /// This option is deprecated because setting `SO_LINGER` on a socket used with Tokio is always
+    /// incorrect as it leads to blocking the thread when the socket is closed. For more details,
+    /// please see:
+    ///
+    /// > Volumes of communications have been devoted the the intricacies of `SO_LINGER` versus
```
martin-g: I understand that this is a copy from somewhere and you may want to have the quote as is!
```suggestion
    /// > Volumes of communications have been devoted the intricacies of `SO_LINGER` versus
```

tokio/src/net/tcp/stream.rs
```diff
@@ -1278,9 +1278,20 @@ impl TcpStream {
         /// If `SO_LINGER` is not specified, and the stream is closed, the system handles the call in a
         /// way that allows the process to continue as quickly as possible.
         ///
+        /// This option is deprecated because setting `SO_LINGER` on a socket used with Tokio is
+        /// always incorrect as it leads to blocking the thread when the socket is closed. For more
+        /// details, please see:
+        ///
+        /// > Volumes of communications have been devoted the the intricacies of `SO_LINGER` versus
```
martin-g:
```suggestion
        /// > Volumes of communications have been devoted the intricacies of `SO_LINGER` versus
```

tokio/src/net/tcp/socket.rs
Darksonn: Looks like one of them should actually be "to".

ADD-SP (approved):

Issue #7753, opened 2025-12-03
tokio-stream: Rustc bug causing trouble when using StreamExt::collect


TethysSvensson:
**Version**
```
tokio-stream v0.1.17
tokio v1.48.0
```

**Description**
There is a bug in rustc, which can cause trouble when using `StreamExt::collect` together types that have an `unsafe impl Send` and which contain a `dyn Any` as a parameter. An example of such a type is `rootcause::Report`.

Here is a minimal proof-of-concept:

```rust
struct MyType<C: ?Sized>(std::marker::PhantomData<C>);

unsafe impl<C: ?Sized> Send for MyType<C> {}
unsafe impl<C: ?Sized> Sync for MyType<C> {}

async fn foo<S>(s: S) -> Vec<MyType<dyn std::any::Any>>
where
    S: 'static + Send + tokio_stream::Stream<Item = MyType<dyn std::any::Any>>,
{
    use tokio_stream::StreamExt;
    s.collect().await
}

fn needs_send<T: Send>(_value: T) {}

fn bar<S>(v: S)
where
    S: 'static + Send + tokio_stream::Stream<Item = MyType<dyn std::any::Any>>,
{
    needs_send(foo(v));
}
```

This gives the following error:

```
error: implementation of `tokio_stream::stream_ext::collect::sealed::FromStreamPriv` is not general enough
  --> src/lib.rs:20:5
   |
20 |     needs_send(foo(v));
   |     ^^^^^^^^^^^^^^^^^^ implementation of `tokio_stream::stream_ext::collect::sealed::FromStreamPriv` is not general enough
   |
   = note: `Vec<MyType<(dyn Any + '0)>>` must implement `tokio_stream::stream_ext::collect::sealed::FromStreamPriv<MyType<(dyn Any + 'static)>>`, for any lifetime `'0`...
   = note: ...but it actually implements `tokio_stream::stream_ext::collect::sealed::FromStreamPriv<MyType<dyn Any>>`

error: could not compile `foo` (lib) due to 1 previous error
```

While this looks to me like it's fairly obviously a bug in rustc, it is probably still something you would want to work-around if possible.

Luckily it is very easy to work around it, you just need to tweak the definition of `Collect` slightly to make it look like this:

```rust
    pub struct Collect<T, U, C>
    {
        #[pin]
        stream: T,
        collection: C,
        _output: PhantomData<U>,
        // Make this future `!Unpin` for compatibility with async trait methods.
        #[pin]
[...]

PR #7754, opened 2025-12-03
stream: work around the rustc bug in `StreamExt::collect`

TethysSvensson:
This is a fix for #7753.

As stated in that issue, there is a rustc bug, which can cause issues when interacting with types that have a manual `unsafe impl Send` and which contains a parameter set to `dyn Any`.

This PR solves the issue by removing the trait bounds on the struct itself, which makes `rustc` able to correctly infer that the combined struct is `Send`.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,7 +1,7 @@
 only_if: $CIRRUS_TAG == '' && ($CIRRUS_PR != '' || $CIRRUS_BRANCH == 'master' || $CIRRUS_BRANCH =~ 'tokio-.*')
 auto_cancellation: $CIRRUS_BRANCH != 'master' && $CIRRUS_BRANCH !=~ 'tokio-.*'
 freebsd_instance:
-  image_family: freebsd-14-2
+  image_family: freebsd-14-3
 env:
   RUST_STABLE: stable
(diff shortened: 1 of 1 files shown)


tokio-stream/src/stream_ext.rs
```diff
@@ -916,7 +916,7 @@ pub trait StreamExt: Stream {
     /// assert_eq!(Err("no"), values);
     /// # }
     /// ```
-    fn collect<T>(self) -> Collect<Self, T>
+    fn collect<T>(self) -> Collect<Self, T, T::InternalCollection>
```
ADD-SP, line 919: Just double-checking: I know the type `Collect` is not public, so I think adding a new generic parameter on it is not a breaking change. Is that correct?

ADD-SP (commented):
Overall look good to me, just a minor question.

tokio-stream/src/stream_ext.rs
TethysSvensson: That is my understanding as well, yes.

TethysSvensson:
> Please sync changes from the base branch to fix CI issues.

Done :)

ADD-SP (approved):
Thanks!

PR #7757, opened 2025-12-04
rt: improve spawn_blocking scalability with sharded queue

alex:
The blocking pool's task queue was protected by a single mutex, causing severe contention when many threads spawn blocking tasks concurrently. This resulted in nearly linear degradation: 16 concurrent threads took ~18x longer than a single thread.

Replace the single-mutex queue with a sharded queue that distributes tasks across 16 lock-protected shards. The implementation adapts to concurrency levels by using fewer shards when thread count is low, maintaining cache locality while avoiding contention at scale.

Benchmark results (spawning 100 batches of 16 tasks per thread):

| Concurrency | Before   | After   | Improvement |
|-------------|----------|---------|-------------|
| 1 thread    | 13.3ms   | 17.8ms  | +34%        |
| 2 threads   | 26.0ms   | 20.1ms  | -23%        |
| 4 threads   | 45.4ms   | 27.5ms  | -39%        |
| 8 threads   | 111.5ms  | 20.3ms  | -82%        |
| 16 threads  | 247.8ms  | 22.4ms  | -91%        |

The slight overhead at 1 thread is due to the sharded infrastructure, but this is acceptable given the dramatic improvement at higher concurrency where the original design suffered from lock contention.

(Notwithstanding that this shows as a commit from claude, every line is human reviewed. If there's a mistake, it's Alex's fault.)

Closes https://github.com/tokio-rs/tokio/issues/2528.

### DIFF
--- a/tokio/src/runtime/blocking/pool.rs
+++ b/tokio/src/runtime/blocking/pool.rs
@@ -412,8 +412,14 @@ impl Spawner {
 
                 // Double-check conditions after acquiring the lock
                 if shared.shutdown {
-                    // Queue was shutdown while we were acquiring the lock
-                    // The task is already in the queue, so it will be drained during shutdown
+                    // Shutdown raced with our push. The task is in the
+                    // sharded queue but workers may have already exited.
+                    // Drain it here so mandatory tasks still run.
+                    drop(shared);
+                    while let Some(task) = self.inner.queue.pop(0) {
+                        self.inner.metrics.dec_queue_depth();
+                        task.shutdown_or_run_if_mandatory();
+                    }
                     return Ok(());
                 }
 
@@ -523,49 +529,31 @@ impl Inner {
                 break;
             }
 
-            // IDLE: Wait for new tasks
+            // IDLE: Wait for new tasks (spurious wakeups handled internally)
             self.metrics.inc_num_idle_threads();
 
-            loop {
-                match self.queue.wait_for_task(preferred_shard, self.keep_alive) {
-                    WaitResult::Task(task) => {
-                        // Got a task, process it
-                        self.metrics.dec_num_idle_threads();
-                        self.metrics.dec_queue_depth();
-                        task.run();
-                        // Go back to busy loop
-                        break;
-                    }
-                    WaitResult::Shutdown => {
-                        // Shutdown initiated
-                        self.metrics.dec_num_idle_threads();
-                        break 'main;
-                    }
-                    WaitResult::Timeout => {
-                        // Timed out, exit this thread
-                        self.metrics.dec_num_idle_threads();
+            match self.queue.wait_for_task(preferred_shard, self.keep_alive) {
+                WaitResult::Task(task) => {
+                    self.metrics.dec_num_idle_threads();
+                    self.metrics.dec_queue_depth();
+                    task.run();
+                }
+                WaitResult::Shutdown => {
+                    self.metrics.dec_num_idle_threads();
+                    break 'main;
+                }
+                WaitResult::Timeout => {
+                    self.metrics.dec_num_idle_threads();
 
-                        // Clean up thread handle
-                        let mut shared = self.shared.lock();
-                        if !shared.shutdown {
-                            let my_handle = shared.worker_threads.remove(&worker_thread_id);
-                            join_on_thread =
-                                std::mem::replace(&mut shared.last_exiting_thread, my_handle);
-                        }
+                    // Clean up thread handle
+                    let mut shared = self.shared.lock();
+                    if !shared.shutdown {
+                        let my_handle = shared.worker_threads.remove(&worker_thread_id);
+                        join_on_thread =
+                            std::mem::replace(&mut shared.last_exiting_thread, my_handle);
+                    }
 
-                        // Exit the main loop and terminate the thread
-                        break 'main;
-                    }
-                    WaitResult::Spurious => {
-                        // Spurious wakeup, check for tasks and continue waiting
-                        if let Some(task) = self.queue.pop(preferred_shard) {
-                            self.metrics.dec_num_idle_threads();
-                            self.metrics.dec_queue_depth();
-                            task.run();
-                            break;
-                        }
-                        // Continue waiting
-                    }
+                    break 'main;
                 }
             }
         }
--- a/tokio/src/runtime/blocking/sharded_queue.rs
+++ b/tokio/src/runtime/blocking/sharded_queue.rs
@@ -23,7 +23,13 @@ use std::time::Duration;
 use super::pool::Task;
 
 /// Number of shards. Must be a power of 2.
+/// Under loom, use a single shard to keep the state space tractable —
+/// the concurrency properties we need to verify (condvar signaling,
+/// shutdown ordering) are independent of shard count.
+#[cfg(not(loom))]
 const NUM_SHARDS: usize = 16;
+#[cfg(loom)]
+const NUM_SHARDS: usize = 1;
 
 /// A single shard containing a queue protected by its own mutex.
 struct Shard {
@@ -94,8 +100,10 @@ pub(super) struct ShardedQueue {
     /// Global condition variable for worker notifications.
     /// We use a single condvar to avoid the complexity of per-shard waiting.
     condvar: Condvar,
-    /// Mutex to pair with the condvar. Only held during wait, not during push/pop.
-    condvar_mutex: Mutex<()>,
+    /// Mutex paired with the condvar. Protects the notification counter
+    /// (`num_notify`), which tracks how many tasks have been pushed and
+    /// need to be picked up by idle workers.
+    condvar_mutex: Mutex<u32>,
 }
 
 impl ShardedQueue {
@@ -106,7 +114,7 @@ impl ShardedQueue {
             push_index: AtomicUsize::new(0),
             shutdown: AtomicBool::new(false),
             condvar: Condvar::new(),
-            condvar_mutex: Mutex::new(()),
+            condvar_mutex: Mutex::new(0),
         }
     }
 
@@ -131,13 +139,17 @@ impl ShardedQueue {
     }
 
     /// Notify one waiting worker that a task is available.
+    ///
+    /// Increments the notification counter under `condvar_mutex` and signals
+    /// the condvar. The counter acts as a persistent notification that cannot
+    /// be lost — even if no worker is currently waiting on the condvar, the
+    /// next worker to enter `wait_for_task` will see the counter and know
+    /// there is work to do.
     pub(super) fn notify_one(&self) {
+        let mut guard = self.condvar_mutex.lock();
+        *guard += 1;
+        drop(guard);
         self.condvar.notify_one();
-    }
-
-    /// Notify all waiting workers (used during shutdown).
-    pub(super) fn notify_all(&self) {
-        self.condvar.notify_all();
     }
 
     /// Try to pop a task, checking the preferred shard first, then others.
@@ -156,8 +168,14 @@ impl ShardedQueue {
 
     /// Set the shutdown flag and wake all workers.
     pub(super) fn shutdown(&self) {
-        self.shutdown.store(true, Release);
-        self.notify_all();
+        // Set the flag while holding condvar_mutex so that any worker
+        // currently inside `wait_for_task` (which also holds condvar_mutex
+        // while checking) is guaranteed to see the flag on its next check.
+        {
+            let _guard = self.condvar_mutex.lock();
+            self.shutdown.store(true, Release);
+        }
+        self.condvar.notify_all();
     }
 
     /// Check if shutdown has been initiated.
@@ -165,49 +183,46 @@ impl ShardedQueue {
         self.shutdown.load(Acquire)
     }
 
-    /// Wait for a task with timeout.
+    /// Wait for a task notification with timeout. Returns when a task has
+    /// been pushed (the caller should then `pop`), shutdown occurs, or the
+    /// wait times out.
+    ///
+    /// Uses a notification counter under `condvar_mutex` to prevent lost
+    /// wakeups — the same pattern as the original single-mutex blocking pool.
     pub(super) fn wait_for_task(&self, preferred_shard: usize, timeout: Duration) -> WaitResult {
-        if self.is_shutdown() {
-            return WaitResult::Shutdown;
-        }
+        let mut guard = self.condvar_mutex.lock();
 
-        // Try to pop without waiting first
-        if let Some(task) = self.pop(preferred_shard) {
-            return WaitResult::Task(task);
-        }
+        loop {
+            if self.is_shutdown() {
+                return WaitResult::Shutdown;
+            }
 
-        // Acquire the condvar mutex before waiting
-        let guard = self.condvar_mutex.lock();
+            if *guard > 0 {
+                // A notification is pending — a task was pushed.
+                *guard -= 1;
+                drop(guard);
+                // Pop outside the condvar_mutex to avoid holding two locks.
+                if let Some(task) = self.pop(preferred_shard) {
+                    return WaitResult::Task(task);
+                }
+                // The task was already consumed in the caller's BUSY loop
+                // (race between push+notify and the worker's pop loop).
+                // Re-enter the wait.
+                guard = self.condvar_mutex.lock();
+                continue;
+            }
 
-        // Double-check shutdown and tasks after acquiring lock, as state may
-        // have changed while we were waiting for the lock
-        if self.is_shutdown() {
-            return WaitResult::Shutdown;
-        }
-        if let Some(task) = self.pop(preferred_shard) {
-            return WaitResult::Task(task);
-        }
+            let (g, timeout_result) = self.condvar.wait_timeout(guard, timeout).unwrap();
+            guard = g;
 
-        // Wait for notification or timeout
-        let (guard, wait_timeout_result) = self.condvar.wait_timeout(guard, timeout).unwrap();
-        let timed_out = wait_timeout_result.timed_out();
-
-        // Drop the lock before doing further work
-        drop(guard);
-
-        if self.is_shutdown() {
-            return WaitResult::Shutdown;
-        }
-
-        // Try to get a task
-        if let Some(task) = self.pop(preferred_shard) {
-            return WaitResult::Task(task);
-        }
-
-        if timed_out {
-            WaitResult::Timeout
-        } else {
-            WaitResult::Spurious
+            if timeout_result.timed_out() && *guard == 0 {
+                // Double-check: shutdown may have raced with the timeout.
+                if self.is_shutdown() {
+                    return WaitResult::Shutdown;
+                }
+                return WaitResult::Timeout;
+            }
+            // Woken by notify or spurious wakeup — loop back to check.
         }
     }
 }
@@ -220,6 +235,4 @@ pub(super) enum WaitResult {
     Timeout,
     /// Shutdown was initiated.
     Shutdown,
-    /// Spurious wakeup, no task found.
-    Spurious,
 }


alex:
(FreeBSD failures look unrelated.)

tokio/src/runtime/blocking/pool.rs
```diff
@@ -403,52 +399,64 @@ impl Spawner {
             return Err(SpawnError::ShuttingDown);
         }
 
-        shared.queue.push_back(task);
+        // Get thread count for adaptive sharding
+        let num_threads = self.inner.metrics.num_threads();
```
ADD-SP: You may want the number of worker thread instead of the number of live thread in the blocking pool.

tokio/src/runtime/blocking/sharded_queue.rs
```diff
@@ -0,0 +1,285 @@
...
+
+use super::pool::Task;
+
+/// Number of shards. Must be a power of 2.
+const NUM_SHARDS: usize = 16;
+
+/// A single shard containing a queue protected by its own mutex.
+struct Shard {
+    /// The task queue for this shard.
+    queue: Mutex<VecDeque<Task>>,
```
ADD-SP, line 37: Compared to `std::sync::mpsc`, what advantages does `Mutex<VecDeque>` have in this scenario?

alex: Hmm, how would you imagine using mpsc? The overall design is to have multiple workers all pulling from a shared queue -- this just shards that queue so there's less contention, but multiple workers can still grab tasks (and multiple threads add them) concurrent.

If there was a mpmc module we could use that to pass tasks around and delete most of this (assuming it had good scalability)

I think if we wanted an alternate design, we could use something like crossbeam-queue's SegQueue.

Would you prefer I tried that?

ADD-SP: I think you mean the `mpsc::Receiver` doesn't allow threads to pull tasks from other shards. If so, I agree with you.

alex: Yes, exactly.

tokio/src/runtime/blocking/sharded_queue.rs
```diff
@@ -0,0 +1,285 @@
...
+
+    /// Check if shutdown has been initiated.
+    pub(super) fn is_shutdown(&self) -> bool {
+        self.shutdown.load(Acquire)
+    }
+
+    /// Wait for a task with timeout.
+    pub(super) fn wait_for_task(&self, preferred_shard: usize, timeout: Duration) -> WaitResult {
+        // Try to pop without waiting first
+        if let Some(task) = self.pop(preferred_shard) {
```
martin-g: Here you prefer the task than the shutdown.
After getting the lock below (line 224+) it is the opposite - first the shutdown and then the task.
I think the shutdown should be checked first in both places.

tokio/src/runtime/blocking/sharded_queue.rs
```diff
@@ -0,0 +1,285 @@
...
+        let index = self.push_index.fetch_add(1, Relaxed) & mask;
+
+        // Update max_shard_pushed BEFORE pushing the task. This ensures that
+        // by the time total_len is incremented (making the task "visible"),
+        // pop() will know to check this shard. Otherwise there's a race where
+        // pop() sees total_len > 0 but max_shard_pushed doesn't include this shard.
+        self.max_shard_pushed.fetch_max(index, Release);
+
+        self.shards[index].push(task);
+        self.total_len.fetch_add(1, Release);
```
martin-g: should this be the last ?!
At https://github.com/tokio-rs/tokio/pull/7757/files#diff-3eec86ec6c78c488331acce56c72f6b6a5c46af0a90c8e7f0aee1244e982565aR160 it is the first thing to check.
It is possible that a task is pushed in the previous line in threadA and another thread (B) tries to pop but since total_len is still 0 it will return None.
If you move the total_len++ before the push then it will have a bigger chance to see the new task and still return None if the push is too late.

tokio/src/runtime/blocking/pool.rs
```diff
@@ -403,52 +399,64 @@ impl Spawner {
...
+                            }
+                            Err(e) => {
+                                // The OS refused to spawn the thread and there is no thread
+                                // to pick up the task that has just been pushed to the queue.
+                                return Err(SpawnError::NoThreads(e));
+                            }
                         }
                     }
                 }
+            } else {
```
martin-g, line 453:
```suggestion
            } else if self.inner.metrics.num_idle_threads() > 0 {
```
Notify only if there is an idle worker ?!

tokio/src/runtime/blocking/sharded_queue.rs
alex: Good call, will fix.

tokio/src/runtime/blocking/sharded_queue.rs
alex: I'm not sure it matters, if pop() has a false negative, we'll go into wait_for_task.

The more I look at it, I'm not sure that it even makes sense to track total_len?

tokio/src/runtime/blocking/pool.rs
alex: We're already inside `num_idle == 0`. Does it matter?

tokio/src/runtime/blocking/sharded_queue.rs
alex: Yeah, benchmarking it shows no benefits, which makes sense. I'm going to drop it entirely :-)

martin-g: > The more I look at it, I'm not sure that it even makes sense to track total_len?

+1
I also thought the same while reading the code.

martin-g:
Please rebase to latest `master` to get the fix for the FreeBSD failures.

tokio/src/runtime/blocking/pool.rs
```diff
@@ -403,52 +399,64 @@ impl Spawner {
...
+                self.inner.queue.notify_one();
             }
         } else {
-            // Notify an idle worker thread. The notification counter
-            // is used to count the needed amount of notifications
-            // exactly. Thread libraries may generate spurious
-            // wakeups, this counter is used to keep us in a
-            // consistent state.
+            // There are idle threads waiting, notify one
             self.inner.metrics.dec_num_idle_threads();
```
martin-g: I wonder whether this is still needed here.
At https://github.com/tokio-rs/tokio/pull/7757/files#diff-c632261b41085469310e4840a71f9582257c348bf23807251dde960fc985b21cR536 the number of idle threads is decremented for all kinds of WaitResult.

AFAIU here the number is decremented, then the sharded queue is notified and then after wait_for_task() finishes it is decremented again.

alex: I think I agree we're double counting, fixed.

tokio/src/runtime/blocking/pool.rs
```diff
@@ -505,90 +510,81 @@ impl Inner {
...
+                        self.metrics.dec_num_idle_threads();
+
+                        // Clean up thread handle
+                        let mut shared = self.shared.lock();
+                        if !shared.shutdown {
+                            let my_handle = shared.worker_threads.remove(&worker_thread_id);
+                            join_on_thread =
+                                std::mem::replace(&mut shared.last_exiting_thread, my_handle);
+                        }
+                        drop(shared);
```
ADD-SP: Emm, why do we have to drop the lock before the `break 'main`?

alex: We don't, it's a bad holdover from the previous code. Fixing!

tokio/src/runtime/blocking/sharded_queue.rs
```diff
@@ -0,0 +1,271 @@
...
+    /// When a task is pushed, we increment this counter and signal the condvar.
+    /// When a worker wakes up, it decrements this counter (if positive) to "claim"
+    /// the notification. This helps workers distinguish between:
+    /// - Waking up because a task was pushed (counter was positive)
+    /// - Spurious wakeup from the condvar (counter was zero)
+    ///
+    /// Multiple workers may race to decrement the counter, which is fine - the
+    /// worker that successfully decrements will likely find the task, while
+    /// others will re-check and either find other tasks or go back to sleep.
+    num_notify: AtomicUsize,
```
ADD-SP: It looks like the current design still works even removing this field. Did I miss anything?

alex: I think you're right -- I naively ported it from the original, but it doesn't make much sense. Will remove assuming tests still pass.

tokio/src/runtime/blocking/pool.rs
```diff
@@ -505,90 +510,80 @@ impl Inner {
...
-                // NOTE: Technically we should also do num_notify++ and notify again,
-                // but since we're shutting down anyway, that won't be necessary.
-                break;
-            }
+        // Drain remaining tasks if shutting down
+        if self.queue.is_shutdown() {
+            self.queue.drain(|task| {
+                self.metrics.dec_queue_depth();
+                task.shutdown_or_run_if_mandatory();
+            });
```
ADD-SP: Why do we need to have each exiting thread drain all shards of the queue?

alex: How else would you do it? No worker owns a shard exclusively, this seems reliable enough. It could probably be simplified to just `pop()` in a loop if that's easier?

(I don't think the perf for shutdown matters much?)

ADD-SP: Just want to know if you've considered other solutions. Yes, shutdown performance isn't a major concern, but if it's not too difficult to improve, it's still worth optimizing.

alex: I didn't think especially hard about it -- this was definitely correct which was my focus.

Because (right now) the workers don't know what any other worker is doing, they have to at least _look_ at each shard. I think maybe you could have the pool do more coordination for shutdown, but I think it'd be more complex.

ADD-SP: Make sense to me.

tokio/src/runtime/blocking/sharded_queue.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// adaptively choose how many shards to distribute across.
+    pub(super) fn push(&self, task: Task, num_threads: usize) {
+        let num_shards = effective_shards(num_threads);
+        let mask = num_shards - 1;
+        let index = self.push_index.fetch_add(1, Relaxed) & mask;
+
+        // Update max_shard_pushed BEFORE pushing the task.
+        self.max_shard_pushed.fetch_max(index, Release);
+
+        self.shards[index].push(task);
```
ADD-SP: With the `Release` ordering, compiler might reorder the `self.shards[index].push(task)` and `fetch_max`, which means that the `.push(task)` might be *sequenced-before* the `fetch_max`.

alex: I think you're right. (I hate atomic orderings :-/) AcqRel I think is what we want.

tokio/src/runtime/blocking/sharded_queue.rs
```diff
@@ -0,0 +1,227 @@
...
+//!
+//! The queue adapts to the current concurrency level by using fewer shards
+//! when there are few threads, which improves cache locality and reduces
+//! lock contention on the active shards.
+
+use crate::loom::sync::{Condvar, Mutex};
+
+use std::collections::VecDeque;
+use std::sync::atomic::Ordering::{AcqRel, Acquire, Relaxed, Release};
+use std::sync::atomic::{AtomicBool, AtomicUsize};
```
mox692: Could we use loom's sync primitive and add a small test for this data structure? That would probably be something like [this](https://github.com/tokio-rs/tokio/blob/97d06ae1a69a38dfe20e65b5972f456781f2e0c1/tokio/src/runtime/tests/loom_multi_thread/queue.rs).

tokio/src/runtime/blocking/sharded_queue.rs
```diff
@@ -0,0 +1,227 @@
...
+            loop {
+                let mut queue = shard.queue.lock();
+                if let Some(task) = queue.pop_front() {
+                    drop(queue);
+                    f(task);
+                } else {
+                    break;
+                }
+            }
+        }
```
mox692:
```suggestion
        for shard in &self.shards {
            let mut queue = shard.queue.lock();
            while let Some(task) = queue.pop_front() {
                f(task);
            }
        }
```

ADD-SP: We should unlock the shard before executing the task.

tokio/src/runtime/blocking/sharded_queue.rs
```diff
@@ -0,0 +1,227 @@
...
+        // have changed while we were waiting for the lock
+        if self.is_shutdown() {
+            return WaitResult::Shutdown;
+        }
+        if let Some(task) = self.pop(preferred_shard) {
+            return WaitResult::Task(task);
+        }
+
+        // Wait for notification or timeout
+        let result = self.condvar.wait_timeout(guard, timeout).unwrap();
```
ADD-SP:
```suggestion
        let (guard, wait_timeout_result) = self.condvar.wait_timeout(guard, timeout).unwrap();
```

tokio/src/runtime/blocking/sharded_queue.rs
```diff
@@ -0,0 +1,227 @@
...
+        let guard = self.condvar_mutex.lock();
+
+        // Double-check shutdown and tasks after acquiring lock, as state may
+        // have changed while we were waiting for the lock
+        if self.is_shutdown() {
+            return WaitResult::Shutdown;
+        }
+        if let Some(task) = self.pop(preferred_shard) {
+            return WaitResult::Task(task);
+        }
```
ADD-SP: Here, we attempt to acquire the shard lock while holding the condvar lock, this is the nested locking pattern. In general, we should avoid this pattern as it is error-prone.

tokio/src/runtime/blocking/sharded_queue.rs
```diff
@@ -0,0 +1,227 @@
...
+
+        // Wait for notification or timeout
+        let result = self.condvar.wait_timeout(guard, timeout).unwrap();
+        let timed_out = result.1.timed_out();
```
ADD-SP:
```suggestion
        let timed_out = wait_timeout_result.timed_out();
```

tokio/src/runtime/blocking/sharded_queue.rs
```diff
@@ -0,0 +1,227 @@
...
+
+        // Drop the lock before doing further work
+        drop(result.0);
```
ADD-SP:
```suggestion
        drop(guard);
```

tokio/src/runtime/blocking/sharded_queue.rs
alex: Hmm, is there a preferred pattern to avoid the nested locking?

In this case we want to ensure that when we wait for a notification, there wasn't already a task that's made pending concurrently.

tokio/src/runtime/blocking/sharded_queue.rs
mox692: That's probably valid, operations from other threads could technically happen at this point.

alex:
(I don't think the netlify failures are related)

alex:
No problem -- rebase was to pick up a fix for the netlify failures.

Let me know if there's other experiments it'd be useful for me to try.

tokio/src/runtime/blocking/sharded_queue.rs
Darksonn: Nested locking is fine as long as locks are always taken in the same order.

tokio/src/runtime/blocking/sharded_queue.rs
```diff
@@ -0,0 +1,209 @@
...
+        queue.pop_front()
+    }
+}
+
+/// A sharded queue that distributes tasks across multiple shards.
+pub(super) struct ShardedQueue {
+    /// The shards - each with its own mutex-protected queue.
+    shards: [Shard; NUM_SHARDS],
+    /// Atomic counter for round-robin task distribution.
+    push_index: AtomicUsize,
```
Darksonn, line 97: Randomness is likely to scale better than a shared integer, as all CPUs have to contend on accessing this integer. You may compare with the watch channel, which has similar logic to use randomness.

tokio/src/runtime/blocking/sharded_queue.rs
```diff
@@ -0,0 +1,209 @@
...
+    condvar: Condvar,
+    /// Mutex to pair with the condvar. Only held during wait, not during push/pop.
+    condvar_mutex: Mutex<()>,
+}
+
+/// Calculate the effective number of shards to use based on thread count.
+/// Uses fewer shards at low concurrency for better cache locality.
+#[inline]
+fn effective_shards(num_threads: usize) -> usize {
+    match num_threads {
```
Darksonn: This logic seems error-prone and likely to lead to missed tasks. Does it actually matter for your benchmark?

tokio/src/runtime/blocking/sharded_queue.rs
```diff
@@ -0,0 +1,209 @@
...
+    pub(super) fn push(&self, task: Task, num_threads: usize) {
+        let num_shards = effective_shards(num_threads);
+        let mask = num_shards - 1;
+        let index = self.push_index.fetch_add(1, Relaxed) & mask;
+
+        // Update max_shard_pushed BEFORE pushing the task.
+        // AcqRel ensures the subsequent push cannot be reordered before this.
+        self.max_shard_pushed.fetch_max(index, AcqRel);
+
+        self.shards[index].push(task);
```
Darksonn, line 138: We can consider a loop that unlocks the mutex when allocation is required to avoid blocking `pop()` calls on `push()` calls that are busy inside `malloc()`.

tokio/src/runtime/blocking/sharded_queue.rs
Darksonn: This is still wrong. Consider the following scenario:
```
Thread A                   Thread B            Thread C
                                               preferred_shard = 0
                           preferred_shard = 1
max_shard_pushed = 0
shards[0].push(_)
condvar.notify_one()
                           wakes up ...
                           max_shard = 0
max_shard_pushed = 1
shards[1].push(_)
condvar.notify_one()
                                               wakes up ...
                                               max_shard = 1
                                               shards[0].pop() = Some
                           shards[0].pop() = None
```
In this case, Thread B does not check `shards[1]` because it read `max_shard` with the value of zero. This means that two tasks were spawned, but only one gets picked up.

Well, I guess in principle Thread C will see the second task after it finishes executing the first one.

alex: Hmm, I think it's always the case that if a push happens concurrently with a pop that the pop might miss it, and we'll have to "fall back" to catching it in the `wait_for_task` loop.

I think in principle we could address this one by reloading `max_shard_pushed`, but of course you can still have a race condition.

In the scenario you've got here, what would happen is that after thread B returns `None` from `pop`, it'll go `wait_for_task` and then the task will get picked up.

tokio/src/runtime/blocking/sharded_queue.rs
alex: Oh I like it, will import that logic.

tokio/src/runtime/blocking/sharded_queue.rs
alex: It matters at N_THREADS=1 -- I don't personally care about that case at all, if we're ok with a small pessimization there (10% iirc?), I'd be delighted to delete this max shard logic and just always use a fixed number.

tokio/src/runtime/blocking/sharded_queue.rs
alex: I'm not sure I understand.

tokio/src/runtime/blocking/sharded_queue.rs
alex: Done

tokio/src/runtime/blocking/sharded_queue.rs
Darksonn: Sorry this was incomplete. Right now, if the `VecDeque` needs to reallocate, you call `malloc` while holding the mutex. We may want to consider avoiding that, as it can block others for a long time.

tokio/src/runtime/blocking/sharded_queue.rs
Darksonn: > Hmm, I think it's always the case that if a push happens concurrently with a pop that the pop might miss it, and we'll have to "fall back" to catching it in the `wait_for_task` loop.

Your `notify_one()` call ensures that for every `push()`, there will be a subsequent call to `pop()` that is not concurrent and hence guaranteed to see the pushed message. So there's at least one thread that's guaranteed to pick up each message.

If we imagine that the `max_shard_pushed` logic was removed, then Thread B would in fact be *guaranteed* to see the message in `shards[1]`.

* `shards[1].push(_)` on A happens-before `shards[0].pop() = Some` on thread C because thread C is the thread waken up by the second `notify_one()` call.
* `shards[0].pop() = Some` on thread C happens-before `shards[0].pop() = None` on thread B, since otherwise thread B would have gotten `Some` when calling `pop()`.
* After `shards[0].pop() = None`, thread B would attempt to call `shards[1].pop()`

So by this logic, the `shards[1].pop()` call would in fact happen after `shards[1].push(_)`, and is hence guaranteed to see the message that was pushed.

tokio/src/runtime/blocking/sharded_queue.rs
```diff
@@ -91,14 +109,33 @@ impl ShardedQueue {
...
+    fn next_push_index(&self, num_shards: usize) -> usize {
+        crate::runtime::context::thread_rng_n(num_shards as u32) as usize
+    }
+
+    /// Select the next shard index for pushing a task -- when the RNG is not
+    /// available.
+    #[cfg(not(all(
+        not(loom),
+        any(feature = "macros", all(feature = "sync", feature = "rt"))
+    )))]
```
Darksonn: I believe the randomness state is only conditional on `#[cfg(any(feature = "rt", feature = "macros"))]`, and this condition is always true when this code is compiled.

tokio/src/runtime/blocking/sharded_queue.rs
alex: Does this sound ok to you? I'd love to delete it because its responsible for a _lot_ of the complexity.

tokio/src/runtime/blocking/sharded_queue.rs
alex: I'm not sure I understand how we could do that -- we have a `Mutex<VecDeque<Task>>`. There's no way to get the `&mut VecDeque<_>` needed to do the malloc without also holding the lock.

(We could switch the whole thing to something like a https://docs.rs/crossbeam-queue/latest/crossbeam_queue/struct.SegQueue.html if we want to avoid the mutex, and don't mind a 3p dependency)

Darksonn: You can note down the required length, unlock, call `VecDeque::with_capacity`, relock, check if still required, and then move items over to the new `VecDeque`.

tokio/src/runtime/blocking/sharded_queue.rs
alex: heh, the conditions were inconsistent between the RNG field, the RNG function, and what watch used. fixed.

tokio/src/runtime/blocking/sharded_queue.rs
alex: Oooh, clever.

tokio/src/runtime/blocking/sharded_queue.rs
alex: I've gone ahead and dropped this behavior.

alex:
rebased and resolved merge conflicts. performance continues to look good! let me know if there's anything I can do to move this forward!

alex:
I just rebased to resolve a conflict and it looks like one of the loom tests is now failing -- will investigate later this week.

alex:
Thank you for the reviews!

Darksonn:
Thank you for pushing through with this even though reviews sometimes took a while. I think it's a great improvement to the runtime, and the PR has turned out really nice.

PR #7761, opened 2025-12-05
io-uring: Support batching

Daksh14:
## Motivation

Batching allows us to submit multiple opcodes in a single io_uring_enter system call, we extend the API to handle a list of CQEs for more efficient I/O

## Solution

Extend uring driver and `op` struct to support batching

### DIFF
--- a/tokio/src/fs/read_uring.rs
+++ b/tokio/src/fs/read_uring.rs
@@ -12,12 +12,12 @@ const PROBE_SIZE: usize = 32;
 const PROBE_SIZE_U32: u32 = PROBE_SIZE as u32;
 
 // batch size for batched reads
-const BATCH_SIZE: usize = 128;
+const BATCH_SIZE: usize = 200;
 
 // Max bytes we can read using io uring submission at a time
 // SAFETY: cannot be higher than u32::MAX for safe cast
 // Set to read max 64 MiB at time
-pub(crate) const MAX_READ_SIZE: usize = 64 * 1024 * 1024;
+pub(crate) const MAX_READ_SIZE: usize = 64 * 1024;
 
 pub(crate) async fn read_uring(path: &Path) -> io::Result<Vec<u8>> {
     let file = OpenOptions::new().read(true).open(path).await?;
@@ -40,11 +40,9 @@ pub(crate) async fn read_uring(path: &Path) -> io::Result<Vec<u8>> {
 }
 
 async fn read_to_end_batch(fd: OwnedFd, buf: Vec<u8>) -> io::Result<Vec<u8>> {
-    let file_len = buf.capacity();
-
     // try reading in batch if we have substantial capacity
-    if file_len > MAX_READ_SIZE {
-        match Op::read_batch_size::<BATCH_SIZE>(fd, buf, file_len).await {
+    if buf.capacity() > MAX_READ_SIZE {
+        match Op::read_batch_size::<BATCH_SIZE>(fd, buf).await {
             Ok(buf) => Ok(buf),
             Err((r_fd, mut r_buf)) => {
                 // clear the buffer before we write to it from start again
--- a/tokio/src/io/uring/read.rs
+++ b/tokio/src/io/uring/read.rs
@@ -67,86 +67,92 @@ impl Op<Read> {
         unsafe { Op::new(read_op, Read { fd, buf }) }
     }
 
-    // Split file read operations by batches of size N. batches will be executed in groups
-    // of N.
-    // This function will return
+    // Split file read operations by batches of size N. batches will be executed
+    // in groups of N.
+    //
+    // This function will return the final buffer of bytes in the file
     pub(crate) async fn read_batch_size<const N: usize>(
         mut fd: OwnedFd,
         mut buf: Vec<u8>,
-        len: usize,
     ) -> Result<Vec<u8>, (OwnedFd, Vec<u8>)> {
-        let mut batches_of_ops = Vec::with_capacity(len / (N * MAX_READ_SIZE));
         // hold batch_size operations and reuse it
         let mut n_ops = [const { MaybeUninit::uninit() }; N];
         let mut last_len = 0;
+        let mut read_len = 0;
+        let len = buf.capacity();
+
+        for i in 0..N {
+            n_ops[i].write(opcode::Nop::new().build());
+        }
+
+        let mut n_ops = unsafe { assume_init(n_ops) };
 
         // total number of batch entries to read the file completly
         for (index, start) in (0..len).step_by(MAX_READ_SIZE).enumerate() {
-            // push a chunk into the batches
-            if (index + 1) % N == 0 {
-                // SAFETY: the index + 1 divides N so we have written N ops.
-                // We can transmute those into a array of sqe entries
-                let entries: [Entry; N] = unsafe { std::mem::transmute_copy(&n_ops) };
+            let end = (start + MAX_READ_SIZE).min(len); // clamp to len for the final chunk
+            let array_idx = index % N;
 
-                batches_of_ops.push(entries);
+            // skip first iteration
+            if array_idx == 0 && index != 0 {
+                // SAFETY: Batches are valid array entries
+                let op = unsafe { Op::batch(n_ops.clone(), Read { fd, buf }) };
+                let (_, r_fd, r_buf) = uring_task(op, &mut read_len).await?;
+
+                fd = r_fd;
+                buf = r_buf;
             } else {
-                let end = (start + MAX_READ_SIZE).min(len); // clamp to len for the final chunk
-                let len = (end - start) as u32;
+                let op = opcode::Read::new(
+                    types::Fd(fd.as_raw_fd()),
+                    buf.spare_capacity_mut()[start..].as_mut_ptr().cast(),
+                    (end - start) as u32,
+                )
+                .offset(start as u64)
+                .build()
+                // link our sqes so cqes arrive in order
+                .flags(Flags::IO_LINK);
 
-                n_ops[index % N].write(
-                    opcode::Read::new(
-                        types::Fd(fd.as_raw_fd()),
-                        buf.spare_capacity_mut()[start..].as_mut_ptr().cast(),
-                        len,
-                    )
-                    .offset(start as u64)
-                    .build()
-                    // link our sqes so cqes arrive in order
-                    .flags(Flags::IO_LINK),
-                );
+                n_ops[array_idx] = op;
             }
 
-            last_len = index % N; // save last len for last batch
-        }
-
-        // Handle all full batches
-        for batches in batches_of_ops {
-            // SAFETY: Batches are valid array entries
-            let op = unsafe { Op::batch(batches, Read { fd, buf }) };
-            let (_, r_fd, r_buf) = uring_task(op).await?;
-
-            fd = r_fd;
-            buf = r_buf;
+            last_len = array_idx; // save last array_idx for last batch
         }
 
         // Handle last partial batch if there is any
         if last_len > 0 {
-            // SAFETY: This should be safe as we will always have valid Entries
-            // in the array at any time. We will slice it by `last_len` to avoid
-            // double counting / wrong cqe.
-            let mut entries: [Entry; N] = unsafe { std::mem::transmute_copy(&n_ops) };
-
-            for (i, entry) in entries.iter_mut().enumerate() {
+            for (i, entry) in n_ops.iter_mut().enumerate() {
+                // no op the double counted entries
                 if i > last_len {
                     *entry = opcode::Nop::new().build();
                 }
             }
 
-            // SAFETY: Because of the loop above, the entries that are repeated
-            // or invalided have been Nop'ed
-            let (_, _, r_buf) = uring_task(unsafe { Op::batch(entries, Read { fd, buf }) }).await?;
+            // SAFETY: Because of the no-op loop above, we can assume double entry
+            // read is not possible
+            let op = unsafe { Op::batch(n_ops, Read { fd, buf }) };
+
+            let (_, _, r_buf) = uring_task(op, &mut read_len).await?;
 
             buf = r_buf;
+        }
+
+        unsafe {
+            buf.set_len(buf.len() + read_len);
         }
 
         Ok(buf)
     }
 }
 
+unsafe fn assume_init<const N: usize>(n_ops: [MaybeUninit<Entry>; N]) -> [Entry; N] {
+    unsafe { std::mem::transmute_copy(&n_ops) }
+}
+
 // Poll the batch operation and get the Output out of it
-async fn uring_task<const N: usize>(op: Op<Read, N>) -> Result<Output<N>, (OwnedFd, Vec<u8>)> {
-    // TODO: Maybe we can put this in a tokio task
-    let (res, r_fd, mut r_buf) = op.await;
+async fn uring_task<const N: usize>(
+    op: Op<Read, N>,
+    read_len: &mut usize,
+) -> Result<Output<N>, (OwnedFd, Vec<u8>)> {
+    let (res, r_fd, r_buf) = op.await;
 
     match res {
         CqeResult::Batch(cqes) => {
@@ -154,18 +160,12 @@ async fn uring_task<const N: usize>(op: Op<Read, N>) -> Result<Output<N>, (Owned
                 let cqe = i32_to_result(cqe);
 
                 match cqe {
-                    Ok(r_size) => {
-                        // SAFETY: We have read `r_size` amount
-                        let len = if let Some(new_len) = r_buf.len().checked_add(r_size as usize) {
-                            new_len
-                        } else {
-                            return Err((r_fd, r_buf));
-                        };
-
-                        unsafe {
-                            r_buf.set_len(len);
+                    Ok(r_size) => match read_len.checked_add(r_size as usize) {
+                        Some(len) => {
+                            *read_len = len;
                         }
-                    }
+                        None => return Err((r_fd, r_buf)),
+                    },
                     Err(e) => {
                         if e.kind() != ErrorKind::Interrupted {
                             return Err((r_fd, r_buf));


tokio/src/fs/read_uring.rs
```diff
@@ -31,12 +31,43 @@ pub(crate) async fn read_uring(path: &Path) -> io::Result<Vec<u8>> {
 
     if let Some(size_hint) = size_hint {
         buf.try_reserve(size_hint)?;
+
+        println!("{:?}", size_hint);
```
martin-g: debug leftover

tokio/src/fs/read_uring.rs
```diff
@@ -31,12 +31,43 @@ pub(crate) async fn read_uring(path: &Path) -> io::Result<Vec<u8>> {
...
+        let (res, r_fd, mut r_buf) = Op::read_batch(fd, buf, file_len).await;
+
+        if let CqeResult::Batch(cqes) = res {
+            let mut written_len = 0;
+
+            for cqe in cqes {
+                if let Ok(entry) = cqe {
+                    written_len += entry as usize;
+                    if entry != MAX_READ_SIZE as u32 {
+                        println!("short read");
```
martin-g: debug leftover

tokio/src/fs/read_uring.rs
```diff
@@ -31,12 +31,43 @@ pub(crate) async fn read_uring(path: &Path) -> io::Result<Vec<u8>> {
...
+                    }
+                } else {
+                    println!("{:?}", "err");
```
martin-g: debug leftover

tokio/tests/fs_uring_read.rs
```diff
@@ -131,6 +131,21 @@ async fn read_small_large_files() {
     assert_eq!(bytes, create_buf(20));
 }
 
+#[tokio::test]
+async fn read_small_large_files() {
```
martin-g, line 120: The name of the new test is exactly the same as the previous one - https://github.com/tokio-rs/tokio/pull/7761/changes#diff-05baf3c39ac57d545c93bd32e8f84f6580a7b5db86c066bc4af8a03e2178ee3fR120

tokio/src/runtime/driver/op.rs
```diff
@@ -34,16 +34,21 @@ pub(crate) enum Lifecycle {
...
     Completed(io_uring::cqueue::Entry),
 }
 
 pub(crate) enum State {
-    Initialize(Option<Entry>),
+    // Single operation state
+    Initialize(Option<squeue::Entry>),
     Polled(usize),
+    // Batch operation state
+    InitalizeBatch(Vec<squeue::Entry>),
```
martin-g:
```suggestion
    InitializeBatch(Vec<squeue::Entry>),
```

tokio/src/runtime/driver/op.rs
```diff
@@ -54,21 +59,41 @@ pub(crate) struct Op<T: Cancellable> {
...
         }
     }
+
+    pub(crate) unsafe fn batch(entries: Vec<squeue::Entry>, data: T) -> Self {
+        let handle = Handle::current();
+
+        Self {
+            handle,
+            data: Some(data),
+            state: State::InitalizeBatch(entries),
```
martin-g:
```suggestion
            state: State::InitializeBatch(entries),
```

tokio/src/runtime/driver/op.rs
```diff
@@ -78,47 +103,49 @@ impl<T: Cancellable> Drop for Op<T> {
     fn drop(&mut self) {
         match self.state {
             // We've already dropped this Op.
-            State::Complete => (),
+            State::Complete => return,
+            // This Op has not been polled yet.
+            // We don't need to do anything here.
+            State::Initialize(_) | State::InitalizeBatch(_) => return,
```
martin-g:
```suggestion
            State::Initialize(_) | State::InitializeBatch(_) => return,
```

tokio/src/runtime/driver/op.rs
```diff
@@ -151,13 +178,91 @@ impl<T: Cancellable + Completable + Send> Future for Op<T> {
...
 
-                        return Poll::Ready(data.complete_with_error(err));
+                        return Poll::Ready(data.complete(CqeResult::InitErr(err)));
                     }
                 };
 
                 Poll::Pending
             }
 
+            State::InitalizeBatch(entries) => {
```
martin-g:
```suggestion
            State::InitializeBatch(entries) => {
```

tokio/src/fs/read_uring.rs
```diff
@@ -31,12 +31,43 @@ pub(crate) async fn read_uring(path: &Path) -> io::Result<Vec<u8>> {
...
+                }
+            }
+
+            unsafe { r_buf.set_len(written_len) }
+        }
```
martin-g, line 52: This needs an `else` for the error results.

tokio/src/io/uring/read.rs
```diff
@@ -58,4 +60,33 @@ impl Op<Read> {
...
+            let mut batch_entries = Vec::new();
+
+            for start in (0..len).step_by(MAX_READ_SIZE) {
+                let end = (start + MAX_READ_SIZE).min(len); // clamp to len for the final chunk
+                                                            // MAX_READ_SIZE is less than u32
+                let len = (end - start) as u32;
+
+                let buf_mut_ptr = buf.spare_capacity_mut().as_mut_ptr().cast();
+
+                let op = opcode::Read::new(types::Fd(fd.as_raw_fd()), buf_mut_ptr, len)
```
martin-g: Shouldn't `buf_mut_ptr` be also offset with `start` ?

tokio/src/runtime/io/driver/uring.rs
```diff
@@ -219,6 +226,55 @@ impl Handle {
...
+            }
+            Ok(())
+        };
+
+        // SAFETY: entry is valid for the entire duration of the operation
+        while unsafe { ctx.ring_mut().submission().push_multiple(entries).is_err() } {
+            // If the submission queue is full, flush it to the kernel
+            submit_or_remove(ctx)?;
+        }
+
```
martin-g, line 263: `register_op()` does the following before submitting:
```suggestion
       // Ensure that the completion queue is not full before submitting the entry.
        while ctx.ring_mut().completion().is_full() {
            ctx.dispatch_completions();
        }
```
isn't it needed here too ?

Daksh14: Yup good catch

tokio/tests/fs_uring_read.rs
Daksh14: It was a ghost test, I will add other tests for this

tokio/src/fs/read_uring.rs
Daksh14: I changed the impl a little, let me know @martin-g

tokio/src/io/uring/read.rs
Daksh14: You are correct, I have pushed out my final improvements

tokio/src/io/uring/read.rs
```diff
@@ -58,4 +60,78 @@ impl Op<Read> {
         // SAFETY: Parameters are valid for the entire duration of the operation
         unsafe { Op::new(read_op, Read { fd, buf }) }
     }
+
+    // Split file read operations by batches of size batch_size. batches will be executed in groupts
```
martin-g:
```suggestion
    // Split file read operations by batches of size batch_size. batches will be executed in groups
```

tokio/src/io/uring/read.rs
```diff
@@ -58,4 +60,78 @@ impl Op<Read> {
...
+        batch_size: usize,
+    ) -> Result<Vec<u8>, (OwnedFd, Vec<u8>)> {
+        // hold multiple >= batch_size length vectors of operations
+        let mut batches_of_ops = Vec::new();
+        // hold batch_size operations and reuse it
+        let mut n_ops = Vec::with_capacity(size_of::<Entry>() * batch_size);
+
+        // total number of batch entries to read the file completly
+        for (index, start) in (0..len).step_by(MAX_READ_SIZE).enumerate() {
+            if (index + 1) % batch_size == 0 {
```
martin-g: This `if` should moved to be the last step of the body of the `else`.
Currently it skips the `op` for the current index.

tokio/src/io/uring/read.rs
```diff
@@ -58,4 +60,78 @@ impl Op<Read> {
...
+    pub(crate) async fn read_batch_size(
+        mut fd: OwnedFd,
+        mut buf: Vec<u8>,
+        len: usize,
+        batch_size: usize,
+    ) -> Result<Vec<u8>, (OwnedFd, Vec<u8>)> {
+        // hold multiple >= batch_size length vectors of operations
+        let mut batches_of_ops = Vec::new();
+        // hold batch_size operations and reuse it
+        let mut n_ops = Vec::with_capacity(size_of::<Entry>() * batch_size);
```
martin-g:
```suggestion
        let mut n_ops = Vec::with_capacity(batch_size);
```
Why do you need `size_of()` here ? The capacity is just the number of the entries, not the occupied size in heap.

tokio/src/io/uring/read.rs
```diff
@@ -58,4 +60,78 @@ impl Op<Read> {
...
+                _ => return Err((r_fd, r_buf)),
+            };
+
+            fd = r_fd;
+            buf = r_buf;
+        }
+
+        // SAFETY: We have read `read_size` amount
+        unsafe {
+            buf.set_len(buf.len() + read_size);
```
martin-g: I think this should be moved to line 114.
That is, the `buf`'s length should be increased after each successful read:
```rust
unsafe {
    buf.set_len(buf.len() + r_size);
}
```

tokio/src/io/uring/read.rs
```diff
@@ -58,4 +60,78 @@ impl Op<Read> {
...
+            let op = unsafe { Op::batch(batches, Read { fd, buf }) };
+            // TODO: Maybe we can put this in a tokio task
+            let (res, r_fd, r_buf) = op.await;
+
+            match res {
+                CqeResult::Batch(cqes) => {
+                    for cqe in cqes {
+                        match cqe {
+                            Ok(r_size) => {
+                                read_size += r_size as usize;
```
martin-g: Let's use `checked_add()` here and fail on overflow.

tokio/src/fs/read_uring.rs
```diff
@@ -33,10 +33,29 @@ pub(crate) async fn read_uring(path: &Path) -> io::Result<Vec<u8>> {
...
     }
 
-    read_to_end_uring(fd, buf).await
+    read_to_end_batch(fd, buf).await
 }
 
-async fn read_to_end_uring(mut fd: OwnedFd, mut buf: Vec<u8>) -> io::Result<Vec<u8>> {
+async fn read_to_end_batch(fd: OwnedFd, buf: Vec<u8>) -> io::Result<Vec<u8>> {
+    let file_len = buf.capacity();
+    let batch_size = 128;
```
martin-g: magic number
Please make it a constant with some docs why this value has been chosen

tokio/src/runtime/io/driver/uring.rs
```diff
@@ -116,6 +118,11 @@ impl UringContext {
         }
     }
 
+    // check if the specified range of slab indexes exist or not
+    pub(crate) fn check_slab_entry(&self, indexes: &[usize]) -> bool {
```
martin-g, line 122: The method name is misleading. It sounds like it checks whether indices exist but actually it returns `true` if they do not exist.

tokio/src/runtime/io/driver/uring.rs
```diff
@@ -210,6 +217,58 @@ impl Handle {
...
+            // If the submission queue is full, flush it to the kernel
+            submit_or_remove(ctx)?;
+        }
+
+        // Ensure that the completion queue is not full before submitting the entry.
+        while ctx.ring_mut().completion().is_full() {
+            ctx.dispatch_completions();
+        }
+
+        submit_or_remove(ctx)?;
```
martin-g, line 269: Should this do any cleanup on error here ?

tokio/src/runtime/driver/op.rs
```diff
@@ -34,16 +34,21 @@ pub(crate) enum Lifecycle {
...
 }
 
 pub(crate) enum State {
-    Initialize(Option<Entry>),
+    // Single operation state
+    Initialize(Option<squeue::Entry>),
     Polled(usize),
+    // Batch operation state
+    InitializeBatch(Vec<squeue::Entry>),
+    PolledBatch(Vec<usize>),
```
Darksonn: If at all possible, we want an architecture where you do not have to allocate these vectors for every operation.

tokio/src/io/uring/read.rs
Daksh14: Correct me if I'm wrong here but wouldn't each vector element will be the Entry? Then each element will occupy size of Entry? Or will the vector be storing pointers?

martin-g: "capacity" is the number of elements of type `T` the Vec can hold without needing to reallocate.
See https://github.com/rust-lang/rust/blob/d1a11b670b617f1370f7b1cdf86e225a4e070f15/library/alloc/src/raw_vec/mod.rs#L125

> Or will the vector be storing pointers?

Usually, a Vec stores elements contiguously in memory. But if `T` is a pointer type, like `Vec<Box<...>>` or `Vec<&str>`, then the Vec stores pointers.

tokio/src/runtime/driver/op.rs
Daksh14: I tried using const generics and static batch sizes

tokio/src/runtime/io/driver/uring.rs
Daksh14: Fixed with cleanup

tokio/src/runtime/io/driver/uring.rs
Daksh14: What else should I name it to? I think there could be a better way then checking if the indices have been cleared off or not.

tokio/src/io/uring/read.rs
Daksh14: Fixed

tokio/src/io/uring/read.rs
Daksh14: I changed it now @martin-g

tokio/src/runtime/driver/op.rs
```diff
@@ -34,109 +36,134 @@ pub(crate) enum Lifecycle {
     Cancelled(
         // This field isn't accessed directly, but it holds cancellation data,
         // so `#[allow(dead_code)]` is needed.
-        #[allow(dead_code)] CancelData,
+        #[allow(dead_code)] Arc<CancelData>,
```
Darksonn, line 39: Why is this Arc needed?

Daksh14: we need to clone it for each op to cancel batch ops

tokio/src/fs/read_uring.rs
```diff
@@ -11,16 +11,19 @@ use std::path::Path;
 const PROBE_SIZE: usize = 32;
 const PROBE_SIZE_U32: u32 = PROBE_SIZE as u32;
 
+// batch size for batched reads
+const BATCH_SIZE: usize = 200;
+
 // Max bytes we can read using io uring submission at a time
 // SAFETY: cannot be higher than u32::MAX for safe cast
 // Set to read max 64 MiB at time
```
martin-g, line 19:
```suggestion
// Set to read max 64 KiB at time
```

tokio/src/io/uring/read.rs
```diff
@@ -1,32 +1,40 @@
...
+            // increase length of buffer on successful
+            // completion
             let new_len = buf.len() + len as usize;
             // SAFETY: Kernel read len bytes
             unsafe { buf.set_len(new_len) };
         }
 
-        (cqe.result, self.fd, buf)
-    }
+        // Handle rest each batch outside
```
martin-g, line 35: Please re-phrase this comment. It is not clear what it means

tokio/src/io/uring/read.rs
```diff
@@ -58,4 +66,116 @@ impl Op<Read> {
         // SAFETY: Parameters are valid for the entire duration of the operation
         unsafe { Op::new(read_op, Read { fd, buf }) }
     }
+
+    // Split file read operations by batches of size N. batches will be executed
+    // in groups of N.
```
martin-g, line 71: So, there will be N groups and each group will have size=N ?!
Why not use two different const parameters ? One for group number and one for size

tokio/src/io/uring/read.rs
```diff
@@ -58,4 +66,116 @@ impl Op<Read> {
...
+        let mut n_ops = unsafe { assume_init(n_ops) };
+
+        // total number of batch entries to read the file completly
+        for (index, start) in (0..len).step_by(MAX_READ_SIZE).enumerate() {
+            let end = (start + MAX_READ_SIZE).min(len); // clamp to len for the final chunk
+            let array_idx = index % N;
+
+            // skip first iteration
+            if array_idx == 0 && index != 0 {
+                // SAFETY: Batches are valid array entries
```
martin-g, line 97: This branch skips a Read operation for the current index.
The body of the `else` should be executed for **all** tasks and then once in a while the batch op should be executed.

tokio/src/io/uring/read.rs
```diff
@@ -58,4 +66,116 @@ impl Op<Read> {
...
+
+                match cqe {
+                    Ok(r_size) => match read_len.checked_add(r_size as usize) {
+                        Some(len) => {
+                            *read_len = len;
+                        }
+                        None => return Err((r_fd, r_buf)),
+                    },
+                    Err(e) => {
+                        if e.kind() != ErrorKind::Interrupted {
```
martin-g, line 170: The cqe should be re-tried after Interrupted error, no ?

tokio/src/runtime/driver/op.rs
```diff
@@ -3,14 +3,16 @@ use crate::io::uring::read::Read;
...
+use io_uring::{cqueue, squeue};
 use std::future::Future;
-use std::io::{self, Error};
+use std::io::{self};
 use std::mem;
 use std::pin::Pin;
+use std::sync::Arc;
 use std::task::{Context, Poll, Waker};
 
+const DEFAULT_BATCH_SIZE: usize = u8::MAX as usize;
```
martin-g, line 14: This batch size leads to allocating `completed:[i32;255]` (~1 KiB), `InitializeBatch([squeue::Entry; 255])` (~16KB) or `State::PolledBatch([usize;255])` (~2 KiB) even for simple (non-batched) operations, like `open file`. I.e. 3-17 KiB of stack memory is allocated without being (fully) used.

tokio/src/io/uring/read.rs
```diff
@@ -58,4 +66,116 @@ impl Op<Read> {
...
+        }
+
+        unsafe {
+            buf.set_len(buf.len() + read_len);
+        }
+
+        Ok(buf)
+    }
+}
+
```
martin-g, line 145: Please add SAFETY doc

tokio/src/runtime/driver/op.rs
```diff
@@ -34,109 +36,134 @@ pub(crate) enum Lifecycle {
...
+
         Self {
             handle,
             data: Some(data),
             state: State::Initialize(Some(entry)),
+            completed: [0; N],
+        }
+    }
+
+    pub(crate) unsafe fn batch(entries: [squeue::Entry; N], data: T) -> Self {
```
martin-g, line 85: Please add SAFETY doc

tokio/src/io/uring/read.rs
```diff
@@ -58,4 +66,116 @@ impl Op<Read> {
...
+        // total number of batch entries to read the file completly
+        for (index, start) in (0..len).step_by(MAX_READ_SIZE).enumerate() {
+            let end = (start + MAX_READ_SIZE).min(len); // clamp to len for the final chunk
+            let array_idx = index % N;
+
+            // skip first iteration
+            if array_idx == 0 && index != 0 {
+                // SAFETY: Batches are valid array entries
+                let op = unsafe { Op::batch(n_ops.clone(), Read { fd, buf }) };
```
martin-g, line 98: Here you clone `n_ops` and later you overwrite its entries one by one for the next batch.
I think there is a problem with the last batch if it has less entries than MAX_READ_SIZE. In that case n_ops will contain several new/fresh entries and some old ones from the previous batch which were not yet overwritten.

PR #7763, opened 2025-12-06
codec: from_parts() for FramedWrite and FramedRead

Mrreadiness:
Implement `from_parts()` for `FramedWrite` and `FramedRead`.

## Motivation

Thanks to #7566, `FramedWrite` and `FramedRead` can be transformed into `FramedParts`, but there is no way to create them back from parts, or to transform `Framed` into `FramedRead` or `FramedWrite` without losing progress in buffers.

My original problem was splitting `Framed` into both `FramedRead` **and** `FramedWrite`, to swap the codec for one of them with `map_codec`. I wish we could have something like `into_split` for `Framed`:
```rust
impl<T, U> Framed<T, U>
where
    T: AsyncRead + AsyncWrite,
    U: Clone,
{
    pub fn into_split(self) -> (FramedRead<ReadHalf<T>, U>, FramedWrite<WriteHalf<T>, U>) {
        let (r, w) = split(self.inner.inner);
        let read = FramedRead::from_parts(FramedParts {
            io: r,
            codec: self.inner.codec.clone(),
            read_buf: self.inner.state.read.buffer,
            write_buf: BytesMut::new(),
            _priv: (),
        });
        let write = FramedWrite::from_parts(FramedParts {
            io: w,
            codec: self.inner.codec,
            read_buf: BytesMut::new(),
            write_buf: self.inner.state.write.buffer,
            _priv: (),
        });
        return (read, write);
    }
}

```
But cloning of the codec doesn't feel right here. So, with `from_parts`, a user will have more control over the reader and writer reconstruction.

## Solution
Implement `from_parts()` for `FramedWrite` and `FramedRead`, which builds those structures from `FramedParts` and preserves internal buffers.

### DIFF
--- a/tokio-util/src/codec/framed.rs
+++ b/tokio-util/src/codec/framed.rs
@@ -249,6 +249,7 @@ impl<T, U> Framed<T, U> {
 
     /// Returns a mutable reference to the read buffer.
     pub fn read_buffer_mut(&mut self) -> &mut BytesMut {
+        self.inner.state.read.is_readable = true;
         &mut self.inner.state.read.buffer
     }
 
--- a/tokio-util/src/codec/framed_read.rs
+++ b/tokio-util/src/codec/framed_read.rs
@@ -147,6 +147,7 @@ impl<T, D> FramedRead<T, D> {
 
     /// Returns a mutable reference to the read buffer.
     pub fn read_buffer_mut(&mut self) -> &mut BytesMut {
+        self.inner.state.is_readable = true;
         &mut self.inner.state.buffer
     }
 


ADD-SP (commented):
Would mind adding some tests for both `from_parts` and `into_parts`?

Mrreadiness:
@ADD-SP Thanks! I added tests as well.

I also noticed that `FramedParts::new()` forces the creation of parts with at least an `Encoder` codec. But it doesn't make much sense, because it could be used for `Decoder` without `Encoder`. Looks like it's just a historical mistake, and there was an attempt to remove this limitation (#5280), but it's a breaking change, so it was reverted. 
That's why I added an alternative constructor: `FramedParts::new_parts`. 

So, solving the task of splitting up Framed will look like:
```rust
let framed = Framed::from_parts(parts);
let FramedParts {
    io,
    read_buf,
    write_buf,
    ..
} = framed.into_parts();
let (io_read, io_write) = tokio::io::split(parts.io);
let mut read_parts = FramedParts::new_parts(io_read, Codec::new());
read_parts.read_buf = read_buf;
let framed_read = FramedRead::from_parts(read_parts);
let mut write_parts = FramedParts::new_parts(io_write, Codec::new());
write_parts.write_buf = write_buf;
let framed_write = FramedWrite::from_parts(read_parts);
```
It's far from perfect, but at least doable now. 
If you see a better way to do such a split or even an ergonomic way to add it as a method of `Framed`, I will be happy to implement it.

ADD-SP:
Thanks for the context! I'm considering deprecating the `tokio_util::codec::FramedParts::new`. What do you think? What should the new constructor be called? `new_parts`? `new2`? `new_v2`?

Mrreadiness:
I end up with `new_parts`. Sounds like not a great naming situation anyway.

tokio-util/tests/framed_read.rs
```diff
@@ -34,6 +34,8 @@ macro_rules! pin {
     };
 }
 
+const INITIAL_CAPACITY: usize = 8 * 1024;
```
martin-g, line 38: Please add a comment that this must be the same as the non-public constant at https://github.com/Mrreadiness/tokio/blob/52fcebcaea1e7f6930d1f21e84b59e36ae22d725/tokio-util/src/codec/framed_impl.rs#L25

tokio-util/src/codec/framed.rs
```diff
@@ -389,4 +389,15 @@ impl<T, U> FramedParts<T, U> {
             _priv: (),
         }
     }
+
+    /// Create a new, default, `FramedParts` without encoder enforcement
```
martin-g: Maybe a little bit clearer:
```suggestion
    /// Create a new, default, `FramedParts` without requiring an encoder.
    ///
    /// This constructor is useful when you only need a decoder (e.g., for `FramedRead`)
    /// and the codec doesn't implement `Encoder`.
```

tokio-util/src/codec/framed_read.rs
```diff
@@ -161,6 +161,18 @@ impl<T, D> FramedRead<T, D> {
...
     }
+
+    /// Creates the `FramedRead` from I/O stream, the buffer
+    /// with unprocessed data, and the codec.
+    pub fn from_parts(parts: FramedParts<T, D>) -> FramedRead<T, D> {
+        FramedRead {
+            inner: FramedImpl {
+                inner: parts.io,
+                codec: parts.codec,
+                state: parts.read_buf.into(),
```
martin-g, line 173: This loses the `eof` and `has_errored` flags - https://github.com/Mrreadiness/tokio/blob/52fcebcaea1e7f6930d1f21e84b59e36ae22d725/tokio-util/src/codec/framed_impl.rs#L76-L77
Is this OK ?

martin-g:
> What should the new constructor be called? `new_parts`? `new2`? `new_v2`?

How about `new_unchecked()` ?!
To make it more clear that `U: Encoder<I>` is not required.

ADD-SP:
> How about new_unchecked() ?!

@martin-g The `_unchecked` suffix is usually for unsafe function.

tokio-util/src/codec/framed_read.rs
Mrreadiness: I replicated the same [behaviour as in Framed](https://github.com/Mrreadiness/tokio/blob/52fcebcaea1e7f6930d1f21e84b59e36ae22d725/tokio-util/src/codec/framed.rs#L161).
As I understood from the [state machine](https://github.com/Mrreadiness/tokio/blob/52fcebcaea1e7f6930d1f21e84b59e36ae22d725/tokio-util/src/codec/framed_impl.rs#L129), it's okay. 

`has_error` - used in 2 cases:
1. Error from "previous" Decoder. A user already saw this error and intentionally created a new `Framed` (probably with another Decoder) to try to decode again. 
2. Error from underlying IO. I'm not 100% sure, but I suppose we get the same error with the next read attempt. A user saw this IO error as well.

For `eof`, we should restore this knowledge with the next call to the underlying IO [here](https://github.com/Mrreadiness/tokio/blob/52fcebcaea1e7f6930d1f21e84b59e36ae22d725/tokio-util/src/codec/framed_impl.rs#L220C32-L220C45). As I understood from [AsyncRead docs](https://docs.rs/tokio/latest/tokio/io/trait.AsyncRead.html), the expected behavior after reaching EOF is Ok(Ready) with 0 new bytes for all subsequent read attempts.

Mrreadiness: I just realised that these 2 flags are related to IO, and the user can change `parts.io`. So we don't have any other way to get this knowledge back, but to interact with IO itself.

Mrreadiness:
Hi @ADD-SP, @martin-g!
Do you have any ideas how to improve the pr?

tokio-util/src/codec/framed_read.rs
ADD-SP: @conradludgate What do you think?

conradludgate: I think it would make sense to add those fields to the FramedParts struct, but I fear using them in Framed::from_parts would be a subtle breaking change, and I am worried about having FramedRead::from_parts behave differently.

My preference would be to mirror the behaviour of Framed::from_parts, and maybe in tokio-util 0.8 we can reconsider. IIRC from_parts also seems to re-allocate if the buffers are deemed too small which upsets me as well.

conradludgate:
I think it would be nice to get this PR merged soon. I just encountered a footgun in the wild:

```rust
let parts = read.into_parts();
let mut read = FramedRead::new(parts.io, parts.codec);
*read.read_buffer_mut() = parts.read_buf;

// times out if no data on the IO socket, even if there's readable data in the buffer
read.next().await
```

If there's data in the read buffer, FramedRead will not decode them as it defaults to `is_readable: false` and never tries to decode the buffer.

This PR does fix this case as `impl From<BytesMut> for ReadFrame` makes sure to set `is_readable: buffer.capacity() > 0` (why capacity and not len?).

To avoid the footgun though we should probably make `read_buffer_mut()` eagerly apply `self.is_readable = true` which as far as I can tell is never incorrect, just a potential perf regression in some cases.

Mrreadiness:
I didn't notice user can change the buffer via `read_buffer_mut()`. Strange that it's possible, but I agree - better default into `framing` state in such cases. Added this to MR.

> why capacity and not len?

I tried to look into history, but couldn't figure out the reasoning. In theory, it should be safe to replace `capacity` to `len` - you can't decode from empty buffer, right? But I'm still afraid to made this change. 

Is there anything else I can do to get this MR merged?

PR #7767, opened 2025-12-08
miri tests: fstat is supported now

RalfJung:
So we can enable the tests that apparently were blocked on that.


RalfJung:
Ah, it still doesn't work...
"unsupported operation: obtaining metadata is only supported on file-backed file descriptors"

PR #7772, opened 2025-12-11
call size_of/align_of directly inside functions

bestgopher:

### DIFF
--- a/tokio/src/runtime/task/raw.rs
+++ b/tokio/src/runtime/task/raw.rs
@@ -82,39 +82,17 @@ pub(super) fn vtable<T: Future, S: Schedule>() -> &'static Vtable {
 /// <https://users.rust-lang.org/t/custom-vtables-with-integers/78508>
 struct OffsetHelper<T, S>(T, S);
 impl<T: Future, S: Schedule> OffsetHelper<T, S> {
-    // Pass `size_of`/`align_of` as arguments rather than calling them directly
-    // inside `get_trailer_offset` because trait bounds on generic parameters
-    // of const fn are unstable on our MSRV.
-    const TRAILER_OFFSET: usize = get_trailer_offset(
-        std::mem::size_of::<Header>(),
-        std::mem::size_of::<Core<T, S>>(),
-        std::mem::align_of::<Core<T, S>>(),
-        std::mem::align_of::<Trailer>(),
-    );
+    const TRAILER_OFFSET: usize = get_trailer_offset::<Header, Core<T, S>, Trailer>();
 
     // The `scheduler` is the first field of `Core`, so it has the same
     // offset as `Core`.
-    const SCHEDULER_OFFSET: usize = get_core_offset(
-        std::mem::size_of::<Header>(),
-        std::mem::align_of::<Core<T, S>>(),
-    );
+    const SCHEDULER_OFFSET: usize = get_core_offset::<Header, Core<T, S>>();
 
-    const ID_OFFSET: usize = get_id_offset(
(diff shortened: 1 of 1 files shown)


ADD-SP (commented):
Could you share the motivation of this PR?

bestgopher:
> Could you share the motivation of this PR?

This change is motivated by a comment that was removed in this PR:

 // Pass `size_of`/`align_of` as arguments rather than calling them directly
 // inside `get_trailer_offset` because trait bounds on generic parameters
 // of const fn are unstable on our MSRV.

ADD-SP:
> > Could you share the motivation of this PR?
> 
> This change is motivated by a comment that was removed in this PR:
> 
> // Pass `size_of`/`align_of` as arguments rather than calling them directly // inside `get_trailer_offset` because trait bounds on generic parameters // of const fn are unstable on our MSRV.

Is there a trait bound in const fn?

bestgopher:
My attention is focused on this sentence: Pass size_of/align_of as arguments rather than calling them directly inside get_trailer_offset. If I’ve misunderstood your point, please feel free to close my PR.

ADD-SP:
> My attention is focused on this sentence: Pass size_of/align_of as arguments rather than calling them directly inside get_trailer_offset. If I’ve misunderstood your point, please feel free to close my PR.

Please don't get me wrong, I just want to make sure we know what we are doing.

The original author did not do this because of "trait bounds on generic parameters of `const fn` are unstable on our MSRV". However, there is no new trait bounds on `const fn` in this PR. So there is a inconsistency, and we have to figure it out. Otherwise, we don't know what we are doing.

tokio/src/runtime/task/raw.rs
```diff
@@ -82,53 +82,29 @@ pub(super) fn vtable<T: Future, S: Schedule>() -> &'static Vtable {
 /// <https://users.rust-lang.org/t/custom-vtables-with-integers/78508>
 struct OffsetHelper<T, S>(T, S);
 impl<T: Future, S: Schedule> OffsetHelper<T, S> {
-    // Pass `size_of`/`align_of` as arguments rather than calling them directly
-    // inside `get_trailer_offset` because trait bounds on generic parameters
-    // of const fn are unstable on our MSRV.
```
Darksonn, line 87: When I wrote this comment, I meant that the const methods would take the same generic parameters as `OffsetHelper` itself, that is, just `T` and `S`. It would then use `size_of::<Core<T,S>>()` directly.

tokio/src/runtime/task/raw.rs
```diff
@@ -188,15 +163,11 @@ const fn get_id_offset(
...
-const fn get_spawn_location_offset(
-    header_size: usize,
-    core_align: usize,
-    scheduler_size: usize,
-    id_align: usize,
-    id_size: usize,
-    spawn_location_align: usize,
-) -> usize {
-    let mut offset = get_id_offset(header_size, core_align, scheduler_size, id_align);
+const fn get_spawn_location_offset<T1, T2, T3, T4, T5>() -> usize {
```
Darksonn, line 166: The types are literally just T1, T2, T3, T4, T5. This change makes readability of this code worse.

bestgopher:
Apologies for the misunderstanding on my part.

Darksonn:
Although we can't yet make this particular change, thanks for taking the time to submit a PR.

Darksonn:
Although we can't yet make this particular change, thanks for taking the time to submit a PR.

Darksonn:
Although we can't yet make this particular change, thanks for taking the time to submit a PR.

PR #7773, opened 2025-12-12
io: always cleanup AsyncFd registration list on deregister

F4RAN:
Fixes memory leak when fd is closed before AsyncFd drop.

Fixes: #7563

## Motivation

When a file descriptor is closed before dropping `AsyncFd`, OS deregistration fails and causes an early return, preventing cleanup of the internal registration list. This leaks `ScheduledIo` objects over time.

## Solution

Always clean up the internal registration list, even if OS deregistration fails. Store the OS result, perform cleanup, then return the error. This is safe because the list is Tokio's internal tracking - if OS deregister fails, the OS already doesn't track it, so cleanup doesn't break memory safety.

Includes a test reproducing the bug scenario.

### DIFF
--- a/tokio/tests/io_async_fd_memory_leak.rs
+++ b/tokio/tests/io_async_fd_memory_leak.rs
@@ -55,7 +55,7 @@ unsafe impl GlobalAlloc for TrackingAllocator {
 static GLOBAL: TrackingAllocator = TrackingAllocator::new();
 
 fn allocated_bytes() -> usize {
-    GLOBAL.load(Ordering::Relaxed)
+    GLOBAL.allocated.load(Ordering::Relaxed)
 }
 
 #[tokio::test]


tokio/tests/io_async_fd.rs
```diff
@@ -956,3 +956,49 @@ async fn try_with_interest() {
 
     assert!(Arc::ptr_eq(&original, &returned));
 }
+
+#[tokio::test]
+async fn memory_leak_when_fd_closed_before_drop() {
+    use std::os::unix::io::{AsRawFd, RawFd};
```
ADD-SP: How does this test cover this issue?

F4RAN:
Hi @ADD-SP,

Thanks for the feedback! I've updated the PR to address your questions.

## Test Infrastructure Added

Added test-only methods to verify the fix:
- **`registration_set.rs`**: `pending_release_count()` and `total_registration_count()` (needed since `LinkedList` has no `len()`)
- **`driver.rs`**: Wrapper methods
- **`handle.rs`**: Methods to expose counts for integration tests

All gated with `#[cfg(feature = "full")]` and `#[doc(hidden)]`. Kept `/// TEST PURPOSE RELATED TO PR #7773` comments for easy removal later.

## Test Evolution

Initial test only exercised the code path. Updated to check registration count after each iteration and assert it doesn't grow (indicating leak).

## Platform-Specific Behavior

**The leak cannot be reproduced on macOS.** On Linux with epoll:
- With buggy code: Test fails, detecting 30 leaked ScheduledIo objects
- With fix: Test passes, count stays at baseline

The test is Linux-specific for leak detection but exercises the code path on macOS too.

## Why Cleanup After OS Failure is Safe
> > if OS deregister fails, the OS already doesn't track it, so cleanup doesn't break memory safety.
> 
> Could you please explain why? I'm curious on it.

When `mio::Registry::deregister()` fails (typically `EBADF` for closed fds), the OS kernel already doesn't track this fd in its polling mechanism (epoll/kqueue). The `ScheduledIo` in Tokio's internal `registrations` list is just our bookkeeping - removing it doesn't affect the kernel or other resources. Since the OS doesn't know about the fd anymore, cleaning up our internal tracking is safe and doesn't break memory safety.


## Remaining Questions
> Only cleaning up the intrusive list may not be a proper fix of this issue. There are some questions we need to answer.
> 
> * Shall we propagate the error to the downstream user?
> * When / When not propagate the error?
> * How to propagate this error?
> * Will our fix cause a breaking change?

Per your comment in #7563, I see two options for into_inner(): (1)add try_into_inner() (non-breaking) or (2) panic on error. I recommend Option 1 for non-breaking behavior and user choice, but I can implement Option 2 if you prefer.

1. Shall we propagate the error to the downstream user?
Yes, but make it optional via try_into_inner() so users can choose.
2. When / When not propagate the error?
Not in Drop (can't return errors), but yes in into_inner() via optional try_into_inner() while keeping into_inner() ignoring errors for compatibility.
3. How to propagate this error?
Add try_into_inner() -> Result<T, (T, io::Error)> (Option 1).
4. Will our fix cause a breaking change?
No.


Thanks again!

tokio/tests/io_async_fd.rs
F4RAN: This is explained in the comment below.

tokio/src/runtime/io/driver.rs
```diff
@@ -317,6 +320,24 @@ impl Handle {
     }
 }
 
+/// TEST PURPOSE RELATED TO PR #7773
+#[cfg(feature = "full")]
```
martin-g: Why not `#[cfg(test)]` ?

tokio/src/runtime/io/registration_set.rs
```diff
@@ -53,6 +53,33 @@ impl RegistrationSet {
...
+
+        // Drain all items
+        while let Some(item) = synced.registrations.pop_back() {
+            items.push(item);
+        }
+
+        let count = items.len();
+
+        // Restore items in reverse order (since we popped from back)
+        for item in items.into_iter().rev() {
```
martin-g: The usage of `.rev()` here breaks the original order.
```suggestion
        for item in items.into_iter() {
```

tokio/tests/io_async_fd.rs
```diff
@@ -947,12 +947,76 @@ async fn try_new() {
...
+///
+/// Reproduces the bug where closing fd before dropping AsyncFd causes
+/// OS deregister to fail, preventing cleanup and leaking ScheduledIo objects.
 #[tokio::test]
-async fn try_with_interest() {
-    let original = Arc::new(InvalidSource);
+async fn memory_leak_when_fd_closed_before_drop() {
+    use std::os::unix::io::{AsRawFd, RawFd};
+    use std::sync::Arc;
+    use tokio::io::unix::AsyncFd;
```
martin-g: These are already imported in the beginning of the file. No need to import them again.

tokio/tests/io_async_fd.rs
```diff
@@ -947,12 +947,76 @@ async fn try_new() {
     assert!(Arc::ptr_eq(&original, &returned));
 }
 
+/// Regression test for issue #7563
+///
+/// Reproduces the bug where closing fd before dropping AsyncFd causes
+/// OS deregister to fail, preventing cleanup and leaking ScheduledIo objects.
 #[tokio::test]
-async fn try_with_interest() {
```
martin-g, line 951: I think the old test should be restored.
It was testing `AsyncFd::try_with_interest()`

tokio/src/runtime/io/driver.rs
```diff
@@ -317,6 +320,24 @@ impl Handle {
...
+impl Handle {
+    /// Returns the number of pending registrations (test-only, not part of public API)
+    #[doc(hidden)]
+    #[allow(unreachable_pub)]
+    pub fn pending_registration_count(&self) -> usize {
+        self.registrations.pending_release_count()
+    }
+    /// Returns the total number of registrations in the main list (test-only)
```
martin-g:
```suggestion
    /// Returns the total number of registrations in the main list (test-only, not part of public API)
```

tokio/tests/io_async_fd.rs
F4RAN: You're right. 
Removed.

tokio/tests/io_async_fd.rs
F4RAN: Sorry my bad.
Done.

tokio/src/runtime/handle.rs
```diff
@@ -488,6 +488,24 @@ impl Handle {
...
+    }
+
+    /// Returns the total number of registrations in the main list (test-only, not part of public API)
+    #[doc(hidden)]
+    #[allow(unreachable_pub)]
+    pub fn io_total_registration_count(&self) -> usize {
+        self.inner.driver().io().total_registration_count()
+    }
+}
+
```
ADD-SP: We'd better avoiding exposing public (even with `#[doc(hidden)]`) interface just for internal testing.

F4RAN: Hi @ADD-SP,
Thanks for the review.
I've addressed your concern by implementing a `__internal_test` feature flag approach. The methods are now gated behind `#[cfg(feature = "__internal_test")]`:

- **I/O driver layer** (`tokio/src/runtime/io/driver.rs`): Uses `pub(crate)` visibility - only accessible within the crate
- **Runtime Handle** (`tokio/src/runtime/handle.rs`): Uses `pub` visibility (wrapped in `cfg_io_driver!` macro) - **these are still public** but only compiled when the `__internal_test` feature is enabled

The `__internal_test` feature flag follows Rust ecosystem patterns (double-underscore prefix signals internal/unstable APIs) and is clearly documented as "Not part of public API. Subject to change without notice." The runtime Handle methods need `pub` visibility because integration tests in `tests/` directory cannot access `pub(crate)` items.

- **Before fix:** Test detects the leak and panics with `Memory leak detected: final count 30 (initial: 0), max seen: 30. With bug, count would be ~30 (30 leaked objects).`
- **After fix:** Test passes successfully, confirming no registrations leak.

**Run command:** 
```sh
cargo test --test io_async_fd --features "full,__internal_test" memory_leak_when_fd_closed_before_drop
```
If you prefer a different approach, I'm ready to implement it. Please let me know your preference.

tokio/Cargo.toml
```diff
@@ -84,6 +84,11 @@ signal = [
 sync = []
 test-util = ["rt", "sync", "time"]
 time = []
+
+# Internal test-only feature. Not part of public API. Subject to change without notice.
+# Used for internal testing infrastructure.
+__internal_test = []
```
martin-g: nit: I was thinking about `integration_test` as a name of this feature. IMO it makes it more clearer that the functionality behind it is used for the IT tests.

F4RAN: Thanks.
agreed. I renamed __internal_test to integration_test.

tokio/Cargo.toml
```diff
@@ -84,6 +84,11 @@ signal = [
 sync = []
 test-util = ["rt", "sync", "time"]
 time = []
+
+# Integration test-only feature. Not part of public API. Subject to change without notice.
+# Used for integration testing infrastructure.
+integration_test = []
```
ADD-SP: Cargo feature is not a good idea. Any tokio dependents can enable it without the end-users knowledge.

F4RAN: I've switched from Cargo feature to `#[cfg(tokio_unstable)]` instead. This requires `RUSTFLAGS="--cfg tokio_unstable"` which cannot be enabled by dependencies through Cargo.toml - it requires explicit compiler flags.

```sh
RUSTFLAGS="--cfg tokio_unstable" cargo test --features full --test io_async_fd memory_leak_when_fd_closed_before_drop
```

mox692: If these methods are only going to be used in tests, how about just gating them with `#[cfg(test)]`? `tokio_unstable` is supposed to be used for public apis, not for these kind of test utility functions.

tokio/tests/io_async_fd.rs
```diff
@@ -956,3 +956,85 @@ async fn try_with_interest() {
...
+    }
+
+    tokio::task::yield_now().await;
+    tokio::time::sleep(Duration::from_millis(100)).await;
+
+    let final_count = rt_handle.io_total_registration_count();
+    max_count_seen = max_count_seen.max(final_count);
+
+    assert!(
+        final_count <= initial_count + 2 && max_count_seen <= initial_count + 2,
```
mox692: Where do these `+2` come from? Can't we just say the final count is the same as initial count? Generally we'd want to avoid range assertion to keep the test deterministic.

tokio/Cargo.toml
martin-g: `#[cfg(test)]` does not help for IT tests (`crate_root/tests/**.rs`).

tokio/tests/io_async_fd.rs
F4RAN: Hi @mox692.
The `+2` is a small tolerance buffer for potential timing variations and background I/O activity. The key assertion is that the count **doesn't grow proportionally with iterations** - with the bug, 30 iterations would leak ~30 objects, not stay within +2 of baseline.

That said, I can make it stricter with `final_count == initial_count` if you prefer. We could also add a small delay before the final count to ensure all cleanup completes. Let me know your preference.

F4RAN:
Hi @martin-g, @ADD-SP, @mox692.
I've tried multiple approaches (`#[doc(hidden)]`, Cargo feature, `tokio_unstable`) and each was rejected. `#[cfg(test)]` doesn't work for integration tests.

What approach would you accept? If the current testing infrastructure doesn't support a proper solution, should I close this PR or keep just the fix without the numerical verification test?

ADD-SP:
Can Miri or address sanitizer capture this memory leak?

F4RAN:
> Can Miri or address sanitizer capture this memory leak?
Hi again @ADD-SP,
Thanks for your idea.

I'll be appreciated if you check my rss-based test:

## Summary

I tested several approaches as suggested:

| Approach | Result |
|----------|--------|
| **LeakSanitizer** | ❌ Cannot detect - objects are still reachable via internal linked list (logical leak, not unreachable memory) |
| **Miri** | ❌ Cannot run - requires system calls (`epoll_ctl`, `libc::close`, sockets) |

Since this is a "logical leak" (objects stuck in a data structure, not truly unreachable), standard leak detectors don't work.

## Solution: RSS-based memory test

I implemented a test that monitors **RSS (Resident Set Size)** before and after running 5000 iterations. No internal API access needed - just reads `/proc/self/statm`.

**Linux results:**
- **Without fix:** FAILED - RSS grew by 1920KB (leak detected!)
- **With fix:** PASSED - RSS stable

The test is gated with `#[cfg(target_os = "linux")]` since it reads `/proc/self/statm`.

Darksonn:
Miri does support epoll.

tokio/src/runtime/io/driver.rs
```diff
@@ -307,6 +308,8 @@ impl Handle {
 
         self.metrics.dec_fd_count();
 
+        os_result?; // Return error after cleanup
+
         Ok(())
```
Darksonn:
```suggestion
        os_result // Return error after cleanup
```

tokio/Cargo.toml
Darksonn, line 1: Unrelated change.

tokio/tests/io_async_fd.rs
Darksonn: This test seems really fragile. What if there's another test running in parallel that consumes a lot of memory? There are other tests in this file, after all.

tokio/Cargo.toml
F4RAN: Remained from integration test approach.
Fixed.
Thanks.

F4RAN:
> Miri does support epoll.

Hi @Darksonn. 
I tested with Miri on Linux. Both with and without the fix, Miri reports no leak:

test memory_leak_miri_check ... ok
Miri detects unreachable memory, but this is a logical leak - the ScheduledIo objects remain reachable through the internal registrations linked list. They're not orphaned allocations, they're just never removed when the fd is closed before drop. Same reason LSan couldn't detect it. The RSS-based test remains the practical way to verify this fix works - it detects unbounded memory growth across multiple phases.

Darksonn:
> Miri detects unreachable memory

Are you saying that miri detects it, and that miri doesn't detect it when the fix is applied? That sounds like a pretty good way to test it.

F4RAN:
> > Miri detects unreachable memory
> 
> Are you saying that miri detects it, and that miri doesn't detect it when the fix is applied? That sounds like a pretty good way to test it.

To clarify: Miri cannot detect this leak - neither with nor without the fix. Miri is designed to detect unreachable memory (orphaned allocations). But this is a logical leak where objects remain reachable through the internal registrations linked list - they're just never removed. Since they're still reachable, Miri sees no problem. Same reason LSan couldn't detect it. The RSS-based test is the practical solution since it measures actual memory growth.

> Both with and without the fix, Miri reports no leak

Darksonn:
Relying on RSS is going to be extremely fragile. Your memory allocator may keep memory around even if the program has freed it, and this still counts in RSS.

If we're going the route of measuring memory, then please declare a `#[global_allocator]` that increments/decrements a static `AtomicUsize` for tracking the memory actually in use (it may forward calls into [the system allocator](https://doc.rust-lang.org/stable/std/alloc/struct.System.html) to perform actual allocation), and use this to measure the amount instead.

The test needs to be moved to its own file so the `#[global_allocator]` does not affect other tests.

F4RAN:
> Relying on RSS is going to be extremely fragile. Your memory allocator may keep memory around even if the program has freed it, and this still counts in RSS.
> 
> If we're going the route of measuring memory, then please declare a `#[global_allocator]` that increments/decrements a static `AtomicUsize` for tracking the memory actually in use (it may forward calls into [the system allocator](https://doc.rust-lang.org/stable/std/alloc/struct.System.html) to perform actual allocation), and use this to measure the amount instead.
> 
> The test needs to be moved to its own file so the `#[global_allocator]` does not affect other tests.

@Darksonn, Implemented a custom #[global_allocator] that tracks allocations via a static AtomicUsize, forwarding to the system allocator. The test is in io_async_fd_memory_leak.rs so the allocator only affects that file.
Verified on Linux: without the fix, the test fails with ~256KB growth per phase; with the fix, it passes with stable memory.

tokio/tests/io_async_fd_memory_leak.rs
```diff
@@ -0,0 +1,188 @@
+//! Regression test for issue #7563 - Memory leak when fd closed before AsyncFd drop
+//!
+//! This test uses a custom global allocator to track actual memory usage,
+//! avoiding false positives from RSS measurements which include freed-but-retained memory.
+
+#![cfg(all(unix, target_os = "linux"))]
```
Darksonn: Please include a `#[cfg()]` for the feature flags required by this test. You may use "full" if you would like.

tokio/tests/io_async_fd_memory_leak.rs
```diff
@@ -0,0 +1,188 @@
...
+
+use std::alloc::{GlobalAlloc, Layout, System};
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+/// A tracking allocator that counts bytes currently allocated
+struct TrackingAllocator;
+
+static ALLOCATED: AtomicUsize = AtomicUsize::new(0);
```
martin-g: nit: this could be a field of the `TrackingAllocator` struct

tokio/tests/io_async_fd_memory_leak.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    impl AsRawFd for ArcFd {
+        fn as_raw_fd(&self) -> RawFd {
+            self.0.as_raw_fd()
+        }
+    }
+
+    fn set_nonblocking(fd: RawFd) {
+        unsafe {
+            let flags = libc::fcntl(fd, libc::F_GETFL);
```
martin-g: Handle negative value as at https://github.com/F4RAN/tokio/blob/fdf771e34b10e9f5ce4acbb96427e2ae331d6fd0/tokio/tests/io_async_fd.rs#L107-L113

F4RAN: Done.

tokio/tests/io_async_fd_memory_leak.rs
F4RAN: Done.

tokio/tests/io_async_fd_memory_leak.rs
F4RAN: Done.

PR #7777, opened 2025-12-16
macros: insert leading colon to avoid unused qualification

djc:
## Motivation

When using the [`unused_qualifications` lint](https://doc.rust-lang.org/stable/nightly-rustc/rustc_lint/builtin/static.UNUSED_QUALIFICATIONS.html) it will currently generate a false positive on the use of the `tokio::` in `tokio::Runtime::builder()` (etc). Because of the use of `last_stmt_start_span` as the span for the ident, the warning ends up pointing at a more or less random piece of language which is surprising.

See for more context here:

https://github.com/rust-lang/rust/issues/149873

## Solution

As pointed out in the Rust issue, using an absolute path should avoid the lint triggering.

### DIFF
--- a/tokio-macros/src/entry.rs
+++ b/tokio-macros/src/entry.rs
@@ -1,7 +1,8 @@
 use proc_macro2::{Span, TokenStream, TokenTree};
 use quote::{quote, quote_spanned, ToTokens};
 use syn::parse::{Parse, ParseStream, Parser};
-use syn::{braced, Attribute, Ident, Path, Signature, Visibility};
+use syn::punctuated::Punctuated;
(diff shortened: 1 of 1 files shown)


Darksonn:
This seems to trigger various build failures. Maybe we should just adjust the span to point at a better location?

ADD-SP:
Hi @djc, are you still working on this PR?

djc:
I think this should get fixed (I don't think just moving the span to a different place is sufficient for that?), but I'm pretty busy with other stuff, so I'm not sure when I'll get back to it and would be happy if someone else wants to take over.

Darksonn:
Then perhaps we should just silence the warning. Regardless, I'll close this so anyone who wishes to pick it up can open a new one.

PR #7779, opened 2025-12-17
ci: Run benchmarks in the CI

mattiapitossi:
## Motivation

Starting from issue #6539, I would like to introduce some consistent benchmarks using Gungraun to be able to run benchmarks on the CI. For the moment we don't run any of the benchmarks due to the fact that the Criterion benchmarks use a wall-time approach and cannot considered reliable for a CI.

## Solution

The `iai-calgrind` library is now called `gungraun` and it's already used by some services (https://github.com/gungraun/gungraun/network/dependents) including Rust's compiler-builtins. I wanted to start with `spawn` first to understand if we want to proceed in this direction.

## Other Considerations

- Gungraun uses `Valgrind` which is only supported by Unix, therefore benchmarks on other system can only be run through Docker
- We are still discussing about whether we want to use Bencher or not

## Results

Multiple runs on M1 (Docker)

```
ct_spawn::group::spawn_rt basic:single_rt()
  Instructions:                        8133|8133                 (No change)
  L1 Hits:                            11844|11844                (No change)
  LL Hits:                               61|61                   (No change)
  RAM Hits:                             433|433                  (No change)
  Total read+write:                   12338|12338                (No change)
  Estimated Cycles:                   27304|27304                (No change)
ct_spawn::group::spawn_rt threaded:multi_rt()
  Instructions:                        7009|7009                 (No change)
  L1 Hits:                             9858|9858                 (No change)
  LL Hits:                              104|104                  (No change)
  RAM Hits:                             273|273                  (No change)
  Total read+write:                   10235|10235                (No change)
  Estimated Cycles:                   19933|19933                (No change)
ct_spawn::group::spawn_rt_100 basic:single_rt()
  Instructions:                      267414|267414               (No change)
  L1 Hits:                           407758|407758               (No change)
  LL Hits:                              340|340                  (No change)
  RAM Hits:                             914|914                  (No change)
  Total read+write:                  409012|409012               (No change)
  Estimated Cycles:                  441448|441448               (No change)
ct_spawn::group::spawn_rt_100 threaded:multi_rt()
  Instructions:                      156427|156427               (No change)
  L1 Hits:                           231658|231659               (-0.00043%) [-1.00000x]
  LL Hits:                              486|485                  (+0.20619%) [+1.00206x]
  RAM Hits:                             699|699                  (No change)
  Total read+write:                  232843|232843               (No change)
  Estimated Cycles:                  258553|258549               (+0.00155%) [+1.00002x]

Gungraun result: Ok. 4 without regressions; 0 regressed; 0 filtered; 4 benchmarks finished in 1.12675s
```

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -27,25 +27,16 @@ jobs:
 
       - name: Run ct benchmarks
         run: cargo bench --bench ${{ matrix.bench }} | tee output.txt
-        working-directory: benches
-
-        # Download previous benchmark result from cache (if exists)
-      - name: Download previous benchmark data
-        uses: actions/cache@v4
-        with:
-          path: ./cache
-          key: ${{ runner.os }}-benchmark
 
       - name: Store benchmark result
         uses: benchmark-action/github-action-benchmark@v1
         with:
           tool: "cargo"
-          output-file-path:
-            output.txt
-            # Where the previous data file is stored
-          external-data-json-path: ./cache/benchmark-data.json
-          # Workflow will fail when an alert happens
-          fail-on-alert: true
+          output-file-path: output.txt
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          comment-on-alert: true
           auto-push: true
+          fail-on-alert: true
+          # Show alert if performance drops > 15% 
+          alert-threshold: '150%' #TODO: TBD
+          comment-on-alert: true
+          #TODO: alert tokio maintainers?
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,0 +1,52 @@
+name: benchmark
+on:
+  push:
+    branches: [master]
+
+jobs:
+  benchmark:
+    name: Benchmark
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        bench:
+          - ct_spawn
+    steps:
+      - uses: actions/checkout@v5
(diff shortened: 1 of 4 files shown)


.github/workflows/bench.yml
```diff
@@ -0,0 +1,34 @@
...
+          version=$(cargo metadata --format-version=1 |\
+            jq '.packages[] | select(.name == "gungraun").version' |\
+            tr -d '"'
+          )
+          cargo binstall --no-confirm gungraun-runner --version $version
+
+      - uses: bencherdev/bencher@main
+      - name: Run Bencher
+        run: |
+          bencher run --adapter rust_gungraun --err "cargo bench --bench ${{ matrix.bench }}"
```
mattiapitossi: TODO: There's an open discussion on Discord whether we want to use Bencher or not

Darksonn: We can try using bencher. I understand that we need to do something special to set it up. Could you share more details on what is required?

mattiapitossi: Sure! Technically, it should be enough to create an account on https://bencher.dev/auth/signup and generate an API token. Then, a new secret called `BENCHER_API_TOKEN` with the API key should be created under repo settings as described [here](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets)

younes-io:
@Darksonn : is this still a topic?

PR #7781, opened 2025-12-19
net: add support for `TCLASS` option on IPv6

jtojnar:
- Rename `{Udp,Tcp}Socket::tos` to `tos_v4` as it is IPv4-specific, following the [rename](https://github.com/rust-lang/socket2/pull/592) from socket2 0.6.0.
- Add `{Tcp,Udp}Socket::tclass_v6` as its IPv6 equivalent.
- Fix affect/effect typos.

## Motivation

I want to be able to set IPv6 [*Traffic Class* field](https://en.wikipedia.org/wiki/IPv6_packet#Traffic_Class_field) as conveniently as the corresponding IPv4 [*Type of Service* field](https://en.wikipedia.org/wiki/Type_of_service), which allows setting the [*Differentiated Services* subfield](https://en.wikipedia.org/wiki/Differentiated_services) necessary for [Quality of Service](https://en.wikipedia.org/wiki/Quality_of_service) requirements.

## Solution

Just addding an extra getter/setter exposing the underlying API.

### DIFF
--- a/tokio/src/net/tcp/socket.rs
+++ b/tokio/src/net/tcp/socket.rs
@@ -489,6 +489,79 @@ impl TcpSocket {
         self.inner.tcp_nodelay()
     }
 
+    /// Gets the value of the `IPV6_TCLASS` option for this socket.
+    ///
+    /// For more information about this option, see [`set_tclass_v6`].
+    ///
+    /// [`set_tclass_v6`]: Self::set_tclass_v6
+    // https://docs.rs/socket2/0.6.1/src/socket2/sys/unix.rs.html#2541
+    #[cfg(any(
+        target_os = "android",
+        target_os = "dragonfly",
+        target_os = "freebsd",
+        target_os = "fuchsia",
+        target_os = "linux",
+        target_os = "macos",
+        target_os = "netbsd",
+        target_os = "openbsd",
+        target_os = "cygwin",
+    ))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(any(
+            target_os = "android",
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "fuchsia",
+            target_os = "linux",
+            target_os = "macos",
+            target_os = "netbsd",
+            target_os = "openbsd",
+            target_os = "cygwin",
+        )))
+    )]
+    pub fn tclass_v6(&self) -> io::Result<u32> {
+        self.inner.tclass_v6()
+    }
+
+    /// Sets the value for the `IPV6_TCLASS` option on this socket.
+    ///
+    /// Specifies the traffic class field that is used in every packet
+    /// sent from this socket.
+    ///
+    /// # Note
+    ///
+    /// This may not have any effect on IPv4 sockets.
+    // https://docs.rs/socket2/0.6.1/src/socket2/sys/unix.rs.html#2566
+    #[cfg(any(
+        target_os = "android",
+        target_os = "dragonfly",
+        target_os = "freebsd",
+        target_os = "fuchsia",
+        target_os = "linux",
+        target_os = "macos",
+        target_os = "netbsd",
+        target_os = "openbsd",
+        target_os = "cygwin",
+    ))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(any(
+            target_os = "android",
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "fuchsia",
+            target_os = "linux",
+            target_os = "macos",
+            target_os = "netbsd",
+            target_os = "openbsd",
+            target_os = "cygwin",
+        )))
+    )]
+    pub fn set_tclass_v6(&self, tclass: u32) -> io::Result<()> {
+        self.inner.set_tclass_v6(tclass)
+    }
+
     /// Gets the value of the `IP_TOS` option for this socket.
     ///
     /// For more information about this option, see [`set_tos_v4`].
--- a/tokio/src/net/udp.rs
+++ b/tokio/src/net/udp.rs
@@ -1955,6 +1955,79 @@ impl UdpSocket {
         self.io.set_multicast_loop_v6(on)
     }
 
+    /// Gets the value of the `IPV6_TCLASS` option for this socket.
+    ///
+    /// For more information about this option, see [`set_tclass_v6`].
+    ///
+    /// [`set_tclass_v6`]: Self::set_tclass_v6
+    // https://docs.rs/socket2/0.6.1/src/socket2/sys/unix.rs.html#2541
+    #[cfg(any(
+        target_os = "android",
+        target_os = "dragonfly",
+        target_os = "freebsd",
+        target_os = "fuchsia",
+        target_os = "linux",
+        target_os = "macos",
+        target_os = "netbsd",
+        target_os = "openbsd",
+        target_os = "cygwin",
+    ))]
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(any(
+            target_os = "android",
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "fuchsia",
+            target_os = "linux",
+            target_os = "macos",
+            target_os = "netbsd",
+            target_os = "openbsd",
+            target_os = "cygwin",
+        )))
+    )]
+    pub fn tclass_v6(&self) -> io::Result<u32> {
+        self.as_socket().tclass_v6()
+    }
+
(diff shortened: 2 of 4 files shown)


Darksonn:
We can't rename functions.

jtojnar:
Not even when keeping the original function as deprecated? I followed the example of https://github.com/tokio-rs/tokio/commit/28c9a14a2e4da99842d41240b301022017b2a04a.

I can drop the rename and keep the comment change, if the policy changed.

tokio/src/net/udp.rs
```diff
@@ -1950,11 +1950,84 @@ impl UdpSocket {
...
+
+    /// Sets the value for the `IPV6_TCLASS` option on this socket.
+    ///
+    /// Specifies the traffic class field that is used in every packet
+    /// sent from this socket.
+    ///
+    /// # Note
+    ///
+    /// This may not have any effect on IPv4 sockets.
+    // https://docs.rs/socket2/0.6.1/src/socket2/sys/unix.rs.html#2566
```
martin-g, line 2001: nit: This points to 0.6.1 but Tokio still uses 0.6.0 - https://github.com/tokio-rs/tokio/blob/064181f386414a543c0819af8c65866bbd879895/tokio/Cargo.toml#L103

jtojnar: Won’t cargo install 0.6.1 since it is semver-compatible?

martin-g: Actually, since it ends on `.0` it actually behaves like "0.6" and the latest patch release will be used.

jtojnar: Right, which is 0.6.1 at the moment. I used a fixed version rather than 0.6, since with the latter the anchor might break if future 0.6.x releases add/remove lines prior to linked lines.

jtojnar:
> I'd suggest to add some tests though!

I do not see any setsockopt tests in https://github.com/tokio-rs/tokio/blob/master/tokio/tests/udp.rs

I guess I could check if getter returns the value passed to a setter like the [underlying library](ttps://github.com/rust-lang/socket2/blob/299c2e511024bfc0651e25becaeff4a9701fd8d1/tests/socket.rs#L1534) does but  not sure if it is worth it duplicating it here.

martin-g:
> not sure if it is worth it duplicating it here.

I think it is's worth it. If Tokio's setter gets broken somehow and stops delegating to the underlying library then the getter will return the wrong value.
Same for the getters and the deprecated functions.

I understand that it is an effort! You could wait for others' opinion on the current changes first! And I could help with the tests if you like !

ADD-SP:
Hi @jtojnar , could you add a simple getter/setter test?

jtojnar:
Added option test for all existing options, mirroring the [tests in socket2 library](https://github.com/rust-lang/socket2/blob/a18be6a302b7f9c127c3593edec5d8d2690839a7/tests/socket.rs#L1345).

For some reason `multicast_ttl_v4` fails on CI for Windows (in `setsockopt`) and on MacOS in `getsockopt`) so made those conditional.

tokio/src/net/tcp/socket.rs
```diff
@@ -516,16 +614,19 @@ impl TcpSocket {
...
     /// sent from this socket.
     ///
-    /// **NOTE:** On Windows, `IP_TOS` is only supported on [Windows 8+ or
-    /// Windows Server 2012+.](https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options)
+    /// # Note
+    ///
+    /// - This may not have any effect on IPv6 sockets.
+    /// - On Windows, `IP_TOS` is only supported on [Windows 8+ or
+    ///   Windows Server 2012+.](https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options)
     // https://docs.rs/socket2/0.5.3/src/socket2/socket.rs.html#1446
```
martin-g: Let's update this link to 0.6.1, like in the other related docs.

tokio/src/net/udp.rs
```diff
@@ -2027,16 +2125,19 @@ impl UdpSocket {
...
     /// sent from this socket.
     ///
-    /// **NOTE:** On Windows, `IP_TOS` is only supported on [Windows 8+ or
-    /// Windows Server 2012+.](https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options)
+    /// # Note
+    ///
+    /// - This may not have any effect on IPv6 sockets.
+    /// - On Windows, `IP_TOS` is only supported on [Windows 8+ or
+    ///   Windows Server 2012+.](https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options)
     // https://docs.rs/socket2/0.5.3/src/socket2/socket.rs.html#1446
```
martin-g: Let's update this link to 0.6.1, like in the other related docs.

tokio/tests/udp.rs
```diff
@@ -644,3 +644,83 @@ async fn poll_ready() {
...
+    };
+}
+
+test!(broadcast, set_broadcast(true));
+
+test!(IPv4 multicast_loop_v4, set_multicast_loop_v4(false));
+
+#[cfg(not(target_os = "windows"))]
+// failed to set option: Os { code: 10022, kind: InvalidInput, message: "An invalid argument was supplied." }
+#[cfg(not(target_os = "macos"))] // failed to get initial value: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
```
martin-g: It seems the same is needed for freebsd too - https://github.com/tokio-rs/tokio/pull/7781/checks?check_run_id=59027543250

jtojnar: ~~Weirdly, it [fails even on Linux](https://github.com/rust-lang/socket2/actions/runs/20551407174/job/59029589312?pr=630) on [CI for socket2 library](https://github.com/rust-lang/socket2/pull/630) so not sure if I should even include it here.~~ **Edit:** Never mind, that was because I accidentally used `SOCK_STREAM` instead of `SOCK_DGRAM` in tests. Now socket2 is all green, even without using `c_uchar`, which is unexpected the other way around.

tokio/src/net/tcp/socket.rs
jtojnar: Done.

tokio/src/net/udp.rs
jtojnar: Done.

martin-g (approved):

tokio/src/net/udp.rs
```diff
@@ -2002,13 +2075,38 @@ impl UdpSocket {
 
     /// Gets the value of the `IP_TOS` option for this socket.
     ///
-    /// For more information about this option, see [`set_tos`].
+    /// For more information about this option, see [`set_tos_v4`].
     ///
-    /// **NOTE:** On Windows, `IP_TOS` is only supported on [Windows 8+ or
-    /// Windows Server 2012+.](https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options)
```
ADD-SP, line 2008: Emm, was this note removed by accident. Or it is intended?

tokio/src/net/tcp/socket.rs
```diff
@@ -516,17 +614,20 @@ impl TcpSocket {
...
+        self.tos_v4()
     }
 
     /// Sets the value for the `IP_TOS` option on this socket.
     ///
     /// This value sets the type-of-service field that is used in every packet
     /// sent from this socket.
     ///
-    /// **NOTE:** On Windows, `IP_TOS` is only supported on [Windows 8+ or
-    /// Windows Server 2012+.](https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options)
```
ADD-SP, line 528: Emm, was this note removed by accident. Or it is intended?

martin-g: It has been moved few lines below - https://github.com/tokio-rs/tokio/pull/7781/changes#diff-dc3a76e08a59d6fb723e2eefc2d1aab18685c22138084f58df8718b3ba1ccfddR628-R629

tokio/src/net/udp.rs
martin-g: It has been moved to the new `set_tos_v4()` method: 
https://github.com/tokio-rs/tokio/pull/7781/changes#diff-044489bdfaedc0e1a180059cd5ade7cb3f0a45f5d6802168a0d80a2cf8d67eadR2139-R2140

The deprecated `set_tos()` points to the new one, so it is assumed that the user will read it too.

PR #7782, opened 2025-12-19
Support as many abstract datagram unix domain socket operations as possible

ntninja:
## Motivation

Currently connecting/binding to the abstract namespace only works with stream Unix sockets.

## Solution

This adds binding to abstract namespace names on datagram Unix sockets and returns error and documents limitations for other datagram socket OPs, as there is no support for specifying a SocketAddr for these in MIO currently unfortunately.

### DIFF
--- a/tokio/tests/uds_datagram.rs
+++ b/tokio/tests/uds_datagram.rs
@@ -8,6 +8,10 @@ use tokio::try_join;
 
 use std::future::poll_fn;
 use std::io;
+#[cfg(any(target_os = "linux", target_os = "android"))]
+use std::os::linux::net::SocketAddrExt;
+#[cfg(any(target_os = "linux", target_os = "android"))]
+use std::os::unix::net as std_net;
 use std::sync::Arc;
 
 async fn echo_server(socket: UnixDatagram) -> io::Result<()> {
@@ -66,6 +70,49 @@ async fn echo_from() -> io::Result<()> {
         let (len, addr) = socket.recv_from(&mut recv_buf[..]).await?;
         assert_eq!(&recv_buf[..len], b"ECHO");
         assert_eq!(addr.as_pathname(), Some(server_path.as_path()));
+    }
+
+    Ok(())
+}
+
+#[tokio::test]
+#[cfg_attr(miri, ignore)] // No `socket` on miri.
+async fn recv_abstract() -> io::Result<()> {
+    // Generate 128-bit random string prefixed by `\0` as socket path
+    let abstract_path = format!(
+        "\0{:016x}{:016x}",
+        rand::random::<u64>(),
+        rand::random::<u64>()
+    );
+
+    // On non-Linux platforms, abstract socket paths are rejected
+    #[cfg(not(any(target_os = "linux", target_os = "android")))]
+    assert_eq!(
+        UnixDatagram::bind(&abstract_path).err().map(|e| e.kind()),
+        Some(io::ErrorKind::Unsupported)
+    );
+
+    // Remaining test is Linux-only
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    {
+        // Create socket in abstract namespace
+        let recv_socket = UnixDatagram::bind(&abstract_path)?;
+
+        // Send message from stdlib
+        //
+        //FIXME: Use `tokio` native here once passing abstract socket paths to
+        //       `send_to`/`connect` is implemented
+        let send_socket = std_net::UnixDatagram::unbound()?;
+        send_socket.send_to_addr(
+            b"READY=1\n",
+            &std_net::SocketAddr::from_abstract_name(&abstract_path[1..])?,
+        )?;
+
+        // Receive and validate message
+        let mut buf = vec![0u8; 32];
+        let (len, addr) = recv_socket.recv_from(&mut buf).await?;
+        assert_eq!(&buf[0..len], b"READY=1\n");
+        assert!(addr.is_unnamed());
     }
 
     Ok(())


ntninja:
This the test failure related to this? (Doesn’t look to me, at least.)

ntninja:
“Fixed” CI by rebasing

Darksonn:
I have a hard time understanding how and whether this changes behavior of Tokio.

Please add some tests that attempt to use these APIs with the abstract namespace, and set up your commits so that I can see the behavior before and after your change.

ntninja:
@Darksonn: Added a test for using `UnixDatagram::bind()` with an appropriately formatted abstract socket path. It tests that this fails in the expected way on non-Linux as well. The usage of the standard library support for abstract datagram unix domain sockets to send a datagram to the bound socket is because it is impossible to implement support for *sending* on the `UnixDatagram` object without prior support in `mio`.

tokio/tests/uds_datagram.rs
```diff
@@ -8,6 +8,10 @@ use tokio::try_join;
 
 use std::future::poll_fn;
 use std::io;
+#[cfg(any(target_os = "linux", target_os = "android"))]
+use std::os::linux::net::SocketAddrExt;
```
martin-g, line 12:
```suggestion
#[cfg(target_os = "android")]
use std::os::android::net::SocketAddrExt;
#[cfg(target_os = "linux")]
use std::os::linux::net::SocketAddrExt;
```

tokio/src/net/unix/datagram/socket.rs
```diff
@@ -605,6 +613,16 @@ impl UnixDatagram {
     /// # }
     /// ```
     pub fn connect<P: AsRef<Path>>(&self, path: P) -> io::Result<()> {
+        //FIXME: Convert this to `mio::net::UnixDatagram::connect_addr` once
+        //       that is supported by MIO
+        #[cfg(any(target_os = "linux", target_os = "android"))]
+        if path.as_ref().as_os_str().as_bytes().starts_with(b"\0") {
+            return Err(io::Error::new(
+                io::ErrorKind::Other,
```
martin-g:
```suggestion
                io::ErrorKind:: InvalidInput,
```
Isn't InvalidInput more appropriate ?
Same for the other checks below. Maybe even extract this snippet of code into a helper method and pass the function name as a parameter.

tokio/tests/uds_datagram.rs
```diff
@@ -71,6 +75,50 @@ async fn echo_from() -> io::Result<()> {
     Ok(())
 }
 
+#[allow(clippy::bool_assert_comparison)]
```
martin-g:
```suggestion
```

tokio/tests/uds_datagram.rs
```diff
@@ -71,6 +75,50 @@ async fn echo_from() -> io::Result<()> {
...
+        send_socket.send_to_addr(
+            b"READY=1\n",
+            &std_net::SocketAddr::from_abstract_name(&abstract_path[1..])?,
+        )?;
+
+        // Receive and validate message
+        let mut buf = vec![0u8; 32];
+        let (len, addr) = recv_socket.recv_from(&mut buf).await?;
+        assert_eq!(&buf[0..len], b"READY=1\n");
+        assert_eq!(addr.is_unnamed(), true);
```
martin-g:
```suggestion
        assert!(addr.is_unnamed());
```

tokio/src/net/unix/stream.rs
```diff
@@ -73,19 +67,7 @@ impl UnixStream {
...
-            if os_str_bytes.starts_with(b"\0") {
-                StdSocketAddr::from_abstract_name(&os_str_bytes[1..])?
-            } else {
-                StdSocketAddr::from_pathname(path)?
-            }
-        };
-        #[cfg(not(any(target_os = "linux", target_os = "android")))]
-        let addr = StdSocketAddr::from_pathname(path)?;
-
+        let addr = SocketAddr::from_path(path)?.0;
```
martin-g: Instead of using `.0` it would be better to introduce `pub(crate) fn into_std(self)` accessor

tokio/src/net/unix/socketaddr.rs
```diff
@@ -7,9 +15,33 @@ use std::path::Path;
 /// can convert to and from the standard library `SocketAddr` type using the
 /// [`From`] trait.
 #[derive(Clone)]
-pub struct SocketAddr(pub(super) std::os::unix::net::SocketAddr);
+pub struct SocketAddr(pub(super) StdSocketAddr);
 
 impl SocketAddr {
+    /// Creates an address from a path
+    ///
+    /// On Linux, allows the path to start with a leading `\0` to declare
```
martin-g:
```suggestion
    /// On Linux/Android, allows the path to start with a leading `\0` to declare
```

tokio/src/net/unix/datagram/socket.rs
```diff
@@ -1225,6 +1278,16 @@ impl UnixDatagram {
     where
         P: AsRef<Path>,
     {
+        //FIXME: Convert this to `mio::net::UnixDatagram::send_to_addr` once
+        //       that is supported by MIO
+        #[cfg(any(target_os = "linux", target_os = "android"))]
+        if target.as_ref().as_os_str().as_bytes().starts_with(b"\0") {
+            return Poll::Ready(Err(io::Error::new(
+                io::ErrorKind::Other,
+                "UnixDatagram::send_to(): abstract paths currently unsupported",
```
martin-g:
```suggestion
                "UnixDatagram::poll_send_to(): abstract paths currently unsupported",
```

tokio/src/net/unix/datagram/socket.rs
```diff
@@ -694,6 +712,11 @@ impl UnixDatagram {
 
     /// Tries to send a datagram to the peer without waiting.
     ///
+    /// # Limitations
+    ///
+    /// This method currently does not allow specifying Linux abstract
```
martin-g:
```suggestion
    /// This method currently does not allow specifying Linux/Android abstract
```

tokio/src/net/unix/datagram/socket.rs
```diff
@@ -1066,6 +1099,11 @@ impl UnixDatagram {
     /// [`tokio::select!`](crate::select) statement and some other branch
     /// completes first, then it is guaranteed that the message was not sent.
     ///
+    /// # Limitations
+    ///
+    /// This method currently does not allow specifying Linux abstract
```
martin-g:
```suggestion
    /// This method currently does not allow specifying Linux/Android abstract
```

tokio/src/net/unix/datagram/socket.rs
```diff
@@ -1205,6 +1253,11 @@ impl UnixDatagram {
     /// `Waker` from the `Context` passed to the most recent call will be scheduled to
     /// receive a wakeup.
     ///
+    /// # Limitations
+    ///
+    /// This method currently does not allow specifying Linux abstract
```
martin-g:
```suggestion
    /// This method currently does not allow specifying Linux/Android abstract
```

tokio/src/net/unix/datagram/socket.rs
```diff
@@ -571,6 +574,11 @@ impl UnixDatagram {
     /// The `send` method may be used to send data to the specified address.
     /// `recv` and `recv_from` will only receive data from that address.
     ///
+    /// # Limitations
+    ///
+    /// This method currently does not allow specifying Linux abstract
```
martin-g:
```suggestion
    /// This method currently does not allow specifying Linux/Android abstract
```

tokio/src/net/unix/datagram/socket.rs
ntninja: I mean, the input is actually correct. Support for it is just not currently implemented by Tokio/MIO, so it’d be `io::ErrorKind::Unsupported`… which turns out to be thing, I’ll change to that!

tokio/tests/uds_datagram.rs
ntninja: There is no `std::os::android::net::SocketAddrExt`. (Android is Linux based and inherits abstract sockets from Linux, btw.)

tokio/src/net/unix/datagram/socket.rs
Darksonn: Can we just add this to mio?

ntninja: Can we take this incremental improvement and then talk about this?

PR #7786, opened 2025-12-23
fs: handle `EINTR` in `fs::write` for io-uring

vrtgs:
internal change fixing a todo in io-uring code

### DIFF
--- a/tokio/src/fs/write.rs
+++ b/tokio/src/fs/write.rs
@@ -72,12 +72,14 @@ async fn write_uring(path: &Path, mut buf: OwnedBuf) -> io::Result<()> {
     let mut buf_offset: usize = 0;
     let mut file_offset: u64 = 0;
     while buf_offset < total {
-        let (n, _buf, _fd) = Op::write_at(fd, buf, buf_offset, file_offset)?.await;
-        // TODO: handle EINT here
-        let n = n?;
-        if n == 0 {
-            return Err(io::ErrorKind::WriteZero.into());
-        }
+        let (res, _buf, _fd) = Op::write_at(fd, buf, buf_offset, file_offset)?.await;
+
+        let n = match res {
+            Ok(0) => return Err(io::ErrorKind::WriteZero.into()),
+            Ok(n) => n,
+            Err(e) if e.kind() == io::ErrorKind::Interrupted => 0,
+            Err(e) => return Err(e),
+        };
 
         buf = _buf;
         fd = _fd;


tokio/src/fs/write.rs
```diff
@@ -73,11 +73,17 @@ async fn write_uring(path: &Path, mut buf: OwnedBuf) -> io::Result<()> {
     let mut file_offset: u64 = 0;
     while buf_offset < total {
         let (n, _buf, _fd) = Op::write_at(fd, buf, buf_offset, file_offset)?.await;
-        // TODO: handle EINT here
-        let n = n?;
-        if n == 0 {
-            return Err(io::ErrorKind::WriteZero.into());
-        }
+
+        let is_eint = |err: &io::Error| {
```
martin-g:
```suggestion
        let is_eintr = |err: &io::Error| {
```

tokio/src/fs/write.rs
```diff
@@ -73,11 +73,17 @@ async fn write_uring(path: &Path, mut buf: OwnedBuf) -> io::Result<()> {
...
+            err.raw_os_error() == Some(libc::EINTR) || err.kind() == io::ErrorKind::Interrupted
+        };
+
+        let n = match n {
+            Ok(0) => return Err(io::ErrorKind::WriteZero.into()),
+            Ok(n) => n,
+            Err(e) if is_eint(&e) => 0,
```
martin-g:
```suggestion
            Err(e) if is_eintr(&e) => 0,
```

tokio/src/fs/write.rs
```diff
@@ -73,11 +73,17 @@ async fn write_uring(path: &Path, mut buf: OwnedBuf) -> io::Result<()> {
...
     while buf_offset < total {
         let (n, _buf, _fd) = Op::write_at(fd, buf, buf_offset, file_offset)?.await;
-        // TODO: handle EINT here
-        let n = n?;
-        if n == 0 {
-            return Err(io::ErrorKind::WriteZero.into());
-        }
+
+        let is_eint = |err: &io::Error| {
+            err.raw_os_error() == Some(libc::EINTR) || err.kind() == io::ErrorKind::Interrupted
```
martin-g: This is the first usage of `libc::EINTR` in Tokio. 
All other places just check for `kind == io::ErrorKind::Interrupted`.
Have you faced cases where the raw error is EINTR but the kind is not `io::ErrorKind::Interrupted` ?

vrtgs: nol; it always resolvees, but when I saw the coment I thought they meant literally EINTR

vrtgs:
I have no idea how my small patch in some io-uring code broke FreeBSD 64 bit (but not FreeBSD 32 bit somehow); either spurious or someone else broke this or somehow not doing `err.raw_os_error() == Some(libc::EINTR)` broke it

Darksonn:
CI can fail spuriously sometimes.

tokio/src/fs/write.rs
```diff
@@ -73,11 +73,13 @@ async fn write_uring(path: &Path, mut buf: OwnedBuf) -> io::Result<()> {
     let mut file_offset: u64 = 0;
     while buf_offset < total {
         let (n, _buf, _fd) = Op::write_at(fd, buf, buf_offset, file_offset)?.await;
-        // TODO: handle EINT here
-        let n = n?;
-        if n == 0 {
-            return Err(io::ErrorKind::WriteZero.into());
-        }
+
+        let n = match n {
```
ADD-SP:
```suggestion
        let n = match res {
```

Since `n` is a `Result`, I prefer to rename it to `res`.

vrtgs: thats fine, I actually agree

I try to keep the original code an edit as little as possible, like for exampl imo casts are bad, either use `<int>::from` or `<int>::try_from` or annotate with a comment why this cast always succeeds, but im pretty sure if I change a codebase in that way 99% its gonna get rejected lol

ADD-SP (approved):
Thanks!

PR #7790, opened 2025-12-24
add more owned buffers to the upgrade list

vrtgs:
## Motivation
`Arc<[u8]>` and `Box<[u8]>` are moderately popular and supporting them takes minimal effort

## Solution
turn `Box<[u8]>` -> `Vec<u8>` and add support for `Arc<[u8]>` as an OwnedBuf

### DIFF
--- a/tokio/src/util/as_ref.rs
+++ b/tokio/src/util/as_ref.rs
@@ -1,8 +1,10 @@
 use super::typeid;
+use std::sync::Arc;
 
 #[derive(Debug)]
 pub(crate) enum OwnedBuf {
     Vec(Vec<u8>),
+    Arc(Arc<[u8]>),
     #[cfg(feature = "io-util")]
     Bytes(bytes::Bytes),
 }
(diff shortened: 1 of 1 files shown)


ADD-SP (commented):
Emm, why do we need this?

vrtgs:
Passing in a `Box<[u8]>` is actually pretty common, at least for me, and so is `Arc<[u8]>`, I'm just wondering why **shouldn't** we support them?
I used to roll my own `tokio::fs::write` and friends, that specifically don't copy buffers

And Its an internal change, no breaking api change with minimal footprint

ADD-SP:
Thanks for your PR! We don't do this kind of internal change unless tokio needs it.

vrtgs:
> Thanks for your PR! We don't do this kind of internal change unless tokio needs it.

I mean I still think this is a decent addition after all look no further than `tokio_util::io::read_exact_arc` and ur gonna realize whilst `Arc<[u8]>` isn't used everywhere its not like it's very niche, and adding support for `Box<[u8]>` reqires as much code as ading support for String, and `Box<[u8]>` is also pretty popular

Darksonn:
Is this the logic used by `tokio::fs::write`?

PR #7791, opened 2025-12-24
fs: Implement io-uring::Op<Statx> and use it to implement fs::try_exists and complete the todo's in read_uring

vrtgs:
## Motivation
get us one step closer to real io-uring

### DIFF
--- a/tokio/Cargo.toml
+++ b/tokio/Cargo.toml
@@ -85,7 +85,7 @@ sync = []
 test-util = ["rt", "sync", "time"]
 time = []
 # Unstable feature. Requires `--cfg tokio_unstable` to enable.
-io-uring = ["dep:io-uring", "libc", "mio/os-poll", "mio/os-ext", "dep:slab"]
+io-uring = ["dep:io-uring", "linux-raw-sys", "libc", "mio/os-poll", "mio/os-ext", "dep:slab"]
 # Unstable feature. Requires `--cfg tokio_unstable` to enable.
 taskdump = ["dep:backtrace"]
 
@@ -111,6 +111,7 @@ tracing = { version = "0.1.29", default-features = false, features = ["std"], op
 # Requires `--cfg tokio_unstable` to enable.
 [target.'cfg(all(tokio_unstable, target_os = "linux"))'.dependencies]
 io-uring = { version = "0.7.6", default-features = false, optional = true }
+linux-raw-sys = { version = "0.12.1", optional = true }
 libc = { version = "0.2.168", optional = true }
 mio = { version = "1.0.1", default-features = false, features = ["os-poll", "os-ext"], optional = true }
 slab = { version = "0.4.9", optional = true }
--- a/tokio/src/fs/read_uring.rs
+++ b/tokio/src/fs/read_uring.rs
@@ -1,4 +1,4 @@
-use crate::fs::OpenOptions;
+use crate::fs::UringOpenOptions;
 use crate::runtime::driver::op::Op;
 
 use std::io;
@@ -17,15 +17,13 @@ const PROBE_SIZE_U32: u32 = PROBE_SIZE as u32;
 const MAX_READ_SIZE: usize = 64 * 1024 * 1024;
 
 pub(crate) async fn read_uring(path: &Path) -> io::Result<Vec<u8>> {
-    let file = OpenOptions::new().read(true).open(path).await?;
+    let fd = UringOpenOptions::new().read(true).open(path).await?;
 
-    // TODO: use io uring in the future to obtain metadata
-    let size_hint: Option<usize> = file.metadata().await.map(|m| m.len() as usize).ok();
+    let (size_hint, fd) = Op::metadata_fd(fd).await;
 
-    let fd: OwnedFd = file
-        .try_into_std()
-        .expect("unexpected in-flight operation detected")
-        .into();
+    let size_hint: Option<usize> = size_hint
+        .ok()
+        .map(|m| usize::try_from(m.len()).unwrap_or(usize::MAX));
 
     let mut buf = Vec::new();
 
--- a/tokio/src/fs/try_exists.rs
+++ b/tokio/src/fs/try_exists.rs
@@ -23,6 +23,39 @@ use std::path::Path;
 /// # }
 /// ```
 pub async fn try_exists(path: impl AsRef<Path>) -> io::Result<bool> {
-    let path = path.as_ref().to_owned();
+    let path = path.as_ref();
+
+    #[cfg(all(
+        tokio_unstable,
+        feature = "io-uring",
+        feature = "rt",
+        feature = "fs",
+        target_os = "linux"
+    ))]
+    {
+        let handle = crate::runtime::Handle::current();
+        let driver_handle = handle.inner.driver().io();
+        if driver_handle.check_and_init()? {
+            return try_exists_uring(path).await;
+        }
+    }
+
+    try_exists_spawn_blocking(path).await
+}
+
+cfg_io_uring! {
+    async fn try_exists_uring(path: &Path) -> io::Result<bool> {
+        use crate::runtime::driver::op::Op;
+
+        match Op::metadata(path)?.await {
+            Ok(_) => Ok(true),
+            Err(error) if error.kind() == io::ErrorKind::NotFound => Ok(false),
+            Err(error) => Err(error),
+        }
+    }
+}
+
+async fn try_exists_spawn_blocking(path: &Path) -> io::Result<bool> {
+    let path = path.to_owned();
     asyncify(move || path.try_exists()).await
 }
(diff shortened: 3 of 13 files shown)


tokio/src/fs/read_uring.rs
```diff
@@ -17,15 +17,13 @@ const PROBE_SIZE_U32: u32 = PROBE_SIZE as u32;
...
-    let size_hint: Option<usize> = file.metadata().await.map(|m| m.len() as usize).ok();
+    let (size_hint, fd) = Op::metadata_fd(fd).await;
 
-    let fd: OwnedFd = file
-        .try_into_std()
-        .expect("unexpected in-flight operation detected")
-        .into();
+    let size_hint: Option<usize> = size_hint
+        .ok()
+        .map(|m| usize::try_from(m.len()).unwrap_or(usize::MAX));
```
Daksh14, line 27: Here if we obtain size_hint `Err` for whatever reason, we should start with a length `0` capacity and gradually increase it

tokio/src/io/uring/statx.rs
```diff
@@ -0,0 +1,136 @@
...
+        // pass in an empty path instead of null as specified by man
+        // https://man7.org/linux/man-pages/man2/statx.2.html
+        let statx_op = opcode::Statx::new(
+            types::Fd(fd.as_raw_fd()),
+            c"".as_ptr(),
+            buffer.as_mut_ptr().cast(),
+        )
+        .flags(flags as i32)
+        .mask(linux_raw_sys::general::STATX_BASIC_STATS)
+        .build();
```
Daksh14, line 131: is there a guarantee that after this line the buffer will be initialized? Should we add a check after it completes? What if the kernel errored and the buffer still remains unint?

tokio/src/io/uring/statx.rs
```diff
@@ -0,0 +1,136 @@
...
+    buffer: Box<MaybeUninit<statx>>,
+}
+
+impl Completable for StatxFd {
+    type Output = (io::Result<Metadata>, OwnedFd);
+
+    fn complete(self, cqe: CqeResult) -> Self::Output {
+        let ret = cqe
+            .result
+            .map(|_| Metadata(*unsafe { box_assume_init(self.buffer) }));
```
Daksh14, line 98: I think we should add a check here to check if the buffer is init or not or a safety comment explaining why this is ok

tokio/src/io/uring/utils.rs
```diff
@@ -1,6 +1,19 @@
...
+// TODO(MSRV 1.82): When bumping MSRV, switch to `Box::<T>::new_uninit()`.
+pub(crate) fn box_new_uninit<T>() -> Box<MaybeUninit<T>> {
+    // Box::<T>::new_uninit()
+    Box::new(MaybeUninit::uninit())
+}
+
+// TODO(MSRV 1.82): When bumping MSRV, switch to `Box::<MaybeUninit<T>>::assume_init()`.
+pub(crate) unsafe fn box_assume_init<T>(boxed: Box<MaybeUninit<T>>) -> Box<T> {
+    let raw = Box::into_raw(boxed);
+    unsafe { Box::from_raw(raw as *mut T) }
```
Daksh14, line 18: nit: can make this function safe and keep the unsafe block in there or remove the unsafe block and keep the unsafe in the function

tokio/src/io/uring/statx.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+impl Op<Statx> {
+    /// Submit a request to open a file.
+    fn statx(path: &Path, follow_symlinks: bool) -> io::Result<Op<Statx>> {
+        let path = cstr(path)?;
+        let mut buffer = box_new_uninit::<statx>();
+
+        let flags: u32 = linux_raw_sys::general::AT_STATX_SYNC_AS_STAT
+            | (linux_raw_sys::general::AT_SYMLINK_NOFOLLOW * u32::from(!follow_symlinks));
```
Daksh14, line 62: You can create an member function that takes in a `flags: u32, path: &Path` then you can have `fn statx_follow_symlinks` and `fn statx_no_symlinks` than have a vague `bool` argument which might confuse future readers and users

Daksh14 (changes requested):
Thanks for this PR! Unfortunately we can't expose this using the std's `Metadata` to users of tokio because we can't construct one but needed internal usage in `read`, `write`, etc.

tokio/src/io/uring/statx.rs
vrtgs: Alright I will check the return to be exactly 0, if it is 0 then yes there is a guarantee

tokio/src/io/uring/utils.rs
vrtgs: I'm pretty sure unsafe is unsafe is like a pretty common `unsafe_op_in_unsafe_fn` is the default now

tokio/src/io/uring/statx.rs
vrtgs: I mean fn state is private, you will see bellow statix can be constructed with Op::metadata or Op::symlink_metadata

vrtgs:
> Thanks for this PR! Unfortunately we can't expose this using the std's `Metadata` to users of tokio because we can't construct one but needed internal usage in `read`, `write`, etc.

Yeah, I did realize that as well which is why I didn't rewrite tokio::fs::metadata 
Never the less as you said internal usage is pretty good, especially with things like `tokio::fs::try_exists` since that is decently common

vrtgs:
I am very sorry for the weird behavior btw, the github app sucks I accidentally closed the pr, when I respond to any comment 2 comment show up, sorry for the inconvenience

tokio/src/io/uring/statx.rs
vrtgs: Simple, if the kernel never returns Ok we just never read the buffer, its not like statx has any drop function

tokio/src/io/uring/utils.rs
Daksh14: yea my bad we can this as is

tokio/src/io/uring/statx.rs
```diff
@@ -0,0 +1,136 @@
...
+
+        let flags: u32 =
+            linux_raw_sys::general::AT_STATX_SYNC_AS_STAT | linux_raw_sys::general::AT_EMPTY_PATH;
+
+        // io-uring was introduced in linux 5.1
+        // pass in an empty path instead of null as specified by man
+        // https://man7.org/linux/man-pages/man2/statx.2.html
+        let statx_op = opcode::Statx::new(
+            types::Fd(fd.as_raw_fd()),
+            c"".as_ptr(),
```
martin-g, line 126: Hm. How does this build with 1.71 (current MSRV) ?!
https://github.com/vrtgs/tokio/blob/972c7c1c10efd2d7e5a7a7b2a273343beab450a2/tokio/Cargo.toml#L11
https://doc.rust-lang.org/edition-guide/rust-2021/c-string-literals.html says that c-string literals are available since Rust 1.77

tokio/src/fs/try_exists.rs
```diff
@@ -23,6 +23,39 @@ use std::path::Path;
...
+        if driver_handle.check_and_init()? {
+            return try_exists_uring(path).await;
+        }
+    }
+
+    try_exists_spawn_blocking(path).await
+}
+
+cfg_io_uring! {
+    async fn try_exists_uring(path: &Path) -> io::Result<bool> {
```
martin-g, line 47: nit: IMO this function can be inlined at line 39 above

tokio/Cargo.toml
```diff
@@ -111,6 +111,7 @@ tracing = { version = "0.1.29", default-features = false, features = ["std"], op
 # Requires `--cfg tokio_unstable` to enable.
 [target.'cfg(all(tokio_unstable, target_os = "linux"))'.dependencies]
 io-uring = { version = "0.7.6", default-features = false, optional = true }
+linux-raw-sys = { version = "0.12.1", optional = true }
```
martin-g, line 114: Is there a particular reason to not use `libc` instead ?
AFAIS it provides all the constants used in this PR from `linux-raw-sys`

martin-g:
BTW there was an earlier attempt for Statx support at https://github.com/tokio-rs/tokio/pull/7616. 
You may check it. There might be something useful to borrow from it.

tokio/src/fs/read_uring.rs
```diff
@@ -17,15 +17,13 @@ const PROBE_SIZE_U32: u32 = PROBE_SIZE as u32;
 const MAX_READ_SIZE: usize = 64 * 1024 * 1024;
 
 pub(crate) async fn read_uring(path: &Path) -> io::Result<Vec<u8>> {
-    let file = OpenOptions::new().read(true).open(path).await?;
```
Daksh14, line 20: I'd like to keep the `OpenOptions`, use it how you would without breaking/adding new changes and opening files with uring should work.

tokio/Cargo.toml
Darksonn: Please do not use `linux-raw-sys`.

Daksh14:
Thanks to @asder8215 for creating https://github.com/tokio-rs/tokio/pull/8080 which is what this PR attempted to do. 
This can be closed now as #8080 is merged

Thanks for your initial attempt at the PR!

PR #7792, opened 2025-12-24
io: add optimizer hint that `memchr` returns in-bounds pointer

vrtgs:
title

### DIFF
--- a/tokio/src/util/memchr.rs
+++ b/tokio/src/util/memchr.rs
@@ -4,22 +4,50 @@
 //! that we only depend on libc on unix.
 
 #[cfg(not(all(unix, feature = "libc")))]
-pub(crate) fn memchr(needle: u8, haystack: &[u8]) -> Option<usize> {
+fn memchr_inner(needle: u8, haystack: &[u8]) -> Option<usize> {
     haystack.iter().position(|val| needle == *val)
 }
 
 #[cfg(all(unix, feature = "libc"))]
-pub(crate) fn memchr(needle: u8, haystack: &[u8]) -> Option<usize> {
+fn memchr_inner(needle: u8, haystack: &[u8]) -> Option<usize> {
     let start = haystack.as_ptr();
 
     // SAFETY: `start` is valid for `haystack.len()` bytes.
-    let ptr = unsafe { libc::memchr(start.cast(), needle as _, haystack.len()) };
+    let ptr = (unsafe { libc::memchr(start.cast(), needle as _, haystack.len()) })
+        .cast::<u8>()
+        .cast_const();
 
     if ptr.is_null() {
         None
     } else {
-        Some(ptr as usize - start as usize)
+        // SAFETY: `ptr` will always be in bounds, since libc guarentees that the ptr will either
+        //          be to an element inside the array or the ptr will be null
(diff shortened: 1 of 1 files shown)


ADD-SP (commented):
Emm, what is the benefit of this change?

vrtgs:
> Emm, what is the benefit of this change?

It removes the panic, I'm on my phone right now, but in readline, it is used right afterwards used to index into the buffer to read the line, and the indexing operation is checked and had a panic branch

Since memchr is a native function that goes through ffi, the rust compiler assumes nothing about the return, and therefore does not know that the index returned is actually in bounds, and will check that the index is in bounds when using it as an index into haystack

The later TODO changes `.addr` honestly changes nothing aside from conforming to provenance which is less meaningful when using ffi, but the ptr offset from with rust msrv 1.87  its just a lot more clear what is happening + compiler can do fancy tricks with it (it doesn't change codegen here tho)

tokio/src/util/memchr.rs
```diff
@@ -4,22 +4,40 @@
...
+
+pub(crate) fn memchr(needle: u8, haystack: &[u8]) -> Option<usize> {
+    let index = memchr_inner(needle, haystack)?;
+
+    // TODO(MSRV 1.81): When bumping MSRV, switch to `std::hint::assert_unchecked(index < haystack.len());`
+    unsafe {
+        if index >= haystack.len() {
+            std::hint::unreachable_unchecked()
+        }
     }
```
Darksonn: New unsafe blocks must have a SAFETY comment that explains why this unsafe code does not result in UB. Also, new unsafe blocks should wrap the unsafe operation as closely as possible.

tokio/src/util/memchr.rs
```diff
@@ -4,22 +4,40 @@
...
+        .cast_const();
 
     if ptr.is_null() {
         None
     } else {
-        Some(ptr as usize - start as usize)
+        unsafe {
+            // TODO(MSRV 1.87): When bumping MSRV, switch to `ptr.byte_offset_from_unsigned(start)`.
+            Some(usize::try_from(ptr.offset_from(start)).unwrap_unchecked())
+        }
```
Darksonn, line 32: New unsafe blocks should wrap only a single unsafe operation and use SAFETY comments.
```suggestion
        // SAFETY: `memchr` returns null or a pointer from same allocation as
        // `start`, and we know it's not null.
        let offset = unsafe { ptr.offset_from(start) } ;
        // TODO(MSRV 1.87): When bumping MSRV, switch to `ptr.byte_offset_from_unsigned(start)`.
        // SAFETY: `memchr` returns a pointer between `start` and `start+len`,
        // so the offset is positive.
        Some(unsafe { usize::try_from(offset).unwrap_unchecked() })
```

Darksonn (commented):
You told me in Discord that you checked that this improves codegen in the callers. Can you share the details?

vrtgs:
> You told me in Discord that you checked that this improves codegen in the callers. Can you share the details?

[all panic branches are removed](https://rust.godbolt.org/z/61z8E6rfv)

aside from that i have also added safety comments to all unsafe blocks

PR #7793, opened 2025-12-24
implement io-uring for read_to_string.rs

vrtgs:


tokio/src/fs/read_to_string.rs
```diff
@@ -25,6 +25,36 @@ use std::{io, path::Path};
...
+        feature = "fs",
+        target_os = "linux"
+    ))]
+    {
+        use crate::fs::read_uring;
+
+        let handle = crate::runtime::Handle::current();
+        let driver_handle = handle.inner.driver().io();
+        if driver_handle.check_and_init()? {
+            return read_uring(path).await.and_then(|vec| {
```
martin-g, line 43:
```suggestion
            return read_uring(path).await.and_then(|bytes| {
```

tokio/src/fs/read_to_string.rs
```diff
@@ -25,6 +25,36 @@ use std::{io, path::Path};
...
+                String::from_utf8(vec).map_err(|_| {
```
martin-g, line 44:
```suggestion
                String::from_utf8(bytes).map_err(|err| {
```

tokio/src/fs/read_to_string.rs
```diff
@@ -25,6 +25,36 @@ use std::{io, path::Path};
...
+                    io::Error::new(
+                        io::ErrorKind::InvalidData,
+                        "stream did not contain valid UTF-8",
```
martin-g, line 47: The bytes might be invalid or incomplete. Return the original error. It has a nice Display impl:
```suggestion
                         err.utf8_error(),
```

tokio/src/fs/read_to_string.rs
martin-g, line 10: This part of the documentation needs to be updated

ADD-SP (commented):
CI is failing, could you investigate it?

mox692:
Hmm, some tests for current thread runtime are probably hitting this.

https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.block_on

> When this is used on a current_thread runtime, only the [Runtime::block_on](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.block_on) method can drive the IO and timer drivers, but the Handle::block_on method cannot drive them. This means that, when using this method on a current_thread runtime, anything that relies on IO or timers will not work unless there is another thread currently calling [Runtime::block_on](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.block_on) on the same runtime.

PR #7796, opened 2025-12-25
util: simplify lock hanlding

Tunglies:
Reduce lock holding time

No benchmark for this part.

### DIFF
--- a/tokio/src/util/idle_notified_set.rs
+++ b/tokio/src/util/idle_notified_set.rs
@@ -368,8 +368,6 @@ impl<'a, T> EntryInOneOfTheLists<'a, T> {
         self.set.length -= 1;
 
         {
-            let mut lock = self.set.lists.lock();
-
             // Safety: We are holding the lock so there is no race, and we will
             // remove the entry afterwards to uphold invariants.
             let old_my_list = self.entry.my_list.with_mut(|ptr| unsafe {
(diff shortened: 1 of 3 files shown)


tokio/src/util/idle_notified_set.rs
```diff
@@ -368,8 +368,6 @@ impl<'a, T> EntryInOneOfTheLists<'a, T> {
         self.set.length -= 1;
 
         {
-            let mut lock = self.set.lists.lock();
-
```
ADD-SP, line 372: We need this lock to synchronize the `self.entry.my_list`.

tokio/src/util/rand/rt.rs
```diff
@@ -34,6 +34,7 @@ impl RngSeedGenerator {
 
         let s = rng.fastrand();
         let r = rng.fastrand();
+        drop(rng);
 
         RngSeed::from_pair(s, r)
```
ADD-SP, line 39: `RngSeed::from_pair(s, r)` is an extremely lightweight operation; releasing the lock prematurely may not yield any benefits.

tokio/src/util/sharded_list.rs
```diff
@@ -63,8 +63,7 @@ impl<L: ShardedListItem> ShardedList<L, L::Target> {
     /// Removes the last element from a list specified by `shard_id` and returns it, or None if it is
     /// empty.
     pub(crate) fn pop_back(&self, shard_id: usize) -> Option<L::Handle> {
-        let mut lock = self.shard_inner(shard_id);
-        let node = lock.pop_back();
+        let node = self.shard_inner(shard_id).pop_back();
```
ADD-SP, line 66: This may not yield better readability to me.

PR #7799, opened 2025-12-26
implement unlink io-uring call to implement remove_file and remove_dir

vrtgs:
implement io-uring unlink.

note that we can't use this to implement remove_dir_all because there is no way to read directory entries uring io-uring and reading entries is a blocking operation

### DIFF
--- a/tokio/src/fs/remove_dir.rs
+++ b/tokio/src/fs/remove_dir.rs
@@ -7,6 +7,33 @@ use std::path::Path;
 ///
 /// This is an async version of [`std::fs::remove_dir`].
 pub async fn remove_dir(path: impl AsRef<Path>) -> io::Result<()> {
-    let path = path.as_ref().to_owned();
+    let path = path.as_ref();
+
+    #[cfg(all(
+        tokio_unstable,
+        feature = "io-uring",
+        feature = "rt",
+        feature = "fs",
+        target_os = "linux",
+    ))]
+    {
+        use crate::io::uring::unlink::Unlink;
+        use crate::runtime::driver::op::Op;
+
+        let handle = crate::runtime::Handle::current();
+        let driver_handle = handle.inner.driver().io();
+
+        type OpUnlink = Op<Unlink>;
+
+        if driver_handle.check_and_init(OpUnlink::CODE)? {
+            return OpUnlink::remove_dir(path)?.await;
+        }
+    }
+
+    remove_dir_blocking(path).await
+}
+
+async fn remove_dir_blocking(path: &Path) -> io::Result<()> {
+    let path = path.to_owned();
     asyncify(move || std::fs::remove_dir(path)).await
 }
--- a/tokio/src/fs/remove_dir.rs
+++ b/tokio/src/fs/remove_dir.rs
@@ -25,7 +25,7 @@ pub async fn remove_dir(path: impl AsRef<Path>) -> io::Result<()> {
 
         type OpUnlink = Op<Unlink>;
 
-        if driver_handle.check_and_init(OpUnlink::CODE)? {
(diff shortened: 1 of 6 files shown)


ADD-SP (commented):
Could you add tests?

tokio/src/io/uring/unlink.rs
```diff
@@ -0,0 +1,57 @@
...
+
+impl Cancellable for Unlink {
+    fn cancel(self) -> CancelData {
+        CancelData::CString(self.path)
+    }
+}
+
+impl Op<Unlink> {
+    /// Submit a request to open a file.
+    pub(crate) fn unlink(path: &Path, directory: bool) -> io::Result<Op<Unlink>> {
```
Daksh14: We can have `remove_dir` and `remove_file` and you can ask for `flags` in the argument, its not apparent what the directory `bool` is going to do

tokio/src/runtime/driver/op.rs
```diff
@@ -16,9 +16,9 @@ use std::task::{Context, Poll, Waker};
 #[allow(dead_code)]
 #[derive(Debug)]
 pub(crate) enum CancelData {
-    Open(Open),
     Write(Write),
     Read(Read),
+    CString(CString),
```
Daksh14: This is a typo, it should be `Unlink(Unlink)`

Daksh14 (changes requested):
Thank you! Tests are welcome

tokio/src/io/uring/unlink.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+impl Cancellable for Unlink {
+    fn cancel(self) -> CancelData {
+        CancelData::CString(self.path)
+    }
+}
+
+impl Op<Unlink> {
+    /// Submit a request to open a file.
```
martin-g:
```suggestion
    /// Submit a request to unlink a file or directory.
```

tokio/src/fs/remove_file.rs
```diff
@@ -11,6 +11,34 @@ use std::path::Path;
...
+        }
+    }
+
+    remove_file_blocking(path).await
+}
+
+cfg_io_uring! {
+    async fn remove_file_uring(path: &Path) -> io::Result<()> {
+        use crate::runtime::driver::op::Op;
+        Op::remove_file(path)?.await
```
martin-g: This is short enough to be inlined at line 27

tokio/src/fs/remove_dir.rs
```diff
@@ -7,6 +7,34 @@ use std::path::Path;
...
+        }
+    }
+
+    remove_dir_blocking(path).await
+}
+
+cfg_io_uring! {
+    async fn remove_dir_uring(path: &Path) -> io::Result<()> {
+        use crate::runtime::driver::op::Op;
+        Op::remove_dir(path)?.await
```
martin-g: This is short enough to be inlined at line 23

martin-g: Alternatively we could introduce `cfg_not_io_uring` macro and use the same function name for both. I.e. only one function will be compiled depending on whether io_uring is enabled or not.
Currently `remove_dir_blocking()` is always compiled, even if it is not used at all when io_uring is enabled.

vrtgs: I am on the path of unifiying all of them as in a macro yes, eventually, just after I get all the pull requests I sent sorted out

vrtgs:
> Thank you! Tests are welcome

I would love to, but I have 0 clue how the testing is structured around here, I was digging around a little, and I couldn't figure it out at a glance, I have finals currently so I ask for your patience on this request

vrtgs:
Sorry for the 6 month delay

PR #7800, opened 2025-12-26
fs: implement rename using io-uring

vrtgs:
## Motivation
more io-uring support


tokio/src/io/uring/rename.rs
```diff
@@ -0,0 +1,55 @@
+use super::utils::cstr;
+
+use crate::runtime::driver::op::{CancelData, Cancellable, Completable, CqeResult, Op};
+
+use io_uring::{opcode, types};
+use std::ffi::CString;
+use std::io;
+use std::io::Error;
```
martin-g: nit: You use `io::Result` below. Maybe do the same for Error too - `io::Error` ?! And remove this import here.

tokio/src/runtime/driver/op.rs
```diff
@@ -118,7 +120,7 @@ pub(crate) trait Completable {
     //
     // The `Op` type that implements this trait can return the passed error
     // upstream by embedding it in the `Output`.
-    fn complete_with_error(self, error: Error) -> Self::Output;
+    fn complete_with_error(self, error: io::Error) -> Self::Output;
```
martin-g, line 154: 👍🏻

martin-g (approved):

ADD-SP (commented):
Could you add tests?

tokio/src/io/uring/rename.rs
```diff
@@ -0,0 +1,55 @@
...
+use std::path::Path;
+
+#[derive(Debug)]
+pub(crate) struct Rename {
+    /// This field will be read by the kernel during the operation, so we
+    /// need to ensure it is valid for the entire duration of the operation.
+    #[allow(dead_code)]
```
Daksh14: nit: Can we apply this lint to the whole struct rather than individual elements?

Daksh14 (changes requested):
Looks good, thanks a lot, last thing we would need is some tests, checkout fs_read_uring.rs tests and other uring tests for examples

mox692 (approved):
The code itself looks good to me.

tokio/src/io/uring/rename.rs
vrtgs: doing that means hey, this whole structure won't be made

tokio/src/io/uring/rename.rs
vrtgs: I 100% agree

vrtgs:
oi somone broke my code not me

vrtgs:
I have no clue why CI failed

vrtgs:
> Looks good, thanks a lot, last thing we would need is some tests, checkout fs_read_uring.rs tests and other uring tests for examples

in my very humble opinion, with only a bit of looking, but tokio tests suck, and have entirely too much code duplication

tokio/src/io/uring/rename.rs
Daksh14: I get that but it still stops the clippy complain and makes the struct more readable by keeping it on top. It's just a nit tho

Daksh14:
> > Looks good, thanks a lot, last thing we would need is some tests, checkout fs_read_uring.rs tests and other uring tests for examples
> 
> in my very humble opinion, with only a bit of looking, but tokio tests suck, and have entirely too much code duplication

You are right but it's just boilerplate code, the test themselves test specific component. Once we add more uring tests this infrastructure can be moved to the common module but note that uring is still early to tokio @vrtgs

tokio/src/io/uring/rename.rs
Darksonn: If there are specific fields that are read through mechanisms such as the kernel, then prefix the name with an underscore.

mattiapitossi:
Hi @vrtgs, are you still interested in finishing this one?

vrtgs:
> Hi @vrtgs, are you still interested in finishing this one?

Yes I am, super sorry I got busy and forgot about this, I'll try to finish it ASAP, hopefully before the end of this week

mattiapitossi:
> > Hi @vrtgs, are you still interested in finishing this one?
> 
> Yes I am, super sorry I got busy and forgot about this, I'll try to finish it ASAP, hopefully before the end of this week

Sure, no worries, take your time!

tokio/src/runtime/driver/op.rs
```diff
@@ -35,6 +36,8 @@ use std::task::{Context, Poll, Waker};
 pub(crate) enum CancelData {
     Open(Open),
     Write(Write),
+    Read(Read),
```
mattiapitossi: Is this needed?

tokio/src/io/uring/rename.rs
vrtgs: alright

mattiapitossi (approved):
Thanks!

Daksh14 (approved):
One nit otherwise LGTM

tokio/src/fs/rename.rs
```diff
@@ -10,8 +10,34 @@ use std::path::Path;
...
+    {
+        use crate::io::uring::rename::Rename;
+        use crate::runtime::driver::op::Op;
+
+        let handle = crate::runtime::Handle::current();
+        let driver_handle = handle.inner.driver().io();
+
+        type RenameOp = Op<Rename>;
+
+        if driver_handle.check_and_init(RenameOp::CODE).await? {
```
Daksh14: nit: for consistency reaons
```suggestion
        if driver_handle.check_and_init(io_uring::opcode::RenameAt).await? {
```

PR #7801, opened 2025-12-27
time: improve the readability of alternative timer

vilgotf:
Replace imperative code with iterators.

### DIFF
--- a/tokio/src/runtime/time_alt/wheel/mod.rs
+++ b/tokio/src/runtime/time_alt/wheel/mod.rs
@@ -115,9 +115,6 @@ impl Wheel {
 
             self.set_elapsed(expiration.deadline);
         }
-        // in this case the poll did not indicate an expiration _and_ we were
-        // not able to find a next expiration in the current list of timers.
-        // Advance to the poll's current time and do nothing else.
         self.set_elapsed(now);
     }
 


tokio/src/runtime/scheduler/util/time_alt.rs
```diff
@@ -139,11 +139,9 @@ pub(crate) fn shutdown_local_timers(
 
         let mut wake_queue = WakeQueue::new();
         // simply wake all unregistered timers
-        for hdl in inject {
-            if !hdl.is_cancelled() {
-                unsafe {
-                    wake_queue.push_front(hdl);
-                }
+        for hdl in inject.into_iter().filter(|hdl| !hdl.is_cancelled()) {
```
ADD-SP, line 142: I'm open to both `for` loop style and iterator style, since style changes are confined to the `time_alt` module. We can maintain consistent style within that module.

Given that `iter.filter` already here, I prefer using `iter.filter().for_each()` over `for elem in iter.filter() {}`.

tokio/src/runtime/time_alt/wheel/mod.rs
```diff
@@ -150,17 +147,10 @@ impl Wheel {
...
-                    res = false;
-                }
-            }
-        }
-
-        res
+        self.levels[start_level..]
+            .iter()
+            .flat_map(|level| level.next_expiration(self.elapsed))
+            .all(|e2| before < e2.deadline)
```
ADD-SP:
```suggestion
            .all(|e2| before <= e2.deadline)
```

Am I right?

tokio/src/runtime/time_alt/wheel/mod.rs
```diff
@@ -107,39 +107,36 @@ impl Wheel {
...
-                debug_assert!(self.no_expirations_before(level_num + 1, expiration.deadline));
-
-                return Some(expiration);
-            }
-        }
+        self.levels
+            .iter()
+            .enumerate()
+            .find_map(|(level_num, level)| {
+                let expiration = level.next_expiration(self.elapsed);
```
ADD-SP:
```suggestion
                let expiration = level.next_expiration(self.elapsed)?;
```

Is this better for readability?

vilgotf: The try operator improves the current code, but makes the `inspect` alternative less obvious. In sum I prefer your suggestion

tokio/src/runtime/scheduler/util/time_alt.rs
vilgotf: Deferring to std, using a `for` loop is more idiomatic than `for_each` in this case. Quoting `for_each`:
> This is equivalent to using a [`for`](https://doc.rust-lang.org/stable/book/ch03-05-control-flow.html#looping-through-a-collection-with-for) loop on the iterator, although `break` and `continue` are not possible from a closure. It's generally more idiomatic to use a `for` loop, but `for_each` may be more legible when processing items at the end of longer iterator chains. In some cases `for_each` may also be faster than a loop, because it will use internal iteration on adapters like `Chain`.

ADD-SP: This makes sense! Thanks!

tokio/src/runtime/time_alt/wheel/mod.rs
```diff
@@ -107,39 +107,34 @@ impl Wheel {
...
-                    // current time and do nothing else.
-                    self.set_elapsed(now);
-                    break;
-                }
-            }
+            self.set_elapsed(expiration.deadline);
         }
+        // in this case the poll did not indicate an expiration _and_ we were
+        // not able to find a next expiration in the current list of timers.
+        // Advance to the poll's current time and do nothing else.
```
ADD-SP: Shall we update this comment?

vilgotf: Deleted; I think the method is simple enough that it no longer requires an additional comment besides its doc comment

ADD-SP (approved):
Thanks!

PR #7802, opened 2025-12-28
rt: clarify the behavior of `Runtime::spawn`

Its-Just-Nans:
## Motivation

This improve the documentation for handle in a `new_current_thread()` runtime

Notably because

```
The provided future will start running in the background immediately
when `spawn` is called, even if you don't await the returned `JoinHandle`.
```

seems incorrect on a blocking runtime `new_current_thread` no?

```rust
use std::{thread::sleep, time::Duration};

fn main() {
    use tokio::runtime;
    use tokio::task;

    let rt = runtime::Builder::new_current_thread().build().unwrap();

    println!("RT");
    let join_handle: task::JoinHandle<_> = rt.spawn(async {
        println!("JOB");
        // some work here
    });

    println!("SLEEP");
    sleep(Duration::from_secs(5));
    println!("SLEEP END");

    rt.block_on(join_handle).unwrap();
    println!("END");
}
```

```
RT
SLEEP
SLEEP END
JOB
END
```

On a new classic runtime, the documentation is correct, the spawn is running in the background immediately

```rust
use std::{thread::sleep, time::Duration};

fn main() {
    use tokio::runtime::Runtime;
    use tokio::task;

    let rt = Runtime::new().unwrap();

    println!("RT");
    let join_handle: task::JoinHandle<_> = rt.spawn(async {
        println!("JOB");
        // some work here
    });

    println!("SLEEP");
    sleep(Duration::from_secs(5));
    println!("SLEEP END");

    rt.block_on(join_handle).unwrap();
    println!("END");
}
```

```
RT
SLEEP
JOB
SLEEP END
END
```

## Solution

Add documentation

~Another solution would be to gate `rt.spawn(` for `rt-multi-thread` no?~ -> no because `new_current_thread` can be called with `rt-multi-thread`

### DIFF
--- a/tokio/src/runtime/runtime.rs
+++ b/tokio/src/runtime/runtime.rs
@@ -206,7 +206,7 @@ impl Runtime {
     /// thread pool. The thread pool is then responsible for polling the future
     /// until it completes.
     ///
-    /// In a non blocking env, the provided future will start running in the background
+    /// In a non blocking environment, the provided future will start running in the background
     /// immediately when `spawn` is called, even if you don't await the returned
     /// [`JoinHandle`].
     ///
@@ -251,7 +251,7 @@ impl Runtime {
     ///     println!("now running on a worker thread");
     /// });
     ///
-    /// rt.block_on(handle);
+    /// rt.block_on(handle).unwrap();
     /// # }
     /// # }
     /// ```


tokio/src/runtime/task/join.rs
```diff
@@ -152,6 +152,49 @@ cfg_rt! {
...
+    ///
+    /// let join_handle: task::JoinHandle<_> = rt.spawn(async {
+    ///     // some work here
+    /// });
+    ///
+    /// rt.block_on(join_handle).unwrap();
+    /// # }
+    /// # }
+    /// ```
+    ///
```
ADD-SP, line 197:
```suggestion
```

The docs of `JoinHandle` is already long enough, adding more trivial docs may not bring significant benefits.

tokio/src/runtime/runtime.rs
```diff
@@ -232,6 +232,29 @@ impl Runtime {
...
+    ///
+    /// // Spawn a future onto the runtime
+    /// let handle = rt.spawn(async {
+    ///     println!("now running on a worker thread");
+    /// });
+    ///
+    /// rt.block_on(handle).unwrap();
+    /// # }
+    /// # }
+    /// ```
```
ADD-SP, line 257: This is too trivial and may not bring significant benefits.

tokio/src/runtime/runtime.rs
```diff
@@ -206,9 +206,9 @@ impl Runtime {
     /// thread pool. The thread pool is then responsible for polling the future
     /// until it completes.
     ///
-    /// The provided future will start running in the background immediately
-    /// when `spawn` is called, even if you don't await the returned
-    /// `JoinHandle`.
+    /// In a non blocking environment, the provided future will start running in the background
+    /// immediately when `spawn` is called, even if you don't await the returned
+    /// [`JoinHandle`].
```
ADD-SP, line 211: > In a non blocking environment

Actually, this is not true. This is not about the "non blocking environment", just because the task queue of current thread runtime can only be polled during the `rt.block_on`.

ADD-SP (commented):
I think the docs of `Runtime::spawn` is a bit confusing. However, this PR may not make it more clear.

While we can add more details to make the docs more clear, it makes the docs too verbose to understand quickly. Thus, I prefer to keep the existing docs.

I will close this PR, and welcome to open another PR if you have a better idea to improve it.

While we don't accept this change, thanks for taking the time to submit a PR.

tokio/src/runtime/runtime.rs
Its-Just-Nans: Yes but the current documation sentence is wrong

Its-Just-Nans:
> I will close this PR, and welcome to open another PR if you have a better idea to improve it.


Thanks for your feedback, I understand that there was a lot of changes. I created #7803 to change the sentence I found problematic

tokio/src/runtime/runtime.rs
ADD-SP: Do you have a better idea to make the docs more clear?

Its-Just-Nans: I changed the sentence on https://github.com/tokio-rs/tokio/pull/7803 

Tell me if you think it's a good change! (or not)

PR #7803, opened 2025-12-29
rt: clarify the documentation of `Runtime::spawn`

Its-Just-Nans:
## Motivation

Following https://github.com/tokio-rs/tokio/pull/7802#pullrequestreview-3614952466

I think clarifiying the doc of `spawn()` is better

[Rendered](https://deploy-preview-7803--tokio-rs.netlify.app/tokio/)

### DIFF
--- a/tokio/src/runtime/runtime.rs
+++ b/tokio/src/runtime/runtime.rs
@@ -206,9 +206,9 @@ impl Runtime {
     /// thread pool. The thread pool is then responsible for polling the future
     /// until it completes.
     ///
-    /// Executes the given future. If the current [Runtime] is
-    /// [running][running-runtime],it will start immediately in the background,
-    /// even if you don't await the returned [`JoinHandle`].
+    /// The provided future will start running in the background immediately
+    /// when `spawn` is called, even if you don't await the returned
+    /// `JoinHandle` (assuming that the runtime [is running][running-runtime]).
     ///
     /// See [module level][mod] documentation for more details.
     ///


tokio/src/runtime/runtime.rs
```diff
@@ -200,15 +200,14 @@ impl Runtime {
         &self.handle
     }
 
-    /// Spawns a future onto the Tokio runtime.
+    /// Spawns a future onto the Tokio runtime and return a [`JoinHandle`].
```
ADD-SP:
```suggestion
    /// Spawns a future onto the Tokio runtime.
```

Downstream user can get this by looking at the function signature.

tokio/src/runtime/runtime.rs
```diff
@@ -200,15 +200,14 @@ impl Runtime {
...
     ///
     /// This spawns the given future onto the runtime's executor, usually a
     /// thread pool. The thread pool is then responsible for polling the future
     /// until it completes.
     ///
-    /// The provided future will start running in the background immediately
-    /// when `spawn` is called, even if you don't await the returned
-    /// `JoinHandle`.
+    /// When this function is called, if the thread pool is active, the provided future
+    /// will start running in the background immediately.
```
ADD-SP: It is recommended to add two sentences explaining the behavior of `Runtime::spawn`: one for current thread runtime and another for multi-threaded runtime.

Please keep it concise—verbose documentation makes it harder for downstream users to understand.

tokio/src/runtime/runtime.rs
Its-Just-Nans: Yes, it was for consistency
https://github.com/tokio-rs/tokio/blob/0a3e3862697d902cffd729506cb28f6007c204d1/tokio/src/task/spawn.rs#L8-L9

tokio/src/runtime/runtime.rs
Its-Just-Nans: I've updated, tell me what I need to change

Its-Just-Nans:
> By the way, there are `Handle::spawn` and `tokio::spawn`, could you also update their docs?

I've updated, tell me what I need to change

tokio/src/runtime/handle.rs
```diff
@@ -164,9 +164,8 @@ impl Handle {
     /// thread pool. The thread pool is then responsible for polling the future
     /// until it completes.
     ///
-    /// The provided future will start running in the background immediately
-    /// when `spawn` is called, even if you don't await the returned
-    /// `JoinHandle`.
+    /// Executes the given future. Depending on the current [`crate::runtime::Runtime`],
+    /// it may start immediately in the background or require blocking on the [`JoinHandle`].
```
ADD-SP: > or require blocking on the [`JoinHandle`]

For the current thread runtime, `join_handle.await` doesn't execute the task if there is no thread is executing the `rt.block_on`. 

Here is an example.

```rust
use std::{thread::sleep, time::Duration};
use tokio::runtime;

fn main() {
    use tokio::task;

    let rt = runtime::Builder::new_current_thread().build().unwrap();

    println!("RT");
    let join_handle: task::JoinHandle<_> = rt.spawn(async {
        println!("JOB");
        // some work here
    });

    println!("SLEEP");
    sleep(Duration::from_secs(5));
    println!("SLEEP END");

    rt.handle().block_on(join_handle).unwrap();
    println!("END");
}
```

[Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e15d5f11d0525eb78152a6a247d7cace)

tokio/src/runtime/runtime.rs
```diff
@@ -206,9 +206,8 @@ impl Runtime {
     /// thread pool. The thread pool is then responsible for polling the future
     /// until it completes.
     ///
-    /// The provided future will start running in the background immediately
-    /// when `spawn` is called, even if you don't await the returned
-    /// `JoinHandle`.
+    /// Executes the given future. Depending on the current [`Runtime`],
```
ADD-SP:
```suggestion
    /// Executes the given future. Depending on the current [`Runtime`] configuration,
```

ADD-SP (commented):
We don't need fully qualified syntax in the docs.

tokio/src/runtime/handle.rs
Its-Just-Nans: Yes by `require blocking on the [`JoinHandle`]` it means, using `block_on(join_handle).unwrap();`

Its depends on the runtime

Its-Just-Nans: I'm not sure to understand the changes you want @ADD-SP  😅

tokio/src/task/spawn.rs
```diff
@@ -6,16 +6,17 @@ use std::future::Future;
 
 cfg_rt! {
     /// Spawns a new asynchronous task, returning a
-    /// [`JoinHandle`](JoinHandle) for it.
+    /// [`JoinHandle`](crate::task::JoinHandle) for it.
```
martin-g:
```suggestion
    /// [JoinHandle] for it.
```

tokio/src/task/spawn.rs
```diff
@@ -6,16 +6,17 @@ use std::future::Future;
...
     ///
     /// The provided future will start running in the background immediately
     /// when `spawn` is called, even if you don't await the returned
-    /// `JoinHandle`.
+    /// [`JoinHandle`](crate::task::JoinHandle).
```
martin-g:
```suggestion
    /// [JoinHandle].
```

tokio/src/task/spawn.rs
```diff
@@ -6,16 +6,17 @@ use std::future::Future;
...
     ///
     /// Spawning a task enables the task to execute concurrently to other tasks. The
     /// spawned task may execute on the current thread, or it may be sent to a
     /// different thread to be executed. The specifics depend on the current
-    /// [`Runtime`](crate::runtime::Runtime) configuration.
+    /// [`Runtime`](crate::runtime::Runtime) configuration - it may start immediately in
+    /// the background or require blocking on the [`JoinHandle`](crate::task::JoinHandle).
```
martin-g:
```suggestion
    /// the background or require blocking on the [JoinHandle].
```

tokio/src/runtime/runtime.rs
```diff
@@ -206,9 +206,8 @@ impl Runtime {
     /// thread pool. The thread pool is then responsible for polling the future
     /// until it completes.
     ///
-    /// The provided future will start running in the background immediately
-    /// when `spawn` is called, even if you don't await the returned
-    /// `JoinHandle`.
+    /// Executes the given future. Depending on the current [`Runtime`] configuration,
```
martin-g:
```suggestion
    /// Executes the given future. Depending on the current [Runtime] configuration,
```

tokio/src/runtime/runtime.rs
```diff
@@ -206,9 +206,8 @@ impl Runtime {
...
+    /// it may start immediately in the background or require blocking on the [`JoinHandle`].
```
martin-g:
```suggestion
    /// it may start immediately in the background or require blocking on the [JoinHandle].
```

tokio/src/runtime/handle.rs
```diff
@@ -164,9 +164,8 @@ impl Handle {
     /// thread pool. The thread pool is then responsible for polling the future
     /// until it completes.
     ///
-    /// The provided future will start running in the background immediately
-    /// when `spawn` is called, even if you don't await the returned
-    /// `JoinHandle`.
+    /// Executes the given future. Depending on the current [`Runtime`](crate::runtime::Runtime),
```
martin-g:
```suggestion
    /// Executes the given future. Depending on the current [Runtime],
```

tokio/src/runtime/handle.rs
```diff
@@ -164,9 +164,8 @@ impl Handle {
...
+    /// it may start immediately in the background or require blocking on the [`JoinHandle`].
```
martin-g:
```suggestion
    /// it may start immediately in the background or require blocking on the [JoinHandle].
```

tokio/src/runtime/handle.rs
martin-g: IMO it will be clearer if the documentation explain when it will be executed immediately and when it needs blocking. I.e. be more specific instead of letting the reader guess and experiment.

martin-g (commented):
I've updated the infra-doc links according to https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html
> Backticks around the link will be stripped, so [`Option`] will correctly link to Option.

tokio/src/runtime/handle.rs
ADD-SP: > IMO it will be clearer if the documentation explain when it will be executed immediately and when it needs blocking. I.e. be more specific instead of letting the reader guess and experiment.

However, this would make the documentation increasingly lengthy, and no one enjoys reading academic papers within `docs.rs`. That's why [I recommend keeping the original documentation as it stands](https://github.com/tokio-rs/tokio/pull/7802#pullrequestreview-3614952466)—the confusion it has caused so far remains within acceptable range.

tokio/src/runtime/handle.rs
Its-Just-Nans: I don't think it's working for this one

image

tokio/src/runtime/handle.rs
Its-Just-Nans: I don’t really see the Tokio docs as an academic paper. I just prefer documentation to be correct for beginner like me.

Tell me what if I should update this MR with more details or to close it

Thanks

tokio/src/task/spawn.rs
Its-Just-Nans: But now the spellcheck is annoying :/

tokio/src/runtime/handle.rs
martin-g: This is because `Runtime` is not imported in this file:
```rust
use crate::runtime;
use crate::runtime::{context, scheduler, RuntimeFlavor, RuntimeMetrics};
```
You need to either add it or use `[Runtime](runtime::Runtime)`

tokio/src/task/spawn.rs
martin-g: Rustdoc vs. spellcheck - the fight of the tools :-)
Maybe this is the reason why backticks are in use here! 
You could try with ``[`JoinHandle`]``. If spellcheck still complains then you will need to update https://github.com/tokio-rs/tokio/blob/4bc2a15d28554426a500ce73be969ecd9ed6dd89/spellcheck.dic:
1) increment the number on the first line (e.g. 311 -> 312)
2) Insert `JoinHandle` (between `iteratively` and `Kotlin's`)

tokio/src/runtime/handle.rs
ADD-SP: > Yes by require blocking on the [JoinHandle] it means, using block_on(join_handle).unwrap();

For the current thread runtime, the `Handle::block_on` is not identical to `Runtime::block_on`. The `Runtime::block_on` polls the task queue, but the `Handle::block_on` does not.

ADD-SP: > IMO it will be clearer if the documentation explain when it will be executed immediately and when it needs blocking. I.e. be more specific instead of letting the reader guess and experiment.

@Its-Just-Nans Could you explain the correct behavior in docs, and then we can evaluate weather it is too long.

tokio/src/runtime/runtime.rs
```diff
@@ -206,9 +206,8 @@ impl Runtime {
     /// thread pool. The thread pool is then responsible for polling the future
     /// until it completes.
     ///
-    /// The provided future will start running in the background immediately
-    /// when `spawn` is called, even if you don't await the returned
-    /// `JoinHandle`.
+    /// Executes the given future. Depending on the current [Runtime] configuration,
+    /// it may start immediately in the background or require blocking on the [JoinHandle].
```
Darksonn: This is wrong!

It never requires blocking on the `JoinHandle`.

As long as `Runtime::block_on()` is called *with any future*, spawned tasks will execute. The `JoinHandle` does not matter. You could even call `rt.block_on(pending())` in a background thread.

Darksonn:
I think it may be less confusing to introduce a concept along of whether or not the runtime is running.

> The provided future will start running in the background immediately when `spawn` is called, even if you don't await the returned `JoinHandle`. (As long as the runtime is running.)

And then the word "running" can be a link to a new section in [the docs for `tokio::runtime`](https://docs.rs/tokio/latest/tokio/runtime/index.html), which clarifies that the multi-threaded runtime is always running, that the current-thread runtime is only running when there is at least on call to `Runtime::block_on()`, and that a `LocalSet` is running when it's being driven by one of more mechanisms.

tokio/src/runtime/handle.rs
Its-Just-Nans: Thanks - fixed, I'm going with the markdown link

tokio/src/runtime/runtime.rs
Its-Just-Nans: Yes
thanks

tokio/src/runtime/handle.rs
Its-Just-Nans: I've updated my branch

Thanks

Its-Just-Nans:
> And then the word "running" can be a link to a new section in [the docs for `tokio::runtime`](https://docs.rs/tokio/latest/tokio/runtime/index.html), which clarifies that the multi-threaded runtime is always running, that the current-thread runtime is only running when there is at least on call to `Runtime::block_on()`, and that a `LocalSet` is running when it's being driven by one of more mechanisms.

I think that's a nice idea! I've updated my branch with this change(s)


ping @ADD-SP - i've updated the docs

tokio/src/runtime/handle.rs
```diff
@@ -164,13 +164,13 @@ impl Handle {
     /// thread pool. The thread pool is then responsible for polling the future
     /// until it completes.
     ///
-    /// The provided future will start running in the background immediately
-    /// when `spawn` is called, even if you don't await the returned
-    /// `JoinHandle`.
+    /// Executes the given future. Depending if the current [`Runtime`](runtime::Runtime)
+    /// is [running][running-runtime] it may start immediately in the background.
```
Darksonn: I don't like this wording. This paragraph was added to make it clear that awaiting the `JoinHandle` is *not* required, but you removed that.

tokio/src/runtime/mod.rs
```diff
@@ -205,6 +205,17 @@
 //! This is done with [`Builder::enable_io`] and [`Builder::enable_time`]. As a
 //! shorthand, [`Builder::enable_all`] enables both resource drivers.
 //!
+//! #### Running runtime
```
Darksonn: You placed this under "Runtime Configurations", but it's not a way you can configure the runtime. Please make it into its own section.

Also, a different section name such as "Driving the runtime" would make more sense to me.

tokio/src/runtime/mod.rs
```diff
@@ -205,6 +205,17 @@
...
+//!
+//! Depending on how the runtime is configured, Tokio may already be running.
+//! If it is not, tasks will not execute unless you explicitly drive the runtime,
+//! for example by calling [`Runtime::block_on`].
+//!
+//! A multi-threaded runtime is always running. The current-thread runtime is only
+//! running when there is at least on call to [`Runtime::block_on`]. A
+//! [`LocalSet`](crate::task::LocalSet) is running when it's being driven by one
+//! of more mechanisms.
```
Darksonn: I think this wording is confusing. Perhaps something along the lines of

> A Tokio runtime can only execute tasks if the runtime is running. Normally this is not an issue as the default configuration of a runtime is always running, but alternate configurations such as the current-thread runtime require that `Runtime::block_on` is called.
>
> * A multi-threaded runtime is always running because it spawns its own worker threads.
> * A current-thread runtime does not spawn any worker threads, so it can only execute tasks when you provide a thread by calling [`Runtime::block_on`].
> * A [`LocalSet`] only executes local tasks spawned on it when the `LocalSet` is `.await`ed or otherwise driven using one of its methods for this purpose.
> 
> Please be aware that [`Handle::block_on`] does *not* drive the runtime. There must be at least one call to [`Runtime::block_on`] when using the current-thread runtime. [`Handle::block_on`] is not enough.

tokio/src/runtime/mod.rs
Its-Just-Nans: Thanks I edited the docs

Its-Just-Nans:
Hi

Should I change something ?

tokio/src/runtime/handle.rs
```diff
@@ -164,13 +164,14 @@ impl Handle {
     /// thread pool. The thread pool is then responsible for polling the future
     /// until it completes.
     ///
-    /// The provided future will start running in the background immediately
-    /// when `spawn` is called, even if you don't await the returned
-    /// `JoinHandle`.
+    /// Executes the given future. Depending if the current [`Runtime`](runtime::Runtime)
+    /// is [running][running-runtime] it may start immediately in the background, even if
+    /// you don't await the returned [`JoinHandle`].
```
Darksonn: This wording is still confusing. All you say is that it *may* start running, but people can't rely on that. They need it to start running, and they need to know when.
```suggestion
    /// The provided future will start running in the background immediately
    /// when `spawn` is called, even if you don't await the returned
    /// `JoinHandle`. (Assuming that the runtime [is running][running-runtime]).
```

tokio/src/task/spawn.rs
```diff
@@ -6,16 +6,20 @@ use std::future::Future;
...
     ///
     /// Spawning a task enables the task to execute concurrently to other tasks. The
     /// spawned task may execute on the current thread, or it may be sent to a
     /// different thread to be executed. The specifics depend on the current
-    /// [`Runtime`](crate::runtime::Runtime) configuration.
+    /// [`Runtime`](crate::runtime::Runtime) configuration. In a
+    /// [running runtime][running-runtime], it may start immediately in the background,
+    /// even if you don't await the returned [`JoinHandle`].
+    /// On a blocked runtime, the user must drive the runtime forward (for example,
+    /// by calling [`Runtime::block_on`](crate::runtime::Runtime::block_on)).
```
Darksonn: This wording is also confusing. You say "in a running runtime, it may start", which implies that even if it's running, it *might not* start. Please avoid the word "may".

Its-Just-Nans: Hi

Thanks for the review, I changed the wording

tokio/src/runtime/runtime.rs
```diff
@@ -206,13 +206,14 @@ impl Runtime {
     /// thread pool. The thread pool is then responsible for polling the future
     /// until it completes.
     ///
-    /// The provided future will start running in the background immediately
-    /// when `spawn` is called, even if you don't await the returned
-    /// `JoinHandle`.
+    /// Executes the given future. Depending if the current [Runtime]
+    /// is [running][running-runtime] it may start immediately in the background,
+    /// even if you don't await the returned [`JoinHandle`].
```
Darksonn: The comments from other files also apply to this file.

tokio/src/task/spawn.rs
```diff
@@ -6,16 +6,20 @@ use std::future::Future;
...
-    /// `JoinHandle`.
+    /// [`JoinHandle`].
     ///
     /// Spawning a task enables the task to execute concurrently to other tasks. The
     /// spawned task may execute on the current thread, or it may be sent to a
     /// different thread to be executed. The specifics depend on the current
-    /// [`Runtime`](crate::runtime::Runtime) configuration.
+    /// [`Runtime`](crate::runtime::Runtime) configuration. In a
+    /// [running runtime][running-runtime], the task will start immediately in the
+    /// background, even if you don't await the returned [`JoinHandle`].
```
Darksonn: The previous paragraph already says the part about await. Perhaps add the same parenthetical here as in `handle.rs`?

Its-Just-Nans: Okay I removed the await part

tokio/src/runtime/runtime.rs
Darksonn: What I meant is that, like the other files also did, this focuses too much on the running runtime. I like the  wording you have in `tokio/src/runtime/handle.rs` better.

Its-Just-Nans: Okay thanks I updated to the same wording

Darksonn (approved):
Thank you.

PR #7806, opened 2025-12-29
task: Better typed RawTask::try_read_output

stepancheg:
## Motivation

Types help reading code.

## Solution

Add more precise type to `RawTask::try_read_output`.

### DIFF
--- a/tokio/src/runtime/task/join.rs
+++ b/tokio/src/runtime/task/join.rs
@@ -339,8 +339,7 @@ impl<T> Future for JoinHandle<T> {
         //
         // The type of `T` must match the task's output type.
         unsafe {
-            self.raw
-                .try_read_output(&mut ret as *mut _ as *mut (), cx.waker());
+            self.raw.try_read_output(&mut ret, cx.waker());
         }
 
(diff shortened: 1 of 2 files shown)


tokio/src/runtime/task/join.rs
```diff
@@ -339,8 +339,7 @@ impl<T> Future for JoinHandle<T> {
         //
         // The type of `T` must match the task's output type.
         unsafe {
-            self.raw
-                .try_read_output(&mut ret as *mut _ as *mut (), cx.waker());
+            self.raw.try_read_output(&mut ret as *mut _, cx.waker());
```
ADD-SP:
```suggestion
            self.raw.try_read_output(std::ptr::addr_of_mut(ret) as _, cx.waker());
```

If I understand correctly, you want to get rid of the double casting (`&mut ret as *mut _ as *mut ()`. If so, is this approach looks good to you for less changes?

stepancheg: Actually, first cast is no longer needed, because &mut is automatically casted to *mut.

Updated the PR.

PR #7807, opened 2025-12-29
Make JoinHandle uninhabited when !rt

stepancheg:
## Motivation

Make more clear that `JoinHandle` is uninhabited when `rt` is not enabled. Mostly to help reading code, but also may make code more efficient.

## Solution

Add empty enum field to `JoinHandle`.

### DIFF
--- a/tokio/src/blocking.rs
+++ b/tokio/src/blocking.rs
@@ -34,8 +34,11 @@ cfg_not_rt! {
         }
     }
 
+    enum Never {}
+
     pub(crate) struct JoinHandle<R> {
         _p: std::marker::PhantomData<R>,
+        never: Never,
     }
 
     unsafe impl<T: Send> Send for JoinHandle<T> {}
@@ -45,7 +48,7 @@ cfg_not_rt! {
         type Output = Result<R, std::io::Error>;
 
         fn poll(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Self::Output> {
-            unreachable!()
+            match self.never {}
         }
     }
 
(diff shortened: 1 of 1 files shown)


tokio/src/blocking.rs
```diff
@@ -45,16 +48,16 @@ cfg_not_rt! {
         type Output = Result<R, std::io::Error>;
 
         fn poll(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Self::Output> {
-            unreachable!()
```
ADD-SP, line 48: Hmm, I think the `unreachable` here is still helpful when this code path is triggered by bugs.

ADD-SP (commented):
Would it be more simple to add comments to `JoinHandle`?

tokio/src/blocking.rs
ADD-SP: I think we can add a simple comment for this `unreachable` instead of declaring a new type.

tokio/src/blocking.rs
```diff
@@ -45,16 +48,16 @@ cfg_not_rt! {
...
     }
 
     impl<T> fmt::Debug for JoinHandle<T>
     where
         T: fmt::Debug,
     {
-        fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
-            fmt.debug_struct("JoinHandle").finish()
+        fn fmt(&self, _fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+            match self.never {}
```
ADD-SP, line 60: Since I think we just need a few comments, we don't need a new `Never` type. Thus I prefer to revert this change.

stepancheg:
> Would it be more simple to add comments to JoinHandle?

IMO, code is faster to read than comment. Not a big deal anyway, I'll close the PR.

PR #7815, opened 2026-01-01
fs: check for opcode support before starting operation

tahmid-23:
## Motivation

`io_uring` exposes additional opcodes with successive kernel versions. Ideally, we can tell which opcodes are supported, so we can know whether to use the `io_uring` version or the threadpool version of an asynchronous function.

This functionality is exposed through `io_uring`'s probing feature, which lists the supported opcodes. This is introduced in kernel 5.6. We fallback to attempting any opcode if probing support is not available.

All of the currently used opcodes (openat, read, write) were introduced before probing support in `io_uring`, so this is unnecessary for the current functionality. However, future additions can benefit from this check.

(Technically, #7799 and #7800 would benefit from this, although running on kernel [5.6, 5.10) is likely rare. But, other opcodes like `mkdirat` are not available on 5.10, which is an LTS.)

I bumped the `io-uring` crate to 0.7.11 to expose `get_opcode`, since it is tokio_unstable.

### DIFF
--- a/tokio/src/runtime/io/driver/uring.rs
+++ b/tokio/src/runtime/io/driver/uring.rs
@@ -1,4 +1,4 @@
-use io_uring::{squeue::Entry, IoUring};
+use io_uring::{squeue::Entry, IoUring, Probe};
 use mio::unix::SourceFd;
 use slab::Slab;
 
@@ -38,6 +38,7 @@ impl State {
 
 pub(crate) struct UringContext {
     pub(crate) uring: Option<io_uring::IoUring>,
+    pub(crate) probe: io_uring::Probe,
     pub(crate) ops: slab::Slab<Lifecycle>,
 }
 
@@ -46,6 +47,7 @@ impl UringContext {
         Self {
             ops: Slab::new(),
             uring: None,
+            probe: Probe::new(),
         }
     }
 
@@ -55,6 +57,10 @@ impl UringContext {
 
     pub(crate) fn ring_mut(&mut self) -> &mut io_uring::IoUring {
         self.uring.as_mut().expect("io_uring not initialized")
+    }
+
+    pub(crate) fn is_opcode_supported(&self, opcode: u8) -> bool {
+        self.probe.is_supported(opcode)
     }
 
     /// Perform `io_uring_setup` system call, and Returns true if this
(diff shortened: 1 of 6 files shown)


tokio/src/runtime/io/driver/uring.rs
```diff
@@ -57,6 +59,17 @@ impl UringContext {
         self.uring.as_mut().expect("io_uring not initialized")
     }
 
+    pub(crate) fn is_opcode_supported(&self, opcode: u8) -> bool {
+        match self.probe.as_ref() {
+            Some(probe) => probe.is_supported(opcode),
+            _ => {
+                // If the kernel does not support IORING_REGISTER_PROBE,
+                // try all opcodes anyway.
+                true
```
Darksonn: It's not clear that this is the correct approach. Perhaps we should simply treat such cases as not supporting io_uring, or hardcode the list opcodes that are present on 5.10, which I understand is the only supported Linux kernel version without IOURING_REGISTER_PROBE?

tokio/src/runtime/io/driver/uring.rs
```diff
@@ -182,12 +208,19 @@ impl Handle {
...
     /// Check if the io_uring context is initialized. If not, it will try to initialize it.
-    pub(crate) fn check_and_init(&self) -> io::Result<bool> {
+    /// Then, check if the provided opcode is supported.
+    ///
+    /// If both the context initialization succeeds and the opcode is supported,
+    /// this returns Ok(true).
+    /// If either io_uring is unsupported or the opcode is unsupported,
+    /// this returns Ok(false).
+    /// An error is returned if an io_uring syscall returns an unexpected error value.
+    pub(crate) fn check_and_init(&self, opcode: u8) -> io::Result<bool> {
```
Darksonn, line 209: I know this is private and not rendered, but please still use markdown syntax for doc comments.

tokio/src/runtime/io/driver/uring.rs
tahmid-23: If probing support isn't present, then none of the currently used opcodes will work (and I assume none of the future opcodes either). So let's go with the first option (5.6+).

vrtgs:
I feel like having to lock a lock when running check_and_init, makes the whole using atomics pointless in my humble opinion

Darksonn:
That's a good point. Does `is_opcode_supported` take `&mut self`?

tahmid-23:
no, just `&self` / it needs to have been initialized.
It's not immediately obvious to me how to avoid a lock though. It needs an `&mut Probe` to store into, which is currently in the `UringContext`.
we could use something like an `ArcSwap<Probe>` in `Handle`, but that would require the `arc-swap` crate.

Darksonn:
I guess we can probe the opcodes that Tokio cares about and store an `AtomicU32` bitmap of the results.

Either way we should open an issue to remove the lock.

PR #7816, opened 2026-01-01
signal: functional storage

vilgotf:


vilgotf:
Is the windows/hyper test spurious?

tokio/src/signal/unix.rs
```diff
@@ -37,15 +39,14 @@ impl Init for OsStorage {
 }
 
 impl Storage for OsStorage {
+    type Iter<'a> = iter::Map<slice::Iter<'a, SignalInfo>, fn(&'a SignalInfo) -> &'a EventInfo>;
```
Darksonn: Now we're introducing dynamic function calls by having to use a function pointer. I'm not so sure this is a great idea.

Darksonn (commented):
I think the current `for_each` is fine.

tokio/src/signal/unix.rs
vilgotf: I might be wrong, but I don't see why it would be dynamic? The `FnMut` of `Map` is not a closure; `fn(&SignalInfo) -> &EventInfo` is its concrete type

Darksonn: Normally `|si| &si.event_info` is going to have an unnameable type (a function item) that implements `FnMut`, but by adding this type annotation you're causing rustc to insert a conversion from function item to function pointer.

vilgotf: So are these two different as well then?

```rust
fn map(item: &SignalInfo) -> &EventInfo {
    &item.event_info
}

self.as_ref().iter().map(map)
self.as_ref().iter().map(|si| &si.event_info)
```

I'm otherwise okay with closing this PR until `type Iter<'a> = iter::Map<slice::Iter<'a, SignalInfo>, impl FnMut(&SignalInfo) -> &EventInfo>;` is stabilized

Darksonn: I think the only way to avoid dynamic dispatch right now is the below, but I think that's too complex.
```rs
pub struct EventInfoIter<'a> {
    inner: std::slice::Iter<'a, SignalInfo>,
}

impl<'a> Iterator for EventInfoIter<'a> {
    type Item = &'a EventInfo;

    #[inline]
    fn next(&mut self) -> Option<Self::Item> {
        self.inner.next().map(|si| &si.event_info)
    }
}

impl Storage for OsStorage {
    // Replace the complex Map type with our clean custom struct
    type Iter<'a> = EventInfoIter<'a>;

    fn event_info(&self, id: EventId) -> Option<&EventInfo> {
        self.get(id).map(|si| &si.event_info)
    }

    fn iter(&self) -> Self::Iter<'_> {
        EventInfoIter {
            inner: self.as_ref().iter(),
        }
    }
}
```

Darksonn:
I'm going to close as per discussion, but thanks for taking the time to submit a PR!

tokio/src/signal/unix.rs
vilgotf: Wait, we could just use RPITIT (once the msrv is 1.75):

```rust
pub(crate) trait Storage {
    fn event_info(&self, id: EventId) -> Option<&EventInfo>;

    fn iter(&self) -> impl Iterator<Item = &EventInfo>;
}
```

Darksonn: Yes that would have worked.

PR #7818, opened 2026-01-01
signal: specialize unix OsStorage

vilgotf:
Noticed that this was unnecessarily boxed, even though that's only required for linux and illumos.

### DIFF
--- a/tokio/src/signal/unix.rs
+++ b/tokio/src/signal/unix.rs
@@ -18,21 +18,33 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Once;
 use std::task::{Context, Poll};
 
-pub(crate) type OsStorage = Box<[SignalInfo]>;
+// There are reliable signals ranging from 1 to 33 available on every Unix platform.
+#[cfg(not(any(target_os = "linux", target_os = "illumos")))]
+pub(crate) type OsStorage = [SignalInfo; 34];
 
+#[cfg(not(any(target_os = "linux", target_os = "illumos")))]
 impl Init for OsStorage {
     fn init() -> Self {
-        // There are reliable signals ranging from 1 to 33 available on every Unix platform.
-        #[cfg(not(any(target_os = "linux", target_os = "illumos")))]
-        let possible = 0..=33;
+        use std::array;
 
-        // On Linux and illumos, there are additional real-time signals
-        // available. (This is also likely true on Solaris, but this should be
-        // verified before being enabled.)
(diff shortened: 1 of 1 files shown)


tokio/src/signal/unix.rs
```diff
@@ -18,21 +18,33 @@ use std::sync::atomic::{AtomicBool, Ordering};
...
+#[cfg(not(any(target_os = "linux", target_os = "illumos")))]
+pub(crate) type OsStorage = [SignalInfo; 34];
 
+#[cfg(not(any(target_os = "linux", target_os = "illumos")))]
 impl Init for OsStorage {
     fn init() -> Self {
-        // There are reliable signals ranging from 1 to 33 available on every Unix platform.
-        #[cfg(not(any(target_os = "linux", target_os = "illumos")))]
-        let possible = 0..=33;
+        use std::array;
```
Darksonn: This import can go on the top, or just do `std::array::from_fn`.

tokio/src/signal/unix.rs
```diff
@@ -18,21 +18,33 @@ use std::sync::atomic::{AtomicBool, Ordering};
...
+// On Linux and illumos, there are additional real-time signals available.
+// (This is also likely true on Solaris, but this should be verified before
+// being enabled.)
+#[cfg(any(target_os = "linux", target_os = "illumos"))]
+pub(crate) type OsStorage = Box<[SignalInfo]>;
+
+#[cfg(any(target_os = "linux", target_os = "illumos"))]
+impl Init for OsStorage {
+    fn init() -> Self {
+        use std::iter;
```
Darksonn: Same here.

tokio/src/signal/unix.rs
```diff
@@ -18,21 +18,33 @@ use std::sync::atomic::{AtomicBool, Ordering};
...
-        // verified before being enabled.)
-        #[cfg(any(target_os = "linux", target_os = "illumos"))]
-        let possible = 0..=libc::SIGRTMAX();
+// On Linux and illumos, there are additional real-time signals available.
+// (This is also likely true on Solaris, but this should be verified before
+// being enabled.)
+#[cfg(any(target_os = "linux", target_os = "illumos"))]
+pub(crate) type OsStorage = Box<[SignalInfo]>;
+
+#[cfg(any(target_os = "linux", target_os = "illumos"))]
```
Darksonn: It may be clearer to move the `#[cfg()]` inside `fn init()` rather than having entirely separate blocks.

Darksonn (approved):
Thanks.

PR #7819, opened 2026-01-01
signal: optimize non-zero unix signal

vilgotf:
0 is not a valid signal, so we can offset signals by 1 in `OsStorage` and save some space.

### DIFF
--- a/tokio/src/signal/unix.rs
+++ b/tokio/src/signal/unix.rs
@@ -531,16 +531,30 @@ mod tests {
 
     #[test]
     fn signal_enable_error_on_invalid_input() {
-        signal_enable(SignalKind::from_raw(-1), &Handle::default()).unwrap_err();
+        let inputs = [-1, 0];
+
+        for input in inputs {
+            assert_eq!(
+                signal_enable(SignalKind::from_raw(input), &Handle::default())
+                    .unwrap_err()
+                    .kind(),
+                ErrorKind::Other,
+            );
+        }
     }
 
     #[test]
     fn signal_enable_error_on_forbidden_input() {
-        signal_enable(
-            SignalKind::from_raw(signal_hook_registry::FORBIDDEN[0]),
-            &Handle::default(),
-        )
-        .unwrap_err();
+        let inputs = signal_hook_registry::FORBIDDEN;
+
+        for &input in inputs {
+            assert_eq!(
+                signal_enable(SignalKind::from_raw(input), &Handle::default())
+                    .unwrap_err()
+                    .kind(),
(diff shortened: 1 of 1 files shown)


tokio/src/signal/unix.rs
```diff
@@ -282,7 +282,7 @@ fn signal_enable(signal: SignalKind, handle: &Handle) -> io::Result<()> {
     handle.check_inner()?;
 
     let globals = globals();
-    let siginfo = match globals.storage().get(signal as EventId) {
+    let siginfo = match globals.storage().get(signal as EventId - 1) {
```
Darksonn: That's too confusing. This code is too far away from the definition, and too easy to forget if this code is ever changed. Can you wrap `OsStorage` in a struct and add a `get` method with the `-1` inside that method internally?

tokio/src/signal/unix.rs
```diff
@@ -271,7 +281,7 @@ fn action(globals: &'static Globals, signal: libc::c_int) {
 /// returning any error along the way if that fails.
 fn signal_enable(signal: SignalKind, handle: &Handle) -> io::Result<()> {
     let signal = signal.0;
-    if signal < 0 || signal_hook_registry::FORBIDDEN.contains(&signal) {
+    if signal <= 0 || signal_hook_registry::FORBIDDEN.contains(&signal) {
```
Darksonn, line 282: Can you add a test for the `signal == 0` case?

tokio/src/signal/unix.rs
```diff
@@ -523,7 +531,11 @@ mod tests {
 
     #[test]
     fn signal_enable_error_on_invalid_input() {
-        signal_enable(SignalKind::from_raw(-1), &Handle::default()).unwrap_err();
+        let inputs = [-1, 0];
+
+        for input in inputs {
+            signal_enable(SignalKind::from_raw(input), &Handle::default()).unwrap_err();
```
Darksonn: Which `ErrorKind` is used in this case?

vilgotf: Both invalid and forbidden inputs translate to `ErrorKind::Other`:

```rust
if signal <= 0 || signal_hook_registry::FORBIDDEN.contains(&signal) {
    return Err(Error::new(
        ErrorKind::Other,
        format!("Refusing to register signal {signal}"),
    ));
}
```

The invalid and forbidden tests did not historically assert the error type, so I kept that behavior. Would you want it changed to assert the error type?

Darksonn: Yes, that'd be great.

PR #7828, opened 2026-01-03
task: improve the docs of `Builder::spawn_local`

mattiapitossi:
## Motivation

Fixes wrong docs on task Builder for `spawn_local` since if used under `current_thread` it panics.

### DIFF
--- a/tokio/src/task/builder.rs
+++ b/tokio/src/task/builder.rs
@@ -20,8 +20,7 @@ use std::{future::Future, io, mem};
 ///   the task
 ///
 /// There are three types of task that can be spawned from a Builder:
-/// - [`spawn_local`] for executing not [`Send`] futures using [`LocalSet`]
-///   or [`LocalRuntime`] (unstable)
+/// - [`spawn_local`] for executing not [`Send`] futures
 /// - [`spawn`] for executing [`Send`] futures on the runtime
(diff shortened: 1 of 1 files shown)


tokio/src/task/builder.rs
```diff
@@ -20,7 +20,8 @@ use std::{future::Future, io, mem};
 ///   the task
 ///
 /// There are three types of task that can be spawned from a Builder:
-/// - [`spawn_local`] for executing futures on the current thread
+/// - [`spawn_local`] for executing not [`Send`] futures using [`LocalSet`]
+///   or [`LocalRuntime`] (unstable)
```
ADD-SP: There is a details doc to cover this, is this newly added doc redundant?

<https://docs.rs/tokio/1.49.0/tokio/task/struct.Builder.html#method.spawn_local>

mattiapitossi: Maybe we could just mention that this is intended for !Send futures as with spawn for Send? Mentioning current_thread only can be misleading

ADD-SP: This makes sense to me.

mattiapitossi: Thanks, fixed!

ADD-SP (approved):
Thanks!

PR #7833, opened 2026-01-04
signal: remember the result of `SetConsoleCtrlHandler`

vilgotf:
The unix implementation remembers whether it failed or not, so the windows implementation should do so as well. This PR also replaces the `(Once, AtomicState)` pair with `OnceLock` and tries to remember the original errno value.

### DIFF
--- a/tokio/src/signal/unix.rs
+++ b/tokio/src/signal/unix.rs
@@ -14,8 +14,7 @@ use crate::sync::watch;
 
 use mio::net::UnixStream;
 use std::io::{self, Error, ErrorKind, Write};
-use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Once;
+use std::sync::OnceLock;
 use std::task::{Context, Poll};
 
 #[cfg(not(any(target_os = "linux", target_os = "illumos")))]
@@ -239,20 +238,10 @@ impl From<SignalKind> for std::os::raw::c_int {
     }
 }
 
+#[derive(Default)]
 pub(crate) struct SignalInfo {
     event_info: EventInfo,
-    init: Once,
-    initialized: AtomicBool,
-}
-
-impl Default for SignalInfo {
-    fn default() -> SignalInfo {
-        SignalInfo {
-            event_info: EventInfo::default(),
-            init: Once::new(),
-            initialized: AtomicBool::new(false),
-        }
-    }
+    init: OnceLock<Result<(), Option<i32>>>,
 }
 
 /// Our global signal handler for all signals registered by this module.
@@ -294,26 +283,20 @@ fn signal_enable(signal: SignalKind, handle: &Handle) -> io::Result<()> {
         Some(slot) => slot,
         None => return Err(io::Error::new(io::ErrorKind::Other, "signal too large")),
     };
-    let mut registered = Ok(());
-    siginfo.init.call_once(|| {
-        registered = unsafe {
-            signal_hook_registry::register(signal, move || action(globals, signal)).map(|_| ())
-        };
-        if registered.is_ok() {
-            siginfo.initialized.store(true, Ordering::Relaxed);
-        }
-    });
-    registered?;
-    // If the call_once failed, it won't be retried on the next attempt to register the signal. In
-    // such case it is not run, registered is still `Ok(())`, initialized is still `false`.
-    if siginfo.initialized.load(Ordering::Relaxed) {
-        Ok(())
-    } else {
-        Err(Error::new(
-            ErrorKind::Other,
-            "Failed to register signal handler",
-        ))
-    }
+
+    siginfo
+        .init
+        .get_or_init(|| {
+            unsafe { signal_hook_registry::register(signal, move || action(globals, signal)) }
+                .map(|_| ())
+                .map_err(|e| e.raw_os_error())
+        })
+        .map_err(|e| {
+            e.map_or_else(
+                || Error::new(ErrorKind::Other, "registering signal handler failed"),
+                Error::from_raw_os_error,
+            )
+        })
 }
 
 /// An listener for receiving a particular type of OS signal.
--- a/tokio/src/signal/windows/sys.rs
+++ b/tokio/src/signal/windows/sys.rs
@@ -1,5 +1,5 @@
 use std::io;
-use std::sync::Once;
+use std::sync::OnceLock;
 
 use crate::signal::registry::{globals, EventId, EventInfo, Storage};
 use crate::signal::RxFuture;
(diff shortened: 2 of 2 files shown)


Darksonn:
I'm confused. You said that unix and windows do different things, but you're changing both.

vilgotf:
The unix change is not meant to modify semantics but to refactor the implementation to mirror (the new) windows implementation. The windows implementation could instead mirror the current unix implementation (adding a `static INITIALIZED: AtomicBool` next to `static INIT: Once`), but `OnceLock` is the modern alternative, so I thought we could modernize both implementations while we're at it.

tokio/src/signal/unix.rs
```diff
@@ -294,26 +283,17 @@ fn signal_enable(signal: SignalKind, handle: &Handle) -> io::Result<()> {
...
-        ))
-    }
+
+    siginfo
+        .init
+        .get_or_init(|| unsafe {
+            match signal_hook_registry::register_unchecked(signal, move |_| action(globals, signal))
+            {
+                Ok(_) => Ok(()),
+                Err(e) => Err(e.raw_os_error().unwrap()),
```
ipetkov: I know this _shouldn't_ fail, but I don't feel good about having an `unwrap()` here. It seems like `signal(2)` can only fail with `EINVAL` so this is probably a safe thing to fall back to just in case

tokio/src/signal/windows/sys.rs
```diff
@@ -85,22 +85,17 @@ impl Storage for OsStorage {
...
 
-    let mut init = None;
-
-    INIT.call_once(|| unsafe {
+    INIT.get_or_init(|| unsafe {
         let rc = console::SetConsoleCtrlHandler(Some(handler), 1);
-        let ret = if rc == 0 {
-            Err(io::Error::last_os_error())
+        if rc == 0 {
+            Err(io::Error::last_os_error().raw_os_error().unwrap())
```
ipetkov: Same comment as on the unix implementation, I'd rather avoid the `.unwrap` here if possible

tokio/src/signal/unix.rs
vilgotf: The unwrap is required to store the errno value, but I changed it to just store an error flag. The error flag then maps to `ErrorKind::Other`.

Darksonn: We could always do `e.raw_os_error().unwrap_or(EINVAL)` to fall back to `EINVAL` when we can't convert it.

tokio/src/signal/unix.rs
```diff
@@ -239,20 +238,10 @@ impl From<SignalKind> for std::os::raw::c_int {
...
-
-impl Default for SignalInfo {
-    fn default() -> SignalInfo {
-        SignalInfo {
-            event_info: EventInfo::default(),
-            init: Once::new(),
-            initialized: AtomicBool::new(false),
-        }
-    }
+    init: OnceLock<Result<(), ()>>,
```
martin-g:
```suggestion
    init: OnceLock<Result<(), io::Error>>,
```
This would make the pattern matching + map_err() below unnecessary, no ?

tokio/src/signal/unix.rs
```diff
@@ -294,26 +283,25 @@ fn signal_enable(signal: SignalKind, handle: &Handle) -> io::Result<()> {
...
-            "Failed to register signal handler",
-        ))
-    }
+
+    let mut error = None;
+    siginfo
+        .init
+        .get_or_init(|| {
+            let result = unsafe {
+                signal_hook_registry::register_unchecked(signal, move |_| action(globals, signal))
```
martin-g: Why `_unchecked()` ?

tokio/src/signal/unix.rs
vilgotf: We can't store `io::Error` directly because `OnceLock<T>` only hands out `&T` and `io::Error` does not implement `Clone`.

tokio/src/signal/unix.rs
vilgotf: I looked through `signal_hook_registry` and found that the only difference between `register` and `register_unchecked` is an assertion in `register` for `!signal_hook_registry::FORBIDDEN.contains(&signal)` (which we already tested).

tokio/src/signal/unix.rs
vilgotf: I actually prefer differing between initial and repeat errors, the errno value is not semantically accurate for repeat errors since they do not try to register the handle.

tokio/src/signal/unix.rs
Darksonn: I don't think we should use `register_unchecked` here. The code is fine with the double check.

vilgotf: The function is badly named (imo), it incurs no additional safety requirements. But I can continue to use `register`, it's not really relevant to this PR either way

tokio/src/signal/unix.rs
```diff
@@ -294,26 +283,25 @@ fn signal_enable(signal: SignalKind, handle: &Handle) -> io::Result<()> {
...
+                error = Some(e);
+                Err(())
+            } else {
+                Ok(())
+            }
+        })
+        .map_err(|_| {
+            error.unwrap_or_else(|| {
+                Error::new(ErrorKind::Other, "registering signal handler failed")
+            })
```
ipetkov: If we're going to make the effort to remember that this has failed in the past, we might as well return the same error that was raised (i.e. we should be capturing the last error in the OnceLock) instead of "minting" our own `Other` error here

tokio/src/signal/windows/sys.rs
```diff
@@ -85,22 +85,23 @@ impl Storage for OsStorage {
...
-    });
-
-    init.unwrap_or_else(|| Ok(()))
+        }
+    })
+    .map_err(|_| {
+        error.unwrap_or_else(|| {
+            io::Error::new(io::ErrorKind::Other, "registering signal handler failed")
+        })
+    })
```
ipetkov: So besides changing the `Once` to a `OnceLock` I'm not really seeing what has changed here, besides throwing away the underlying error? Which I'm also not sure I agree with, why are we making all errors with `io::ErrorKind::Other`?

ipetkov (changes requested):

tokio/src/signal/unix.rs
vilgotf: I feel like we are going in a circle. To store the original error we need the errno value, i.e. `raw_os_error()` (which returns `Option<i32>)`. Also what do you think about my previous comment?

> I actually prefer differing between initial and repeat errors, the errno value is not semantically accurate for repeat errors since they do not try to register the handle.

tokio/src/signal/windows/sys.rs
vilgotf: The original error is captured in `error`. The change is that subsequent calls now return an error instead of erroneously returning `Ok`

tokio/src/signal/unix.rs
ipetkov: My view is that we should pass through the original error even in repeat calls because that _is_ the error which led to the operation failing. The fact that we only try once happens to be an abstraction we've chosen; I don't think its any more "wrong" to return the same error (pretending we tried again but failed)

Only if `raw_os_error()` returns `None` we can fall back to `Error::other("...")`

vilgotf: Okay I will update the code to store `Result<(), Option<i32>>`

martin-g (approved):

PR #7836, opened 2026-01-05
fix: require tokio_unstable cfg for io-uring and taskdump features

innobead:
## Motivation

To resolve the build failure on Mac.

## Solution

As the commit message indicated.


mattiapitossi:
This is a bit of annoying on Mac when using all features, I think what you can try is manually selecting features (or use full/stable), not sure if there's another way to fix this

PR #7838, opened 2026-01-06
rt: `event_interval` panics when set to 0

mattiapitossi:
## Motivation

The following configuration fails:

```rust
    let rt = Builder::new_multi_thread()
        .enable_all()
        .event_interval(0)
        .build()
        .unwrap();
```

with

```
thread 'tokio-runtime-worker' (8988985) panicked at src/runtime/scheduler/multi_thread/worker.rs:740:12:
attempt to calculate the remainder with a divisor of zero
```

Due to this part of the code in the multi_threaded worker:

```rust
   fn maintenance(&self, mut core: Box<Core>) -> Box<Core> {
        if core.tick % self.worker.handle.shared.config.event_interval == 0 {
```

It seems that event_interval=0 also does not work with current_thread runtime, for example with channel notification that are never handled in the block_on loop:

```rust
 fn block_on<F: Future>(self, future: F) -> F::Output {
...
   for _ in 0..handle.shared.config.event_interval {
... handles notify and run the task
```

While I'm not sure if Tokio should support `event_interval` to be 0, at the current moment it panics on the multi_thread runtime and it loops forever in the current_thread (please let me know if this requires an issue).

## Solution

I'm proposing a similar solution that was applied to the global_queue_interval to avoid unintended behavior when this value is zero.

### DIFF
--- a/tokio/tests/rt_panic.rs
+++ b/tokio/tests/rt_panic.rs
@@ -94,7 +94,6 @@ fn builder_event_interval_interval_panic_caller() -> Result<(), Box<dyn Error>> 
     Ok(())
 }
 
-
 fn current_thread() -> Runtime {
     tokio::runtime::Builder::new_current_thread()
         .enable_all()


Darksonn (approved):
Thanks.

PR #7839, opened 2026-01-06
docs: update outdated unstable features section

mattiapitossi:
## Motivation

Some docs for `unstable features` section were outdated, there was also a leftover config after the stabilization of the task id

[Rendered](https://deploy-preview-7839--tokio-rs.netlify.app/tokio/)

### DIFF
--- a/tokio/src/lib.rs
+++ b/tokio/src/lib.rs
@@ -320,8 +320,7 @@
 //! Beware though that this will pull in many extra dependencies that you may not
 //! need.
 //!
-//! - `full`: Enables all features listed below except `test-util`, `tracing`,
-//!   `taskdump`, and `io-uring`.
+//! - `full`: Enables all features listed below except `test-util` and unstable features.
 //! - `rt`: Enables `tokio::spawn`, the current-thread scheduler,
 //!   and non-scheduler utilities.
 //! - `rt-multi-thread`: Enables the heavier, multi-threaded, work-stealing scheduler.


tokio/src/lib.rs
```diff
@@ -351,16 +352,18 @@
...
+//! - `io-uring`: Enables io-uring (Linux only).
+//! - `taskdump`: Enables taskdump (Linux only).
 //!
-//! Likewise, some parts of the API are only available with the same flag:
+//! Likewise, this flag enables access to specific unstable APIs, such as:
 //!
 //! - [`task::Builder`]
-//! - Some methods on [`task::JoinSet`]
-//! - [`runtime::RuntimeMetrics`]
+//! - `build_task` method on [`task::JoinSet`]
```
mattiapitossi: not sure keeping this list adds much value since there are lots of unstable APIs, as alternative we can just mention that the FF enables also access to specific unstable APIs

Darksonn: That makes sense to me.

mattiapitossi: Thanks, removed!

tokio/src/lib.rs
```diff
@@ -320,7 +320,8 @@
 //! Beware though that this will pull in many extra dependencies that you may not
 //! need.
 //!
-//! - `full`: Enables all features listed below except `test-util` and `tracing`.
+//! - `full`: Enables all features listed below except `test-util`, `tracing`,
+//!   `taskdump`, and `io-uring`.
```
Darksonn: what about:

> except `test-util` and unstable features.

mattiapitossi: agree, updated

PR #7843, opened 2026-01-08
fs: avoid lock acquisition after uring init

tahmid-23:
## Motivation
#7815 added a lock acquisition after the uring had been initialized. This negates the point of the atomic state variable. However, we need to mutably initialize the `Probe` once. This sounds exactly like a use case for `OnceCell` / `OnceLock`.

Instead of an atomic indicating the uring state, we use a `OnceCell<Option<Probe>>` to indicate the uring state: `None` indicating it being unsupported, and the `Probe` indicating successful initialization. (This `Probe` could later be replaced with any other necessary readonly uring state, but this is currently the only necessary type.)

Although we would like to use std's `OnceLock`, its `get_or_try_init` method is currently unstable. We cannot implement it ourselves (as previously present), since `OnceState::poison` is not public and is necessary to implement `get_or_try_init`. Hence, I added `once_cell` as a dependency under the `io_uring` feature. The 1.65 MSRV should be compatible anyway.

### DIFF
--- a/tokio/src/runtime/io/driver/uring.rs
+++ b/tokio/src/runtime/io/driver/uring.rs
@@ -188,7 +188,9 @@ impl Handle {
             }
         })?;
 
-        Ok(probe.as_ref().is_some_and(|probe| probe.is_supported(opcode)))
+        Ok(probe
+            .as_ref()
+            .is_some_and(|probe| probe.is_supported(opcode)))
     }
 
     /// Initialize the io_uring context if it hasn't been initialized yet.


tokio/src/runtime/io/driver/uring.rs
```diff
@@ -207,41 +173,35 @@ impl Handle {
...
-            State::Unsupported => Ok(false),
-            State::Initialized => Ok(self.get_uring().lock().is_opcode_supported(opcode)),
-        }
+            }
+        })?;
+
+        Ok(match probe {
+            Some(probe) => probe.is_supported(opcode),
+            None => false,
+        })
```
ADD-SP:
```suggestion
        Ok(probe.is_some_and(|x| x.is_supported(opcode)))
```

tokio/src/runtime/io/driver/uring.rs
```diff
@@ -207,41 +173,35 @@ impl Handle {
...
     }
 
     /// Initialize the io_uring context if it hasn't been initialized yet.
-    /// Then, check whether the given opcode is supported.
```
ADD-SP, line 237: unintended change?

tahmid-23: nope, I reverted this function because we checks opcode support above, after acquiring the probe. so this comment no longer makes sense.

tokio/src/runtime/io/driver/uring.rs
```diff
@@ -207,41 +173,34 @@ impl Handle {
...
                 Err(e) => Err(e),
-            },
-            State::Unsupported => Ok(false),
-            State::Initialized => Ok(self.get_uring().lock().is_opcode_supported(opcode)),
-        }
+            }
+        })?;
+
+        Ok(probe
+            .as_ref()
```
ADD-SP, line 192: Maybe a dumb question, is this `.as_ref` required? It should be ok to consume the `probe`.

tahmid-23:
```
error[E0507]: cannot move out of `*probe` which is behind a shared reference
   --> tokio/src/runtime/io/driver/uring.rs:191:12
    |
191 |         Ok(probe.is_some_and(|probe| probe.is_supported(opcode)))
    |            ^^^^^ ----------------------------------------------- `*probe` moved due to this method call
    |            |
    |            help: consider calling `.as_ref()` or `.as_mut()` to borrow the type's contents
    |            move occurs because `*probe` has type `Option<Probe>`, which does not implement the `Copy` trait
    |
note: `Option::<T>::is_some_and` takes ownership of the receiver `self`, which moves `*probe`
   --> /home/as-threadripper/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:655:30
    |
655 |     pub const fn is_some_and(self, f: impl [const] FnOnce(T) -> bool + [const] Destruct) -> bool {
    |                              ^^^^
help: you could `clone` the value and consume it, if the `Probe: Clone` trait bound could be satisfied
    |
191 |         Ok(<Option<Probe> as Clone>::clone(&probe).is_some_and(|probe| probe.is_supported(opcode)))
    |            +++++++++++++++++++++++++++++++++     +

For more information about this error, try `rustc --explain E0507`.
error: could not compile `tokio` (lib) due to 1 previous error
```

PR #7845, opened 2026-01-09
signal: use eventfd instead of UnixStream for signal notification

winter-loo:
## Motivation

According to the Linux [eventfd](https://man7.org/linux/man-pages/man2/eventfd.2.html) doc:

> Applications can use an eventfd file descriptor instead of a pipe
       (see [pipe(2)](https://man7.org/linux/man-pages/man2/pipe.2.html)) in all cases where a pipe is used simply to signal
       events.  The kernel overhead of an eventfd file descriptor is much
       lower than that of a pipe, and only one file descriptor is
       required (versus the two required for a pipe).

So, I made an attempt to replace `mio::UnixStream` with `libc::eventfd`.

## Solution
The benchmark(`RUSTFLAGS="" cargo bench -p benches --bench signal`) shows some improvements:

1. On arch linux, the performance improvement is about 5%.
2. On ubuntu inside WSL 2, the performance gain is about 15%
3. On centos 7, the performance gain is about 10%

image

### DIFF
--- a/tokio/src/signal/windows/sys.rs
+++ b/tokio/src/signal/windows/sys.rs
@@ -29,7 +29,7 @@ pub(super) fn ctrl_shutdown() -> io::Result<RxFuture> {
 
 fn new(signum: u32) -> io::Result<RxFuture> {
     global_init()?;
-    let rx = globals()?.register_listener(signum as EventId)?;
+    let rx = globals()?.register_listener(signum as EventId);
     Ok(RxFuture::new(rx))
 }
 


tokio/src/signal/unix.rs
```diff
@@ -61,20 +60,211 @@ impl Storage for OsStorage {
...
+        self.sender.receiver()
+    }
+}
 
+#[cfg(not(any(target_os = "linux", target_os = "illumos")))]
+#[derive(Debug)]
+pub(crate) struct OsExtraData {
+    sender: pipe::Sender,
+    receiver: pipe::Receiver,
+}
```
Darksonn: It's hard to read this code with two separate implementations next to each other. Can you maybe put them in separate eventfd / pipe modules?

tokio/src/runtime/signal/mod.rs
```diff
@@ -4,9 +4,9 @@
 
 use crate::runtime::{driver, io};
 use crate::signal::registry::globals;
+use crate::signal::unix::pipe;
 
-use mio::net::UnixStream;
-use std::io::{self as std_io, Read};
+use std::io::{self as std_io};
```
ipetkov:
```suggestion
use std::io as std_io;
```

tokio/src/signal/unix.rs
ipetkov: +1 please move each of the modules into their own file

tokio/src/signal/unix.rs
```diff
@@ -61,20 +60,211 @@ impl Storage for OsStorage {
...
+            SourceFd(&self.fd.as_raw_fd()).deregister(registry)
+        }
+    }
+
+    impl Sender {
+        pub(crate) fn new() -> Self {
+            let fd = unsafe { libc::eventfd(0, libc::EFD_NONBLOCK | libc::EFD_CLOEXEC) };
+            if fd < 0 {
+                panic!("eventfd failed: {}", io::Error::last_os_error());
+            }
```
ipetkov: Instead of panicking can we return an error here?

tokio/src/signal/unix.rs
```diff
@@ -61,20 +60,211 @@ impl Storage for OsStorage {
...
+            Sender {
+                fd: unsafe { OwnedFd::from_raw_fd(fd) },
+            }
+        }
+
+        pub(crate) fn receiver(&self) -> Receiver {
+            Receiver {
+                fd: self.fd.try_clone().unwrap(),
```
ipetkov: We should not unwrap here, if this is fallible operation (duplicating file descriptors _is_ fallible) we should return an error

tokio/src/signal/unix.rs
```diff
@@ -61,20 +60,211 @@ impl Storage for OsStorage {
...
+    }
+
+    impl Clone for Receiver {
+        fn clone(&self) -> Self {
+            let receiver_fd = self.inner.as_raw_fd();
+            let original = ManuallyDrop::new(unsafe {
+                std::os::unix::net::UnixStream::from_raw_fd(receiver_fd)
+            });
+            let inner =
+                UnixStream::from_std(original.try_clone().expect("failed to clone UnixStream"));
```
ipetkov: Same comment, this is a fallible operation, we should return an error here, not panic

tokio/src/signal/unix.rs
```diff
@@ -61,20 +60,211 @@ impl Storage for OsStorage {
...
+            // Drain the pipe completely so we can receive a new readiness event
+            // if another signal has come in.
+            let mut buf = [0; 128];
+            #[allow(clippy::unused_io_amount)]
+            loop {
+                match self.inner.read(&mut buf) {
+                    Ok(0) => panic!("EOF on self-pipe"),
+                    Ok(_) => continue, // Keep reading
+                    Err(e) if e.kind() == std::io::ErrorKind::WouldBlock => break,
+                    Err(e) => panic!("Bad read on self-pipe: {e}"),
```
ipetkov: Lets replace these panics with returning an error

tokio/src/signal/unix.rs
```diff
@@ -61,20 +60,211 @@ impl Storage for OsStorage {
...
+                }
+            }
+            0
+        }
+    }
+
+    pub(crate) fn channel() -> (Sender, Receiver) {
+        let (sender, receiver) = UnixStream::pair().expect("failed to create UnixStream");
```
ipetkov: Again, return an error here

ipetkov (changes requested):

winter-loo:
Thanks for review. It's time to make it better.

tokio/src/signal/pipe/eventfd.rs
Darksonn, line 1: Comparing with [what mio does](https://github.com/tokio-rs/mio/blob/master/src/sys/unix/waker/eventfd.rs), this seems to not have a bunch of logic to reset the eventfd and so on... Is it missing here?

winter-loo: A read on fd returned by eventfd resets eventfd. Here's `libc::eventfd_read` in Receiver::read() function.

tokio/src/signal/registry.rs
```diff
@@ -144,18 +144,18 @@ impl Globals {
 
 fn globals_init() -> Globals
 where
-    OsExtraData: 'static + Send + Sync + Default,
+    OsExtraData: 'static + Send + Sync,
     OsStorage: 'static + Send + Sync + Default,
 {
     Globals {
-        extra: OsExtraData::default(),
+        extra: OsExtraData::new().expect("failed to initialize OsExtraData"),
```
ipetkov: Can we also turn this initialization fallible instead of panicking? I took a quick peek and every caller of `globals()` returns an `io::Result` so we should be able to propagate that instead of panicking

winter-loo: I followed your suggestion, but ran into a complication with [signal::Driver::process](https://github.com/winter-loo/tokio/blob/e7e66d051dcd85c9fe1fceda7f9fe5e84f986275/tokio/src/runtime/signal/mod.rs#L106) since it doesn't return a std::io::Result. To proceed with the refactor, I’d also need to modify the park and park_timeout methods.

I think the original approach is sufficient for now, as `globals_init` is a one-time call during the program's initialization.

ipetkov: We could do a best-effort broadcast there (i.e. `if let Ok(g) = globals() { g.broadcast() }`)

If by that point we failed to initialize then there's nothing to broadcast anyway

winter-loo: Hi, I have commited my changes. However, the CI failed to pass the Windows full future test and **I have no clue to find the cause** as the CI said only "job failed". I have also tested on Windows with command: `cargo test -p tokio --features full` and the test passed: 

> test result: ok. 702 passed; 0 failed; 62 ignored; 0 measured; 0 filtered out; finished in 213.97s

Could you help me pinpoint the problem?

mattiapitossi: I don't think it's related to your changes. I'm having the same issue where some Windows related steps fail after 3s.

martin-g: I've restarted the CI jobs and now they passed!

tokio/src/signal/registry.rs
```diff
@@ -142,25 +142,28 @@ impl Globals {
...
 
-pub(crate) fn globals() -> &'static Globals
+pub(crate) fn globals() -> std::io::Result<&'static Globals>
 where
-    OsExtraData: 'static + Send + Sync + Default,
+    OsExtraData: 'static + Send + Sync,
     OsStorage: 'static + Send + Sync + Default,
 {
-    static GLOBALS: OnceLock<Globals> = OnceLock::new();
+    static GLOBALS: OnceLock<std::io::Result<Globals>> = OnceLock::new();
```
vilgotf, line 161: Drive by review: I think you should store `Result<Globals, i32>` like I did for the per-signal result storage. You can then convert the i32 to an error via `Error::from_raw_os_error` and avoid allocating.

tokio/src/signal/pipe/eventfd.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    pub(crate) fn receiver(&self) -> std::io::Result<Receiver> {
+        Ok(Receiver {
+            fd: self.fd.try_clone()?,
+        })
+    }
+}
+
+impl Sender {
```
martin-g, line 62: This could be merged with the `impl` above.

tokio/src/signal/pipe/eventfd.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub(crate) fn read(&mut self) -> std::io::Result<libc::c_int> {
+        let fd = self.fd.as_raw_fd();
+        let mut value: libc::eventfd_t = 0;
+
+        // SAFETY: it's ok to call libc API
+        let r = unsafe { libc::eventfd_read(fd, &mut value as *mut libc::eventfd_t) };
+        if r == 0 {
+            Ok(0)
+        } else {
+            Err(std::io::Error::last_os_error())
```
martin-g, line 84: Since it is non-blocking it may return EAGAIN (ErrorKind::WouldBlock) which is not an error.

```rust
    let err = std::io::Error::last_os_error();
    // On a non-blocking eventfd, it is expected to get an EAGAIN
    // when the counter is 0.
    if err.kind() == std::io::ErrorKind::WouldBlock {
        Ok(0)
    } else {
        Err(err)
    }
```

tokio/src/signal/pipe/unixstream.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+impl OsExtraData {
+    pub(crate) fn new() -> std::io::Result<Self> {
+        let (sender, receiver) = channel()?;
+        Ok(Self { sender, receiver })
+    }
+}
+
+impl OsExtraData {
```
martin-g, line 82: This impl could be merged with the one above

tokio/src/signal/pipe/eventfd.rs
```diff
@@ -0,0 +1,107 @@
...
+        })
+    }
+}
+
+impl Sender {
+    pub(crate) fn write(&self) -> std::io::Result<usize> {
+        // SAFETY: it's ok to call libc API
+        let r = unsafe { libc::eventfd_write(self.fd.as_raw_fd(), 1) };
+        if r == 0 {
+            Ok(0)
```
martin-g, line 67: There is a small different between the eventfd and unixstream impls here.
The eventfd's Sender::write() returns 0 on success. The UnixStream returns the number of written bytes (1).
AFAIS the successful result is not used. 
Maybe change eventfd to return 1 too or change both to return `Result<()>` ?!

PR #7847, opened 2026-01-09
 fs: implement tokio::fs::create_dir[_all] via io_uring

tahmid-23:
io_uring + mkdirat()

As opposed to std's implementation of recursive directory creation, this does not use a recursive function to create parent directories. Instead, it iterates over positions of `/` (the unix / linux path separator), first failing to create parent directories until success, and then creating all children.

This prevents us from allocating a boxed future for each recursive function call.
It does have the minimal overhead of needing to find the `/` positions an additional time in the second forward iterator. We could save this by allocating a `Vec` of `/` positions; I don't necessarily know what's better.

### DIFF
--- a/tokio/src/fs/create_dir.rs
+++ b/tokio/src/fs/create_dir.rs
@@ -57,7 +57,10 @@ pub async fn create_dir(path: impl AsRef<Path>) -> io::Result<()> {
     {
         let handle = crate::runtime::Handle::current();
         let driver_handle = handle.inner.driver().io();
-        if driver_handle.check_and_init(io_uring::opcode::MkDirAt::CODE).await? {
+        if driver_handle
+            .check_and_init(io_uring::opcode::MkDirAt::CODE)
+            .await?
+        {
             return crate::runtime::driver::op::Op::mkdir(&path)?.await;
         }
     }
--- a/tokio/src/fs/create_dir_all.rs
+++ b/tokio/src/fs/create_dir_all.rs
@@ -60,7 +60,10 @@ pub async fn create_dir_all(path: impl AsRef<Path>) -> io::Result<()> {
 
         let handle = crate::runtime::Handle::current();
         let driver_handle = handle.inner.driver().io();
-        if driver_handle.check_and_init(io_uring::opcode::MkDirAt::CODE).await? {
+        if driver_handle
+            .check_and_init(io_uring::opcode::MkDirAt::CODE)
+            .await?
+        {
             return create_dir_all_uring(&path).await;
         }
     }


tokio/src/fs/create_dir.rs
```diff
@@ -46,5 +46,21 @@ use std::path::Path;
...
+        tokio_unstable,
+        feature = "io-uring",
+        feature = "rt",
+        feature = "fs",
+        target_os = "linux"
+    ))]
+    {
+        let handle = crate::runtime::Handle::current();
+        let driver_handle = handle.inner.driver().io();
+        if driver_handle.check_and_init(io_uring::opcode::MkDirAt::CODE)? {
```
martin-g:
```suggestion
        if driver_handle.check_and_init(io_uring::opcode::MkDirAt::CODE).await? {
```

tokio/src/fs/create_dir_all.rs
```diff
@@ -47,5 +47,23 @@ use std::path::Path;
...
+        feature = "rt",
+        feature = "fs",
+        target_os = "linux"
+    ))]
+    {
+        use crate::fs::create_dir_all_uring;
+
+        let handle = crate::runtime::Handle::current();
+        let driver_handle = handle.inner.driver().io();
+        if driver_handle.check_and_init(io_uring::opcode::MkDirAt::CODE)? {
```
martin-g:
```suggestion
        if driver_handle.check_and_init(io_uring::opcode::MkDirAt::CODE).await? {
```

martin-g (commented):
The build fails with:
```
error[E0277]: the `?` operator can only be applied to values that implement `std::ops::Try`
  --> tokio/src/fs/create_dir.rs:60:12
   |
60 |         if driver_handle.check_and_init(io_uring::opcode::MkDirAt::CODE)? {
   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the `?` operator cannot be applied to type `impl std::future::Future<Output = std::result::Result<bool, std::io::Error>>`
   |
   = help: the trait `std::ops::Try` is not implemented for `impl std::future::Future<Output = std::result::Result<bool, std::io::Error>>`
help: consider `await`ing on the `Future`
```

tokio/src/fs/create_dir_all_uring.rs
```diff
@@ -0,0 +1,90 @@
...
+    let path_bytes = path.as_os_str().as_bytes();
+    for (separator_pos, _) in path_bytes
+        .iter()
+        .enumerate()
+        .rev()
+        .filter(|(_, &byte)| byte == b'/')
+    {
+        let parent_bytes = &path_bytes[..separator_pos];
+        let parent_str = OsStr::from_bytes(parent_bytes);
+        let parent_path = Path::new(parent_str);
```
martin-g, line 42: Why not use [Path::parent()](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.parent) ?

tokio/src/fs/create_dir_all_uring.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    // Once we have found the correct /,
+    // we can iterate the remaining components in the forward direction.
+    //
+    // In the example /a/b/c/d path, there is only one remaining /, after c.
+    // Hence, we first create /a/b/c.
+    //
+    // TODO: We're attempting to create all directories sequentially.
+    // This would benefit from batching.
+    for (separator_pos, _) in path_bytes
```
martin-g, line 58: And here you can use [Path::components()](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.components)

tokio/src/fs/create_dir_all_uring.rs
```diff
@@ -0,0 +1,90 @@
...
+    mkdir_parent_created(path).await?;
+    Ok(())
+}
+
+async fn mkdir_parent_missing(path: &Path) -> io::Result<bool> {
+    match Op::mkdir(path)?.await {
+        Ok(()) => Ok(true),
+        Err(ref e) if e.kind() == io::ErrorKind::NotFound => Ok(false),
+        // TODO: replace with uring-based statx
+        Err(_) if crate::fs::metadata(path).await?.is_dir() => Ok(true),
```
martin-g, line 78: `metadata(path).await?` may hide the original error (the ignored one in `Err(_)`)

```suggestion
        Err(ref e) if e.kind() == io::ErrorKind::AlreadyExists => {
            if crate::fs::metadata(path).await?.is_dir() {
                Ok(true)
            } else {
                Err(e)
            }
        }
```
something like this should be better

tokio/src/fs/create_dir_all_uring.rs
```diff
@@ -0,0 +1,90 @@
...
+        Err(e) => Err(e),
+    }
+}
+
+async fn mkdir_parent_created(path: &Path) -> io::Result<()> {
+    match Op::mkdir(path)?.await {
+        Ok(()) => Ok(()),
+        // TODO: replace with uring-based statx
+        Err(_) if crate::fs::metadata(path).await?.is_dir() => Ok(()),
```
martin-g, line 87: Same as above

tokio/src/fs/create_dir_all_uring.rs
```diff
@@ -0,0 +1,90 @@
+#[cfg(all(
+    tokio_unstable,
+    feature = "io-uring",
+    feature = "rt",
+    feature = "fs",
+    target_os = "linux"
+))]
```
martin-g, line 7: This cfg guard is not really needed.
This module is already conditionally used with `cfg_io_uring!` in mod.rs

tokio/src/io/uring/mkdir.rs
```diff
@@ -0,0 +1,49 @@
+use super::utils::cstr;
+
+use crate::runtime::driver::op::{CancelData, Cancellable, Completable, CqeResult, Op};
+
+use io_uring::{opcode, types};
+use std::ffi::CString;
+use std::io;
+use std::io::Error;
```
martin-g, line 8:
```suggestion
use std::io::{self, Error};
```

tokio/tests/fs_uring_mkdir.rs
```diff
@@ -0,0 +1,168 @@
...
+use std::future::Future;
+use std::pin::pin;
+use std::sync::mpsc;
+use std::task::Poll;
+use std::time::Duration;
+use tokio::runtime::{Builder, Runtime};
+use tokio_test::assert_pending;
+use tokio_util::task::TaskTracker;
+
+fn multi_rt(n: usize) -> Box<dyn Fn() -> Runtime> {
```
martin-g, line 22: Those seem copied from one of the other tests.
Let's extract them into a helper module.

Issue #7832, opened 2026-01-04
Release steps say to remove path deps


Darksonn:
We used to remove path dependencies in the `Cargo.toml` when releasing a crate. This was to ensure that the released crate does not depend on unreleased changes in other crates, as removing the path dependency makes CI compile it using the crates.io version of the dependency.

However, we've stopped doing this, and the big reason is that it doesn't work anymore. CI triggers errors such as this one when you try it:
```
error: There are multiple `tokio` packages in your project, and the specification `tokio` is ambiguous.
Please re-run this command with `-p <spec>` where `<spec>` is one of the following:
  file:///home/runner/work/tokio/tokio/tokio#1.49.0
  https://github.com/rust-lang/crates.io-index#tokio@1.49.0
Error: Process completed with exit code 101.
```
We should either fix CI so that we can continue doing it, or we should remove this instruction from the docs.

ADD-SP:
> Cargo only looks at the patch settings in the `Cargo.toml` manifest at the root of the workspace. Patch settings defined in dependencies will be ignored.
>
> <https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section>

We could try to patch the `tokio-*` crates in the root `Cargo.toml` for daily development, and remove it before the release.

```toml
[patch.crates-io]
tokio = { path = "tokio" }
tokio-macros = { path = "tokio-macros" }
tokio-test = { path = "tokio-test" }
tokio-stream = { path = "tokio-stream" }
tokio-util = { path = "tokio-util" }
```

Darksonn:
I'm not sure that would work:

* If there are two crates that both depend on unreleased changes, then removing the path dep to release one may break the other.
* When we *do* remove the patch entries, I think that still results in the same ambiguity error as we get today.

martin-g:
> or we should remove this instruction from the docs.

Where is the document explaining the release steps ?

ADD-SP:
After removing this release step, how to capture the issue caused by depending on unreleased crate?

Darksonn:
> Where is the document explaining the release steps ?

It's in the `Cargo.toml` files. E.g.:

https://github.com/tokio-rs/tokio/blob/41d1877689f8669902b003a6affce60bdfeb3025/tokio/Cargo.toml#L3-L9

Darksonn:
> After removing this release step, how to capture the issue caused by depending on unreleased crate?

Right now we simply do not have any mechanism to catch such mistakes.

[...]

Issue #7732, opened 2025-11-12
Tokio-util should have its tokio dependency bumped


sarsko:
Got the following build error:

```
error[E0433]: failed to resolve: could not find `coop` in `task`
   --> /.../tokio-util-0.7.17/src/task/join_queue.rs:197:46
    |
197 |         let jh = std::pin::pin!(tokio::task::coop::unconstrained(jh));
    |  
```

It looks to be due to tokio-util because the tokio-util tokio dependency is on version 1.28.0 (https://docs.rs/crate/tokio-util/0.7.17/source/Cargo.toml#32), which doesn't have `task::coop` (published in 1.44 — It exists in https://docs.rs/tokio/1.44.0/tokio/task/coop/index.html, but not in https://docs.rs/tokio/1.43.3/tokio/task/coop/index.html). 

This applies only to tokio-util 0.7.17, which introduced `join_queue` (there in https://docs.rs/crate/tokio-util/0.7.17/source/src/task/, not in https://docs.rs/crate/tokio-util/0.7.16/source/src/task/).

I'll put up an issue instead of a fix because I'm not sure if the desired solution is to bump the tokio dependency or to gain `Unconstrained` from `task` for pre 1.44 versions, or to gate `join_queue`.

ADD-SP:
Yes, this is not the only issue. I also encountered another similar issue on my devbox. I think we should bump the tokio version in the `tokio-util/Cargo.toml`.

<img src="https://github.com/user-attachments/assets/f911efe4-869d-4e31-a996-6f4103a726a4" />

PR #7848, opened 2026-01-09
ci: fix ambiguity issue during tokio releases

mattiapitossi:
## Motivation

see #7832 

## Solution

More context here: https://github.com/tokio-rs/tokio/issues/7832#issuecomment-3728051001

### DIFF
--- a/tokio/src/runtime/builder.rs
+++ b/tokio/src/runtime/builder.rs
@@ -516,7 +516,7 @@ impl Builder {
 
     /// Sets name of threads spawned by the `Runtime`'s thread pool.
     ///
-    /// The default name is "tokio-runtime-worker".
+    /// The default name is "tokio-rt-worker".
     ///
     /// # Examples
     ///
@@ -540,7 +540,7 @@ impl Builder {
 
     /// Sets a function used to generate the name of threads spawned by the `Runtime`'s thread pool.
     ///
-    /// The default name fn is `|| "tokio-runtime-worker".into()`.
+    /// The default name fn is `|| "tokio-rt-worker".into()`.
     ///
     /// # Examples
     ///
--- a/tokio/tests/tcp_socket.rs
+++ b/tokio/tests/tcp_socket.rs
@@ -123,10 +123,10 @@ const SET_BUF_SIZE: u32 = 4096;
 // Linux doubles the buffer size for kernel usage, and exposes that when
 // retrieving the buffer size.
 
-#[cfg(not(target_os = "linux"))]
+#[cfg(not(any(target_os = "android", target_os = "linux")))]
 const GET_BUF_SIZE: u32 = SET_BUF_SIZE;
 
-#[cfg(target_os = "linux")]
+#[cfg(any(target_os = "android", target_os = "linux"))]
 const GET_BUF_SIZE: u32 = 2 * SET_BUF_SIZE;
 
 test!(keepalive, set_keepalive(true));


tokio-util/Cargo.toml
```diff
@@ -35,7 +35,7 @@ join-map = ["rt", "hashbrown"]
 __docs_rs = ["futures-util"]
 
 [dependencies]
-tokio = { version = "1.44.0", path = "../tokio", features = ["sync"] }
+tokio = { version = "1.44.0", features = ["sync"] }
```
martin-g: This is interesting!

Do I understand it correctly ?
Until now for local build it was using the in-development version of Tokio (`path = "../tokio"`), i.e. it may use functionalities of Tokio 1.45+. But once released `path` is no more in play and it is built against latest Tokio 1.44.x

martin-g: why 1.44 ? Why not an older or a newer version ?

tokio-util/Cargo.toml
```diff
@@ -47,9 +47,9 @@ tracing = { version = "0.1.29", default-features = false, features = ["std"], op
 hashbrown = { version = "0.15.0", default-features = false, optional = true }
 
 [dev-dependencies]
-tokio = { version = "1.0.0", path = "../tokio", features = ["full"] }
-tokio-test = { version = "0.4.0", path = "../tokio-test" }
-tokio-stream = { version = "0.1", path = "../tokio-stream" }
+tokio = { version = "1.0.0", features = ["full"] }
```
martin-g, line 50: 1.0.0 resolves to the latest 1.x, i.e. 1.49.0 at the moment, right ?
Why do we use 1.44 above and 1.0.0 here ?!

tokio-util/Cargo.toml
mattiapitossi: Yes, that's correct, I've added additional context on https://github.com/tokio-rs/tokio/issues/7832#issuecomment-3728051001.

I don't think there's a specific reason on why we still use 1.44, I guess it wasn't just updated since then, maybe it can be bumped to 1.49.0 now?

Darksonn: Tokio-util should specify the oldest version of Tokio that it may be used with. In this case, Tokio 1.44 introduced the `tokio::task::coop` module, which tokio-util uses. (ref #7732)

tokio-util/Cargo.toml
Darksonn: It's probably more useful to specify the same version both places, but when a dep appears in both deps and dev-deps, then the largest of the two is the real number.

Darksonn:
We should also document this procedure somewhere in `CONTRIBUTING.md`. We can explain that each crate (tokio-util, tokio-stream, etc) should specify dependencies according to these rules:

1. The listed version should be the oldest version that the crate works with.
2. When a crate starts using a newer feature in a dependency, the version should be bumped to the version that introduced it.
3. If a crate depends on an unreleased feature in a dependency, it may use a `path =` dependency to specify this. Since path deps must be removed on release, this ensures that it can't be released until the dependency has a new version.

tokio-util/Cargo.toml
Darksonn: > maybe it can be bumped to 1.49.0 now?

No do not bump it unnecessarily. If tokio-util works with releases 1.44, 1.45, 1.46, 1.47, 1.48, 1.49, then we should not *require* the user to use 1.49. Instead, we should accurately state that the minimum required version of Tokio is 1.44.

mattiapitossi: I added some docs in the `CONTRIBUTING.md` mentioning this, and also the suggestion below

.github/workflows/ci.yml
```diff
@@ -53,6 +53,35 @@ jobs:
     steps:
       - run: exit 0
 
+  # Checks that crates do not depend on unpublished versions of their deps
+  # by stripping path deps and verifying they compile with the latest version
+  # published on crates.io.
+  # see: (https://github.com/tokio-rs/tokio/issues/7832)
+  check-release:
+    name: check release ${{ matrix.crate }}
```
mattiapitossi: Maybe this could use https://github.com/marketplace/actions/continue-on-error-comment so we could still add unreleased features on crates without blocking

Darksonn: I do not think that is required. We can add dependencies on unreleased features using the following strategy:

1. In the PR where tokio-util starts using an unreleased feature in Tokio, add a path dependency from tokio-util to tokio.
2. After releasing Tokio, remove the path dependency again, as it is no longer needed.

Darksonn: So this means that:
```toml
tokio = "1.47.0"
```
means "depends only on released features and the oldest compatible version is 1.47.0", whereas
```toml
tokio = { version = "1.47.0", path = "../tokio" }
```
means "depends on unreleased features".

Once Tokio 1.48.0 is released, we can change it:
```diff
-tokio = { version = "1.47.0", path = "../tokio" }
+tokio = "1.48.0"
```

mattiapitossi: So, If I got correctly: 
- I can remove this `release-check` step in the CI
- Remove all the path dependencies across crates (tokio, tokio-util,..)
- Maybe also update the contributing mentioning this?

Darksonn: Yes.

docs/contributing/how-to-specify-crates-dependencies-versions.md
```diff
@@ -0,0 +1,29 @@
+# How to specify crates dependencies versions
+
+Each crate (e.g., `tokio-util`, `tokio-stream`, etc.) should specify dependencies
+according to these rules:
+
+1. The listed version should be the oldest version that the crate works with
+(e.g., if `tokio-util` works with `1.44`, this should be the version specified.
```
Darksonn:
```suggestion
(e.g., if `tokio-util` works with `tokio` version `1.44` but not `1.43`, then `tokio-util` should specify version `1.44` for its `tokio` dependency.)
```

docs/contributing/how-to-specify-crates-dependencies-versions.md
```diff
@@ -0,0 +1,29 @@
...
+[dependencies]
+futures-core = { version = "0.3.0" }
+pin-project-lite = "0.2.11"
+tokio = { version = "1.15.0", path = "../tokio", features = ["sync"] }
+```
+
+In this case, local development of `tokio-stream` uses the local version
+of `tokio` via the `path` dependency. During CI checks, the path will be
+removed to check if the crate can be released and does not depend on unreleased
+changes. As mentioned before, this version should only be bumped when adding a new
```
Darksonn:
```suggestion
of `tokio` via the `path` dependency. This means that it's currently not possible to release `tokio-stream`. Once a new version of `tokio` is released, the path dependency will be removed from `tokio-stream`.
As mentioned before, this version should only be bumped when adding a new
```

examples/Cargo.toml
Darksonn, line 1: Is it possible to keep path deps for examples and benches?

mattiapitossi: I was trying the solution to remove path deps from the crates but we have issues with cyclical dependencies. For example in Tokio we have a dev dependency on tokio-util, and tokio-util has a dependency on tokio:

```bash
cd tokio && cargo tree

tokio v1.49.0

tokio v1.49.0
[dev-dependencies]
....
└── tokio-util v0.7.18
    ├── bytes v1.11.0
    ├── futures-core v0.3.31
    ├── futures-sink v0.3.31
    ├── futures-util v0.3.31 (*)
    ├── pin-project-lite v0.2.16
    └── tokio v1.49.0 (*)
```

when running checks this can easily be fixed with either cargo hack or by specifying the manifest-path, but when running tests this is a bit hard to handle. To make this work in the CI, I patched the dependency to be a path only when running tests on tokio since the other dependencies don't have this problem.

I’m not sure if there are other solutions to make this work (I guess breaking the cycle), I see that there are many open issues on cargo related to this.

Darksonn: What you described is not a dependency cycle. Tokio's tests are considered a separate crate from tokio itself, so you have the dependency change `tokio (tests)` -> `tokio-util` -> `tokio`, which is not cyclic.

mattiapitossi: The only way I found to fix the dependency change issue is to patch the tests crate with '--config patch.crates-io.tokio.path="tokio”’ in the CI.

The root cause of the issue seems to be in some tests where we use both tokio and tokio-test crates, one indeed coming from local crate and the other one coming from crates.io:

cd tokio && cargo test —features full,test-util leads to 
```
  ::: tokio/tests/io_read_to_string.rs:5:5
   |
 5 | use tokio::io::AsyncReadExt;
   |     ----- one version of crate `tokio` used here, as a direct dependency of the current crate
 6 | use tokio_test::assert_ok;
   |     ---------- one version of crate `tokio` used here, as a dependency of crate `tokio_test`
```

So patching this to be local, solves the issue so you don’t have different sources.

Even if tokio/tests is considered a separate crate I’m not aware if there’s a global patch we can apply to the workspace to the specific crate tests (unless it becomes a separate crate with its own Cargo.toml).

Would you consider the approach I used in the CI acceptable, or maybe you had something different in mind?

Darksonn: Hmm, ok.

This is just a suggestion, but would it be easier to add the [patch] entry to our top-level `Cargo.toml`, and then have the few CI steps that don't want it remove it, e.g. by having a step that modifies the `Cargo.toml` file, or by having a second `Cargo.toml` and having those CI jobs swap the tomls.

mattiapitossi: I tested the patch/CI modifies approach and I think it's better for this case since my approach had the trade-off that running some checks or tests in tokio wouldn't be possible anymore unless you changed the configs locally.

The main idea is that when we run tests on tokio (and other crates) we use the patch defined in the workspace, then we perform another test by removing the patch from the workspace using sed and test every crate without path dependency (unless defined differently in the crate itself) except for Tokio since it won't work without the patch.

One note: I had to use the path dependency between tokio and tokio-macros because #7866 introduced some code that is not released yet on Tokio macros and this was causing some failures in the CI. So, when the new version of tokio-macros is released we can remove the path dependency from Tokio.

Everything should be ready now, CI checks are passing.

mattiapitossi:
I believe all review comments are now resolved, please let me know if I missed anything

.github/workflows/ci.yml
```diff
@@ -112,6 +112,14 @@ jobs:
...
           cargo nextest run --workspace --features $TOKIO_STABLE_FEATURES
+          # Removing workspace patches to run tests without path dependencies
+          # (if not specified differently in the crate)
+          sed -i.bak '/^\[patch\.crates-io\]$/d; /^tokio = { path = "tokio"}$/d' Cargo.toml && rm -f Cargo.toml.bak
+          cargo nextest run \
+            --workspace \
+            --exclude tokio \
+            --exclude examples \
+            --features $TOKIO_STABLE_FEATURES
           cargo test --doc --workspace --features $TOKIO_STABLE_FEATURES
```
Darksonn, line 125: Why does this introduce an extra `nextest run` invocation when there is already a `cargo test` invocation? Please document the reason with a comment (or remove if mistake).

.github/workflows/ci.yml
```diff
@@ -146,6 +154,16 @@ jobs:
...
+          # Removing the tokio workspace patch to run tests without path dependencies
+          # (if not specified differently in the crate)
+          sed -i.bak '/^\[patch\.crates-io\]$/d; /^tokio = { path = "tokio"}$/d' Cargo.toml && rm -f Cargo.toml.bak
+          RUSTFLAGS="$RUSTFLAGS -C panic=abort -Zpanic-abort-tests" cargo nextest run \
+            --workspace \
+            --exclude tokio \
+            --exclude tokio-macros \
+            --exclude tests-build \
+            --features $TOKIO_STABLE_FEATURES \
+            --tests
```
Darksonn: I see things are tested twice here, though this time it's with the `[patch]` removed the second time.

Although it's good to have a job doing this at least once somewhere, I don't think we need to test both with and without `[patch]` for every single CI job. It's fine for most CI jobs to not carry out the `sed` removal.

.github/workflows/ci.yml
```diff
@@ -728,7 +760,7 @@ jobs:
 
             cargo check -p tokio --features $TOKIO_STABLE_FEATURES
             # Other crates doesn't have unstable features, so we can use --all-features.
-            cargo check -p tokio-macros -p tokio-stream -p tokio-util -p tokio-test --all-features
+            cargo hack check -p tokio-macros -p tokio-stream -p tokio-util -p tokio-test --all-features
```
Darksonn, line 751: What is the effect of using `cargo hack` here?

.github/workflows/ci.yml
```diff
@@ -1061,7 +1105,8 @@ jobs:
           RUSTDOCFLAGS: -C link-args=--max-memory=67108864
 
       - name: WASI test tokio-stream
-        run: cargo test -p tokio-stream --target ${{ matrix.target }} --features time,net,io-util,sync
+        run: cargo test --manifest-path=tokio-stream/Cargo.toml --target ${{ matrix.target }} --features time,net,io-util,sync
+
```
Darksonn: Spurious added newline.

docs/contributing/how-to-specify-crates-dependencies-versions.md
```diff
@@ -0,0 +1,31 @@
+# How to specify crates dependencies versions
+
+Each crate (e.g., `tokio-util`, `tokio-stream`, etc.) should specify dependencies
+according to these rules:
+
+1. The listed version should be the oldest version that the crate works with
+(e.g., if `tokio-util` works with `tokio` version `1.44` but not `1.43`, then
+`tokio-util` should specify version `1.44` for its `tokio` dependency.)
+We don't require users to use the latest version unnecessarily).
```
Darksonn: There is only one opening brace, but two closing braces here.

mattiapitossi: fixed thanks!

.github/workflows/ci.yml
mattiapitossi: removed

.github/workflows/ci.yml
mattiapitossi: In this case, running cargo hack is the same as running `cargo check --manifest-path=..` for each crate, since now we have both path and crates.io in the project with -p there's an ambiguity

.github/workflows/ci.yml
mattiapitossi: cargo test runs only docs tests in this case, it seems that nextest crate does not support running them so we need also this step. I added a comment linking the related issue, see https://github.com/tokio-rs/tokio/pull/7848/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR123

.github/workflows/ci.yml
mattiapitossi: As for the tests, I only left one CI step that removes the patches as I agree it's not needed to do it multiple times. For cargo check and clippy, I think it might still be useful to do it since it shouldn't be as heavy as running tests multiple times, and with tests only we might miss some dependencies that are included during the full check.  Let me know if this make sense to you.

.github/workflows/ci.yml
```diff
@@ -112,6 +112,16 @@ jobs:
         run: |
           set -euxo pipefail
           cargo nextest run --workspace --features $TOKIO_STABLE_FEATURES
+          # Removing workspace patches to run tests without path dependencies
+          # (if not specified differently in the crate)
+          sed -i.bak '/^\[patch\.crates-io\]$/d; /^tokio = { path = "tokio"}$/d' Cargo.toml && rm -f Cargo.toml.bak
```
Darksonn:
```suggestion
          sed -i '/^\[patch\.crates-io\]$/d; /^tokio = { path = "tokio"}$/d' Cargo.toml
```

Cargo.toml
```diff
@@ -15,6 +15,9 @@ members = [
   "tests-integration",
 ]
 
+[patch.crates-io]
+tokio = { path = "tokio"}
+
```
Darksonn, line 20: I believe also tokio-stream depends on tokio-util, so shouldn't tokio-util be mentioned here too?

.github/workflows/ci.yml
mattiapitossi: I used .bak here because this step runs also on MacOS, which does not support `sed -i`. If you try to run this without it you will get an error.

Cargo.toml
mattiapitossi: tokio-stream seems to works fine with all the features even without the patch, I think there are no ambiguity across the crate

.github/workflows/ci.yml
Darksonn: Lol what, ok.

martin-g:
I'll take a look now!

.github/workflows/ci.yml
martin-g: I think the workaround for this is `-i''` but I cannot verify it at the moment. 
Using `.bak` is OK too!

.github/workflows/ci.yml
```diff
@@ -716,6 +732,8 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: ${{ env.rust_min }}
+      - name: Install cargo-hack
+        uses: taiki-e/install-action@cargo-hack
```
martin-g:
```
- name: Install cargo-hack
  uses: taiki-e/install-action@v2
  with:
      tool: cargo-hack
```
as done below

.github/workflows/ci.yml
martin-g:
```suggestion
          sed -i.bak '/^\[patch\.crates-io\]$/d; /^tokio = { path = "tokio"\s*}$/d' Cargo.toml && rm -f Cargo.toml.bak
```
Maybe it is just me but I'd add a space character before the closing `}` to make it symmetric with the opening curly brace and this will break the sed removal here.
```diff
- tokio = { path = "tokio"}
+ tokio = { path = "tokio" }
```
So, let's use `\s*` to make it more robust.

docs/contributing/how-to-specify-crates-dependencies-versions.md
```diff
@@ -0,0 +1,31 @@
...
+during the release of the crate, this ensures that it can't be released until
+the dependency has a new version.
+
+Consider the following example from `tokio-stream`:
+
+```toml
+[dependencies]
+futures-core = { version = "0.3.0" }
+pin-project-lite = "0.2.11"
+tokio = { version = "1.15.0", path = "../tokio", features = ["sync"] }
```
martin-g:
```suggestion
tokio = { path = "../tokio", features = ["sync"] }
```

IMO we could use just `path = "../tokio"`, without `version = ...` until the next release is cut. The release manager will remove `path =` and add `version = "x.y.z"`, where `x.y.z` is the currently released version.

tokio-stream/Cargo.toml
```diff
@@ -38,14 +38,14 @@ signal = ["tokio/signal"]
...
 futures-core = { version = "0.3.0" }
 pin-project-lite = "0.2.11"
-tokio = { version = "1.38.0", path = "../tokio", features = ["sync"] }
-tokio-util = { version = "0.7.0", path = "../tokio-util", optional = true }
+tokio = { version = "1.38.0", features = ["sync"] }
+tokio-util = { version = "0.7.0", optional = true }
 
 [dev-dependencies]
-tokio = { version = "1.2.0", path = "../tokio", features = ["full", "test-util"] }
+tokio = { version = "1.2.0", features = ["full", "test-util"] }
```
martin-g:
```suggestion
tokio = { version = "1.38.0", features = ["full", "test-util"] }
```
let's sync the `dep` with the `dev-dep` version, since Cargo resolves 1.38.0 anyway

Cargo.toml
martin-g: Why exactly do we need the patch here ?
The new `how-to-specify-crates-dependencies.md` suggests to use a local `path = "../tokio"` when needed to use a non-released functionality

mattiapitossi: This is mostly needed because tokio tests depend on both tokio and tokio-util, one being local and the other coming from crates.io. In this context, you end up in having some conflicts because of two difference sources of tokio.

In order to avoid in having such conflicts, I added the patch which is going to be removed during CI checks. If a crates specifies the path (which means that depends on unreleased changes) then nothing changes, while if it does not specify the path as described in the contributing the CI will checks against the crates.io version.

(see also for additional context: https://github.com/tokio-rs/tokio/pull/7848#discussion_r2715989917)

.github/workflows/ci.yml
mattiapitossi: Agree, thanks! updated

.github/workflows/ci.yml
mattiapitossi: Thanks, updated! Actually I realized that the other one was a leftover from a previous check that I removed, so it's no longer needed

mattiapitossi:
Thanks @martin-g, I believe all suggestions are resolved

PR #7849, opened 2026-01-09
Revert "fs: avoid lock acquisition after uring init"

Darksonn:
Please don't add new dependencies without a very strong justification. In the case of once_cell, this is actually a crate we have removed previously when it bumped its MSRV and broke ours.

Reverts tokio-rs/tokio#7843

cc @tahmid-23

### DIFF
--- a/tokio/Cargo.toml
+++ b/tokio/Cargo.toml
@@ -85,7 +85,7 @@ sync = []
 test-util = ["rt", "sync", "time"]
 time = []
 # Unstable feature. Requires `--cfg tokio_unstable` to enable.
-io-uring = ["dep:io-uring", "libc", "mio/os-poll", "mio/os-ext", "dep:slab", "once_cell"]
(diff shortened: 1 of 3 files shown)


tahmid-23:
no leniency for unstable dep? :slightly_frowning_face: unfortunate that other solutions are uglier

Darksonn:
No, sorry. But you can reintroduce our custom `OnceCell` implementation from #5048 / #7208.

tahmid-23:
it's insufficient without reimplementing OnceState, requires a private poisoning method.

Darksonn:
Could you use our async implementation from `tokio::sync::OnceCell`? It doesn't seem like it would be an issue to make `check_and_init` into an async function. If you need to block on init in the IO driver, you can use `crate::future::block_on` for those cases (but my guess is that the non-async `get` is sufficient outside of the checks in async fns in `tokio::fs`).

PR #7850, opened 2026-01-09
runtime: avoid lock acquisition after uring init (via async OnceCell)

tahmid-23:
#7843, again, but with tokio's async OnceCell instead of once_cell's OnceCell. Thanks for suggestion.

Notably, `register_op` tries to `check_and_init`. I didn't want to invasively change the code to polling (can I even do it without boxing?) so I changed it to an assert, since `check_and_init` should generally always be called before submission.

cc @ADD-SP @Darksonn

### DIFF
--- a/tokio/src/runtime/io/driver/uring.rs
+++ b/tokio/src/runtime/io/driver/uring.rs
@@ -175,7 +175,7 @@ impl Handle {
     ///
     /// TODO: This would like to be a synchronous function,
     /// but we require `OnceLock::get_or_try_init`.
-    /// https://github.com/rust-lang/rust/issues/109737
+    /// <https://github.com/rust-lang/rust/issues/109737>
     pub(crate) async fn check_and_init(&self, opcode: u8) -> io::Result<bool> {
         let probe = self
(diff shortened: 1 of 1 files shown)


tokio/src/runtime/io/driver/uring.rs
```diff
@@ -255,10 +221,7 @@ impl Handle {
     /// Callers must ensure that parameters of the entry (such as buffer) are valid and will
     /// be valid for the entire duration of the operation, otherwise it may cause memory problems.
     pub(crate) unsafe fn register_op(&self, entry: Entry, waker: Waker) -> io::Result<usize> {
-        // Note: Maybe this check can be removed if upstream callers consistently use `check_and_init`.
-        if !self.check_and_init(entry.get_opcode() as u8)? {
-            return Err(io::Error::from_raw_os_error(libc::ENOSYS));
-        }
+        assert!(self.uring_probe.initialized());
```
Darksonn, line 224: If you want to preserve the opcode check, you can access the `OnceCell` without async using e.g. `get`.

PR #7851, opened 2026-01-10
runtime: add comments to `schedule_option_task_without_yield`

IgorErin:
## Motivation

Strange, magical function, seems like manual hack.

## Solution

Add comment about LLVM codegen reduction.

### DIFF
--- a/tokio/src/runtime/scheduler/multi_thread/worker.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/worker.rs
@@ -1280,6 +1280,7 @@ impl Handle {
         });
     }
 
+    // Separated case to reduce LLVM codegen in `Handle::bind_new_task`.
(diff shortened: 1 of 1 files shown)


ADD-SP:
The `schedule_option_task_without_yield` is for reducing the LLVM code generation (see <https://github.com/tokio-rs/tokio/pull/5859>), I prefer to keep it.

However, it would be better adding an simple comment on `schedule_option_task_without_yield` to explain "why".

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -1280,6 +1280,7 @@ impl Handle {
         });
     }
 
+    // Separated to reduce LLVM codegen.
```
ADD-SP: Hmm, this is too simple to understand. We could add more context here.

IgorErin:
is that better?

Issue #7292, opened 2025-04-29
IntervalStream does not implement FusedStream


leif-hesperides:
Why is there no implementation of FusedStream for IntervalStream? It seems like we could have `is_terminated(self) { false }`.

stormshield-fabs:
This was fixed in https://github.com/tokio-rs/tokio/pull/7854 and can be closed.

PR #7854, opened 2026-01-13
time: implement FusedStream for IntervalStream

robcholz:
## Motivation

See #7292 

## Solution

implement FusedStream for IntervalStream

### DIFF
--- a/tokio-stream/tests/stream_fuse.rs
+++ b/tokio-stream/tests/stream_fuse.rs
@@ -50,6 +50,7 @@ async fn basic_usage() {
 }
 
 #[tokio::test]
+#[cfg(feature = "time")]
 async fn interval_stream_is_never_terminated() {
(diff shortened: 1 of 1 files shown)


tokio-stream/tests/stream_fuse.rs
```diff
@@ -48,3 +48,14 @@ async fn basic_usage() {
     assert_eq!(stream.next().await, None);
     assert_eq!(stream.size_hint(), (0, Some(0)));
 }
+
+#[tokio::test]
```
martin-g, line 52:
```suggestion
#[tokio::test]
#[cfg(feature = "time")]
```

IntervalStream is available only if the `time` feature is enabled
https://github.com/robcholz/tokio/blob/7a2135f426ab560a9efb320620f80e1935c43d31/tokio-stream/src/wrappers.rs#L36-L39

PR #7855, opened 2026-01-13
Add AbortOnDrop complement type to AbortOnDropHandle

ultrabear:
## Motivation

`AbortOnDropHandle<T>` exists for cases where you wish to "forward cancel" a `JoinHandle<T>`, but frequently in code I find myself writing a wrapper around `AbortHandle`. This is for the same reasons someone would use `AbortHandle` (no generic param to carry, can hold multiple in different places, don't need the return value).

## Solution

This PR adds a complement to `AbortOnDropHandle<T>`; `AbortOnDrop`. This simply wraps `AbortHandle` and otherwise mimics the API/function of `AbortOnDropHandle<T>`. The name is perhaps unfortunate and open to bikeshedding, but I couldn't come up with anything that felt great considering that `AbortOnDropHandle` is already taken.

### DIFF
--- a/tokio-util/src/task/abort_on_drop.rs
+++ b/tokio-util/src/task/abort_on_drop.rs
@@ -87,11 +87,12 @@ impl<T> AsRef<JoinHandle<T>> for AbortOnDropHandle<T> {
 /// A wrapper around a [`tokio::task::AbortHandle`],
 /// which [aborts] the task when it is dropped.
 ///
-/// Unlike [`AbortOnDropHandle`], [`AbortOnDrop`] does not represent the permission to await a
-/// tasks completion, and in exchange can be a concrete type.
+/// Unlike [`AbortOnDropHandle`], [`AbortOnDrop`] cannot be `.await`ed for a result.
 ///
+/// It has no generic parameter, making it suitable when you only need to keep
+/// a task handle in a struct and do not care about the output.
 ///
-/// [aborts]: tokio::task::JoinHandle::abort
+/// [aborts]: tokio::task::AbortHandle::abort
 #[must_use = "Dropping the handle aborts the task immediately"]
 pub struct AbortOnDrop(AbortHandle);
 
--- a/tokio-util/src/task/mod.rs
+++ b/tokio-util/src/task/mod.rs
@@ -12,7 +12,7 @@ cfg_rt! {
     pub use task_tracker::TaskTracker;
 
     mod abort_on_drop;
-    pub use abort_on_drop::{AbortOnDropHandle, AbortOnDrop};
+    pub use abort_on_drop::{AbortOnDrop, AbortOnDropHandle};
 
     mod join_queue;
     pub use join_queue::JoinQueue;
--- a/tokio-util/tests/abort_on_drop.rs
+++ b/tokio-util/tests/abort_on_drop.rs
@@ -70,7 +70,7 @@ async fn handle_does_not_abort_after_detach() {
         let _ = rx.await;
     });
     let handle = AbortOnDrop::new(handle.abort_handle());
-    handle.detach(); // returns and drops the original join handle
+    handle.detach(); // returns and drops the original abort handle
     yield_now().await;
     assert!(!tx.is_closed()); // the task is still alive
 }


ADD-SP (commented):
Hmm, the name `AbortOnDrop` is very similar to the existing `AbortOnDropHandle` and is likely to cause confusion for downstream users. Do you have better idea?

tokio-util/tests/abort_on_drop.rs
```diff
@@ -37,3 +37,40 @@ async fn does_not_abort_after_detach() {
...
+#[tokio::test]
+async fn handle_does_not_abort_after_detach() {
+    let (tx, rx) = oneshot::channel::<bool>();
+    let handle = tokio::spawn(async move {
+        let _ = rx.await;
+    });
+    let handle = AbortOnDrop::new(handle.abort_handle());
+    handle.detach(); // returns and drops the original join handle
+    yield_now().await;
+    assert!(!tx.is_closed()); // task is still live
```
martin-g:
```suggestion
    assert!(!tx.is_closed()); // the task is still alive
```

tokio-util/src/task/mod.rs
```diff
@@ -12,7 +12,7 @@ cfg_rt! {
     pub use task_tracker::TaskTracker;
 
     mod abort_on_drop;
-    pub use abort_on_drop::AbortOnDropHandle;
+    pub use abort_on_drop::{AbortOnDropHandle, AbortOnDrop};
```
martin-g:
```suggestion
    pub use abort_on_drop::{AbortOnDrop, AbortOnDropHandle};
```
nit: just to keep the imports order consistent with https://github.com/tokio-rs/tokio/pull/7855/changes#diff-25762358a8cae400f354bde1b017ea6e726ce1d62b07c1e288d9c278abbefc00R2

tokio-util/src/task/abort_on_drop.rs
```diff
@@ -79,6 +82,59 @@ impl<T> AsRef<JoinHandle<T>> for AbortOnDropHandle<T> {
...
+/// tasks completion, and in exchange can be a concrete type.
+///
+///
+/// [aborts]: tokio::task::JoinHandle::abort
+#[must_use = "Dropping the handle aborts the task immediately"]
+pub struct AbortOnDrop(AbortHandle);
+
+impl Drop for AbortOnDrop {
+    fn drop(&mut self) {
+        self.0.abort()
```
martin-g:
```suggestion
        self.abort()
```

tokio-util/src/task/abort_on_drop.rs
```diff
@@ -79,6 +82,59 @@ impl<T> AsRef<JoinHandle<T>> for AbortOnDropHandle<T> {
...
+
+impl AbortOnDrop {
+    /// Create an [`AbortOnDrop`] from a [`AbortHandle`].
+    pub fn new(handle: AbortHandle) -> Self {
+        Self(handle)
+    }
+
+    /// Abort the task associated with this handle,
+    /// equivalent to [`AbortHandle::abort`].
+    pub fn abort(&self) {
```
martin-g, line 114:
```suggestion
    #[inline]
    pub fn abort(&self) {
```

tokio-util/src/task/abort_on_drop.rs
```diff
@@ -79,6 +82,59 @@ impl<T> AsRef<JoinHandle<T>> for AbortOnDropHandle<T> {
...
+        self.0.abort()
+    }
+
+    /// Checks if the task associated with this handle is finished,
+    /// equivalent to [`AbortHandle::is_finished`].
+    pub fn is_finished(&self) -> bool {
```
martin-g, line 121:
```suggestion
    #[inline]
    pub fn is_finished(&self) -> bool {
```

tokio-util/src/task/abort_on_drop.rs
```diff
@@ -79,6 +82,59 @@ impl<T> AsRef<JoinHandle<T>> for AbortOnDropHandle<T> {
     }
 }
 
+/// A wrapper around a [`tokio::task::AbortHandle`],
+/// which [aborts] the task when it is dropped.
+///
+/// Unlike [`AbortOnDropHandle`], [`AbortOnDrop`] does not represent the permission to await a
+/// tasks completion, and in exchange can be a concrete type.
```
martin-g:
```suggestion
/// task's completion, and in exchange can be a concrete type.
```

tokio-util/src/task/abort_on_drop.rs
```diff
@@ -79,6 +82,59 @@ impl<T> AsRef<JoinHandle<T>> for AbortOnDropHandle<T> {
...
+///
```
martin-g, line 94:
```suggestion
```

tokio-util/src/task/abort_on_drop.rs
```diff
@@ -79,6 +82,59 @@ impl<T> AsRef<JoinHandle<T>> for AbortOnDropHandle<T> {
...
+///
+/// [aborts]: tokio::task::JoinHandle::abort
```
martin-g:
```suggestion
/// [aborts]: tokio::task::AbortHandle::abort
```

tokio-util/tests/abort_on_drop.rs
```diff
@@ -37,3 +37,40 @@ async fn does_not_abort_after_detach() {
...
+}
+
+#[tokio::test]
+async fn handle_does_not_abort_after_detach() {
+    let (tx, rx) = oneshot::channel::<bool>();
+    let handle = tokio::spawn(async move {
+        let _ = rx.await;
+    });
+    let handle = AbortOnDrop::new(handle.abort_handle());
+    handle.detach(); // returns and drops the original join handle
```
martin-g:
```suggestion
    handle.detach(); // returns and drops the original abort handle
```

conradludgate:
> Hmm, the name `AbortOnDrop` is very similar to the existing `AbortOnDropHandle` and is likely to cause confusion for downstream users. Do you have better idea?

Maybe we can do
```rust
#[deprecated]
type AbortOnDropHandle<T> = AbortOnDropJoinHandle<T>;

struct AbortOnDropJoinHandle<T> { ... };
struct AbortOnDropAbortHandle<T> { ... };
```

Darksonn:
I'm not too worried about the name confusion actually. It seems like even if the two get confused, if the code compiles, then it will still do the right thing.

tokio-util/src/task/abort_on_drop.rs
ultrabear: While I'm not opposed to this I want to point out that the AbortOnDropHandle type uses an identical construction (as this was framed as a copy paste), so I'd like reconsideration here 
(that ends in;
 "yes still change this the inconsistency is ok", 
 "no don't change this to keep consistency", or 
 "yes still change this *and also* change AbortOnDropHandle's impl to match"
)

This applies to the remaining suggestions, I will reference this comment in reply to those

tokio-util/src/task/abort_on_drop.rs
ultrabear: See reply from `self.0.abort() -> self.abort()`

tokio-util/src/task/abort_on_drop.rs
ultrabear: See reply from `self.0.abort() -> self.abort()`

tokio-util/tests/abort_on_drop.rs
ultrabear: See reply from `self.0.abort() -> self.abort()`

ultrabear (commented):
Want to make sure codestyle consistency is considered before proceeding with the rest of the proposed changes

Darksonn:
Changing `AbortOnDropHandle` to match sounds good to me.

tokio-util/src/task/abort_on_drop.rs
```diff
@@ -79,6 +82,58 @@ impl<T> AsRef<JoinHandle<T>> for AbortOnDropHandle<T> {
...
+        self.0.is_finished()
+    }
+
+    /// Cancels aborting on drop and returns the original [`AbortHandle`].
+    pub fn detach(self) -> AbortHandle {
+        // Avoid invoking `AbortOnDrop`'s `Drop` impl
+        let this = ManuallyDrop::new(self);
+        // SAFETY: `&this.0` is a reference, so it is certainly initialized, and
+        // it won't be double-dropped because it's in a `ManuallyDrop`
+        unsafe { std::ptr::read(&this.0) }
```
ADD-SP, line 131: To minimize the unsafe scope, would using `this.0.clone()` be a better option?

tokio-util/src/task/abort_on_drop.rs
```diff
@@ -79,6 +82,58 @@ impl<T> AsRef<JoinHandle<T>> for AbortOnDropHandle<T> {
     }
 }
 
+/// A wrapper around a [`tokio::task::AbortHandle`],
+/// which [aborts] the task when it is dropped.
+///
+/// Unlike [`AbortOnDropHandle`], [`AbortOnDrop`] does not represent the permission to await a
+/// task's completion, and in exchange can be a concrete type.
```
ADD-SP:
```suggestion
/// Unlike [`AbortOnDropHandle`], [`AbortOnDrop`] cannot be `.await`ed for a result.
///
/// It has no generic parameter, making it suitable when you only need to keep
/// a task handle in a struct and do not care about the output.
```

It would be better to share the usecase to the downstream user.

tokio-util/src/task/abort_on_drop.rs
Darksonn: Using `clone` here leads to leaking a refcount, so we can't do that.

ultrabear:
> Changing `AbortOnDropHandle` to match sounds good to me.

I acknowledge reading this, but as its not from the original reviewer (or a project maintainer) I did not want to move forward with it until receiving further confirmation, it appears the latest review is moving forward with option 2 of keeping consistency with existing code. 

However, the latest review is not from the same author as the initial review, so from my limited understanding of tokio's review process I am in limbo on what the correct path here is

Going to assume that option 2 has been picked for now

Darksonn:
Sorry which options is 1 and which is 2?

@ADD-SP and I are both Tokio maintainers.

ultrabear:
> Sorry which options is 1 and which is 2?
> 
> @ADD-SP and I are both Tokio maintainers.

referencing:
```
 "yes still change this the inconsistency is ok",
"no don't change this to keep consistency", or
"yes still change this and also change AbortOnDropHandle's impl to match"
```
from the comments i left from the first review

Also, did not know you were a maintainer, sorry, I have been going off of the github badges where you are listed as a "contributor", not "member"
will move forward with option 3 in that case

Darksonn:
Huh, apparently it is set to private for some reason:

image

I did not know that was even an option.

ultrabear:
Applied the last of the feedback that I'm aware of, please let me know if I missed a spot :)

PR #7856, opened 2026-01-14
(macros): improve error message for return type mismatch in #[tokio::main]

### LINKED ISSUE #7784: Error message is inaccurate in tokio::main macro
**Version**
tokio v1.48.0

**Platform**
Win11(x64)

**Description**
The error message is inaccurate.

I tried this code:

```rust
#[tokio::main]
async fn main() -> () {
    if true {
        return 1;
    }
    println!("Hello, world!");
    return ();
}
```

I expected to see this happen: Report errors accurately.

Instead, this happened: 

```
E:\Projects\projects_rust\test_tokio>cargo build
   Compiling test_tokio v0.1.0 (E:\Projects\projects_rust\test_tokio)
 --> src\main.rs:7:12
  |
7 |     return ();
  |            ^^ expected integer, found `()`
  |
note: return type inferred to be `{integer}` here
 --> src\main.rs:4:16
  |
4 |         return 1;
  |                ^

 --> src\main.rs:7:5
  |
2 | async fn main() -> () {
  |                    -- expected `()` because of return type
...
7 |     return ();
  |     ^^^^^^^^^^ expected `()`, found integer

error: could not compile `test_tokio` (bin "test_tokio") due to 2 previous errors

E:\Projects\projects_rust\test_tokio>
```

daschinmoy21:
### Motivation
Fixes #7784
When using `#[tokio::main]` with conflicting return statements, the error message was misleading:
```rust
#[tokio::main]
async fn main() -> () {
    if true {
        return 1;
    }
    return ();
}
```
The compiler would report:
```rust
 --> src/main.rs:7:12
  |
7 |     return ();
  |            ^^ expected integer, found `()`
```
This blames return () (which matches the declared -> ()) instead of return 1.

### Solution
Added a check_output identity function that explicitly constrains the async block's output type to match the declared function return type:

```rust
fn check_output<F: Future<Output = #output_type>>(f: F) -> F { f }
check_output(body)
```
This ensures the compiler reports type mismatches against the declared return type rather than an inferred type. The function has zero runtime overhead since it just returns its input unchanged.

After this change, the error is clearer:

```rust
 --> src/main.rs:6:12
  |
6 |     return ();
  |            ^^ expected integer, found `()`
  |
note: return type inferred to be `{integer}` here
 --> src/main.rs:4:16
  |
4 |         return 1;
  |                ^

error[E0271]: expected `{async block@src/main.rs:1:1: 1:15}` to be a future that resolves to `()`, but it resolves to `{integer}`
 --> src/main.rs:1:1
  |
1 | #[tokio::main]
  | ^^^^^^^^^^^^^^ expected `()`, found integer
  |
note: required by a bound in `check_output`
```

### DIFF
--- a/tokio/src/time/clock.rs
+++ b/tokio/src/time/clock.rs
@@ -124,8 +124,39 @@ cfg_test_util! {
     /// other timer-backed primitives can cause the runtime to advance the
     /// current time when awaited.
     ///
+    /// # Preventing auto-advance
+    ///
+    /// In some testing scenarios, you may want to keep the clock paused without
+    /// auto-advancing, even while waiting for I/O or other asynchronous operations.
+    /// This can be achieved by using [`spawn_blocking`] to wrap your I/O operations.
+    ///
+    /// When a blocking task is running, the clock's auto-advance is temporarily
+    /// inhibited. This allows you to wait for I/O to complete while keeping the
+    /// paused clock stationary:
+    ///
+    /// ```ignore
+    /// use tokio::time::{Duration, Instant};
+    /// use tokio::task;
+    ///
+    /// #[tokio::test(start_paused = true)]
+    /// async fn test_with_io() {
+    ///     let start = Instant::now();
+    ///
+    ///     // The clock will NOT auto-advance while this blocking task runs
+    ///     let result = task::spawn_blocking(|| {
+    ///         // Perform I/O operations here
+    ///         std::thread::sleep(std::time::Duration::from_millis(10));
+    ///         42
+    ///     }).await.unwrap();
+    ///
+    ///     // Time has not advanced
+    ///     assert_eq!(start.elapsed(), Duration::ZERO);
+    /// }
+    /// ```
+    ///
     /// [`Sleep`]: crate::time::Sleep
     /// [`advance`]: crate::time::advance
+    /// [`spawn_blocking`]: crate::task::spawn_blocking
     #[track_caller]
     pub fn pause() {
         with_clock(|maybe_clock| {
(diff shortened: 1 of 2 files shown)


taiki-e:
As said in https://github.com/tokio-rs/tokio/issues/6306#issuecomment-2027319790 (and https://github.com/tokio-rs/tokio/pull/3039#issuecomment-716039701 and https://github.com/tokio-rs/tokio/pull/5619#issuecomment-1509579919 and https://github.com/tokio-rs/tokio/pull/6438#issuecomment-2027199138), this needs to handle cases where the impl trait is included in Output (`tokio::main` can be used for functions other than the `main` function). Alternatively, for now, you could restrict its use to functions that return a unit.

(I've mentioned this point repeatedly, so we should probably add a test case like the following.)

```rs
#[tokio::main]
async fn never() -> ! {
    loop {}
}
#[tokio::main]
async fn impl_trait() -> impl Iterator<Item = impl core::fmt::Debug> {
    [()].into_iter()
}
```

daschinmoy21:
Updated the PR as requested by @taiki-e 
All checks pass.including test cases added in #7857 ,thanks to @Darksonn

tokio-macros/src/entry.rs
```diff
@@ -494,22 +494,42 @@ fn parse_knobs(mut input: ItemFn, is_test: bool, config: FinalConfig) -> TokenSt
...
         quote! {
             let body = async #body;
             #crate_path::pin!(body);
             let body: ::core::pin::Pin<&mut dyn ::core::future::Future<Output = #output_type>> = body;
         }
     } else {
+        // force typecheck without runtime overhead
+        let check_block = match &input.sig.output {
+            syn::ReturnType::Type(_, t)
+                if matches!(**t, syn::Type::Never(_) | syn::Type::ImplTrait(_)) =>
```
taiki-e: Sorry, my test case above wasn't sufficient. impl trait can also be placed inside other types, so this check isn't sufficient.
```rs
#[tokio::main]
async fn impl_trait2() -> Result<(), impl core::fmt::Debug> {
    Err(())
}
```

daschinmoy21: oh ok i will add this to the test cases and also modify my code.

daschinmoy21:
Added another test case where impl trait can also be placed inside other types, so previous check wasn't sufficient.
```rs
#[tokio::main]
async fn impl_trait2() -> Result<(), impl core::fmt::Debug> {
    Err(())
}
```

updated code and test case to cover this

daschinmoy21:
any update on this @taiki-e ?

PR #7857, opened 2026-01-15
Add `#[tokio::main]` test for `impl Trait` and `!` return types

Darksonn:
See https://github.com/tokio-rs/tokio/pull/7856#issuecomment-3753356639


taiki-e (approved):
Thanks!

PR #7858, opened 2026-01-15
docs(time): enhance time manipulation documentation

### LINKED ISSUE #7841: [Docs] Enhancements to time manipulation docs
Pulling over some doc update requests from discord.

Specifically:
- Document how to avoid advancing a paused clock even if waiting on i/o
- Document pitfalls of `tokio::time::advance`

### Avoiding advancing a paused clock:

arielby@:
> I want a tokio::time::stand_still that makes a paused clock not advance even if the runtime is waiting for some IO, maybe with a real-time timeout
> Currently I am using a yield_now spin loop for that which is unfortunate

Alice Rhys@:
> I think time stays still while there are spawn blocking tasks. Maybe you can abuse that

### `tokio::time::advance` footguns:

arielby@:
> What is the actual use case of tokio::time::advance other than being a chaos monkey?

Alice Rhys@:
> I think it's just broken. You should use sleep when you think you want advance.

arielby@:
> Should you give it a big box warning then ?
> ״simulates your program freezing for ‘duration’. Many programs have timeouts when they freeze”
> It’s almost but not quite reliable enough to force timeouts by “simulating freezing”

AndreaBozzo:
## Motivation

Closes #7841

Users have been discovering through trial-and-error that:
1. `spawn_blocking` prevents auto-advance of paused time (mentioned in Discord, documented behavior but not in public docs)
2. `time::advance` has pitfalls when used for forcing timeouts or simulating freezes (see [twilight-rs/twilight#2509](https://github.com/twilight-rs/twilight/pull/2509) which replaced `advance` with `sleep`)

This information should be documented to help users understand the correct patterns for time manipulation in tests.

## Solution

Added two new documentation sections:

1. **"Preventing auto-advance"** section in `pause()` docs:
   - Explains that `spawn_blocking` inhibits auto-advance while blocking tasks run
   - Provides an example showing how to wait for I/O without advancing time

2. **"When to use `sleep` instead"** section in `advance()` docs:
   - Warns about limitations when forcing timeouts (timers not guaranteed to be processed)
   - Explains the batch behavior issue when simulating freezes
   - Recommends using `sleep` with auto-advance for reliable timeout testing

### DIFF
--- a/tokio/src/time/clock.rs
+++ b/tokio/src/time/clock.rs
@@ -124,8 +124,39 @@ cfg_test_util! {
     /// other timer-backed primitives can cause the runtime to advance the
     /// current time when awaited.
     ///
+    /// # Preventing auto-advance
+    ///
+    /// In some testing scenarios, you may want to keep the clock paused without
+    /// auto-advancing, even while waiting for I/O or other asynchronous operations.
+    /// This can be achieved by using [`spawn_blocking`] to wrap your I/O operations.
+    ///
+    /// When a blocking task is running, the clock's auto-advance is temporarily
+    /// inhibited. This allows you to wait for I/O to complete while keeping the
+    /// paused clock stationary:
+    ///
+    /// ```ignore
+    /// use tokio::time::{Duration, Instant};
+    /// use tokio::task;
+    ///
+    /// #[tokio::test(start_paused = true)]
+    /// async fn test_with_io() {
+    ///     let start = Instant::now();
+    ///
+    ///     // The clock will NOT auto-advance while this blocking task runs
+    ///     let result = task::spawn_blocking(|| {
+    ///         // Perform I/O operations here
+    ///         std::thread::sleep(std::time::Duration::from_millis(10));
+    ///         42
+    ///     }).await.unwrap();
+    ///
+    ///     // Time has not advanced
+    ///     assert_eq!(start.elapsed(), Duration::ZERO);
+    /// }
+    /// ```
+    ///
     /// [`Sleep`]: crate::time::Sleep
     /// [`advance`]: crate::time::advance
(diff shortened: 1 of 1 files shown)


tokio/src/time/clock.rs
```diff
@@ -124,8 +124,39 @@ cfg_test_util! {
...
     ///
+    /// # Preventing auto-advance
+    ///
+    /// In some testing scenarios, you may want to keep the clock paused without
+    /// auto-advancing, even while waiting for I/O or other asynchronous operations.
+    /// This can be achieved by using [`spawn_blocking`] to wrap your I/O operations.
+    ///
+    /// When a blocking task is running, the clock's auto-advance is temporarily
+    /// inhibited. This allows you to wait for I/O to complete while keeping the
+    /// paused clock stationary:
```
ADD-SP, line 135: Hmm, I know it works for now, but it abuses the `spawn_blocking`. Since we don't have a better solution for this use case, I'm ok to add this, but I don't want to make this as a part of the semantic version guarantee.

Shall we add a statement to claim that this might be changed in the future without a major release? What do you think?

AndreaBozzo: The idea i had at first was to add a note stating that this behaviour Is being looked into and might change, but i wasn't sure 'couse it made the docstring more verbose.

Maybe a one-liner addition so we can track It  properly? Open to feedbacks on this

Darksonn: We may add a better way of doing it in the future, but I don't think we can take away the ability to use `spawn_blocking` like this. That'd be a breaking change.

ADD-SP (approved):

AndreaBozzo:
Thanks everyone!

PR #7860, opened 2026-01-16
docs: fix broken links of `select!`

mattiapitossi:
## Motivation

While working on #7848 I noticed that select! intra-links are broken across crates. See for example: https://docs.rs/tokio-stream/latest/tokio_stream/struct.StreamMap.html#method.next_many.

When building docs locally and with the CI it works when using paths due to feature unification across dependency targets. But this is not true when we publish crates since paths are removed and both `tokio-util` and `tokio-stream` do not have a dependency to "macros" features, therefore not resolving the link.

## Solution

As did with other macros, instead of adding the macros features, I added a direct link to select macro

[Rendered](https://deploy-preview-7860--tokio-rs.netlify.app/tokio/)

### DIFF
--- a/tokio-util/src/io/simplex.rs
+++ b/tokio-util/src/io/simplex.rs
@@ -172,7 +172,7 @@ impl AsyncRead for Receiver {
 ///
 /// See [`Sender::poll_shutdown`].
 ///
-/// [`tokio::select!`]: https://docs.rs/tokio/1.0/tokio/macro.select.html
+/// [`tokio::select!`]: https://docs.rs/tokio/latest/tokio/macro.select.html
 #[derive(Debug)]
 pub struct Sender {
     inner: Arc<Mutex<Inner>>,


tokio-stream/src/stream_map.rs
```diff
@@ -579,9 +579,11 @@ where
...
     ///
     /// This method is cancel safe. If `next_many` is used as the event in a
-    /// [`tokio::select!`](tokio::select) statement and some other branch
-    /// completes first, it is guaranteed that no items were received on any of
-    /// the underlying streams.
+    /// [`tokio::select!`] statement and some other branch completes first,
+    /// it is guaranteed that no items were received on any of the underlying
+    /// streams.
+    ///
+    /// [`tokio::select!`]: https://docs.rs/tokio/1.0/tokio/macro.select.html
```
martin-g:
```suggestion
    /// [`tokio::select!`]: https://docs.rs/tokio/latest/tokio/macro.select.html
```
Using `latest` will still work even when Tokio 2.x is released.

Using `1.0` has the benefit (?!) that it redirects to the latest released full version, e.g. `1.0.49` at the moment. Some people may find it a bit more user friendly.

tokio-util/src/io/simplex.rs
```diff
@@ -88,13 +88,15 @@ impl Inner {
...
-/// completes first, it is guaranteed that no bytes were received on this
+/// [`tokio::select!`] statement and some other branch completes
+/// first, it is guaranteed that no bytes were received on this
 /// channel.
 ///
 /// You can still read the remaining data from the buffer
 /// even if the write half has been dropped.
 /// See [`Sender::poll_shutdown`] and [`Sender::drop`] for more details.
+///
+/// [`tokio::select!`]: https://docs.rs/tokio/1.0/tokio/macro.select.html
```
martin-g:
```suggestion
/// [`tokio::select!`]: https://docs.rs/tokio/latest/tokio/macro.select.html
```

tokio-util/src/io/simplex.rs
```diff
@@ -163,13 +165,14 @@ impl AsyncRead for Receiver {
...
-/// completes first, it is guaranteed that no bytes were sent on this
-/// channel.
+/// [`tokio::select!`] statement and some other branch completes
+/// first, it is guaranteed that no bytes were sent on this channel.
 ///
 /// # Shutdown
 ///
 /// See [`Sender::poll_shutdown`].
+///
+/// [`tokio::select!`]: https://docs.rs/tokio/1.0/tokio/macro.select.html
```
martin-g:
```suggestion
/// [`tokio::select!`]: https://docs.rs/tokio/latest/tokio/macro.select.html
```

ADD-SP (approved):
Thanks!

PR #7863, opened 2026-01-18
test: add error case tests for fs::hard_link

kevinWangSheng:
## Motivation

  The `tokio::fs::hard_link` function currently lacks test coverage for error scenarios. The existing test only
  covers the success case, leaving error handling untested.

  ## Solution

  This PR adds three test cases to verify proper error handling:

  1. **test_hard_link_error_source_not_found** - Verifies that `NotFound` error is returned when the source file
  does not exist
  2. **test_hard_link_error_destination_already_exists** - Verifies that `AlreadyExists` error is returned when the
  destination file already exists
  3. **test_hard_link_error_source_is_directory** - Verifies that an error is returned when attempting to hard link
  a directory (which is not allowed on most systems)

  These tests improve coverage by ensuring `hard_link` properly handles common error scenarios and returns
  appropriate `io::ErrorKind` values.

  ## Testing

  All new tests pass:
  4. running 5 tests
  test test_hard_link ... ok
  test test_hard_link_error_destination_already_exists ... ok
  test test_hard_link_error_source_is_directory ... ok
  test test_hard_link_error_source_not_found ... ok
  test test_symlink ... ok

  4. test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

### DIFF
--- a/tokio/tests/fs_link.rs
+++ b/tokio/tests/fs_link.rs
@@ -68,10 +68,7 @@ async fn test_hard_link_error_source_not_found() {
     let src = dir.path().join("nonexistent.txt");
     let dst = dir.path().join("dst.txt");
 
-    let result = fs::hard_link(&src, &dst).await;
-
-    assert!(result.is_err());
-    let err = result.unwrap_err();
+    let err = fs::hard_link(&src, &dst).await.unwrap_err();
     assert_eq!(err.kind(), std::io::ErrorKind::NotFound);
 }
 
@@ -83,22 +80,13 @@ async fn test_hard_link_error_destination_already_exists() {
     let dst = dir.path().join("dst.txt");
 
     // Create source file
-    std::fs::File::create(&src)
-        .unwrap()
-        .write_all(b"source content")
-        .unwrap();
+    std::fs::write(&src, b"source content").unwrap();
 
     // Create destination file
-    std::fs::File::create(&dst)
-        .unwrap()
-        .write_all(b"destination content")
-        .unwrap();
+    std::fs::write(&dst, b"destination content").unwrap();
 
     // Attempt to create hard link when destination already exists
-    let result = fs::hard_link(&src, &dst).await;
-
-    assert!(result.is_err());
-    let err = result.unwrap_err();
+    let err = fs::hard_link(&src, &dst).await.unwrap_err();
     assert_eq!(err.kind(), std::io::ErrorKind::AlreadyExists);
 }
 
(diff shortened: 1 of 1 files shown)


tokio/tests/fs_link.rs
```diff
@@ -60,3 +60,62 @@ async fn test_symlink() {
...
+#[cfg_attr(miri, ignore)] // No `linkat` in miri.
+async fn test_hard_link_error_source_not_found() {
+    let dir = tempdir().unwrap();
+    let src = dir.path().join("nonexistent.txt");
+    let dst = dir.path().join("dst.txt");
+
+    let result = fs::hard_link(&src, &dst).await;
+
+    assert!(result.is_err());
+    let err = result.unwrap_err();
```
Darksonn: I believe `unwrap_err()` will give a better error message than your assert if the `Result` is `Ok`.
```suggestion
    let err = result.unwrap_err();
```

tokio/tests/fs_link.rs
```diff
@@ -60,3 +60,62 @@ async fn test_symlink() {
...
+async fn test_hard_link_error_destination_already_exists() {
+    let dir = tempdir().unwrap();
+    let src = dir.path().join("src.txt");
+    let dst = dir.path().join("dst.txt");
+
+    // Create source file
+    std::fs::File::create(&src)
+        .unwrap()
+        .write_all(b"source content")
+        .unwrap();
```
Darksonn:
```suggestion
    std::fs::write(src, b"source content").unwrap();
```

tokio/tests/fs_link.rs
```diff
@@ -60,3 +60,62 @@ async fn test_symlink() {
...
+
+    // Create source directory
+    fs::create_dir(&src_dir).await.unwrap();
+
+    // Attempt to create hard link from a directory
+    let result = fs::hard_link(&src_dir, &dst).await;
+
+    // On most systems, hard linking directories is not allowed
+    assert!(result.is_err());
+    // The exact error kind may vary by platform, but it should fail
```
Darksonn: I'm not sure it is worth it, but if it's just windows vs others, perhaps we could add per-platform logic here?
```rs
#[cfg(windows)]
assert_eq!(err.kind(), A);
#[cfg(not(windows))]
assert_eq!(err.kind(), A);
```
If it's more complex, then I wouldn't bother.

kevinWangSheng: Thanks for the review! I've applied all the suggestions:
  - Removed redundant error checks before unwrap_err()
  - Simplified file creation using std::fs::write()
  - Added platform-specific error assertions

PR #7866, opened 2026-01-20
Use call_site hygiene to avoid unused qualification

mu001999:
## Motivation

Previous attempt: https://github.com/tokio-rs/tokio/pull/7777
More context: https://github.com/rust-lang/rust/issues/149873

Direct using user's span here will lead to unexpected warning of unused qualification for user code.

## Solution

As @taiki-e suggests (https://github.com/rust-lang/rust/issues/149873#issuecomment-3739121206), use `Span::call_site` hygiene here to help the lint can use `Span::from_expansion` to check it.

Have tested this solution in https://github.com/mu001999-contrib/tokio-unexpected-warning, the output is:
```
~/r/unexpected-warning @ main » cd base 
~/r/u/base @ main » cargo check --tests
    Checking base v0.1.0 (/Users/mu00/repos/unexpected-warning/base)
error: unnecessary qualification
 --> base/src/lib.rs:5:5
  |
5 |     if true {}
  |     ^^
  |
  = note: requested on the command line with `-D unused-qualifications`

error: could not compile `base` (lib test) due to 1 previous error
~/r/u/base @ main [101] » cd ../new          
~/r/u/new @ main » cargo check --tests
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
```

The difference is that the new crate uses the dependency `tokio = { git = "https://github.com/mu001999-contrib/tokio", branch = "fix/use-callsite-span", features = ["full"] }`

### DIFF
--- a/tests-build/tests/pass/use_builder_outer.rs
+++ b/tests-build/tests/pass/use_builder_outer.rs
@@ -1,0 +1,9 @@
+#![deny(unused_qualifications)]
+
+use tests_build::tokio;
+pub use tokio::runtime;
+
+#[tokio::main]
+async fn main() {
+    if true {}
+}
--- a/tests-build/tests/macros.rs
+++ b/tests-build/tests/macros.rs
@@ -16,6 +16,9 @@ fn compile_fail_full() {
     t.pass("tests/pass/impl_trait.rs");
 
     #[cfg(feature = "full")]
+    t.pass("tests/pass/use_builder_outer.rs");
+
+    #[cfg(feature = "full")]
     t.compile_fail("tests/fail/macros_invalid_input.rs");
 
     #[cfg(feature = "full")]


Darksonn:
It sounds like the output is still incorrect. Am I mistaken? What is the output prior to this change?

mu001999:
> It sounds like the output is still incorrect. Am I mistaken? What is the output prior to this change?

The above output have two parts, base is the output prior to this change:
```
~/r/unexpected-warning @ main » cd base 
~/r/u/base @ main » cargo check --tests
    Checking base v0.1.0 (/Users/mu00/repos/unexpected-warning/base)
error: unnecessary qualification
 --> base/src/lib.rs:5:5
  |
5 |     if true {}
  |     ^^
  |
  = note: requested on the command line with `-D unused-qualifications`

error: could not compile `base` (lib test) due to 1 previous error
```

The new is the output after this change:
```
~/r/u/base @ main [101] » cd ../new          
~/r/u/new @ main » cargo check --tests
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
```

Darksonn:
Ah, so the error went away. That's great. Could we test this under tests-build?

mu001999:
> Could we test this under tests-build?

Yes, I have added one.

tests-build/tests/pass/use_builder_outer.rs
```diff
@@ -0,0 +1,9 @@
+#![deny(unused_qualifications)]
+
+use tests_build::tokio;
+pub use tokio::runtime;
+
+#[tests_build::tokio::main]
```
Darksonn: Other tests just do:
```suggestion
#[tokio::main]
```
Is this necessary to trigger the error somehow?

tokio-macros/src/entry.rs
```diff
@@ -470,6 +477,8 @@ fn parse_knobs(mut input: ItemFn, is_test: bool, config: FinalConfig) -> TokenSt
         #[cfg(all(#(#checks),*))]
         #[allow(clippy::expect_used, clippy::diverging_sub_expression, clippy::needless_return, clippy::unwrap_in_result)]
         {
+            #use_builder
```
Darksonn, line 480: This will cause the `use` statement to be in scope for the body. Please don't import anything into the scope of the body.

tests-build/tests/pass/use_builder_outer.rs
mu001999: removed

tokio-macros/src/entry.rs
mu001999: why? this is inside the last block. I think it won't be imported into the body

mu001999: And without this, I think we cannot fix this. If we just change the span of the following:
```
return #rt
    .enable_all()
    .#build
    .block_on(#body_ident);
```

the expr may emit unexpected suggestion when type mismatch, like:
```
#[tokio::main].expect("REASON")
```

mu001999: Other way like `let rt = #rt;` will also introduce new variable.

Darksonn: Hmm, I didn't realize that `#body_ident` was not the entire body. Still, we've had several macro bugs due to `use` before. Is it possible to avoid it?

mu001999: Using new variable to split the last stmt is another way I can think of.

djc:
Awesome, thanks!

PR #7869, opened 2026-01-25
signal: guarantee that listeners never return `None`

vilgotf:
As discussed on Discord, having the listeners return `Option<()>` was an accident due to them previously being implemented through `Stream::poll_next`.

Notably this does not guarantee a new shutdown behavior for listeners. I did not find a documented behavior for `time`, but I suppose both modules could have similar guarantees.

### DIFF
--- a/tokio/src/signal/unix.rs
+++ b/tokio/src/signal/unix.rs
@@ -422,6 +422,7 @@ impl Signal {
     /// Receives the next signal notification event.
     ///
     /// Although this returns `Option<()>`, it will never actually return `None`.
+    /// This was accidentally exposed and would be a breaking change to be removed.
     ///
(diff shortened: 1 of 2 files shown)


tokio/src/signal/unix.rs
```diff
@@ -421,7 +421,7 @@ pub(crate) fn signal_with_handle(
 impl Signal {
     /// Receives the next signal notification event.
     ///
-    /// `None` is returned if no more events can be received by this stream.
+    /// Although this returns `Option<()>`, it will never actually return `None`.
```
ADD-SP, line 424: nit: we could also say that the `Option` is an accident here to give more context to downstream users.

Issue #4996, opened 2022-09-09
Second call of UdpSocket::send seems to fail even after calling UdpSocket::connect


Andful:
**Version**
tokio 1.21.0

**Platform**
Linux \<redacted\> 5.15.0-46-generic #49~20.04.1-Ubuntu SMP Thu Aug 4 19:15:44 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

**Description**
After calling `UdpSocket::connect`, the first call to `UdpSocket::send` works, but the second call to `UdpSocket::send` fails.

Here is some code that fails.
```rust
use tokio::net::UdpSocket;

#[tokio::main]
async fn main() {
    let socket =  UdpSocket::bind(("0.0.0.0", 12345)).await.unwrap();
    socket.connect(("127.0.0.1", 1122)).await.unwrap();
    socket.send(b"test 123").await.unwrap();
    socket.send(b"test 123").await.unwrap();
}
```
and a rust playground link to the same code.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b522081daa9f88ba0bfb22080fd3d87e.

The code should work, and it should not matter if any program is listening to the specific port.

Noah-Kennedy:
This is actually how UDP is supposed to work. Because nothing is listening on that port, an ICMP Unreachable packet is sent as a response to the initial datagram, causing the second attempted send to fail with "connection refused". This logic is in the kernel, not in tokio, which is why std::net::UdpSocket has the same behavior.

hydra:
thanks for posting this question and the response, I bumped into this today too, it was entirely non-obvious from the UdpSocket documentation and really weird unexpected behavior.  On Windows this doesn't happen, so when I tried my Tokio UDP code on Linux it failed in really weird ways...

Darksonn:
In general, you should never expect that a successfully sent UDP packet was received. UDP packets may be lost, duplicated, or reordered on the way to the recipient.

You may even encounter a situation where it fails on, say, the fifth packet. This can happen if the first three were lost and never made it to the recipient, and the fourth made the recipient reply with ICMP unreachable. Or even if they're not lost, you may have sent the first four packets quickly enough that the reply to the first one arrived after the fourth packet.

hydra:
@Darksonn yes indeed, however 'connection refused' on the second message seemed somewhat odd, given that UDP is lossy to begin with.

[...]

PR #7870, opened 2026-01-26
net: add docs on ConnectionRefused errors with udp sockets

Darksonn:
Refs: #4996 

cc @hydra

### DIFF
--- a/spellcheck.dic
+++ b/spellcheck.dic
@@ -1,4 +1,4 @@
-315
+317
 &
 +
 <
@@ -32,6 +32,7 @@
 8MB
 ABI
 accessors
+adaptively
 adaptor
 adaptors
 Adaptors
@@ -224,6 +225,7 @@ reregistering
 resize
 resized
 RMW
+RNG
 runtime
 runtime's
 runtimes
--- a/tokio/Cargo.toml
+++ b/tokio/Cargo.toml
@@ -125,7 +125,7 @@ libc = { version = "0.2.168", optional = true }
 
 [target.'cfg(unix)'.dev-dependencies]
 libc = { version = "0.2.168" }
-nix = { version = "0.29.0", default-features = false, features = ["aio", "fs", "socket"] }
+nix = { version = "0.31.0", default-features = false, features = ["aio", "fs", "socket"] }
 
 [target.'cfg(windows)'.dependencies.windows-sys]
 version = "0.61"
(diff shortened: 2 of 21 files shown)


ADD-SP (approved):

tokio/src/net/udp.rs
```diff
@@ -529,9 +529,11 @@ impl UdpSocket {
     /// connected to.
     ///
     /// The [`connect`] method will connect this socket to a remote address.
-    /// This method will fail if the socket is not connected.
+    /// This method will fail with a [`ConnectionRefused`] error if the remote
+    /// address has replied with ICMP Unreachable to a previously sent packet.
```
the8472: s/will/may/ ? The issue mentions that windows doesn't do this.

Though the [sendto docs](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-sendto) do mention 

> WSAECONNRESET
>	[...] For UDP sockets, the remote host was unable to deliver a previously sent UDP datagram and responded with a "Port Unreachable" ICMP packet. The application should close the socket as it is no longer usable.

OpenBSD [send manpage](https://man.openbsd.org/sendmsg.2#ERRORS) says it only returns that error for connected sockets.

Darksonn: Hmm, this is on the `send` method docs, which you can only use if the socket is connected, so that seems ok.

hydra: @the8472 correct, on windows it never seems to fail for me with the same code. but on linux it fails on the second send every time.  so 'may' is more appropriate based on my experience.

also:

> The application should close the socket as it is no longer usable.

is not really correct and depends on the use-case; you can happily spam out udp packets so that when the host comes back online or is re-plugged or reconnected to the network it will immedately start receiving the messages again.

Darksonn: Do you want me to change the wording, or?

hydra: yes please, how about:

```
    /// This method may fail with a [`ConnectionRefused`] error if the remote
    /// address has replied with ICMP Unreachable to a previously sent packet.
    /// the behavior is operating system dependant.
```

tokio/src/net/udp.rs
```diff
@@ -529,9 +529,13 @@ impl UdpSocket {
     /// connected to.
     ///
     /// The [`connect`] method will connect this socket to a remote address.
-    /// This method will fail if the socket is not connected.
```
mattiapitossi, line 532: But this is still true right? probably we should merge this with the one below?

Darksonn: Good catch, thanks.

PR #7871, opened 2026-01-26
Implement vectored writes for `write_buf`

Darksonn:


Darksonn:
Verified in CI that the new test fails without the implementation change.

tokio/tests/io_write_buf.rs
```diff
@@ -54,3 +54,63 @@ async fn write_all() {
...
+        }
+
+        fn poll_write_vectored(
+            mut self: Pin<&mut Self>,
+            _cx: &mut Context<'_>,
+            bufs: &[io::IoSlice<'_>],
+        ) -> Poll<Result<usize, io::Error>> {
+            let mut n = 0;
+            for buf in bufs {
+                let amt = cmp::min(4 - n, buf.len());
```
martin-g: nit: Maybe extract `4` as a local variable/constant, so it does not look like a magic number ?

tokio/src/io/util/write_buf.rs
```diff
@@ -42,13 +42,22 @@ where
...
+        const MAX_VECTOR_ELEMENTS: usize = 64;
+
         let me = self.project();
 
         if !me.buf.has_remaining() {
             return Poll::Ready(Ok(0));
         }
 
-        let n = ready!(Pin::new(me.writer).poll_write(cx, me.buf.chunk()))?;
+        let n = if me.writer.is_write_vectored() {
```
paolobarbolini, line 53: I have a project in which I checked if `Buf::chunk().len()` < `Buf::remaining()` before using vectored writes and it resulted into a small performance boost. Could be useful here too.

tokio/tests/io_write_buf.rs
```diff
@@ -54,3 +54,63 @@ async fn write_all() {
...
+                self.buf.extend_from_slice(&buf[..amt]);
+                n += amt;
+                if n == 4 {
+                    break;
+                }
+            }
+            self.cnt += 1;
+            Ok(n).into()
+        }
```
ADD-SP, line 94: I had to spend quite a bit of time to understand what this `for` loop is doing.

To verify that `poll_write_vectored` actually works, it seems sufficient to simply set `Wr::is_called = true` and check whether `Wr::buf` equals `b"hello world!"`.

If I understand correctly, `Wr::cnt` is used to track how many times poll_write_vectored is called, while the local variable n limits the maximum number of bytes written per call. Why did we choose this relatively more complicated approach here instead of a simpler one? Did I miss something?

Darksonn: I simplified this test, PTAL.

tokio/tests/io_write_buf.rs
Darksonn: Obsolete as per simplification.

Issue #7775, opened 2025-12-14
CancellationToken is_cancelled locks on a mutex, caused occasional >2ms delays


KarstenB:
**Version**
tokio v1.48.0
tokio-util v0.7.17

**Platform**
Linux machine 6.12.57+deb13-rt-amd64 #1 SMP PREEMPT_RT Debian 6.12.57-1 (2025-11-05) x86_64 GNU/Linux

The machine is tuned for running RT applications.

**Description**
I have an async method that spawns a regular thread. The async method has a CancellationToken that is used in a loop like this:

```rust
while !token.is_cancelled() && !tx_thread.is_finished() {
    tick.tick().await;
    //print statistics
}
```
And the sync thread is doing almost the same:
```rust
adjust_prio(51, SCHED_RR);
while !token.is_cancelled() {
   do_work();
   thread::sleep(Duration::from_micros(100));
}
```

The sync thread is running on an isol_cpu, the IRQs have been moved away from it. Nothing else is running on this particular core. So I ran FTrace to figure out what was going on, and the results looked something like this:
```
MyProg-442147  [013] ..... 242755.807682: common_nsleep <-__x64_sys_clock_nanosleep
MyProg-442147  [013] ..... 242755.807793: common_nsleep <-__x64_sys_clock_nanosleep
MyProg-442147  [013] ..... 242755.807916: common_nsleep <-__x64_sys_clock_nanosleep
MyProg-442147  [013] ..... 242755.808037: common_nsleep <-__x64_sys_clock_nanosleep
MyProg-442147  [013] ..... 242755.808150: common_nsleep <-__x64_sys_clock_nanosleep
MyProg-442147  [013] ..... 242755.808270: common_nsleep <-__x64_sys_clock_nanosleep
MyProg-442147  [013] ..... 242755.808381: common_nsleep <-__x64_sys_clock_nanosleep
MyProg-442147  [013] ...1. 242755.808492: sys_futex(uaddr: 7f9f3011b810, op: 89, val: 2, utime: 0, uaddr2: 0, val3: 7f9effffffff)
MyProg-442147  [013] ...1. 242755.810545: sys_futex -> 0x0
MyProg-442147  [013] ...1. 242755.810547: sys_futex(uaddr: 7f9f3011b810, op: 81, val: 1, utime: 0, uaddr2: 7f9effffffff, val3: 0)
MyProg-442147  [013] ...1. 242755.810558: sys_futex -> 0x1
[...]

PR #7874, opened 2026-01-31
util: add AtomicBool for lock-free is_cancelled check

HueCodes:
## Motivation

`CancellationToken::is_cancelled()` acquires a mutex on every call, causing >2ms delays on real-time systems.

Fixes: #7775

## Solution

Add an `AtomicBool` field to `TreeNode` that mirrors the `is_cancelled` state. The atomic is updated with `Release` ordering when cancellation occurs, and read with `Acquire` ordering in `is_cancelled()`, providing lock-free reads without mutex contention.

### DIFF
--- a/tokio-util/src/sync/cancellation_token/tree_node.rs
+++ b/tokio-util/src/sync/cancellation_token/tree_node.rs
@@ -74,7 +74,7 @@ impl TreeNode {
 /// The data contained inside a `TreeNode`.
 ///
 /// This struct exists so that the data of the node can be wrapped
-/// in a Mutex.
+/// in a `Mutex`.
 struct Inner {
     parent: Option<Arc<TreeNode>>,
     parent_idx: usize,
@@ -88,7 +88,7 @@ pub(crate) fn is_cancelled(node: &Arc<TreeNode>) -> bool {
 }
 
 /// Returns whether or not the node is cancelled, using a mutex lock for proper
-/// synchronization. This is used in WaitForCancellationFuture::poll() to ensure
+/// synchronization. This is used in `WaitForCancellationFuture::poll()` to ensure
 /// correct memory ordering with the cancellation process.
 pub(crate) fn is_cancelled_with_lock(node: &Arc<TreeNode>) -> bool {
     let _guard = node.inner.lock().unwrap();
--- a/tokio-util/src/sync/cancellation_token.rs
+++ b/tokio-util/src/sync/cancellation_token.rs
@@ -207,7 +207,7 @@ impl CancellationToken {
     }
 
     /// Internal method that checks cancellation with mutex lock for proper
-    /// synchronization in WaitForCancellationFuture::poll().
+    /// synchronization in `WaitForCancellationFuture::poll()`.
     fn is_cancelled_with_lock(&self) -> bool {
         tree_node::is_cancelled_with_lock(&self.inner)
     }


tokio-util/src/sync/cancellation_token/tree_node.rs
Darksonn: Please delete this field to avoid duplication.

tokio-util/src/sync/cancellation_token/tree_node.rs
```diff
@@ -74,13 +79,12 @@ struct Inner {
...
     parent_idx: usize,
     children: Vec<Arc<TreeNode>>,
-    is_cancelled: bool,
     num_handles: usize,
 }
 
 /// Returns whether or not the node is cancelled
 pub(crate) fn is_cancelled(node: &Arc<TreeNode>) -> bool {
-    node.inner.lock().unwrap().is_cancelled
+    node.is_cancelled.load(Ordering::Acquire)
```
Darksonn, line 87: Hmm. Acquire is not actually strong enough here. When `WaitForCancellationFuture::poll()` returns pending in parallel with being woken up, we currently rely on the following logic:

1. Inside `WaitForCancellationFuture`, we create a `Notified` future.
2. Inside `WaitForCancellationFuture`, we read `false` from `is_cancelled`.
3. Inside `CancellationToken::cancel()`, we write `true` to `is_cancelled`.
4. Inside `CancellationToken::cancel()`, we call `notify_waiters()`.

And here it is really important that `notify_waiters()` will wake up the `Notified` future created in step 1. Otherwise the `WaitForCancellationFuture` might sleep forever. With a mutex, that's fine, but with the current atomics, it's not actually satisfied!

That's because what we need is that step 2 happens-before step 3. However, for that to be true, operation 2 must be Release, and operation 3 must be Acquire. Unfortunately, it's reversed! Operation 2 is Acquire, and operation 3 is Release.

Unfortunately, a `load()` can't be Release and a `store()` can't be `Acquire`, so we have to upgrade them to read-modify-write operations for this to work:

1. Use `AtomicBool::swap(true, AcqRel)` to write `true`.
2. Use `AtomicBool::compare_exchange(false, false, AcqRel, Acquire)` to read. (`fetch_and(true, AcqRel)` would also work)

Darksonn: Or we can just take the mutex in `WaitForCancellationFuture::poll()`, and leave the `AtomicBool` optimization for users of `is_cancelled()` only.

martin-g: IMO using both AtomicBool and Mutex for the same purpose but for different use cases is not a good idea. It makes the reasoning harder - should I use the atomics based solution or the mutex one for use case 3 (a use case from the future) ?! 
If it could be implemented fully with atomics (`swap` + `compare_exchange`) it would be better!

A good start would be to add an integration (or bench) test that reproduces the problem with Acquire+Release and then fixing it by using swap+compare_exchange.

Darksonn: What do you mean by a use-case from the future? As #7775 indicates, there are users today that call `is_cancelled()` instead of `.await`ing a future, and such users would get the benefit today.

martin-g: > and such users would get the benefit today

This is clear!

What bothers me is that some use cases could use `is_cancelled()` (lock-free, for both external and internal users) and other use cases (internal user) like `WaitForCancellationFuture::poll()` can't and they need to depend on the Mutex. 
With a "use case from the future" I mean the second - a new Tokio internal functionality that needs to know whether a token is cancelled or not. What should the developer use - `is_cancelled()` or `is_cancelled_with_lock()` ?! I guess the answer is `it depends`.
But if by using swap+compare_exchange there won't be a need of `is_cancelled_with_lock()` then the decision will be simple.

PR #7882, opened 2026-02-04
io: fix race in `Mock::poll_write`

ADD-SP:
It is hard to add a test for this bug, so I have tested it manually.

_fix #7881_

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,3 +1,6 @@
+R-loom-blocking:
+- tokio/src/runtime/blocking/*
+- tokio/src/runtime/blocking/**/*
 
 R-loom-sync:
 - tokio/src/sync/*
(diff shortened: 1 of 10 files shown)


edef1c:
Hm, this seems like it might change behaviour somewhat: I was using `Duration::ZERO` in the mocks to force another poll cycle, to ensure I'm testing code properly against `Poll::Pending` cases. AFAICT, the loop simply continuing pretends the timeout "never happened". Though I realise my usage is a bit of a hack; plausibly I should be using a 1ns timeout and the deterministic clock mode for this.

ADD-SP:
> Hm, this seems like it might change behaviour somewhat: I was using `Duration::ZERO` in the mocks to force another poll cycle, to ensure I'm testing code properly against `Poll::Pending` cases. AFAICT, the loop simply continuing pretends the timeout "never happened". Though I realise my usage is a bit of a hack; plausibly I should be using a 1ns timeout and the deterministic clock mode for this.

Hmm, force yielding using `Duration::ZERO` is a bit hacky, I think we can consider adding a new feature for `Mock` instead of relying on buggy behavior.

PR #7886, opened 2026-02-05
Drop rx_task when oneshot receiver is dropped

### LINKED ISSUE #7884: oneshot retains rx_task waker even after rx is closed, leads to RawTask 'leak'
https://github.com/tokio-rs/tokio/blob/master/tokio/src/sync/oneshot.rs#L1301 

The rx_task waker is only used if the oneshot is not closed. Therefore, I think it would be reasonable to uninitialise the rx_task waker on close. Because we do not currently do this, we hold a reference to the receiving-end task waker for no reason, and this can lead to a RawTask memory leak; even after cancellation, the RawTask containing the receive-side future is kept around as long as the oneshot sender still exists. Obviously this can be fixed by dropping any closed senders, but it is very unintuitive that this can lead to a task leak on the receiving side (and tasks are quite large, while the oneshot inner is typically small, so the task leak would dominate).

If we agree this is a sensible change I would love to make the PR!

jackkleeman:
## Motivation

Fixes #7884

### DIFF
--- a/tokio/src/sync/oneshot.rs
+++ b/tokio/src/sync/oneshot.rs
@@ -1247,15 +1247,6 @@ impl<T> Drop for Receiver<T> {
         if let Some(inner) = self.inner.as_ref() {
             let state = inner.close();
 
-            // We need to unset RX_TASK_SET and drop the waker now, rather than
-            // waiting for completion, to avoid an unnecessary ref to the rx task
-            if state.is_rx_task_set() && !state.is_complete() {
-                State::unset_rx_task(&inner.state);
-                unsafe {
-                    inner.rx_task.drop_task();
-                }
-            }
-
             if state.is_complete() {
                 // SAFETY: we have ensured that the `VALUE_SENT` bit has been set,
                 // so only the receiver can access the value.
@@ -1398,6 +1389,19 @@ impl<T> Inner<T> {
             unsafe {
                 self.tx_task.with_task(Waker::wake_by_ref);
             }
+        }
+
+        if prev.is_rx_task_set() && !prev.is_complete() {
+            State::unset_rx_task(&self.state);
+            // SAFETY: The sender only accesses `rx_task` (via
+            // `wake_by_ref`) in `complete()` after successfully setting
+            // `VALUE_SENT`. But `set_complete` will not set `VALUE_SENT`
+            // if `CLOSED` is already set (its CAS loop breaks early).
+            // Since `prev` shows that `VALUE_SENT` was not set before we
+            // set `CLOSED`, the sender can no longer set `VALUE_SENT` and
+            // will never access `rx_task`. Therefore, we have exclusive
+            // access here.
+            unsafe { self.rx_task.drop_task() };
         }
 
         prev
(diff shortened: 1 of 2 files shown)


tokio/src/sync/oneshot.rs
```diff
@@ -1247,6 +1247,15 @@ impl<T> Drop for Receiver<T> {
         if let Some(inner) = self.inner.as_ref() {
             let state = inner.close();
 
+            // We need to unset RX_TASK_SET and drop the waker now, rather than
+            // waiting for completion, to avoid an unnecessary ref to the rx task
+            if state.is_rx_task_set() && !state.is_complete() {
+                State::unset_rx_task(&inner.state);
+                unsafe {
+                    inner.rx_task.drop_task();
```
Darksonn: So what I was worried about in https://github.com/tokio-rs/tokio/issues/7884#issuecomment-3853413058 is that the sender might be calling `wake_by_ref()` on this field in parallel with `drop_task()`, which would be UB. But I guess this works because the sender only touches `rx_task` if `VALUE_SENT` is set before `CLOSED`, and here we know that `CLOSED` was set first.

Can you add a safety comment to the unsafe block explaining the above?

Also, moving the semicolon like this changes the formatting in a way I like better:
```rs
unsafe { inner.rx_task.drop_task() };
```

tokio/src/sync/tests/loom_oneshot.rs
```diff
@@ -185,3 +185,30 @@ fn checking_tx_send_ok_not_drop() {
...
+
+        // Spawn a thread that will poll the receiver once (setting rx_task),
+        // then drop it.
+        let rx_thread = thread::spawn(move || {
+            // Poll once to set rx_task, then drop
+            let _ = block_on(poll_fn(|cx| {
+                let _ = Pin::new(&mut rx).poll(cx);
+                // Always return Ready to exit the poll, then rx will be dropped
+                Ready(())
+            }));
```
Darksonn: I think we can move the `poll` outside the thread, and have the only operation in the thread be `drop(rx)`.

The parallel part of Loom tests is really really expensive, so we should perform as few operations as possible in the parallel part of the test.

tokio/src/sync/oneshot.rs
```diff
@@ -1247,6 +1247,21 @@ impl<T> Drop for Receiver<T> {
...
+                // SAFETY: The sender only accesses `rx_task` (via
+                // `wake_by_ref`) in `complete()` after successfully setting
+                // `VALUE_SENT`. But `set_complete` will not set `VALUE_SENT`
+                // if `CLOSED` is already set (its CAS loop breaks early).
+                // Since `state` (returned by `close()`) shows that
+                // `VALUE_SENT` was not set before we set `CLOSED`, the sender
+                // can no longer set `VALUE_SENT` and will never access
+                // `rx_task`. Therefore, we have exclusive access here.
+                unsafe { inner.rx_task.drop_task() };
+            }
```
Darksonn: Here's an idea: perhaps this logic should be inside `close()`? That way, it takes effect right away if `Receiver::close()` is called too.

PR #7889, opened 2026-02-05
tests: fix constants for `target_os = "android"`

gburgessiv:
## Motivation

We recently tried to upgrade Android to the newest tokio, but tests that use these constants fail, since non-`target_os = "linux"` values are selected.

## Solution

Opt Android into the Linux constants. :)


gburgessiv:
Testing on Android looks good to me

PR #7891, opened 2026-02-06
task: fix task module feature flags

mattiapitossi:
## Motivation

Currently, if you look at the Tokio docs the task module it's not under the rt feature flag:

image

But every public module and struct inside task is under `rt` feature flag (except for coop which is only available for the crate).

This also leads to confusion because items that have multiple feature flags don't display the rt flag:

image

## Solution

The solution is to gate the task module under rt feature flag:

image

image

## Docs

[rendered docs](https://deploy-preview-7891--tokio-rs.netlify.app/tokio/)

### DIFF
--- a/tokio/src/lib.rs
+++ b/tokio/src/lib.rs
@@ -555,7 +555,6 @@ cfg_not_sync! {
 // of the task module.
 #[cfg_attr(docsrs, doc(cfg(feature = "rt")))]
 pub mod task;
-
 cfg_rt! {
     pub use task::spawn;
 }


tokio/src/lib.rs
```diff
@@ -549,11 +549,15 @@ cfg_not_sync! {
...
 }
 
-pub mod task;
 cfg_rt! {
+    pub mod task;
     pub use task::spawn;
 }
 
+cfg_not_rt! {
+    pub(crate) mod task;
```
mattiapitossi: Inside the task module only crate coop is available under `not_rt`

Darksonn:
I believe `tokio::task::coop` is public and available when `rt` is not enabled, so this change seems wrong.

mattiapitossi:
> I believe `tokio::task::coop` is public and available when `rt` is not enabled, so this change seems wrong.

Do you mean before this change? coop should be only visible inside the crate when the `rt` is not enabled:

inside task/mod is defined as:

```rust
cfg_not_rt! {
    pub(crate) mod coop;
}
```

So for example when importing tokio without rt it's private:
image

tokio/src/lib.rs
```diff
@@ -549,11 +549,15 @@ cfg_not_sync! {
     mod sync;
 }
 
-pub mod task;
 cfg_rt! {
+    pub mod task;
     pub use task::spawn;
 }
```
Darksonn, line 560: Even if the module is empty this is a breaking change because some consumers may write:
```rs
use tokio::task;
```
in their codebase even if `rt` is not enabled.

Can we make this a doc-only change?
```rs
#[doc(cfg(feature = "rt"))]
pub mod task;
```

mattiapitossi: Make sense to me if we don't want to introduce such breaking change. I added the ff with a comment to explain, let me know if it makes sense to you to have it.

PR #7896, opened 2026-02-08
runtime: correct the default thread name in docs

DaniPopes:
## Motivation

Missed in https://github.com/tokio-rs/tokio/pull/7880, sorry about that.

## Solution

Update docs.


ADD-SP:
Thanks!

PR #7899, opened 2026-02-09
Document possible `Mutex` deadlock in `block_in_place`

### LINKED ISSUE #7892: `block_in_place` + `block_on` + `tokio::sync::Mutex` may lead to a deadlock
In https://github.com/eclipse-zenoh/zenoh/issues/2409 I have a scenario roughly equivalent to to the following `Future` implementation:

```rust
/// A [`Future`] that polls a single mutex in two branches.
///
/// On the first poll, we poll a [`Mutex::lock_owned`] (i.e. [`Self::lock_fut`]) on the mutex while
/// it is held (see [`main`]). Thus we return [`Poll::Pending`].
///
/// On the second poll we use [`task::block_in_place`] and [`Handle::block_on`] to await
/// [`Mutex::lock`] on the mutex.
///
/// Because [`Self::lock_fut`] is assigned the only available permit by Tokio's semaphore logic, we
/// are unable to acquire the mutex on the second poll.
struct BlockOnSecondPoll {
    polled: AtomicBool,
    mutex: Arc<Mutex<()>>,
    lock_fut: Pin<Box<dyn Future<Output = tokio::sync::OwnedMutexGuard<()>>>>,
}

impl Future for BlockOnSecondPoll {
    type Output = ();

    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
        let was_polled = self.polled.fetch_or(true, atomic::Ordering::SeqCst);

        if was_polled {
            tracing::info!("will block forever");
            let () = *task::block_in_place(|| Handle::current().block_on(self.mutex.lock()));
            unreachable!()
        } else {
            match self.lock_fut.as_mut().poll(cx) {
                Poll::Ready(_) => unreachable!(),
                Poll::Pending => {
                    tracing::info!("pending");
                    Poll::Pending
                }
            }
        }
    }
}
```

(see https://github.com/fuzzypixelz/tokio-inaccessible-permit for the full code)

`BlockOnSecondPoll` reliably deadlocks the process. Tokio's behavior is perfectly correct here. But I believe that the warning section of the `block_in_place` documentation:

> Be aware that although this function avoids starving other independently spawned tasks, any other code running concurrently in the same task will be suspended during the call to block_in_place. This can happen e.g. when using the [join!](https://docs.rs/tokio/latest/tokio/macro.join.html) macro. To avoid this issue, use [spawn_blocking](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) instead of block_in_place.

Could be a bit more strongly worded ;)

fuzzypixelz:
Resolves #7892.

### DIFF
--- a/tokio/src/task/blocking.rs
+++ b/tokio/src/task/blocking.rs
@@ -65,6 +65,21 @@ cfg_rt_multi_thread! {
     /// # }
     /// ```
     ///
+    /// <div class="warning">
+    ///
+    /// If a future comprises different branches that attempt to acquire the same async [`Mutex`]
+    /// (e.g. [`select`] branches), then calling `lock` via `block_in_place` and `block_on` in one
+    /// branch can **deadlock** if another branch had previously yielded while waiting for the same
+    /// mutex. This happens because the yielded branch holds a position in the mutex's (FIFO) wait
+    /// queue that cannot be fulfilled. Hence the future fails to make progress.
+    ///
+    /// See [#7892](https://github.com/tokio-rs/tokio/issues/7892) for more information.
+    ///
+    /// [`Mutex`]: struct@crate::sync::Mutex
+    /// [`select`]: macro@crate::select
+    ///
+    /// </div>
+    ///
     /// # Panics
     ///
     /// This function panics if called from a [`current_thread`] runtime.


fuzzypixelz:
Rendered:

image

All links are valid.

I found the MRE too complicated to include in the Rustdoc string, so I linked the issue instead. I couldn't find any existing instances where documentation links to closed issues so I don't know how appropriate this is.

cc @Darksonn.

tokio/src/task/blocking.rs
Darksonn, line 19: We already have this paragraph.

The issue is not really specific to `Mutex`. It happens with lots of stuff, but you make it sound like the only problem is `Mutex`. If you want to write something that talks about `Mutex`, it needs to be clear that it's an *example* and that it applies to any other code running in the same task.

fuzzypixelz: > We already have this paragraph.

I understand that the two paragraphs are perhaps too similar, my intension was for the added paragraph to have the extra context of `block_on` use. I don't mind merging the second paragraph into the first, I'm just not yet sure if `spawn_blocking` is a good workaround for this issue as well (if it actually does solve it at all?).

> The issue is not really specific to Mutex. It happens with lots of stuff, but you make it sound like the only problem is Mutex.

Right, it seems to me that it happens with `Semaphore` and `RwLock` as well. Is there a concept that encompasses all/most possible cases? In either case, I agree `Mutex` should be marked as an example.

I'll rework this once I have satisfactory answers to both questions.

fuzzypixelz: > We already have this paragraph.

I understand that the two paragraphs are perhaps too similar, my intension was for the added paragraph to have the extra context of `block_on` use. I don't mind merging the second paragraph into the first, I'm just not yet sure if `spawn_blocking` is a good workaround for this issue as well (if it actually does solve it at all?).

> The issue is not really specific to Mutex. It happens with lots of stuff, but you make it sound like the only problem is Mutex.

Right, it seems to me that it happens with `Semaphore` and `RwLock` as well. Is there a concept that encompasses all/most possible cases? In either case, I agree `Mutex` should be marked as an example.

I'll rework this once I have satisfactory answers to both questions.> We already have this paragraph.

I understand that the two paragraphs are perhaps too similar, my intension was for the added paragraph to have the extra context of `block_on` use. I don't mind merging the second paragraph into the first, I'm just not yet sure if `spawn_blocking` is a good workaround for this issue as well (if it actually does solve it at all?).

> The issue is not really specific to Mutex. It happens with lots of stuff, but you make it sound like the only problem is Mutex.

Right, it seems to me that it happens with `Semaphore` and `RwLock` as well. Is there a concept that encompasses all/most possible cases? In either case, I agree `Mutex` should be marked as an example.

I'll rework this once I have satisfactory answers to both questions.

Darksonn: > I'm just not yet sure if `spawn_blocking` is a good workaround for this issue as well

The most generic thing I can say is "to avoid this, do not use `block_in_place`". There does not exist one single alternative I can recommend that works in all cases.

I recommend `spawn_blocking`, which is sometimes a suitable alternative, but it's not identical to `block_in_place` and so it does not work in all cases where `block_in_place` can be used.

> Is there a concept that encompasses all/most possible cases?

The way I would describe the general problem is "intra-task concurrency". That is to say, the task calling `block_in_place` is also running any other future concurrently within the same task, using any intra-task concurrency primitive such as `select!`, `join!`, `FuturesUnordered`, `buffer_unordered`, etc

The bug is a variation of what has been recently dubbed [futurelock](https://rfd.shared.oxide.computer/rfd/0609).

ADD-SP:
@fuzzypixelz Hi, are you still going to work on this PR?

PR #7900, opened 2026-02-10
sync: add `closed` method to `broadcast::WeakSender`

mladedav:
## Motivation

We have a broadcast channel that is filled by a background task that is expensive to start. We want to stop the task if there were no active subscribers for a given time.

Right now, we clone the `Sender` and call `closed` on it along with watching for a channel that notifies us of new subscribers. However, this means that the strong count is incremented by this new sender instance and if we use it wrong, concurrent calls to `Sender::closed` and `Receiver::recv` could deadlock.

This method can also be used to react to dropping the last receiver after dropping the sender, something along the lines of
```rust
sender.send(msg);
let weak = sender.downgrade();
// This was the last strong sender
drop(sender);
println!("Waiting for all receivers to process the last message...");
weak.closed().await;
println!("All receivers have finished.");
``` 

## Solution

Reuse the method already present on `broadcast::Sender`.

### DIFF
--- a/tokio/src/sync/broadcast.rs
+++ b/tokio/src/sync/broadcast.rs
@@ -366,8 +366,8 @@ struct Tail {
     /// Number of active receivers.
     rx_cnt: usize,
 
-    /// True if the channel is closed.
-    closed: bool,
+    /// True if there are any strong senders.
+    has_senders: bool,
 
     /// Receivers waiting for a value.
     waiters: LinkedList<Waiter, <Waiter as linked_list::Link>::Target>,
@@ -566,7 +566,7 @@ impl<T> Sender<T> {
             tail: Mutex::new(Tail {
                 pos: 0,
                 rx_cnt: receiver_count,
-                closed: receiver_count == 0,
+                has_senders: true,
                 waiters: LinkedList::new(),
             }),
             num_tx: AtomicUsize::new(1),
@@ -887,24 +887,12 @@ impl<T> Sender<T> {
     /// # }
     /// ```
     pub async fn closed(&self) {
-        loop {
-            let notified = self.shared.notify_last_rx_drop.notified();
-
-            {
-                // Ensure the lock drops if the channel isn't closed
-                let tail = self.shared.tail.lock();
-                if tail.closed {
-                    return;
-                }
-            }
-
-            notified.await;
-        }
+        self.shared.closed_for_senders().await;
     }
 
     fn close_channel(&self) {
         let mut tail = self.shared.tail.lock();
-        tail.closed = true;
+        tail.has_senders = false;
 
         self.shared.notify_rx(tail);
     }
@@ -925,13 +913,6 @@ fn new_receiver<T>(shared: Arc<Shared<T>>) -> Receiver<T> {
     let mut tail = shared.tail.lock();
 
     assert!(tail.rx_cnt != MAX_RECEIVERS, "max receivers");
-
-    if tail.rx_cnt == 0 {
-        // Potentially need to re-open the channel, if a new receiver has been added between calls
-        // to poll(). Note that we use rx_cnt == 0 instead of is_closed since is_closed also
-        // applies if the sender has been dropped
-        tail.closed = false;
-    }
 
     tail.rx_cnt = tail.rx_cnt.checked_add(1).expect("overflow");
     let next = tail.pos;
@@ -1053,6 +1034,27 @@ impl<T> Shared<T> {
 
         wakers.wake_all();
     }
+
+    async fn closed_for_senders(&self) {
+        cooperative(async {
+            crate::trace::async_trace_leaf().await;
+
+            loop {
+                let notified = self.notify_last_rx_drop.notified();
+
+                {
+                    // Ensure the lock drops if the channel isn't closed
+                    let tail = self.tail.lock();
+                    if tail.rx_cnt == 0 {
+                        return;
+                    }
+                }
+
+                notified.await;
+            }
+        })
+        .await;
+    }
 }
 
 impl<T> Clone for Sender<T> {
@@ -1100,6 +1102,37 @@ impl<T> WeakSender<T> {
                 Err(prev_count) => tx_count = prev_count,
             }
         }
+    }
+
+    /// A future which completes when the number of [Receiver]s subscribed to this channel reaches
+    /// zero, regardless of whether strong senders still exist.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use futures::FutureExt;
+    /// use tokio::sync::broadcast;
+    ///
+    /// # #[tokio::main(flavor = "current_thread")]
+    /// # async fn main() {
+    /// let (tx, mut rx1) = broadcast::channel::<u32>(16);
+    /// let mut rx2 = tx.subscribe();
+    ///
+    /// let _ = tx.send(10);
+    /// let weak = tx.downgrade();
+    /// drop(tx);
+    ///
+    /// assert_eq!(rx1.recv().await.unwrap(), 10);
+    /// drop(rx1);
+    /// assert!(weak.closed().now_or_never().is_none());
+    ///
+    /// assert_eq!(rx2.recv().await.unwrap(), 10);
+    /// drop(rx2);
+    /// assert!(weak.closed().now_or_never().is_some());
+    /// # }
+    /// ```
+    pub async fn closed(&self) {
+        self.shared.closed_for_senders().await;
     }
 
     /// Returns the number of [`Sender`] handles.
@@ -1256,7 +1289,7 @@ impl<T> Receiver<T> {
                     // At this point the channel is empty for *this* receiver. If
                     // it's been closed, then that's what we return, otherwise we
                     // set a waker and return empty.
-                    if tail.closed {
+                    if !tail.has_senders {
                         return Err(TryRecvError::Closed);
                     }
 
@@ -1555,7 +1588,6 @@ impl<T> Drop for Receiver<T> {
 
         if remaining_rx == 0 {
             self.shared.notify_last_rx_drop.notify_waiters();
-            tail.closed = true;
         }
 
         drop(tail);
--- a/tokio/tests/sync_broadcast.rs
+++ b/tokio/tests/sync_broadcast.rs
@@ -674,11 +674,13 @@ fn broadcast_sender_closed() {
 }
 
 #[test]
-fn broadcast_sender_closed_with_extra_subscribe() {
+fn broadcast_weak_sender_closed() {
     let (tx, rx) = broadcast::channel::<()>(1);
     let rx2 = tx.subscribe();
+    let weak = tx.downgrade();
+    drop(tx);
 
-    let mut task = task::spawn(tx.closed());
+    let mut task = task::spawn(weak.closed());
     assert_pending!(task.poll());
 
     drop(rx);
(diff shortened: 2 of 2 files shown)


tokio/src/sync/broadcast.rs
```diff
@@ -1102,6 +1106,36 @@ impl<T> WeakSender<T> {
         }
     }
 
+    /// A future which completes when the number of [Receiver]s subscribed to this `Sender` reaches
```
martin-g:
```suggestion
    /// A future which completes when the number of [Receiver]s subscribed to this channel reaches
```

tokio/src/sync/broadcast.rs
```diff
@@ -1053,6 +1041,22 @@ impl<T> Shared<T> {
 
         wakers.wake_all();
     }
+
+    async fn closed(&self) {
+        loop {
```
martin-g: Not caused by this PR but `watch::Sender::closed()` uses `cooperative()` + `async_trace_leaf()` around the loop.
Should the same be done here too ?

Darksonn:
Actually looks like there are failing tests. We'll need those fixed first.

tokio/src/sync/broadcast.rs
```diff
@@ -1102,6 +1106,36 @@ impl<T> WeakSender<T> {
...
+    ///     drop(rx1);
+    ///     assert!(weak.closed().now_or_never().is_none());
+    ///
+    ///     assert_eq!(rx2.recv().await.unwrap(), 10);
+    ///     drop(rx2);
+    ///     assert!(weak.closed().now_or_never().is_some());
+    /// }
+    /// ```
+    pub async fn closed(&self) {
+        self.shared.closed().await;
```
martin-g: I am not sure that `WeakSender#closed()` can delegate to `Shared::close()` because of its usage of `tail.closed` (line 1052).
`tail.closed = true` is done at two places:
1) https://github.com/mladedav/tokio/blob/22dcbdde18347b1c3c14478fc4662f35defcdc0b/tokio/src/sync/broadcast.rs#L1589 - when the last receiver is dropped
2) https://github.com/mladedav/tokio/blob/22dcbdde18347b1c3c14478fc4662f35defcdc0b/tokio/src/sync/broadcast.rs#L895 - at `close_channel()` which is called when the last strong Sender is dropped (https://github.com/mladedav/tokio/blob/22dcbdde18347b1c3c14478fc4662f35defcdc0b/tokio/src/sync/broadcast.rs#L1074)

This means `Shared::closed()` will return immediately whenever all strong senders have been dropped, even if receivers are still alive.
IMO the WeakSender::closed() should depend on `tail.rx_cnt == 0`

mladedav: You're right, that's also what the failed test was about.

I've split `Sahred::closed` into two parts - closed for senders and closed for receivers. When I read through all the usages it seems like it's these two concepts together. Basically, `closed` meant either that there were no senders and then only receivers could observe the value or that there were no receivers and then only senders could observe the value.

Furthermore, after this split, I think we could also just not have `closed_for_senders` at all and use `rx_cnt == 0` in all places instead. I did not do that yet in case you think the boolean should be kept for whatever reason, but I don't see any.

tokio/src/sync/broadcast.rs
mladedav: Applied, thanks.

tokio/src/sync/broadcast.rs
mladedav: Sure, I've added it similarly to what is in `watch.rs`.

tokio/src/sync/broadcast.rs
```diff
@@ -1102,6 +1115,36 @@ impl<T> WeakSender<T> {
...
+    /// use futures::FutureExt;
+    /// use tokio::sync::broadcast;
+    ///
+    /// # #[tokio::main(flavor = "current_thread")]
+    /// # async fn main() {
+    /// let (tx, mut rx1) = broadcast::channel::<u32>(16);
+    /// let mut rx2 = tx.subscribe();
+    ///
+    /// let _ = tx.send(10);
+    /// let weak = tx.downgrade();
```
martin-g, line 1122: This variable is not used below

tokio/src/sync/broadcast.rs
```diff
@@ -1053,6 +1045,27 @@ impl<T> Shared<T> {
 
         wakers.wake_all();
     }
+
+    async fn closed_for_senders(&self) {
+        crate::trace::async_trace_leaf().await;
```
martin-g: In watch.rs the order is: 1) cooperative(); 2) async_trace_leaf(); 3) loop
https://github.com/tokio-rs/tokio/blob/dce1c9709a91cb18f3628adb0daec3ebb62d4413/tokio/src/sync/watch.rs#L1313-L1317

Let's keep them consistent!

tokio/src/sync/broadcast.rs
```diff
@@ -1555,7 +1598,7 @@ impl<T> Drop for Receiver<T> {
 
         if remaining_rx == 0 {
             self.shared.notify_last_rx_drop.notify_waiters();
-            tail.closed = true;
+            tail.closed_for_senders = true;
```
martin-g: IMO it will be more correct to set the new state before notifying the waiters.
Currently it works because of the acquired tail lock.

tokio/src/sync/broadcast.rs
```diff
@@ -1102,6 +1115,36 @@ impl<T> WeakSender<T> {
...
+    /// # async fn main() {
+    /// let (tx, mut rx1) = broadcast::channel::<u32>(16);
+    /// let mut rx2 = tx.subscribe();
+    ///
+    /// let _ = tx.send(10);
+    /// let weak = tx.downgrade();
+    ///
+    /// assert_eq!(rx1.recv().await.unwrap(), 10);
+    /// drop(rx1);
+    /// assert!(tx.closed().now_or_never().is_none());
```
martin-g:
```suggestion
    /// assert!(weak.closed().now_or_never().is_none());
```

tokio/src/sync/broadcast.rs
```diff
@@ -1102,6 +1115,36 @@ impl<T> WeakSender<T> {
...
+    ///
+    /// assert_eq!(rx2.recv().await.unwrap(), 10);
+    /// drop(rx2);
+    /// assert!(tx.closed().now_or_never().is_some());
```
martin-g:
```suggestion
    /// assert!(weak.closed().now_or_never().is_some());
```

tokio/src/sync/broadcast.rs
mladedav: Sorry, the test should be now correct again.

tokio/src/sync/broadcast.rs
mladedav: Fixed.

tokio/src/sync/broadcast.rs
mladedav: Sure. I even though so too but after noticing the  lock decided not to change things that were not necessary or could just cause confusion about why I'm changing them.

mladedav:
I don't think the netlify CI failures are caused by the changes here, when I tried to look for more information it gave me an internal server error so maybe they have some problems.

tokio/src/sync/broadcast.rs
```diff
@@ -887,24 +891,12 @@ impl<T> Sender<T> {
...
-                // Ensure the lock drops if the channel isn't closed
-                let tail = self.shared.tail.lock();
-                if tail.closed {
-                    return;
-                }
-            }
-
-            notified.await;
-        }
+        self.shared.closed_for_senders().await;
```
Darksonn, line 890: I was initially going to say this changes behavior of the existing function, but I guess it doesn't because if you can call this method then there is a sender, so the closed for receivers condition does not matter here.

tokio/src/sync/broadcast.rs
```diff
@@ -366,8 +366,11 @@ struct Tail {
     /// Number of active receivers.
     rx_cnt: usize,
 
-    /// True if the channel is closed.
-    closed: bool,
+    /// True if there are no receivers.
+    closed_for_senders: bool,
+
+    /// True if the last sender has been dropped.
+    closed_for_receivers: bool,
```
Darksonn: If we're going to split this into two booleans, then please use better names. For example, what about `has_senders` and `has_receivers`?

Darksonn (changes requested):

tokio/src/sync/broadcast.rs
mladedav: Yes, the closed boolean was a bit overloaded - if there were no receivers, senders could observe that there are no receivers; if there were no senders, receivers could observe that there are no senders.

I'll just add that `has_receivers` is now the same as `rx_cnt != 0` and `has_senders` is the same as `num_tx.load(Ordering::Relaxed) != 0`. The first one can be safely deleted because they are both in a locked `Tail`. I'm not entirely sure about the other one since `num_tx` is in `Shared` and not in `Tail` so it's not behind the same lock as the boolean currently is. I'm just mentioning this if you'd want me to delete `has_receivers` so that it cannot diverge and if I should look closer into deleting `has_senders` if it's safe.

tokio/src/sync/broadcast.rs
mladedav: Sure, renamed, thanks for the suggestion. I didn't think of flipping the logic hence the awkward names I came up with.

mladedav: I kept just the `closed_for_senders` name for the (private) method. We could do something like `wait_for_no_receivers` instead, but at that level this feels better to me.

tokio/src/sync/broadcast.rs
Darksonn: If you can delete the fields entirely, then that's great. But don't do it if it's too complex.

mladedav: Ok, I've deleted just `has_receivers`. I'm pretty sure that's safe. I'm not so sure about the other one so I'll leave it alone.

tokio/src/sync/broadcast.rs
```diff
@@ -366,8 +366,8 @@ struct Tail {
     /// Number of active receivers.
     rx_cnt: usize,
 
-    /// True if the channel is closed.
-    closed: bool,
+    /// True if there are no senders.
```
martin-g:
```suggestion
    /// True if there are strong senders.
```

tokio/src/sync/broadcast.rs
```diff
@@ -1102,6 +1104,37 @@ impl<T> WeakSender<T> {
         }
     }
 
+    /// A future which completes when the number of [Receiver]s subscribed to this channel reaches
+    /// zero.
```
martin-g:
```suggestion
    /// zero, regardless of whether strong senders still exist.
```
nit: Maybe clarify that strong senders are not taken into account for WeakSender::closed() ?!

tokio/src/sync/broadcast.rs
mladedav: Thanks, fixed.

tokio/src/sync/broadcast.rs
mladedav: I tried to have the same method as `Sender` but now that I think about it, that might be confusing because `Sender` can really only be interested in all receivers being done while someone might want to use `WeakSender` to checking either that there are no receivers or no senders.

So do you prefer this method to be named `closed` same as on `Sender` and behave in the same way or do you think it would be better to name it something different (though I can't think of a good name) to better signify that the receivers are gone and have an option to add a method that all strong senders are gone later?

For now applied exactly as suggested.

Issue #7897, opened 2026-02-08
Add unstable io-uring uring_cmd support


SidongYang:
## Summary
Tokio already has unstable `io-uring` support for `OpenAt`, `Read`, and `Write` in `tokio::fs` paths. I’d like to propose adding unstable support for `IORING_OP_URING_CMD`.

Prototype branch: `SidongYang:sidong/uring-cmd`
Prototype commit: `2d3c8cff`

## Motivation
Some Linux file/device operations are exposed as uring commands rather than regular read/write syscalls. Supporting `IORING_OP_URING_CMD` in Tokio’s existing unstable io-uring integration would make those operations available without leaving Tokio’s runtime/driver model.

## Proposed API (initial)
A minimal unstable API on `tokio::fs::File`:

- `async fn uring_cmd(&self, cmd_op: u32, cmd: [u8; 16], buf_index: Option<u16>) -> io::Result<u32>`

Behavior:
- Same unstable/feature gating as current io-uring support (`tokio_unstable`, `io-uring`, `rt`, `fs`, Linux).
- Runtime probe via `check_and_init(opcode::UringCmd16::CODE)`.
- Returns `ErrorKind::Unsupported` if unavailable.

## Scope
- Initial support for 16-byte payload (`UringCmd16`) only.
- Reuses existing op lifecycle + cancellation plumbing.

## Open questions
- Is `tokio::fs::File` the right public API location, or should this be exposed elsewhere (e.g. a lower-level io-uring module)?
- Should Tokio expose a more generic/extensible command payload API up front (or keep initial scope narrow)?

If this direction looks good, I can open/convert a PR with tests and docs aligned to maintainer guidance.

RajMandaliya:
Thanks for sharing this proposal!

I like that the API keeps things narrow with `UringCmd16`. One thing I’m curious about is how the cancellation and async runtime integration will be handled for these commands — for example, if a user awaits a long-running uring command and wants to cancel it via `tokio::sync` primitives, would that fit naturally into the existing op lifecycle?

I’m excited to try this out once the PR with tests is available.

SidongYang:
Implemented in #7901.

This wires `File::uring_cmd` through Tokio’s existing io_uring op lifecycle (`Op::uring_cmd16` + cancellable op path), and includes tests. Feedback welcome.

PR #7901, opened 2026-02-11
fs: add unstable io_uring uring_cmd support for File

SidongYang:
Implements File::uring_cmd using UringCmd16 with io_uring runtime/cancellation integration and tests.

Refs #7897

### DIFF
--- a/tokio/src/fs/file.rs
+++ b/tokio/src/fs/file.rs
@@ -451,6 +451,104 @@ impl File {
         asyncify(move || std.metadata()).await
     }
 
+    /// Executes an `IORING_OP_URING_CMD` operation on this file descriptor.
+    ///
+    /// This submits a 16-byte device/file-specific command payload that is
+    /// handled by the kernel subsystem backing this file descriptor.
+    /// Since the commands and their payloads are device specific, there is no
+    /// central list of possible commands. See the relevant kernel subsystem
+    /// documentation for the device you are interacting with.
+    ///
+    /// # io_uring support
+    ///
+    /// To use this API, enable `--cfg tokio_unstable`, the `io-uring` feature,
+    /// `Builder::enable_io`, and `Builder::enable_io_uring` on Linux.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because one could send privileged commands
+    /// or commands with payloads "telling" the kernel to read/write random
+    /// memory addresses specific to the command, which could fail due to
+    /// use-after-free or cause data corruption.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// # #[tokio::main]
+    /// # async fn main() -> std::io::Result<()> {
+    /// use tokio::fs::File;
+    ///
+    /// // Suppose you have a character device that supports a specific `uring_cmd`.
+    /// let file = File::open("/dev/my_custom_device").await?;
+    ///
+    /// // The `cmd_op` is defined by the device driver (similar to an ioctl number).
+    /// const MY_DRIVER_CMD_OP: u32 = 0x1234;
+    ///
+    /// // The 16-byte payload is also defined by the device driver.
+    /// // Often, this is used to pass small serialized C structs or pointers.
+    /// #[repr(C)]
+    /// struct MyCmdData {
+    ///     device_id: u32,
+    ///     size_param: u32,
+    ///     _pad: [u8; 8], // Pad to exactly 16 bytes
+    /// }
+    ///
+    /// let data = MyCmdData {
+    ///     device_id: 42,
+    ///     size_param: 100,
+    ///     _pad: [0; 8],
+    /// };
+    ///
+    /// // Safely transmute the 16-byte struct into a byte array
+    /// let cmd_payload: [u8; 16] = unsafe { std::mem::transmute(data) };
+    ///
+    /// // SAFETY: The command and payload must be exactly what the specific
+    /// // device driver expects to prevent undefined behavior in the kernel.
+    /// let _ = unsafe { file.uring_cmd(MY_DRIVER_CMD_OP, cmd_payload, None).await };
+    /// # Ok(())
+    /// # }
+    /// ```
+    ///
+    /// # Errors
+    ///
+    /// Returns [`std::io::ErrorKind::Unsupported`] if `io_uring` or
+    /// `IORING_OP_URING_CMD` is not available at runtime.
+    #[cfg(all(
+        tokio_unstable,
+        feature = "io-uring",
+        feature = "rt",
+        feature = "fs",
+        target_os = "linux"
+    ))]
+    pub async unsafe fn uring_cmd(
+        &self,
+        cmd_op: u32,
+        cmd: [u8; 16],
+        buf_index: Option<u16>,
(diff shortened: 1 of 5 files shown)


tokio/src/fs/file.rs
```diff
@@ -446,6 +446,58 @@ impl File {
...
+    /// Executes an `IORING_OP_URING_CMD` operation on this file descriptor.
+    ///
+    /// This submits a 16-byte device/file-specific command payload that is
+    /// handled by the kernel subsystem backing this file descriptor.
+    ///
+    /// # io_uring support
+    ///
+    /// To use this API, enable `--cfg tokio_unstable`, the `io-uring` feature,
+    /// and `Builder::enable_io_uring` on Linux.
+    ///
```
martin-g, line 511: Maybe add an `# Example` section ?!

tokio/tests/fs_uring.rs
```diff
@@ -145,6 +145,14 @@ async fn cancel_op_future() {
     assert!(res.is_cancelled());
 }
 
+#[tokio::test]
+async fn uring_cmd_is_available_via_file() {
+    let (tmp_file, _path): (Vec<NamedTempFile>, Vec<PathBuf>) = create_tmp_files(1);
+    let file = File::open(tmp_file[0].path()).await.unwrap();
+
+    let _ = file.uring_cmd(0, [0; 16], None).await;
```
martin-g: What is cmd=0 ? Maybe extract it into a local variable with a meaningful name ?!
My quick Google search didn't return anything useful ...

tokio/src/fs/file.rs
```diff
@@ -446,6 +446,58 @@ impl File {
...
+    /// # Errors
+    ///
+    /// Returns [`std::io::ErrorKind::Unsupported`] if `io_uring` or
+    /// `IORING_OP_URING_CMD` is not available at runtime.
```
martin-g, line 515: Maybe add an external link to the respective io_uring docs with the possible commands (`cmd_op`) and their respective payloads (`cmd`) ?!
Or there is no such one place where the cmds and their payloads are listed because they are device specific ?

tokio/tests/fs_uring.rs
martin-g: Assert that the result is Ok/Err ?!

tokio/src/fs/file.rs
```diff
@@ -446,6 +446,58 @@ impl File {
...
+    #[cfg(all(
+        tokio_unstable,
+        feature = "io-uring",
+        feature = "rt",
+        feature = "fs",
+        target_os = "linux"
+    ))]
+    pub async fn uring_cmd(
```
martin-g: I wonder whether this function should be marked as `unsafe`.
Reading more about IORING_OP_URING_CMD I understand that one could send privileged commands or commands with payloads "telling" the kernel to read random memory addresses specific to the command and this could fail due use-after-free or data corruption ...

tokio/src/fs/file.rs
```diff
@@ -446,6 +446,58 @@ impl File {
...
+        cmd_op: u32,
+        cmd: [u8; 16],
+        buf_index: Option<u16>,
+    ) -> io::Result<u32> {
+        use crate::runtime::driver::op::Op;
+        use std::os::fd::OwnedFd;
+
+        let mut inner = self.inner.lock().await;
+        inner.complete_inflight().await;
+        drop(inner);
```
martin-g: Is it OK to start other File ops while uring_cmd16() is still executing ?

tokio/src/fs/file.rs
```diff
@@ -446,6 +446,58 @@ impl File {
...
 
+    /// Executes an `IORING_OP_URING_CMD` operation on this file descriptor.
+    ///
+    /// This submits a 16-byte device/file-specific command payload that is
+    /// handled by the kernel subsystem backing this file descriptor.
+    ///
+    /// # io_uring support
+    ///
+    /// To use this API, enable `--cfg tokio_unstable`, the `io-uring` feature,
+    /// and `Builder::enable_io_uring` on Linux.
```
martin-g: It also depends on `Builder::enable_io` because of `let driver_handle = handle.inner.driver().io();`

tokio/src/fs/file.rs
SidongYang: Sure, I'll add an example for this.

tokio/tests/fs_uring.rs
SidongYang: cmd=0 is invalid uring-cmd. Usually it expects the file returns `EOPNOTSUPP` for invalid uring-cmd. So I'll add assertion for check the error.

tokio/src/fs/file.rs
SidongYang: Actually, I know the linux kernel code that returns `-EOPNOTSUPP` ([link](https://github.com/torvalds/linux/blob/7dff99b354601dd01829e1511711846e04340a69/io_uring/uring_cmd.c#L247)). Or it depends on its drivers implementation if driver has the uring_cmd callback.

tokio/src/fs/file.rs
SidongYang: I think it should be marked as unsafe. Typically, User uses a struct or pointer that would be transmuted to cmd args `[u8; 16]`. And the args points some other memory that should be pinned during uring_cmd process.

tokio/src/fs/file.rs
SidongYang: Thanks, I'll fix the comments.

tokio/src/fs/file.rs
SidongYang: Yes, the driver should guarantee its safety.

SidongYang:
@martin-g Thanks for the review! I've updated the PR and squashed the changes into a single commit.

Here are the updates based on your feedback and some self-review:
- **`unsafe` & Safety docs**: Marked `uring_cmd` as `unsafe` and added a `# Safety` section, as it allows arbitrary payload submission which can lead to use-after-free or memory corruption.
- **Realistic Example**: Added an `# Examples` section in the docs that demonstrates how to safely pack a `#[repr(C)]` struct into the 16-byte payload.
- **API Documentation**: Explicitly added `Builder::enable_io` to the required features in the docstring.
- **Unit Test Improvements**: Extracted the dummy payload into local variables, wrapped the call in `unsafe`, and added a specific assertion to check for `std::io::ErrorKind::Unsupported` (since a standard file without a `uring_cmd` implementation returns `-EOPNOTSUPP`).
- **Locking Refactor (Self-review)**: Simplified the `inner` locking logic (`complete_inflight`) to be more idiomatic and match the style used in `try_clone()`.
- **Concurrency**: Regarding your question about whether it's OK to start other File ops while `uring_cmd16()` is still executing: Yes, it is! The operation creates an independent `OwnedFd` via `self.std.try_clone()`, so it won't block the inner state of the Tokio file descriptor for subsequent operations.

Let me know if there's anything else that needs adjusting!

martin-g (approved):

SidongYang:
Rebased onto the latest `master` to resolve the merge conflict (in `tokio/src/runtime/driver/op.rs`).

PR #7902, opened 2026-02-11
loom: remove `StaticAtomicU64`

walker84837:
This PR addresses the TODO in tokio/src/loom/mocked.rs:65: <https://github.com/tokio-rs/tokio/blob/9dacb1c53e341b1183a89800d46ee33c6eabf7ad/tokio/src/loom/mocked.rs#L65-L66>

## Motivation

When running under `cfg(loom)`, `StaticAtomicU64` previously fell back to `std::sync::atomic::AtomicU64`. Because this bypasses loom's concurrency model, static atomic operations were not fully exercised by loom, which reduced coverage for code paths such as thread and task ID allocation.

## Solution

This PR replaces the std fallback with a loom-compatible implementation of `StaticAtomicU64`, allowing loom to explore all relevant interleavings while preserving the existing API and usage patterns.

The reason this wasn't replaced with [loom::sync::atomic::AtomicU64](https://docs.rs/loom/latest/loom/sync/atomic/struct.AtomicU64.html) is because it doesn't have const initialization.

Implementation notes:
- Accept `Ordering` arguments to match the API, though actual ordering semantics are not modeled, but operations are serialized via a Mutex.
- `compare_exchange_weak` behaves like `compare_exchange` (no spurious failures), which should be sufficient for loom testing.
- Back the static storage with `OnceLock<Mutex<u64>>` (under `cfg(loom)`) to support safe const initialization and deterministic simulation.
- Implement the full set of required atomic operations: `load`, `store`, `fetch_add`, `compare_exchange`, and `compare_exchange_weak`.
- Match the behavior and public API of the existing non-loom `StaticAtomicU64` variants to avoid changes at call sites.
- Note that while this implementation uses std::sync::OnceLock internally, this is compatible with loom testing as loom itself [wraps std::sync::Mutex](https://docs.rs/loom/latest/src/loom/sync/mutex.rs.html#8-11). The only limitation would be if StaticAtomicU64 needed to hold non-std types where loom doesn't provide equivalents.

Tests:
- Add loom tests in `sync::tests::loom_atomic_static_u64` covering basic atomic semantics, concurrent access patterns, const initialization, and compare-exchange behavior.
- All existing loom tests continue to pass.

### DIFF
--- a/tokio/src/runtime/thread_id.rs
+++ b/tokio/src/runtime/thread_id.rs
@@ -5,18 +5,12 @@ pub(crate) struct ThreadId(NonZeroU64);
 
 impl ThreadId {
     pub(crate) fn next() -> Self {
-        use crate::loom::sync::atomic::Ordering::Relaxed;
-
-        #[cfg(all(test, loom))]
-        use loom::sync::atomic::AtomicU64;
+        use crate::loom::sync::atomic::{AtomicU64, Ordering::Relaxed};
 
         #[cfg(all(test, loom))]
         crate::loom::lazy_static! {
             static ref NEXT_ID: AtomicU64 = AtomicU64::new(0);
         }
-
-        #[cfg(not(all(test, loom)))]
-        use crate::loom::sync::atomic::AtomicU64;
 
         #[cfg(not(all(test, loom)))]
         static NEXT_ID: AtomicU64 = AtomicU64::new(0);


tokio/src/sync/tests/loom_atomic_static_u64.rs
```diff
@@ -0,0 +1,75 @@
...
+        let result = atomic.compare_exchange_weak(150, 500, Relaxed, Relaxed);
+        assert_eq!(result, Err(400));
+        assert_eq!(atomic.load(Relaxed), 400);
+    });
+}
+
+#[test]
+fn static_atomic_u64_concurrent_access() {
+    loom::model(|| {
+        let atomic = Arc::new(StaticAtomicU64::new(0));
```
Darksonn: Please try changing this test to using a `static` instead of a local variable.

walker84837: I tried switching to a static as suggested, but under loom the const init doesn't work and tests panicked at <https://github.com/tokio-rs/tokio/blob/833cff8ba19dbdf1728b9a1947546c6eb8e1a59b/tokio/src/runtime/thread_id.rs#L10-L12>

To fix this in a Loom-compatible way, I added `#[cfg(all(test, loom))]` lazy_static for `NEXT_ID`. All Loom concurrency tests pass with the new static setup - I saw no failures.

tokio/src/runtime/thread_id.rs
```diff
@@ -7,6 +7,12 @@ impl ThreadId {
     pub(crate) fn next() -> Self {
         use crate::loom::sync::atomic::{Ordering::Relaxed, StaticAtomicU64};
 
+        #[cfg(all(test, loom))]
+        crate::loom::lazy_static! {
+            static ref NEXT_ID: StaticAtomicU64 = StaticAtomicU64::new(0);
+        }
```
Darksonn, line 13: But if we're using `lazy_static!`, can't we just use the normal `AtomicU64` here?

walker84837:
FYI, I rebased to avoid conflicts, and I realize the guidelines recommend merging `master` instead. Won't do it next time.

tokio/src/runtime/thread_id.rs
walker84837: Thanks for the comment! I agree, the initialization doesn't need to be const-safe in this specific use, so we can just use `loom::sync::atomic::AtomicU64` directly. Switched it to AtomicU64 as suggested.

Darksonn: If we no longer need `StaticAtomicU64`, can we remove it?

walker84837: Yes, we could remove `StaticAtomicU64`, but doing so would require using `loom::lazy_static!` (or similar lazy initialization) outside `#[cfg(all(test, loom))]`.

Currently, everything in the `loom` crate (including `lazy_static!`) is re-exported under `#[cfg(all(test, loom))]` (as defined in loom/mod.rs): <https://github.com/tokio-rs/tokio/blob/8a4f82869f7e3968f7f717d6572e06ad3aed98a7/tokio/src/loom/mocked.rs#L1>

If we'd rather remove the abstraction entirely, I'm open to editing the PR accordingly.

Darksonn: Outside of `#[cfg(all(test, loom))]`, you should be able to create a `AtomicU64` from a const fn, so the lazy static can be used only for `#[cfg(all(test, loom))]`.

walker84837: I'm happy to update the PR to remove `StaticAtomicU64` while keeping the correct initialization strategy for both loom and non-loom builds, if that is the preferred solution.

tokio/src/loom/mocked.rs
```diff
@@ -61,9 +61,100 @@ pub(crate) mod sync {
...
+            ///
+            /// Note: Ordering parameter is ignored; Loom uses Mutex to simulate atomic behavior.
+            pub(crate) fn fetch_add(&self, val: u64, _order: Ordering) -> u64 {
+                let mut lock = self.inner().lock();
+                let prev = *lock;
+                *lock = prev + val;
+                prev
+            }
+
+            /// Atomically compares and exchanges the value, but may fail spuriously.
```
martin-g: > but may fail spuriously.

This contradicts with the PR description:
```
compare_exchange_weak behaves like compare_exchange (no spurious failures), which should be sufficient for loom testing.
```
Should the docstring be updated ?!

tokio/src/loom/mocked.rs
```diff
@@ -61,9 +61,100 @@ pub(crate) mod sync {
 
     pub(crate) mod atomic {
         pub(crate) use loom::sync::atomic::*;
+        use std::sync::OnceLock;
 
-        // TODO: implement a loom version
-        pub(crate) type StaticAtomicU64 = std::sync::atomic::AtomicU64;
+        pub(crate) struct StaticAtomicU64 {
+            init: u64,
+            cell: OnceLock<crate::loom::sync::Mutex<u64>>,
```
martin-g: The PR description says that it is OK to use non-Loom OnceLock here:
```
Note that while this implementation uses std::sync::OnceLock internally, 
this is compatible with loom testing as loom itself [wraps std::sync::Mutex]
(https://docs.rs/loom/latest/src/loom/sync/mutex.rs.html#8-11). The only 
limitation would be if StaticAtomicU64 needed to hold non-std types where loom 
doesn't provide equivalents.
```
What bothers me is that the `init` part of `get_or_init()` won't be tracked. Is this OK ?

tokio/src/runtime/thread_id.rs
martin-g: BTW task/id.rs still uses StaticAtomicU64 but slightly differently:
https://github.com/tokio-rs/tokio/blob/a48c3269519ce8f88f1a6b5adebd07fcfa86128e/tokio/src/runtime/task/id.rs#L73-L79

tokio/src/loom/mocked.rs
```diff
@@ -61,9 +61,100 @@ pub(crate) mod sync {
...
+            /// Note: Ordering parameters are ignored; Loom uses Mutex to simulate atomic behavior.
+            pub(crate) fn compare_exchange(
+                &self,
+                current: u64,
+                new: u64,
+                success: Ordering,
+                failure: Ordering,
+            ) -> Result<u64, u64> {
+                self.compare_exchange_weak(current, new, success, failure)
+            }
```
martin-g: Is there a reason why `fetch_or()` is not implemented here ?
atomic_u64_as_mutex.rs provides it: 
https://github.com/tokio-rs/tokio/blob/a48c3269519ce8f88f1a6b5adebd07fcfa86128e/tokio/src/loom/std/atomic_u64_as_mutex.rs#L37-L42

tokio/src/loom/mocked.rs
```diff
@@ -61,9 +61,100 @@ pub(crate) mod sync {
...
+                *self.inner().lock()
+            }
+
+            /// Atomically adds to the current value, returning the previous value.
+            ///
+            /// Note: Ordering parameter is ignored; Loom uses Mutex to simulate atomic behavior.
+            pub(crate) fn fetch_add(&self, val: u64, _order: Ordering) -> u64 {
+                let mut lock = self.inner().lock();
+                let prev = *lock;
+                *lock = prev + val;
```
martin-g: This will panic on overflow.
Same for https://github.com/tokio-rs/tokio/blob/a48c3269519ce8f88f1a6b5adebd07fcfa86128e/tokio/src/loom/std/atomic_u64_as_mutex.rs#L30-L35

https://doc.rust-lang.org/std/sync/atomic/struct.AtomicI64.html#method.fetch_add always wraps on overflow.

tokio/src/loom/mocked.rs
```diff
@@ -61,9 +61,100 @@ pub(crate) mod sync {
 
     pub(crate) mod atomic {
         pub(crate) use loom::sync::atomic::*;
+        use std::sync::OnceLock;
 
-        // TODO: implement a loom version
-        pub(crate) type StaticAtomicU64 = std::sync::atomic::AtomicU64;
+        pub(crate) struct StaticAtomicU64 {
```
martin-g: What is the difference with tokio/src/loom/std/atomic_u64_static_once_cell.rs ?

tokio/src/loom/mocked.rs
walker84837: Good catch - that doc comment is stale from an earlier draft. I'll update the docstring and any relevant documentation to match the current behavior.

tokio/src/loom/mocked.rs
walker84837: That's a good point. The initialization path of the `std::sync::OnceLock` in `StaticAtomicU64` would not be modeled by loom.

I chose it since loom doesn't currently provide a checked `OnceLock` equivalent, and reimplementing it here felt out of scope.

tokio/src/loom/mocked.rs
walker84837: I didn't implement `fetch_or` in `StaticAtomicU64` since there are currently no call sites, and adding it would introduce unused code warnings in CI unless I add `#[allow(dead_code)]` to the function. If we prefer API parity with the underlying atomic type, I can add it.

tokio/src/loom/mocked.rs
walker84837: You're right - I'll update the implementation to use `wrapping_add` to preserve the same semantics. For consistency, we should also consider aligning `atomic_u64_as_mutex.rs` accordingly.

tokio/src/loom/mocked.rs
walker84837: Originally, `StaticAtomicU64` was intended as a unified abstraction for both loom and non-loom builds.

With the current setup (like in `thread_id.rs`) `crate::loom::lazy_static!` is used for `#[cfg(all(test, loom))]`, so Loom tests use that initialization, while non-Loom builds continue to use `StaticAtomicU64` with `static`.

tokio/src/runtime/thread_id.rs
Darksonn: If we can avoid adding a new atomic type, we should do so. Especially since my impression from https://github.com/tokio-rs/tokio/pull/7902#discussion_r2805587233 is that the current implementation doesn't actually work when used in a `static`.

walker84837: I want to make sure I'm aligned properly before continuing, since this would affect the structure of the changes.

My understanding of the `const fn` suggestion is that outside `#[cfg(all(test, loom))]`, we rely on `AtomicU64::new` being `const` (via the std implementation), and use a plain `static`, removing `StaticAtomicU64` entirely.

Under the loom test configuration, we'd instead use `loom::lazy_static!` since the loom `AtomicU64` doesn't support const initialization.

That would look something like:

```rust
#[cfg(not(all(test, loom)))]
static NEXT_ID: AtomicU64 = AtomicU64::new(0);

#[cfg(all(test, loom))]
crate::loom::lazy_static! {
    static ref NEXT_ID: AtomicU64 = AtomicU64::new(0);
}
```

This means we'd use a plain `static` in non-loom builds and `lazy_static!` only under loom.

Could you confirm that this is the direction you'd like me to take, or if you have a different structure in mind? I'd like to settle on the final solution before implementing broader changes.

Darksonn: Yes, that looks correct.

tokio/src/runtime/task/id.rs
```diff
@@ -68,15 +68,20 @@ impl fmt::Display for Id {
...
 
         #[cfg(all(test, loom))]
         crate::loom::lazy_static! {
-            static ref NEXT_ID: StaticAtomicU64 = StaticAtomicU64::new(1);
+            static ref NEXT_ID: AtomicU64 = AtomicU64::new(1);
         }
 
         #[cfg(not(all(test, loom)))]
-        static NEXT_ID: StaticAtomicU64 = StaticAtomicU64::new(1);
+        use crate::loom::sync::atomic::AtomicU64;
```
Darksonn: You seem to be importing this twice. Please just import it once.

walker84837: Good catch, thanks. I'm a bit tied up at the moment, but I'll fix the duplicate import soon.

Darksonn: Same ptten remains in tokio/src/runtime/thread_id.rs

PR #7904, opened 2026-02-12
io: Explain how to flush stdout/stderr

stepancheg:
## Motivation

Help other people debugging lost stdout.

## Solution

Explain behavior of `tokio::io::stdout` and `tokio::io::stderr`.

## Also

Fixed typo in `stderr.rs`.

### DIFF
--- a/tokio/src/io/stderr.rs
+++ b/tokio/src/io/stderr.rs
@@ -29,7 +29,7 @@ cfg_io_std! {
     ///
     /// #[tokio::main]
     /// async fn main() -> io::Result<()> {
-    ///     let mut stderr = io::stdout();
+    ///     let mut stderr = io::stderr();
     ///     stderr.write_all(b"Print some error here.").await?;
     ///     Ok(())
     /// }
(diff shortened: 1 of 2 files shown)


tokio/src/io/stderr.rs
```diff
@@ -50,6 +50,20 @@ cfg_io_std! {
     /// to occur as a single write, so multiple threads writing data with
     /// [`write_all`] may result in interleaved output.
     ///
+    /// Note that unlike [`std::io::stderr`], each call to this `stderr()`
+    /// produces a new writer, so for example, this program does not flush stderr:
```
martin-g:
```suggestion
    /// produces a new writer, so for example, this program does **not** flush stderr:
```

tokio/src/io/stdout.rs
```diff
@@ -74,6 +74,20 @@ cfg_io_std! {
     /// to occur as a single write, so multiple threads writing data with
     /// [`write_all`] may result in interleaved output.
     ///
+    /// Note that unlike [`std::io::stdout`], each call to this `stdout()`
+    /// produces a new writer, so for example, this program does not flush stdout:
```
martin-g:
```suggestion
    /// produces a new writer, so for example, this program does **not** flush stdout:
```

stepancheg:
Applied bold for "not".

PR #7907, opened 2026-02-13
Implement `AsyncRead` for files with io uring

threecgreen:
## Motivation

Add an alternative for reading files through the `AsyncRead` trait without using `spawn_blocking`. `fs::read` can already use io-uring, so I used as much of the existing work there as possible.

## Solution

I used #7713 as a reference for the changes required. I changed `fs::read` from an `OwnedFd` to an `ArcFd` to share more logic with the `AsyncRead` implementation and the overhead should be minimal. I refactored `uring::Read` to be generic over the buffer type so `fs::read` could continue to use a `Vec`. 

Like the other places where experimental io-uring support has been added, it'll fall back on `spawn_blocking` if  `io-uring` isn't enabled. Unlike the other implementations, we're checking if it's initialized within a poll method and `check_and_init` is now async. `poll_read_inner` uses a three-way check: `is_uring_ready` for the fast path when io uring is already initialized (e.g. via `File::open`), a lazy-init path that calls `check_and_init` inside an async task when io uring hasn't been probed  yet (covering `File::from_std`), and a direct `spawn_blocking` fallback when io uring has been probed but is unsupported.

## Testing

Because the existing fs unit tests were built around `spawn_blocking`, reworking them and the mock `JoinHandle` to support both blocking and async tasks would have been an invasive change for functionality still gated behind the `tokio_unstable` flag. Instead, I wrote integration tests that use the actual io-uring driver, which more fully test the new functionality while limiting the impact on existing, stable code.

### DIFF
--- a/tokio/src/fs/file.rs
+++ b/tokio/src/fs/file.rs
@@ -987,14 +987,7 @@ impl Inner {
         }
 
         // Fallback: spawn_blocking
-        let join = spawn_blocking(move || {
-            let mut buf = buf;
-            // SAFETY: the `Read` implementation of `std` does not
-            // read from the buffer it is borrowing and correctly
-            // reports the length of the data written into the buffer.
-            let res = unsafe { buf.read_from(&mut &*std, max_buf_size) };
-            (Operation::Read(res), buf)
-        });
+        let join = Self::spawn_blocking_read(buf, std, max_buf_size);
         Ok(join)
     }
 
@@ -1052,16 +1045,7 @@ impl Inner {
             let fd: crate::io::uring::utils::ArcFd = std;
             Self::uring_read(fd, buf, max_buf_size).await
         } else {
-            match spawn_blocking(move || {
-                let mut buf = buf;
-                // SAFETY: the `Read` implementation of `std` does not
-                // read from the buffer it is borrowing and correctly
-                // reports the length of the data written into the buffer.
-                let res = unsafe { buf.read_from(&mut &*std, max_buf_size) };
-                (Operation::Read(res), buf)
-            })
-            .await
-            {
+            match Self::spawn_blocking_read(buf, std, max_buf_size).await {
                 Ok(result) => result,
                 Err(e) => (
                     Operation::Read(Err(io::Error::new(io::ErrorKind::Other, e))),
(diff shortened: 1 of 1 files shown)


.helix/languages.toml
ADD-SP, line 1: please don't check in irrelevant files.

tokio/src/fs/read_uring.rs
```diff
@@ -110,13 +114,13 @@ async fn small_probe_read(
...
-    mut fd: OwnedFd,
+    mut fd: ArcFd,
     mut buf: Vec<u8>,
     offset: &mut u64,
     read_len: u32,
-) -> io::Result<(OwnedFd, Vec<u8>, bool)> {
+) -> io::Result<(ArcFd, Vec<u8>, bool)> {
     loop {
-        let (res, r_fd, r_buf) = Op::read(fd, buf, read_len, *offset).await;
+        let (res, r_fd, r_buf) = Op::read_at(fd, buf, read_len as usize, *offset).await;
```
Darksonn, line 119: I don't really want to make `tokio::fs::read()` worse by forcing it to use `ArcFd` when it does not really need it, just because `File` wants it.

tokio/src/fs/file.rs
```diff
@@ -952,6 +951,116 @@ cfg_windows! {
...
+        {
+            let fd: crate::io::uring::utils::ArcFd = std;
+            Self::uring_read(fd, buf, max_buf_size).await
+        } else {
+            let mut buf = buf;
+            // SAFETY: the `Read` implementation of `std` does not
+            // read from the buffer it is borrowing and correctly
+            // reports the length of the data written into the buffer.
+            let res = unsafe { buf.read_from(&mut &*std, max_buf_size) };
+            (Operation::Read(res), buf)
```
Darksonn: This must be in `spawn_blocking`, right?

threecgreen: yes, I've fixed that

tokio/src/fs/read_uring.rs
threecgreen: understood. I initially did that because I ran into some trait bound issues with `Arc<dyn AsRawFd>` not implementing `AsRawFd` as the generic `Arc<T: AsRawFd>` impl requires `T` to be sized. I created a `UringFd` helper trait to get around that. Let me know if that makes sense

tokio/src/fs/file.rs
```diff
@@ -952,6 +951,126 @@ cfg_windows! {
...
+
+        // Fallback: spawn_blocking
+        let join = spawn_blocking(move || {
+            let mut buf = buf;
+            // SAFETY: the `Read` implementation of `std` does not
+            // read from the buffer it is borrowing and correctly
+            // reports the length of the data written into the buffer.
+            let res = unsafe { buf.read_from(&mut &*std, max_buf_size) };
+            (Operation::Read(res), buf)
+        });
```
Darksonn: This code is duplicated in `lazy_init_read`, please pull out to separate function.

threecgreen:
thank you for the help and maintaining tokio

Darksonn:
Thanks for the PR, and sorry about the delay.

PR #7910, opened 2026-02-15
runtime: fix double-increment of num_idle_threads on blocking pool shutdown

### LINKED ISSUE #6439: BlockingPool num_idle_threads is wrongly double-increased when shutting down

veeceey:
When a blocking pool worker thread hits the shutdown path, `num_idle_threads` gets incremented twice but only decremented once on thread exit, leaving the metric over-counted.

**What happens:**

The worker loop increments `num_idle_threads` when entering IDLE state (before the condvar wait). During shutdown, `condvar.notify_all()` wakes all idle threads, but unlike a normal task spawn, it does *not* decrement `num_idle_threads` or increment `num_notify`. The woken thread sees `shared.shutdown == true`, exits the inner while loop, and enters the `if shared.shutdown` block which had a second `inc_num_idle_threads()` call. On thread exit, only one `dec_num_idle_threads()` happens, so the net effect is +1 per idle thread at shutdown time.

**The fix:**

Remove the extra `inc_num_idle_threads()` in the shutdown path. The thread is already counted as idle from the increment at the top of the IDLE section, so the balance is: +1 on entering IDLE, -1 on thread exit.

Closes #6439

### DIFF
--- a/tokio/tests/rt_unstable_metrics.rs
+++ b/tokio/tests/rt_unstable_metrics.rs
@@ -63,6 +63,38 @@ fn num_idle_blocking_threads() {
 }
 
 #[test]
+fn num_idle_blocking_threads_is_zero_after_shutdown() {
+    let rt = current_thread();
+    let handle = rt.handle().clone();
+
+    // Spawn a blocking task to create a worker thread.
+    let _ = rt.block_on(rt.spawn_blocking(move || {}));
+
+    // Wait for the thread to become idle.
+    rt.block_on(async {
+        time::sleep(Duration::from_millis(5)).await;
+    });
+    if handle.metrics().num_idle_blocking_threads() == 0 {
+        rt.block_on(async {
+            time::sleep(Duration::from_secs(1)).await;
+        });
+    }
+    assert_eq!(1, handle.metrics().num_idle_blocking_threads());
+
+    // Drop the runtime, which triggers shutdown and joins all blocking
+    // threads. Before the fix for #6439, the shutdown path incremented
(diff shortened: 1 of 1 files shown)


Darksonn:
Thanks! Can you add a test for this?

veeceey:
Added a regression test in `rt_unstable_metrics` — it grabs a `Handle` before dropping the runtime, spawns a blocking task, waits for the thread to go idle, then drops the runtime and asserts `num_idle_blocking_threads() == 0`. Verified it fails with the double-increment present and passes with the fix.

veeceey:
@Darksonn I've added the regression test as requested. Would you mind taking another look? Thanks!

ADD-SP (approved):
Thanks!

PR #7911, opened 2026-02-15
feat(net): add Unix socket ancillary data API for fd passing (SCM_RIGHTS)

MohibShaikh:
## Summary
Adds `send_vectored_with_ancillary` and `recv_vectored_with_ancillary` methods to `UnixStream` and `UnixDatagram`, enabling file descriptor passing over Unix sockets via `sendmsg`/`recvmsg` with `SCM_RIGHTS` control messages.

## New types
- `SocketAncillary` – buffer for sending/receiving ancillary data
- `AncillaryData` – enum for received control messages
- `ScmRights` – iterator over received file descriptors
- `Messages` – iterator over control messages

## Platform support
- `MSG_CMSG_CLOEXEC` on Linux, FreeBSD, Android, DragonFly, NetBSD, OpenBSD
- `fcntl` fallback on macOS and other Unix platforms

## Testing
- Ran `cargo test -p tokio --test uds_stream --test uds_datagram` in CodeSpaces (Linux)

Closes tokio-rs#2975

### DIFF
--- a/tokio/src/net/unix/ancillary.rs
+++ b/tokio/src/net/unix/ancillary.rs
@@ -1,0 +1,347 @@
+use std::fmt;
+use std::marker::PhantomData;
+use std::mem;
+use std::os::unix::io::RawFd;
+
+/// Error returned when an ancillary data operation fails.
+///
+/// This typically occurs when the buffer provided to [`SocketAncillary`]
+/// is too small to hold the requested control messages.
+#[derive(Debug)]
+pub struct AncillaryError;
+
+impl fmt::Display for AncillaryError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "ancillary buffer too small")
+    }
+}
+
+impl std::error::Error for AncillaryError {}
+
+/// Ancillary data received from a Unix socket control message.
+///
+/// This enum is yielded by the [`Messages`] iterator when parsing
+/// received ancillary data from [`SocketAncillary`].
+pub enum AncillaryData<'a> {
+    /// A set of file descriptors received via `SCM_RIGHTS`.
+    ///
+    /// File descriptors received this way are new descriptors in the
+    /// receiving process. The caller is responsible for closing them
+    /// (e.g., by wrapping in [`OwnedFd`](std::os::unix::io::OwnedFd)).
+    ScmRights(ScmRights<'a>),
+}
+
+impl fmt::Debug for AncillaryData<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            AncillaryData::ScmRights(_) => f.debug_tuple("ScmRights").finish(),
+        }
+    }
+}
+
+/// An iterator over file descriptors received via `SCM_RIGHTS`.
+///
+/// This struct is created by [`AncillaryData::ScmRights`].
+/// Each call to [`next()`](Iterator::next) yields a [`RawFd`].
+/// The caller is responsible for managing the lifetime of these
+/// file descriptors.
+pub struct ScmRights<'a> {
+    data: &'a [u8],
+    offset: usize,
+}
+
+impl fmt::Debug for ScmRights<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("ScmRights").finish()
+    }
+}
+
+impl<'a> ScmRights<'a> {
+    fn new(data: &'a [u8]) -> Self {
+        ScmRights { data, offset: 0 }
+    }
+}
+
+impl Iterator for ScmRights<'_> {
+    type Item = RawFd;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let fd_size = mem::size_of::<RawFd>();
+        if self.offset + fd_size > self.data.len() {
+            return None;
+        }
+        let mut fd_bytes = [0u8; mem::size_of::<RawFd>()];
+        fd_bytes.copy_from_slice(&self.data[self.offset..self.offset + fd_size]);
+        self.offset += fd_size;
+        Some(RawFd::from_ne_bytes(fd_bytes))
+    }
+}
+
+/// An iterator over control messages in a [`SocketAncillary`] buffer.
+///
+/// Created by [`SocketAncillary::messages`].
+pub struct Messages<'a> {
+    current: *const libc::cmsghdr,
+    msg: libc::msghdr,
+    _marker: PhantomData<&'a [u8]>,
+}
+
+impl fmt::Debug for Messages<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Messages").finish()
+    }
+}
+
+// SAFETY: Messages only reads from a shared reference to the buffer.
+unsafe impl Send for Messages<'_> {}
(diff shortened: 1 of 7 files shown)


tokio/src/net/unix/ancillary.rs
```diff
@@ -0,0 +1,347 @@
+use std::fmt;
+use std::marker::PhantomData;
+use std::mem;
+use std::os::unix::io::RawFd;
+
+/// Error returned when an ancillary data operation fails.
+///
+/// This typically occurs when the buffer provided to [`SocketAncillary`]
```
martin-g, line 8: _typically_ !
Would it make sense to change AncillaryError to an enum:
```rust
pub enum AncillaryError {
  BUFFER_TOO_SMALL
}
```
?
This way it could be extended with other kinds of causes in the future, if needed.

tokio/src/net/unix/ancillary.rs
```diff
@@ -0,0 +1,347 @@
...
+}
+
+// SAFETY: Messages only reads from a shared reference to the buffer.
+unsafe impl Send for Messages<'_> {}
+// SAFETY: Messages only reads from a shared reference to the buffer.
+unsafe impl Sync for Messages<'_> {}
+
+impl<'a> Messages<'a> {
+    fn new(buffer: &'a [u8], length: usize) -> Self {
+        let mut msg: libc::msghdr = unsafe { mem::zeroed() };
```
martin-g, line 102: Please add a `// SAFETY: ...` comment to each `unsafe {...}`

tokio/src/net/unix/ancillary.rs
```diff
@@ -0,0 +1,347 @@
...
+        target_os = "linux",
+        target_os = "android",
+        target_os = "freebsd",
+        target_os = "dragonfly",
+        target_os = "netbsd",
+        target_os = "openbsd",
+    )))]
+    pub(crate) fn set_cloexec(&self) -> std::io::Result<()> {
+        for msg in self.messages() {
+            if let AncillaryData::ScmRights(fds) = msg {
```
martin-g, line 339:
```suggestion
            let AncillaryData::ScmRights(fds) = msg {
```

tokio/src/net/unix/ancillary.rs
```diff
@@ -0,0 +1,347 @@
...
+
+        self.length += space;
+        Ok(())
+    }
+
+    /// Returns `true` if the ancillary data was truncated during receive.
+    ///
+    /// This happens when the buffer provided was too small to hold all
+    /// control messages. Any file descriptors that did not fit are lost.
+    pub fn is_truncated(&self) -> bool {
```
martin-g, line 283: nit: Consider annotating the method with `must_use`
```suggestion
    #[must_use]
    pub fn is_truncated(&self) -> bool {
```
Same for `messages()`, `len()` and `is_empty()` below.

tokio/src/net/unix/ancillary.rs
```diff
@@ -0,0 +1,347 @@
...
+            if cmsg.cmsg_level == libc::SOL_SOCKET && cmsg.cmsg_type == libc::SCM_RIGHTS {
+                let data_ptr = libc::CMSG_DATA(cmsg as *const _ as *mut _);
+                #[allow(clippy::unnecessary_cast)]
+                let data_len =
+                    cmsg.cmsg_len as usize - (data_ptr as usize - cmsg as *const _ as usize);
+                let data = std::slice::from_raw_parts(data_ptr, data_len);
+                Some(AncillaryData::ScmRights(ScmRights::new(data)))
+            } else {
+                // Skip unknown control message types, try next
+                self.next()
```
martin-g, line 139: This recursion is prone to stack overflow when fed with malicious input (many unknown message types).
Maybe re-work it to a `loop` ?!

tokio/src/net/unix/ancillary.rs
```diff
@@ -0,0 +1,347 @@
...
+                }
+            };
+
+            if cmsg.is_null() {
+                return Err(AncillaryError);
+            }
+
+            (*cmsg).cmsg_level = libc::SOL_SOCKET;
+            (*cmsg).cmsg_type = libc::SCM_RIGHTS;
+            (*cmsg).cmsg_len = libc::CMSG_LEN(fd_bytes_len as libc::c_uint) as _;
```
martin-g, line 269: Here a `usize` is casted to c_uint (u32) and may lead to an overflow.

tokio/src/net/unix/ancillary.rs
```diff
@@ -0,0 +1,347 @@
...
+
+            let data_ptr = libc::CMSG_DATA(cmsg);
+            std::ptr::copy_nonoverlapping(fds.as_ptr() as *const u8, data_ptr, fd_bytes_len);
```
martin-g, line 272: Here this still uses the original usize (fd_bytes_len) but cmsg_len might be a unexpected value if there was an overflow.

tokio/src/net/unix/ancillary.rs
```diff
@@ -0,0 +1,347 @@
...
+            buffer,
+            length: 0,
+            truncated: false,
+        }
+    }
+
+    /// Returns the minimum buffer size needed to send or receive
+    /// `num_fds` file descriptors via `SCM_RIGHTS`.
+    pub fn buffer_size_for_rights(num_fds: usize) -> usize {
+        unsafe { libc::CMSG_SPACE((num_fds * mem::size_of::<RawFd>()) as libc::c_uint) as usize }
```
martin-g, line 215: This may lead to a integer overflow if `num_fds` is a big number (malicious input).

tokio/tests/uds_datagram.rs
```diff
@@ -423,3 +423,65 @@ async fn poll_ready() -> io::Result<()> {
...
+
+    let n = b
+        .recv_vectored_with_ancillary(
+            &mut [io::IoSliceMut::new(&mut recv_buf)],
+            &mut recv_ancillary,
+        )
+        .await
+        .unwrap();
+    assert_eq!(n, 5);
+    assert_eq!(&recv_buf[..5], b"dgram");
```
martin-g, line 467:
```suggestion
    assert_eq!(&recv_buf[..5], b"dgram");
    assert!(!recv_ancillary.is_truncated());
```
as the new test does in uds_stream.rs

tokio/src/net/unix/ancillary.rs
```diff
@@ -0,0 +1,347 @@
...
+    }
+}
+
+// SAFETY: Messages only reads from a shared reference to the buffer.
+unsafe impl Send for Messages<'_> {}
+// SAFETY: Messages only reads from a shared reference to the buffer.
+unsafe impl Sync for Messages<'_> {}
+
+impl<'a> Messages<'a> {
+    fn new(buffer: &'a [u8], length: usize) -> Self {
```
martin-g, line 101: Is there a use case when the `buffer.len()` would be different than the passed `length` ?
Maybe remove the `length` argument ?!

Darksonn:
I'm sorry but this is not going to happen. You're exposing a lot of API surface, and once we add it we can never change it. I'm quite sure the API as proposed is not designed correctly, as it operates on `RawFd` rather than using owned fd types, which are the modern way to go about this kind of thing. Even if you suggest a different API, I need to see the API be used in practice in a separate crate to feel confident that it's right.

Also the implementation is really complex, and I'd prefer if this was implemented by std, or at least a separate crate such as socket2. We don't really want to maintain it directly in Tokio.

For the above reasons, I am closing this PR.

PR #7912, opened 2026-02-16
tokio-util: fix is_readable should be buffer empty or not

cuiweixie:


tokio-util/src/codec/framed_impl.rs
```diff
@@ -65,14 +65,15 @@ impl Default for WriteFrame {
...
+        let is_readable = !buffer.is_empty();
         let size = buffer.capacity();
         if size < INITIAL_CAPACITY {
             buffer.reserve(INITIAL_CAPACITY - size);
         }
 
         Self {
             buffer,
-            is_readable: size > 0,
+            is_readable,
```
Darksonn, line 76: Is this fixing a bug? If so, please add a test for it.

cuiweixie: Done, add unittests!

Darksonn: These tests just check the private field. Does this affect the public behavior of the module? I'd rather have a test showing how this impacts the public API of tokio-util. And in the `tests/` directory next to `src/`.

cuiweixie: add to tests

martin-g (approved):

PR #7916, opened 2026-02-17
chore: unify workspace dependencies with cargo-rail

loadingalias:
## Summary
Supersedes closed #7914 and #7915.

This PR reapplies workspace dependency unification using `cargo-rail v0.10.7` after the Tokio specific workspace member handling fixes.

## Why
Tokio has many workspace members that depend on each other. Unification needs to avoid local vs registry split resolution and keep member dependency behavior stable across CI targets.

## What changed
- Unified **13 deps** into root `[workspace.dependencies]`
- Updated member manifests to consume workspace deps where applicable
- Fixed **7 undeclared feature** borrow issues across 5 crates
- Wrote workspace `rust-version = "1.85.0"` based on dependency graph analysis
- Manifest-only maintenance change (`Cargo.toml` updates)

## Unified dependencies
- bytes
- futures
- libc
- parking_lot
- rand
- slab
- tempfile
- tokio
- tokio-macros
- tokio-stream
- tokio-test
- tokio-util
- tracing

## Validation
Generated with:
- `cargo rail unify --check --explain`
- `cargo rail unify`

Recommended CI validation surfaces:
- `cargo check --workspace --all-targets`
- Tokio matrix jobs (including wasm/fuzz/semver lanes)

## Notes
- This PR replaces earlier attempts to keep history clean after branch resets (#7914 & #7915 - sorry!)
- No runtime Rust source changes are included.

### DIFF
--- a/.config/rail.toml
+++ b/.config/rail.toml
@@ -1,0 +1,65 @@
+# cargo-rail configuration for tokio fork
+# Documentation: https://github.com/loadingalias/cargo-rail
+
+# Targets
+targets = [
+  "aarch64-pc-windows-msvc",
+  "aarch64-unknown-linux-gnu",
+  "arm-linux-androideabi",
+  "armv5te-unknown-linux-gnueabi",
+  "armv7-sony-vita-newlibeabihf",
+  "armv7-unknown-linux-gnueabihf",
+  "i686-unknown-linux-gnu",
+  "powerpc-unknown-linux-gnu",
+  "powerpc64-unknown-linux-gnu",
+  "wasm32-unknown-unknown",
+  "wasm32-wasip1",
+  "wasm32-wasip1-threads",
+  "x86_64-fortanix-unknown-sgx",
+  "x86_64-unknown-haiku",
+  "x86_64-unknown-linux-gnu",
+  "x86_64-unknown-linux-musl",
+  "x86_64-unknown-redox",
+]
+
+[unify]
+include_paths = true
+include_renamed = false
+
+pin_transitives = false  # enable for hakari/workspace-hack users
+transitive_host = "root"  # only used if pin_transitives = true
+
+strict_version_compat = true
+exact_pin_handling = "warn"
+major_version_conflict = "warn"  # warn = skip, bump = force highest
+
+msrv = true
+msrv_source = "workspace"  # deps | workspace | max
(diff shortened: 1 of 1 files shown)


ADD-SP (commented):
Thanks for you PR!

Since it is possible that `tokio-util` and `tokio-stream` require different minimal version of dependencies like `tokio`, moving such dependencies to workspace-level won't work. So i will closed this PR.

loadingalias:
You can simply exclude [tokio] and still solve the undeclared features bugs.

Cargo.toml
```diff
@@ -33,3 +33,21 @@ unexpected_cfgs = { level = "warn", check-cfg = [
...
+futures = { version = "^0.3.32", features = ["async-await", "thread-pool"] } #unified
+tokio-stream = { path = "tokio-stream", version = "^0.1.18" } #unified
+tokio-test = { path = "tokio-test", version = "^0.4.5" } #unified
+bytes = "^1.11.1" #unified
+tempfile = "^3.25.0" #unified
+tokio-macros = { path = "tokio-macros", version = "^2.6.0" } #unified
+parking_lot = "^0.12.5" #unified
+
+[workspace.package]
+rust-version = "1.85.0"
```
martin-g, line 53: @loadingalias What leads to this bump ?
The current MSRV is 1.71.0 and the CI is OK

loadingalias: Thanks. The bump was tool-generated, not intentional... or rather, I misunderstood the meaning of 'MSRV' in the workspace. 

I ran `cargo-rail unify` with `msrv = true` and `msrv_source = "max"`. In that mode, cargo-rail computes MSRV as the maximum `rust-version` in the resolved dependency graph. On this branch it evaluated to:

  Computed MSRV: 1.85.0 (from 132 deps with rust-version)
  Contributors: getrandom

The full output came from `cargo rail unify --explain` in my local Tokio checkout... which will explain via the planner each choice made, deterministically.

I understand this is not the right source of truth for Tokio now. Tokio’s MSRV is the project compatibility policy (`1.71`), so this PR should not change `rust-version`. 

I’ll revert the MSRV edit and rerun with MSRV writing disabled for this repo (`msrv = false`) so no policy drift is introduced. I’ll also exclude Tokio-family crates from unification in this PR scope. That would unify the graph, resolve the undeclared bugs breaking isolation, and address both concerns?

PR #7921, opened 2026-02-19
runtime: add `worker_index()` getter

rcoh:
## Motivation

I am working on some runtime tracing instrumentation and one of the current pain points is being able to figure out what worker is currently running so that it can be properly correlated with the runtime metrics. I am working around this currently by enumerating `RuntimeMetrics::worker_thread_id` but that's obviously not ideal.

## Solution

Add `tokio::runtime::worker_id()` which returns the 0-based index of the current runtime worker thread. The index matches the worker indices used by `RuntimeMetrics` (e.g. `worker_total_busy_duration`), making it possible to correlate per-worker metrics with specific tasks.

Returns `None` from blocking threads, the `block_on` caller thread (on the multi-thread runtime), and non-Tokio threads. For the current-thread runtime, always returns `Some(0)`.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -16,7 +16,7 @@ env:
 task:
   name: FreeBSD 64-bit
   setup_script:
-    - pkg install -y bash
+    - pkg install -y bash ca_root_nss
     - curl https://sh.rustup.rs -sSf --output rustup.sh
     - sh rustup.sh -y --profile minimal --default-toolchain $RUST_STABLE
     - . $HOME/.cargo/env
@@ -42,7 +42,7 @@ task:
     RUSTFLAGS: --cfg docsrs --cfg tokio_unstable
     RUSTDOCFLAGS: --cfg docsrs --cfg tokio_unstable -Dwarnings
   setup_script:
-    - pkg install -y bash
+    - pkg install -y bash ca_root_nss
     - curl https://sh.rustup.rs -sSf --output rustup.sh
     - sh rustup.sh -y --profile minimal --default-toolchain $RUST_NIGHTLY
     - . $HOME/.cargo/env
@@ -58,7 +58,7 @@ task:
 task:
   name: FreeBSD 32-bit
   setup_script:
-    - pkg install -y bash
+    - pkg install -y bash ca_root_nss
     - curl https://sh.rustup.rs -sSf --output rustup.sh
     - sh rustup.sh -y --profile minimal --default-toolchain $RUST_STABLE
     - . $HOME/.cargo/env
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1057,6 +1057,8 @@ jobs:
         target:
           - wasm32-wasip1
           - wasm32-wasip1-threads
+    env:
+      rust_stable: '1.93.1'
     steps:
       - uses: actions/checkout@v4
       - name: Install Rust ${{ env.rust_stable }}
--- a/README.md
+++ b/README.md
@@ -1,3 +1,7 @@
+*[TokioConf 2026 program and tickets are now available!](https://tokioconf.com)*
+
+---
+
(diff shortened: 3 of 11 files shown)


ADD-SP (commented):
Shall we use `worker_index()` instead of `worker_id()`? It looks like methods like `worker_local_schedule_count` are using the term "worker index".

rcoh:
Yes I like that:
- `Id` => opaque id
- `Index` => numeric monotonic etc.

Will change, thanks for review!

tokio/tests/rt_worker_index.rs
```diff
@@ -0,0 +1,101 @@
...
+    let index = rt.block_on(async { runtime::worker_index() });
+    assert_eq!(
+        index, None,
+        "block_on thread is not a worker thread on multi-thread runtime"
+    );
+}
+
+#[tokio::main(flavor = "current_thread")]
+#[test]
+async fn worker_index_tokio_main_current_thread() {
```
ADD-SP: Does this test duplicate the `worker_index_current_thread`?

rcoh: yes thanks, will delete

tokio/src/runtime/mod.rs
```diff
@@ -569,6 +569,40 @@ cfg_rt! {
...
+        /// [`worker_total_busy_duration`](RuntimeMetrics::worker_total_busy_duration).
+        ///
+        /// Returns `None` when called from outside a runtime worker thread
+        /// (for example, from a blocking thread or a non-Tokio thread). On the
+        /// multi-thread runtime, the thread that calls [`Runtime::block_on`] is
+        /// not a worker thread, so this also returns `None` there.
+        ///
+        /// For the current-thread runtime, this always returns `Some(0)`
+        /// (including inside `block_on`, since the calling thread *is* the
+        /// worker thread).
```
martin-g: Maybe add a line mentioning that this method may return different results across `.await` points.

tokio/tests/rt_worker_index.rs
```diff
@@ -0,0 +1,94 @@
+#![warn(rust_2018_idioms)]
+#![cfg(all(
+    feature = "full",
+    tokio_unstable,
+    not(target_os = "wasi"),
+    target_has_atomic = "64"
```
martin-g: This flag is only needed for the Metrics, no ?
Maybe move it to `worker_index_matches_metrics_worker_thread_id()` and allow the other tests to run without this requirement.

tokio/tests/rt_worker_index.rs
```diff
@@ -0,0 +1,94 @@
...
+        let index = index.expect("should be Some on worker thread");
+        assert!(
+            index < num_workers,
+            "worker_index {index} >= num_workers {num_workers}"
+        );
+    });
+}
+
+#[test]
+fn worker_index_current_thread() {
```
martin-g, line 7: Please also add a test for LocalRuntime.

tokio/src/runtime/mod.rs
```diff
@@ -569,6 +569,40 @@ cfg_rt! {
...
+        /// The returned value is a 0-based index matching the worker indices
+        /// used by [`RuntimeMetrics`] methods such as
+        /// [`worker_total_busy_duration`](RuntimeMetrics::worker_total_busy_duration).
+        ///
+        /// Returns `None` when called from outside a runtime worker thread
+        /// (for example, from a blocking thread or a non-Tokio thread). On the
+        /// multi-thread runtime, the thread that calls [`Runtime::block_on`] is
+        /// not a worker thread, so this also returns `None` there.
+        ///
+        /// For the current-thread runtime, this always returns `Some(0)`
```
martin-g: maybe mention LocalRuntime too ?!

ADD-SP:
Thanks!

PR #7923, opened 2026-02-20
task: clarify when to use spawn_blocking vs dedicated threads

0rlych1kk4:
## Motivation

The current spawn_blocking documentation states that the function is intended for operations that eventually finish and that thread::spawn should be used for ordinary threads.
However, the practical distinction between bounded blocking work and long-lived or persistent workers is not explicit. This often leads to confusion about when long-running tasks should use spawn_blocking versus a dedicated thread.
This clarification was requested following discussion #7878 .

## Solution

Add a short guidance section to the spawn_blocking documentation explaining:
  - spawn_blocking is intended for short-lived, bounded blocking work
  - Long-lived or persistent blocking tasks reduce the effective capacity of the blocking thread pool
  - For such workloads, a dedicated thread via std::thread::spawn is preferred
The change is documentation-only and does not modify behavior.

### DIFF
--- a/tokio/src/task/blocking.rs
+++ b/tokio/src/task/blocking.rs
@@ -103,21 +103,6 @@ cfg_rt! {
     /// their own. If you want to spawn an ordinary thread, you should use
     /// [`thread::spawn`] instead.
     ///
-    /// # When to use `spawn_blocking` vs dedicated threads
-    ///
-    /// `spawn_blocking` is intended for *bounded* blocking work that eventually finishes.
-    /// Each call occupies a thread from the runtime's blocking thread pool for the duration
-    /// of the task. Long-lived tasks therefore reduce the pool's effective capacity, which
-    /// can delay other blocking operations once the pool is saturated and work is queued.
-    ///
-    /// For workloads that run indefinitely or for extended periods (for example,
-    /// background workers or persistent processing loops), prefer a dedicated thread created with
-    /// [`thread::spawn`].
-    ///
-    /// As a rule of thumb:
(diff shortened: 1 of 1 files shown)


tokio/src/task/blocking.rs
```diff
@@ -103,6 +103,21 @@ cfg_rt! {
     /// their own. If you want to spawn an ordinary thread, you should use
     /// [`thread::spawn`] instead.
```
Darksonn, line 104: Seems like a duplicate of this paragraph. Can we include it somehow?

tokio/src/task/blocking.rs
Darksonn, line 120: You added a heading above, so this text is now part of "When to use `spawn_blocking` vs dedicated threads". Please move it above the heading or add another heading here.

tokio/src/task/blocking.rs
0rlych1kk4: Good catch.  I’ve consolidated the wording to avoid duplication and aligned it with the existing paragraph.
Please let me know if this reads better.

tokio/src/task/blocking.rs
0rlych1kk4: Thanks for the review. good point. I’ll move this section above the "When to use spawn_blocking vs dedicated threads" heading so the structure reads:
 - General behavior and limitations
 - Usage guidance

I’ll push an update shortly. Please let me know if this organization looks better.

0rlych1kk4:
Thanks for the review and merge. Appreciate the feedback!

PR #7924, opened 2026-02-20
rt: introduce runtime name

### LINKED ISSUE #5545: Runtime identifier and name

mattiapitossi:
closes #5545 

[rendered docs](https://deploy-preview-7924--tokio-rs.netlify.app/tokio/)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -4,7 +4,7 @@ freebsd_instance:
   image_family: freebsd-14-3
 env:
   RUST_STABLE: stable
-  RUST_NIGHTLY: nightly-2025-01-25
+  RUST_NIGHTLY: nightly-2025-10-12
   RUSTFLAGS: -D warnings
   # This excludes unstable features like io_uring, which require '--cfg tokio_unstable'.
   TOKIO_STABLE_FEATURES: full,test-util
@@ -16,7 +16,7 @@ env:
 task:
   name: FreeBSD 64-bit
   setup_script:
-    - pkg install -y bash ca_root_nss
+    - pkg install -y bash
     - curl https://sh.rustup.rs -sSf --output rustup.sh
     - sh rustup.sh -y --profile minimal --default-toolchain $RUST_STABLE
     - . $HOME/.cargo/env
@@ -42,7 +42,7 @@ task:
     RUSTFLAGS: --cfg docsrs --cfg tokio_unstable
     RUSTDOCFLAGS: --cfg docsrs --cfg tokio_unstable -Dwarnings
   setup_script:
-    - pkg install -y bash ca_root_nss
+    - pkg install -y bash
     - curl https://sh.rustup.rs -sSf --output rustup.sh
     - sh rustup.sh -y --profile minimal --default-toolchain $RUST_NIGHTLY
     - . $HOME/.cargo/env
@@ -58,7 +58,7 @@ task:
 task:
   name: FreeBSD 32-bit
   setup_script:
-    - pkg install -y bash ca_root_nss
+    - pkg install -y bash
     - curl https://sh.rustup.rs -sSf --output rustup.sh
     - sh rustup.sh -y --profile minimal --default-toolchain $RUST_STABLE
     - . $HOME/.cargo/env
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ env:
   RUST_BACKTRACE: 1
   RUSTUP_WINDOWS_PATH_ADD_BIN: 1
   # Change to specific Rust release to pin
-  rust_stable: '1.93.1'
+  rust_stable: stable
   rust_nightly: nightly-2025-10-12
   # Pin a specific miri version
   rust_miri_nightly: nightly-2025-11-13
@@ -114,7 +114,7 @@ jobs:
           cargo nextest run --workspace --features $TOKIO_STABLE_FEATURES
           # Removing workspace patches to run tests without path dependencies
           # (if not specified differently in the crate)
-          sed -i.bak '/^\[patch\.crates-io\]$/d; /^tokio = { path = "tokio"\s*}$/d' Cargo.toml && rm -f Cargo.toml.bak
+          perl -0 -i -pe 's/\[patch\.crates-io\].+\n\[/[/s' Cargo.toml
           cargo nextest run \
             --workspace \
             --exclude tokio \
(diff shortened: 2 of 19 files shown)


tokio/src/runtime/scheduler/multi_thread/mod.rs
```diff
@@ -54,6 +54,7 @@ pub(crate) struct MultiThread;
 // ===== impl MultiThread =====
 
 impl MultiThread {
+    #[allow(clippy::too_many_arguments)]
     pub(crate) fn new(
```
mattiapitossi, line 58: Not sure what to do here, I don't feel like name should be in Config, maybe we should moving some other field? Or maybe allowing too_many_arguments in this case might be OK

tokio/src/runtime/builder.rs
```diff
@@ -538,6 +544,28 @@ impl Builder {
...
+    /// # use tokio::runtime;
+    ///
+    /// # pub fn main() {
+    /// let rt = runtime::Builder::new_multi_thread()
+    ///     .name("my-runtime")
+    ///     .build();
+    /// # }
+    /// # }
+    /// ```
+    pub fn name(&mut self, val: impl Into<String>) -> &mut Self {
```
mattiapitossi, line 568: Do you think this new API should be gated behind the unstable ff?

tokio-macros/src/entry.rs
```diff
@@ -53,6 +53,7 @@ impl UnhandledPanic {
 }
 
 struct FinalConfig {
+    name: Option<String>,
```
mattiapitossi, line 56: probably the name could be the first in the order, and so for the diagnostics

mattiapitossi:
This is ready for review. I've left some comments regarding few things that might need to be discussed.

tokio/src/runtime/builder.rs
martin-g, line 1875:
```suggestion
            .field("name", &self.name)
            .field("worker_threads", &self.worker_threads)
```
Since `name` is an Option this should probably in `if self.name.is_some()` and use [finish_non_exhaustive](https://doc.rust-lang.org/stable/std/fmt/struct.DebugStruct.html#method.finish_non_exhaustive) if it is None

tokio/src/runtime/scheduler/current_thread/mod.rs
```diff
@@ -34,6 +34,9 @@ pub(crate) struct CurrentThread {
 
 /// Handle to the current thread scheduler
 pub(crate) struct Handle {
+    /// The name of the runtime
+    pub(crate) name: Option<String>,
```
martin-g: Is there a need to be `pub(crate)` ? There is a getter

tokio/src/runtime/scheduler/multi_thread/handle.rs
```diff
@@ -23,6 +23,9 @@ use crate::loom::sync::atomic::{AtomicBool, Ordering::SeqCst};
 
 /// Handle to the multi thread scheduler
 pub(crate) struct Handle {
+    /// The name of the runtime
+    pub(crate) name: Option<String>,
```
martin-g: Is there a need to be pub(crate) ? There is a getter

tokio/src/runtime/builder.rs
```diff
@@ -538,6 +544,28 @@ impl Builder {
...
+        let val = val.into();
+        self.name = Some(val);
```
martin-g, line 571: nit: `""` is allowed at the moment. IMO it does not make much sense to allow it

tokio/src/runtime/scheduler/current_thread/mod.rs
mattiapitossi: good point, removed!

tokio/src/runtime/scheduler/multi_thread/handle.rs
mattiapitossi: In this case yes, because the Handler is instantiated inside the worker. But I changed it to super instead of crate

tokio/src/runtime/builder.rs
mattiapitossi: There were already many different option fields, should we check if there's at least one option field to use the finish_non_exhaustive?

martin-g: Here is an example of similar use case: https://github.com/apache/avro-rs/blob/f18cf9573574d4a4a54a4bedd620776527e8a7cd/avro/src/schema/record/schema.rs#L60-L64

mattiapitossi: Thanks, this is interesting, I've applied your suggestion to use the non exhaustive when the name is not set. I'm not totally sure about the already existing fields, maybe changing those (that now are displayed to None) could cause some issue to downstreams?

tokio/src/runtime/builder.rs
mattiapitossi: I added an assertion that prevents empty strings, is that what you had in mind? I was also thinking to convert empty strings to None, but I don't think is the best approach since it's an hidden logic

martin-g: Yes! This is what I meant!
Thank you!

PR #7927, opened 2026-02-23
time: skip lock in timer cancel when not registered

p-player:
## Motivation

Relates to #6504.

The timer driver's `Handle::clear_entry` always acquires the driver mutex lock, even when the timer entry is definitely not registered with the driver. Similarly, `TimerEntry::cancel` (called by `Drop`) always invokes [clear_entry](cci:1://file:///Users/surunqian/Desktop/META-MATH-V4/tokio/tokio/src/runtime/time/mod.rs:369:4-403:5) even if the timer was never polled and thus never registered.

This is a common pattern with timeouts: a `tokio::time::timeout` wraps an operation, the operation completes before the timeout fires, and the inner `Sleep` is dropped without ever being polled. Under high timer load, these unnecessary lock acquisitions cause significant mutex contention.

## Solution

Add two fast-path checks:

1. **`TimerEntry::cancel()`**: If `!self.registered` (the timer was never submitted to the driver), return immediately. The timer was never visible to the driver thread, so no acq/rel fence or deregistration is needed.

2. **`Handle::clear_entry()`**: If [might_be_registered()](cci:1://file:///Users/surunqian/Desktop/META-MATH-V4/tokio/tokio/src/runtime/time/entry.rs:448:4-454:5) returns `false`, skip the lock and fire directly. This check is conservative — if it returns `false`, the timer is *definitely* not in the wheel. In the deregistered state, [fire()](cci:1://file:///Users/surunqian/Desktop/META-MATH-V4/tokio/tokio/src/runtime/time/entry.rs:690:4-703:5) will see `STATE_DEREGISTERED` and return `None`, making it safe to call without the lock.

## Future Work

This is the first step toward reducing timer driver contention (see #6504). Potential follow-ups include:
- Sharding the timer wheel lock per worker thread
- Differentiating cancel-dominant (timeout) vs complete-dominant (sleep) timer strategies

### DIFF
--- a/tokio/src/runtime/time/mod.rs
+++ b/tokio/src/runtime/time/mod.rs
@@ -379,20 +379,6 @@ impl Handle {
     /// `add_entry` must not be called concurrently.
     pub(self) unsafe fn clear_entry(&self, entry: NonNull<TimerShared>) {
         unsafe {
-            // Fast path: if the timer is definitely not registered with the
-            // driver, we can skip acquiring the lock entirely. This is the
-            // common case for timers that are dropped before being polled
-            // (e.g., timeouts that complete before firing).
-            //
-            // `might_be_registered()` uses relaxed ordering but is conservative:
-            // if it returns false, the timer is *definitely* not in the wheel.
-            // In the deregistered state, `fire()` is a no-op (returns None),
-            // so calling it without the lock is safe.
-            if !entry.as_ref().might_be_registered() {
-                entry.as_ref().handle().fire(Ok(()));
-                return;
-            }
-
             let mut lock = self.inner.lock();
 
             if entry.as_ref().might_be_registered() {


ADD-SP:
Thanks for your PR, I will review it soon. 

By the way, for "Future Work", we have completed further optimization in another approach, please see <https://github.com/tokio-rs/tokio/issues/7745> for more details.

tokio/src/runtime/time/entry.rs
```diff
@@ -543,6 +543,18 @@ impl TimerEntry {
...
+        // lock. The timer was never visible to the driver thread, so no
+        // acq/rel fence is needed for memory ordering.
+        //
+        // This is a common case for timeouts: a timeout wraps an operation,
+        // the operation completes before the timeout fires, and the Sleep is
+        // dropped without ever being polled.
+        if !self.registered {
+            return;
+        }
+
```
ADD-SP: If the timer never be polled, the `TimerEntry::inner` will be `None`. So we will never hit this branch.

https://github.com/tokio-rs/tokio/blob/25e7f2641ef2555d688c267059431a2802805f1d/tokio/src/runtime/time/entry.rs#L297

In other words, the intrusive node has already been registered as long as the `TimerEntry::inner` is `Some`.

ADD-SP (commented):
Cond

p-player:
Closing this PR as the contention issue is being addressed more fundamentally by #7745. Thanks for the review!

Issue #7928, opened 2026-02-23
`tokio::select!` shadows useful names


iriri:
**Version**
`v1.49.0`

**Platform**
Any

**Description**

The definition of `tokio::select!` contains local `use` declarations that shadow `Future`, `Pin`, `Ready`, and `Pending`. The first two aren't a big deal in practice, but `Ready` and `Pending` are generic names that have a good chance of showing up in user code.

```rust
enum Ready { A }                                                                                     
enum RReady { A }                                                                                    
                                                                                                     
async fn foo(x: impl Future<Output = Ready>, y: impl Future<Output = RReady>) {                      
   tokio::select! {                                                                                                                                                                                                                           
      x = x => match x {                                                                             
         Ready::A => (), // failed to resolve: `Ready` is a variant, not a module
      },                                                                                             
      y = y => match y {                                                                             
         RReady::A => (),                                                                            
      },                                                                                             
   }                                                                                                 
}
```

I expected to see this happen: my code working.

Instead, this happened: I got a mildly confusing error message and then remembered that I've hit this issue before.

PR #7929, opened 2026-02-23
Remove (most) local `use` declarations in `tokio::select!`

iriri:
## Motivation

Some of the `use` declarations in `tokio::select!` shadow common names. See https://github.com/tokio-rs/tokio/issues/7928 for a reproduction of the issue.

## Solution

Using fully qualified paths avoids unnecessary imports. We unfortunately have to leave the `Pin` import untouched, however, as some downstream crates (likely unknowingly) depend on it being implicitly brought into scope.

### DIFF
--- a/tokio/src/macros/select.rs
+++ b/tokio/src/macros/select.rs
@@ -615,9 +615,7 @@ doc! {macro_rules! select {
 
         // `tokio::macros::support` is a public, but doc(hidden) module
         // including a re-export of all types needed by this macro.
-        use $crate::macros::support::Future as __Future;
-        use $crate::macros::support::Pin as __Pin;
-        use $crate::macros::support::Poll::{Ready as __Ready, Pending as __Pending};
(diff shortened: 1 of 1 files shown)


martin-g:
> Using fully qualified paths is the more principled solution, but it results in some very long lines. I am perfectly happy to switch to that solution however.

Looking at the diff I think it won't lead to very long lines in this case.
+1 for fully qualified paths!

iriri:
```
   Compiling hyper v1.8.1 (/home/runner/work/tokio/tokio/hyper)
error[E0433]: failed to resolve: use of undeclared type `Pin`
   --> examples/upgrades.rs:166:40
    |
166 |                         let mut conn = Pin::new(&mut conn);
    |                                        ^^^ use of undeclared type `Pin`
    |
help: consider importing this struct
```

Wow, there's downstream code that depends on these declarations existing. That's unfortunate to say the least. We may have to leave `Pin` unchanged.

PR #7943, opened 2026-02-28
chore: prepare tokio-macros v2.6.1

ADD-SP:
# 2.6.1 (Mar 1st, 2026)

- macros: improve error message for return type mismatch in `#[tokio::main]` ([#7856])
- macros: use `call_site` hygiene to avoid unused qualification ([#7866])

[#7856]: https://github.com/tokio-rs/tokio/pull/7856
[#7866]: https://github.com/tokio-rs/tokio/pull/7866

### DIFF
--- a/tokio-macros/CHANGELOG.md
+++ b/tokio-macros/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 2.6.1 (Mar 1st, 2026)
+# 2.6.1 (Mar 2nd, 2026)
 
 - macros: improve error message for return type mismatch in #[tokio::main] ([#7856])
 - macros: use call_site hygiene to avoid unused qualification ([#7866])


tokio/Cargo.toml
```diff
@@ -90,7 +90,7 @@ io-uring = ["dep:io-uring", "libc", "mio/os-poll", "mio/os-ext", "dep:slab"]
 taskdump = ["dep:backtrace"]
 
 [dependencies]
-tokio-macros = { version = "~2.6.0", path = "../tokio-macros", optional = true }
+tokio-macros = { version = "~2.6.1", path = "../tokio-macros", optional = true }
```
Darksonn: Hmm ... do we actually require 2.6.1 here?

Most likely this should either stay 2.6.0, or tokio-macros should be released as 2.7.0.

Thoughts @taiki-e ?

taiki-e: This release contains changes to avoid false positives of rustc lint and improvements to error messages.


As long as we are not compiling tests using minimal-versions dependencies without path dependencies in CI, it is fine to leave this at 2.6.0.

ADD-SP: Initially, it doesn't compile, I just tried it again locally, and compilation success. I pushed this change again.

ADD-SP: Maybe my local IDE was broken as that time.

PR #7944, opened 2026-02-28
bench: add remote_spawn benchmark for inject queue contention

alex:
Add a benchmark that measures contention on the scheduler's inject queue mutex (push_remote_task) when multiple external threads spawn tasks into the runtime simultaneously. Every rt.spawn() from a non-worker thread unconditionally goes through push_remote_task, so this directly measures the scalability of the inject queue.

Results on an M1 Max MacBook Pro (10 cores), spawning 12,800 total tasks:

  threads/1:  3.39 ms  (265 ns/task, 1.00x)
  threads/2:  4.74 ms  (370 ns/task, 1.40x)
  threads/4:  5.89 ms  (460 ns/task, 1.74x)
  threads/8:  8.10 ms  (633 ns/task, 2.39x)

Wall-clock time increases with more threads despite constant total work, confirming the single mutex serializes producers.

### DIFF
--- a/benches/remote_spawn.rs
+++ b/benches/remote_spawn.rs
@@ -19,6 +19,7 @@ use tokio::runtime::{self, Runtime};
 /// Total number of tasks spawned across all threads per iteration.
 /// Must be divisible by the largest parallelism level (64).
 const TOTAL_TASKS: usize = 12_800;
+const _: () = assert!(TOTAL_TASKS % 64 == 0, "TOTAL_TASKS must be divisible by 64");
 
 fn remote_spawn_contention(c: &mut Criterion) {
     let parallelism_levels = parallelism_levels();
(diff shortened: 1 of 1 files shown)


alex:
I also tried a version of this that had remote threads pushing onto an mpsc unbounded queue -- it displayed much worse scaling behavior, but it wasn't clear how much of that was about the mpcs queue vs. the injector queue.

benches/remote_spawn.rs
```diff
@@ -0,0 +1,95 @@
...
+        let num_threads = *num_threads;
+        group.bench_with_input(
+            BenchmarkId::new("threads", num_threads),
+            &num_threads,
+            |b, &num_threads| {
+                let rt = rt();
+                let tasks_per_thread = TOTAL_TASKS / num_threads;
+
+                b.iter(|| {
+                    let barrier = Barrier::new(num_threads);
```
martin-g: You can move this before `b.iter()` to avoid measuring its construction.
https://doc.rust-lang.org/stable/std/sync/struct.Barrier.html#method.wait says:
```
Barriers are re-usable after all threads have rendezvoused once
```

benches/remote_spawn.rs
```diff
@@ -0,0 +1,95 @@
...
+//! - All threads are synchronized with a barrier to maximize contention
+//! - Tasks are trivial no-ops to isolate the push overhead
+
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use std::sync::Barrier;
+use tokio::runtime::{self, Runtime};
+
+/// Total number of tasks spawned across all threads per iteration.
+/// Must be divisible by the largest parallelism level (64).
+const TOTAL_TASKS: usize = 12_800;
```
martin-g, line 21:
```suggestion
const TOTAL_TASKS: usize = 12_800;
const _: () = assert!(TOTAL_TASKS % 64 == 0, "TOTAL_TASKS must be divisible by 64");
```

benches/remote_spawn.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    [1, 2, 4, 8, 16, 32, 64]
+        .into_iter()
+        .filter(|&n| n <= max_parallelism)
+        .collect()
+}
+
+fn rt() -> Runtime {
+    runtime::Builder::new_multi_thread()
+        .enable_all()
```
martin-g:
```suggestion
```
No need to start IO and timers.

alex:
Thanks, will implement this feedback later this afternoon!

Do you agree that this is measuring the right thing? (That's my biggest concern, that I might accidentally be measuring the wrong contention!)

alex:
Feedback incorporated, thanks!

martin-g:
> Do you agree that this is measuring the right thing?

I think so, yes!

alex:
FWIW, once this lands I've got an implementation of a sharded queue (similar to https://github.com/tokio-rs/tokio/pull/7757) that improves scalability considerably (tested up to 64 cores).

alex:
Is anything else needed to merge this? Happy to take any work.

PR #7948, opened 2026-03-04
Add RateLimiter to tokio::sync

ebokoo:
Implements a token bucket rate limiter that controls the rate of operations. Supports both blocking and non-blocking acquire.

Features:
- Token bucket algorithm for rate limiting
- Configurable rate and burst capacity
- try_acquire() for non-blocking acquire
- acquire() for blocking acquire with await

### DIFF
--- a/tokio/src/sync/mod.rs
+++ b/tokio/src/sync/mod.rs
@@ -476,23 +476,11 @@ cfg_sync! {
     mod semaphore;
     pub use semaphore::{Semaphore, SemaphorePermit, OwnedSemaphorePermit};
 
-    mod rwlock;
-    pub use rwlock::RwLock;
-    pub use rwlock::owned_read_guard::OwnedRwLockReadGuard;
-    pub use rwlock::owned_write_guard::OwnedRwLockWriteGuard;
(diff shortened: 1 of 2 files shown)


tokio/src/sync/mod.rs
```diff
@@ -476,23 +476,11 @@ cfg_sync! {
...
-    pub use rwlock::write_guard_mapped::RwLockMappedWriteGuard;
-
     mod task;
     pub(crate) use task::AtomicWaker;
 
-    mod once_cell;
-    pub use self::once_cell::{OnceCell, SetError};
-
-    mod set_once;
-    pub use self::set_once::{SetOnce, SetOnceError};
```
martin-g, line 495: Why all these re-exports are removed ?

Darksonn:
I do not think this implementation is correct - for example, if `acquire` cannot immediately succeed, then it should make use of `tokio::time::sleep` to wait until the next permit becomes available, right?

Either way, we're not interested in adding a token bucket to `tokio::sync`. Please prototype your implementation in a separate crate first.

PR #7954, opened 2026-03-06
stream: impl FromStream for BTreeSet

figsoda:
## Motivation

This allows streams to be `collect`ed into `BTreeSet`s. If this looks good, I can also work on other collections supported by `FromIterator`

## Solution

the code should be trivial enough


figsoda:
CI failures seem unrelated

PR #7960, opened 2026-03-08
feat: add SQPOLL support for io_uring

SidongYang:
This adds `uring_setup_sqpoll` to the Runtime `Builder`, allowing the io_uring driver
to be configured with SQPOLL enabled. This offloads submission queue polling to a
kernel thread, which can significantly reduce syscall overhead.

- Added `uring_setup_sqpoll` to `tokio::runtime::Builder`.
- Updated `UringContext` to sync the submission queue and check `need_wakeup`.
- Added tests to verify `sqpoll` in both single-thread and multi-thread runtimes.

### DIFF
--- a/tokio/src/runtime/builder.rs
+++ b/tokio/src/runtime/builder.rs
@@ -63,6 +63,16 @@ pub struct Builder {
     enable_io: bool,
     nevents: usize,
 
+    /// Idle timeout (in milliseconds) for the io_uring SQPOLL kernel thread.
+    #[cfg(all(
+        tokio_unstable,
+        feature = "io-uring",
+        feature = "rt",
+        feature = "fs",
+        target_os = "linux"
+    ))]
+    uring_setup_sqpoll: Option<u32>,
+
     /// Whether or not to enable the time driver
     enable_time: bool,
 
@@ -286,6 +296,15 @@ impl Builder {
             // I/O defaults to "off"
             enable_io: false,
             nevents: 1024,
+
+            #[cfg(all(
+                tokio_unstable,
+                feature = "io-uring",
+                feature = "rt",
+                feature = "fs",
+                target_os = "linux"
+            ))]
+            uring_setup_sqpoll: None,
 
             // Time defaults to "off"
             enable_time: false,
@@ -1672,6 +1691,19 @@ impl Builder {
         cfg.timer_flavor = TimerFlavor::Traditional;
         let (driver, driver_handle) = driver::Driver::new(cfg)?;
 
+        #[cfg(all(
+            tokio_unstable,
+            feature = "io-uring",
+            feature = "rt",
+            feature = "fs",
+            target_os = "linux"
+        ))]
+        if let Some(idle_timeout) = self.uring_setup_sqpoll {
+            if let Some(io) = driver_handle.io.as_ref() {
+                io.setup_uring_sqpoll(idle_timeout);
+            }
+        }
+
         // Blocking pool
         let blocking_pool = blocking::create_blocking_pool(self, self.max_blocking_threads);
         let blocking_spawner = blocking_pool.spawner().clone();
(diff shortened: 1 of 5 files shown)


tokio/src/runtime/builder.rs
```diff
@@ -1742,6 +1755,28 @@ cfg_io_uring! {
...
+        /// use tokio::runtime;
+        ///
+        /// let rt = runtime::Builder::new_multi_thread()
+        ///     .enable_io_uring()
+        ///     .uring_setup_sqpoll(2000)
+        ///     .build()
+        ///     .unwrap();
+        /// ```
+        #[cfg_attr(docsrs, doc(cfg(feature = "io-uring")))]
+        pub fn uring_setup_sqpoll(&mut self, idle: u32) -> &mut Self {
```
```suggestion
        pub fn uring_setup_sqpoll(&mut self, idle_timeout: u32) -> &mut Self {
```

tokio/src/runtime/builder.rs
```diff
@@ -1742,6 +1755,28 @@ cfg_io_uring! {
...
+            self.uring_setup_sqpoll = Some(idle);
```
```suggestion
            self.uring_setup_sqpoll = Some(idle_timeout);
```

tokio/src/runtime/builder.rs
```diff
@@ -1742,6 +1755,28 @@ cfg_io_uring! {
             self.enable_io = true;
             self
         }
+
+        /// Enables SQPOLL for the io_uring driver and sets the idle timeout.
```
```suggestion
        /// Enables SQPOLL for the io_uring driver and sets the idle timeout (in milliseconds).
```
if the timeout is in millis. Otherwise, please mention the correct time unit.

tokio/src/runtime/builder.rs
```diff
@@ -1742,6 +1755,28 @@ cfg_io_uring! {
...
+        ///
+        /// When SQPOLL is enabled, a kernel thread is created to poll the
+        /// submission queue. This can reduce syscall overhead.
+        ///
```
martin-g, line 1861: Maybe mention what privileges are required by SQPOLL, e.g. CAP_SYS_ADMIN, minimum Linux version, etc. ?!

tokio/src/runtime/builder.rs
```diff
@@ -1598,6 +1604,13 @@ impl Builder {
         cfg.timer_flavor = TimerFlavor::Traditional;
         let (driver, driver_handle) = driver::Driver::new(cfg)?;
 
+        #[cfg(all(tokio_unstable, feature = "io-uring", target_os = "linux"))]
```
martin-g: Here the list of features is missing `rt` and `fs`. This may fail to compile due to missing symbols. Please list all required features as at other places.

tokio/src/runtime/io/driver/uring.rs
```diff
@@ -44,7 +46,11 @@ impl UringContext {
             return Ok(false);
         }
 
-        let uring = IoUring::new(DEFAULT_RING_SIZE)?;
+        let uring = if let Some(idle) = self.sqpoll_idle {
+            IoUring::builder().setup_sqpoll(idle).build(DEFAULT_RING_SIZE)?
```
martin-g: What would happen if SQPOLL is not allowed but regular io_uring is allowed on the system ?
Should it fallback to regular io_uring? Or to blocking ?!

Darksonn: SQPOLL cannot be the default because of its CAP_SYS_ADMIN requirement, but if it's explicitly enabled, I think it's fair to fail when it's not available.

SidongYang:
I have updated the PR to address the remaining review comments:

- **Naming**: Renamed the `idle` parameter to `idle_timeout` in `uring_setup_sqpoll` and its internal implementation for better clarity.
- **Documentation**: 
  - Explicitly specified that the `idle_timeout` unit is in milliseconds.
  - Added prerequisites regarding required Linux kernel versions (5.1+) and `CAP_SYS_ADMIN` privileges (required before Linux 5.11).
  - Clarified that the first I/O operation will fail if SQPOLL is enabled but not supported by the system.
- **Feature Flags**: Standardized `cfg` attributes for `io-uring` to consistently include `rt`, `fs`, and `target_os = "linux"` to match workspace conventions and prevent compilation failures.
- **Validation**: Verified that the changes compile and that the `fs_uring_sqpoll` tests pass successfully under `tokio_unstable`.

tokio/src/runtime/builder.rs
SidongYang: Thanks, Applied.

tokio/src/runtime/builder.rs
SidongYang: Thanks, Applied.

tokio/src/runtime/builder.rs
SidongYang: Thanks, Applied.

tokio/src/runtime/builder.rs
SidongYang: I've added prerequisites below. Thanks!

martin-g:
Please fix the formatting issues and update to master to fix the FreeBSD build

tokio/src/runtime/builder.rs
SidongYang: Thanks, applied!

SidongYang:
@martin-g Thanks, I've fixed formatting and rebased this branch.

martin-g:
Now the spellchecker is complaining.
Please add `SQPOLL` to https://github.com/tokio-rs/tokio/blob/master/spellcheck.dic and increment the number on the first line in this file.

SidongYang:
Thanks, @martin-g . I've added SQPOLL to the spellcheck file and incremented the number on the first line.

tokio/src/runtime/builder.rs
```diff
@@ -58,6 +58,15 @@ pub struct Builder {
     enable_io: bool,
     nevents: usize,
 
+    #[cfg(all(
+        tokio_unstable,
+        feature = "io-uring",
+        feature = "rt",
+        feature = "fs",
+        target_os = "linux"
+    ))]
```
martin-g, line 73: Please add some rustdoc here

martin-g (approved):

tokio/src/runtime/builder.rs
SidongYang: Sure, I've added a line for rustdoc above. Thanks.

SidongYang:
Rebased onto the latest `master` to resolve the merge conflicts (in `spellcheck.dic` and `tokio/src/runtime/io/driver/uring.rs`).

PR #7961, opened 2026-03-08
ci: freeze rustc on 1.93.1

ADD-SP:
This is to unblock the daily PR review/merge works.

See <https://github.com/rust-lang/rust/issues/153486> for more context.


ADD-SP:
Emm, this is weird.

taiki-e (approved):

PR #7962, opened 2026-03-08
examples: add graceful shutdown example

Sim-hu:
## Summary

- Adds a `graceful-shutdown` example demonstrating `tokio::signal::ctrl_c()` for clean server shutdown
- TCP echo server that notifies active connections via `broadcast` channel before exiting
- Shows the `tokio::select!` pattern for combining accept loops with shutdown signals

### Motivation

Graceful shutdown is one of the most common patterns in production async applications, but none of the existing 20 examples cover `tokio::signal` or shutdown coordination. This fills that gap with a minimal, self-contained example.

### Checklist

- [x] `cargo build --example graceful-shutdown`
- [x] `cargo run --example graceful-shutdown` — tested with `nc` + Ctrl+C
- [x] `cargo clippy --example graceful-shutdown`
- [x] `cargo fmt --check`

### DIFF
--- a/examples/graceful-shutdown.rs
+++ b/examples/graceful-shutdown.rs
@@ -78,19 +78,17 @@ async fn handle_connection(mut socket: TcpStream, addr: SocketAddr, token: Cance
     tokio::select! {
         _ = echo(&mut socket) => {}
         _ = token.cancelled() => {
-            // Optionally notify the client before closing. Use a timeout
-            // so a slow or unresponsive client doesn't hold up shutdown.
-            let _ = time::timeout(
-                Duration::from_secs(1),
-                socket.write_all(b"server shutting down\n"),
-            )
-            .await;
+            notify_shutdown(&mut socket).await;
         }
     }
 
     println!("connection from {addr} closed");
 }
 
+/// Reads lines from the client and writes them back.
+///
+/// Called for every accepted connection. Runs until the client disconnects
+/// or a read/write error occurs.
 async fn echo(socket: &mut TcpStream) {
     let (reader, mut writer) = socket.split();
     let mut reader = BufReader::new(reader);
@@ -107,4 +105,16 @@ async fn echo(socket: &mut TcpStream) {
             }
         }
     }
+}
+
+/// Sends a shutdown notice to the client before closing the connection.
+///
+/// Called when the cancellation token fires. Uses a timeout so that a
+/// slow or unresponsive client cannot hold up the server shutdown.
+async fn notify_shutdown(socket: &mut TcpStream) {
+    let _ = time::timeout(
+        Duration::from_secs(1),
+        socket.write_all(b"server shutting down\n"),
+    )
+    .await;
 }


examples/graceful-shutdown.rs
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+    }
+
+    // Notify all active connections.
+    let _ = shutdown_tx.send(());
+
+    // Drop the sender and wait briefly for connections to finish.
+    drop(shutdown_tx);
+    tokio::time::sleep(std::time::Duration::from_millis(100)).await;
```
martin-g: Wouldn't it be better to do `shutdown_tx.closed().await;` (i.e. wait for all subscribers to be dropped) instead of dropping the sender and hoping that the subscribers will finish in 100ms ? This is not really graceful.

Another option is to collect the JoinHandles from `tokio::spawn()` and .await them.

examples/graceful-shutdown.rs
```diff
@@ -0,0 +1,88 @@
...
+                    let mut line = String::new();
+
+                    loop {
+                        tokio::select! {
+                            // Echo lines back to the client.
+                            result = reader.read_line(&mut line) => {
+                                match result {
+                                    Ok(0) | Err(_) => break,
+                                    Ok(_) => {
+                                        let _ = writer.write_all(line.as_bytes()).await;
```
martin-g:
```suggestion
                                        if writer.write_all(line.as_bytes()).await.is_err() {
                                            // Client disconnected.
                                            break;
                                        }
```

examples/graceful-shutdown.rs
```diff
@@ -0,0 +1,88 @@
...
+    println!("listening on 127.0.0.1:6142");
+
+    // A broadcast channel to notify all tasks when shutdown is requested.
+    let (shutdown_tx, _) = broadcast::channel::<()>(1);
+
+    loop {
+        tokio::select! {
+            // Wait for a new connection.
+            result = listener.accept() => {
+                let (socket, addr) = result?;
```
martin-g: In production a failed accept() is not a reason to stop listening.
But since this is an example for a graceful shutdown this is OK-ish.

examples/graceful-shutdown.rs
```diff
@@ -0,0 +1,88 @@
...
+
+                tokio::spawn(async move {
+                    let (reader, mut writer) = socket.into_split();
+                    let mut reader = BufReader::new(reader);
+                    let mut line = String::new();
+
+                    loop {
+                        tokio::select! {
+                            // Echo lines back to the client.
+                            result = reader.read_line(&mut line) => {
```
martin-g: This is another bad practice for production. A bad client can DoS you by sending a huge line. Better use tokio_util::codec::LinesCodec or similar.
Since this is an example of graceful shutdown it is OK but it would be good to mention it as a comment or in the documentation at the top.

examples/graceful-shutdown.rs
```diff
@@ -0,0 +1,88 @@
...
+
+use std::error::Error;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let listener = TcpListener::bind("127.0.0.1:6142").await?;
+    println!("listening on 127.0.0.1:6142");
+
+    // A broadcast channel to notify all tasks when shutdown is requested.
+    let (shutdown_tx, _) = broadcast::channel::<()>(1);
```
Darksonn: We have a tutorial page on this topic: [Graceful Shutdown][1]. If we have an example, it should follow the same strategies as the tutorial page - anything else would be confusing.

In particular, we should use cancellation tokens to signal completion, and task tracker to wait for them to exit (a sleep is definitely not right).

[1]: https://tokio.rs/tokio/topics/shutdown

Sim-hu:
Reworked the example based on the feedback. Now uses `CancellationToken` + `TaskTracker` from `tokio-util` to match the [Graceful Shutdown tutorial](https://tokio.rs/tokio/topics/shutdown).

Changes:
- Replaced `broadcast` channel with `CancellationToken` for shutdown signaling
- Replaced `sleep` with `TaskTracker` to properly wait for in-flight connections
- `accept()` errors now log and continue instead of crashing
- `write_all` errors break the loop (client disconnect)

Sim-hu:
Makes sense — since there's already a tutorial page covering this, do you think a standalone example is still worth having alongside it? Happy to rework it with CancellationToken + TaskTracker if so, otherwise I'll close this out.

Sim-hu:
Good point — for a real server you'd definitely want to break or back off on fatal accept errors. I kept `continue` here since transient errors like fd exhaustion should recover on their own, but agreed it could get noisy if the listener is actually wedged. I'll add a comment noting the trade-off.

examples/graceful-shutdown.rs
```diff
@@ -0,0 +1,100 @@
...
+                        // backoff or break on fatal errors to avoid a busy loop.
+                        eprintln!("failed to accept: {e}");
+                        continue;
+                    }
+                };
+                println!("accepted connection from {addr}");
+
+                let token = token.clone();
+
+                tracker.spawn(async move {
```
Darksonn: Please move this to a separate async fn.

examples/graceful-shutdown.rs
```diff
@@ -0,0 +1,100 @@
...
+                                            break;
+                                        }
+                                        line.clear();
+                                    }
+                                }
+                            }
+                            _ = token.cancelled() => {
+                                let _ = writer.write_all(b"server shutting down\n").await;
+                                break;
+                            }
```
Darksonn: Current code means that we cannot be cancelled during `write_all` call. Is that what you intended?

examples/graceful-shutdown.rs
Sim-hu: Done — extracted it into `handle_connection`.

examples/graceful-shutdown.rs
Sim-hu: No, that wasn't intentional. Wrapped the write_all in a select! so cancellation is noticed immediately even during a slow write.

Darksonn: I think you should probably split the function in two. The outer one just uses cancellation token on the inner async fn, and then the inner code can call `read` and `write_all` normally. These multiple selects is not very clean.

Sim-hu: Good idea, done. Split into an outer `handle_connection` that does `select!` between the cancellation token and an inner `echo` fn that just does normal read/write.

examples/graceful-shutdown.rs
```diff
@@ -0,0 +1,107 @@
...
+async fn handle_connection(
+    mut socket: TcpStream,
+    addr: SocketAddr,
+    token: CancellationToken,
+) {
+    tokio::select! {
+        _ = echo(&mut socket) => {}
+        _ = token.cancelled() => {
+            let _ = socket.write_all(b"server shutting down\n").await;
+        }
```
Darksonn, line 82: Including logic like this is interesting for the example, but I think it should have a timeout on it. And there should be comment discussing this option.

Sim-hu: Good call — wrapped the write_all in a 1-second timeout so a slow client can't block shutdown. Also added a comment explaining the trade-off.

Darksonn: Since this is an example, it might be cleaner to move this into a separate method, so that you have `async fn echo` and `async fn cancel` with this just calling one then the other. And with doc-comments on each method explaining when they are called.

Sim-hu: Done — extracted the shutdown notification into a separate `notify_shutdown` function with doc-comments on both `echo` and `notify_shutdown` explaining their roles.

PR #7966, opened 2026-03-11
impl `FromStream` for `std::collections::*`

figsoda:
## Motivation

Make `Stream::collect` closer to feature parity with `Iterator::collect`, follow up to #7954

## Solution

the code should be trivial enough if you read commit by commit

### DIFF
--- a/tokio-stream/tests/stream_collect.rs
+++ b/tokio-stream/tests/stream_collect.rs
@@ -92,18 +92,19 @@ async fn collect_linkedlist_items() {
 
     assert_pending!(fut.poll());
 
-    tx.send(1).unwrap();
-    assert!(fut.is_woken());
-    assert_pending!(fut.poll());
-
-    tx.send(2).unwrap();
-    assert!(fut.is_woken());
-    assert_pending!(fut.poll());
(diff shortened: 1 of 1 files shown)


tokio-stream/src/stream_ext/collect.rs
```diff
@@ -136,6 +137,44 @@ impl<T> sealed::FromStreamPriv<T> for Vec<T> {
...
+    type InternalCollection = LinkedList<T>;
+
+    fn initialize(_: sealed::Internal, _lower: usize, _upper: Option<usize>) -> LinkedList<T> {
+        LinkedList::new()
+    }
+
+    fn extend(_: sealed::Internal, collection: &mut LinkedList<T>, item: T) -> bool {
+        collection.push_back(item);
+        true
+    }
```
Darksonn, line 171: For ordered collections, can you add a test that they appear in the same order as when using normal `Iterator::collect`?

figsoda: did so for VecDeque and LinkedList, lmk if I should rebase and force push

PR #7967, opened 2026-03-11
ci: patch workspace members to disambiguate `--package`

figsoda:
## Motivation

Before the change, testing workspace members fails with `specification is ambiguous`

```console
$ cargo test -p tokio-stream # cargo nextest run -p tokio-stream
error: specificationm `tokio-stream` is ambiguous
help: re-run this command with one of the following specifications
  path+file:///home/figsoda/tokio/tokio-stream#0.1.18
  registry+https://github.com/rust-lang/crates.io-index#tokio-stream@0.1.18
```

This can be fixed with `cargo test -p file://$PWD/tokio-stream`, but it seems unnecessarily complicated

## Solution

Patch the other workspace members, just like what we are doing with `tokio`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,7 +114,7 @@ jobs:
           cargo nextest run --workspace --features $TOKIO_STABLE_FEATURES
           # Removing workspace patches to run tests without path dependencies
           # (if not specified differently in the crate)
-          sed -i.bak '/^\[patch\.crates-io\]$/d; /^tokio = { path = "tokio"\s*}$/d' Cargo.toml && rm -f Cargo.toml.bak
+          perl -0 -i -pe 's/\[patch\.crates-io\].+\n\[/[/s' Cargo.toml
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -17,6 +17,10 @@ members = [
 
 [patch.crates-io]
 tokio = { path = "tokio" }
+tokio-macros = { path = "tokio-macros" }
```
mattiapitossi, line 20: This won't work, the CI performs some checks and removes the tokio patch. Locally you can use --manifest-path with the Cargo.toml you need instead of specifying the package

figsoda: not sure how CI was passing before, but the new commit should make it so CI removes the tokio-* patches too

mattiapitossi: The CI uses --manifest-path (with cargo hack) to the avoid ambiguity

mattiapitossi:
Thanks!

PR #7955, opened 2026-03-06
Add freeze_last_access_time/freeze_last_write_time for Windows

martin-g:
## Motivation

The Windows build started failing with Rust 1.94.0:
```
error[E0046]: not all trait items implemented, missing: `freeze_last_access_time`, `freeze_last_write_time`
  --> tokio\src\fs\open_options\mock_open_options.rs:12:1
   |
12 | / mock! {
13 | |     #[derive(Debug)]
14 | |     pub OpenOptions {
15 | |         pub fn append(&mut self, append: bool) -> &mut Self;
...  |
31 | |     #[cfg(windows)]
32 | |     impl OpenOptionsExt for OpenOptions {
   | |_______________________________________^ missing `freeze_last_access_time`, `freeze_last_write_time` in implementation
   |
   = note: this error originates in the macro `mock` (in Nightly builds, run with -Z macro-backtrace for more info)
   = help: implement the missing item: `fn freeze_last_access_time(&mut self, _: bool) -> &mut Self { todo!() }`
   = help: implement the missing item: `fn freeze_last_write_time(&mut self, _: bool) -> &mut Self { todo!() }`
```

https://github.com/tokio-rs/tokio/actions/runs/22751306508/job/65986377619?pr=7954

Caused by: https://github.com/rust-lang/rust/commit/d772c3d99d41d1f44a23ab5256e2fbf62b589b03
Rust issue: https://github.com/rust-lang/rust/issues/149715

## Solution

Add the missing methods

PR #7968, opened 2026-03-12
ci: workaround for OpenOptionsExt in 1.94.0

ChrisDenton:
## Motivation

There was an accidental breaking change in `std::os::windows::OpenOptionsExt` in rust 1.94 (see #7955). The fix seems like it might take awhile to land and be released so I made this PR to allow tokio's CI to keep working on stable in the meantime in case it's useful to the project.

## Solution

Instead of using extension traits for mock_open_options, implement the extension methods on OpenOptions itself. This should be robust against any future issues too (I did it for both Windows and Unix options).

### DIFF
--- a/tokio/src/fs/open_options/mock_open_options.rs
+++ b/tokio/src/fs/open_options/mock_open_options.rs
@@ -15,6 +15,9 @@ mock! {
         pub fn read(&mut self, read: bool) -> &mut Self;
         pub fn truncate(&mut self, truncate: bool) -> &mut Self;
         pub fn write(&mut self, write: bool) -> &mut Self;
+
+        // Not mocking OpenOptionsExt trait due to:
+        // https://github.com/rust-lang/rust/issues/153486
         #[cfg(unix)]
         pub fn custom_flags(&mut self, flags: i32) -> &mut Self;
         #[cfg(unix)]


ChrisDenton:
Oh, this seems to be hitting an unrelated issue with `wasm32-wasip1-thread` on 1.94.0 https://github.com/rust-lang/rust/issues/153475 😔

ChrisDenton:
Given that, I've changed this PR to not bump CI just yet (though all tests except `wasip1-threads` passed). But I'm not sure how worth it this is without that. It does make it so tokio won't be affected if something like this happens again. And tokio won't be able to implement sealed traits (like `OpenOptionsEx2`) so if they do want them in the future a solution like this would be required for them,

Darksonn:
We can always freeze rustc for the wasm target specifically. But based [on zulip](https://rust-lang.zulipchat.com/#narrow/channel/241545-t-release/topic/1.2E94.2E1/with/578221626) we may get a point release that fixes it.

ChrisDenton:
Ok, I've locked only wasip1 to 1.93.1 and now all tests are passing.

martin-g (approved):

xtqqczze:
This workaround should no longer be necessary with [Rust 1.94.1](https://releases.rs/docs/1.94.1/).

PR #7971, opened 2026-03-13
stream: add `StreamExt::filter_map_async`

figsoda:
## Motivation

This is like the async closure version of `StreamExt::filter_map`, similar to `StreamExt::then` for `StreamExt::map`.

## Solution

The code is largely adapted from `StreamExt::then` and `StreamExt::filter_map`. I'm not super sure about the name `filter_map_async`, as this will probably be what other async closure variants will be named after.

The only test is the doctest, since I couldn't find any standalone tests for the surrounding functions. I have tested it for my use case, but it might still be helpful to have more tests in tree.

I also wanted to implement `StreamExt::filter_async`, but that seemed to require `AsyncFnMut` due to lifetime issues, which hasn't been in stable Rust for 6 months yet which is the MSRV policy. So I will hold on to that until then.

### DIFF
--- a/tokio-stream/src/stream_ext/filter_map_async.rs
+++ b/tokio-stream/src/stream_ext/filter_map_async.rs
@@ -70,6 +70,12 @@ where
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        (0, self.stream.size_hint().1)
+        let future_len = usize::from(self.future.is_some());
+        let upper = self
+            .stream
+            .size_hint()
+            .1
+            .and_then(|upper| upper.checked_add(future_len));
+        (future_len, upper)
     }
 }
--- a/tokio-stream/src/stream_ext/filter_map_async.rs
+++ b/tokio-stream/src/stream_ext/filter_map_async.rs
@@ -76,6 +76,6 @@ where
             .size_hint()
             .1
             .and_then(|upper| upper.checked_add(future_len));
-        (future_len, upper)
+        (0, upper)
     }
 }
--- a/tokio-stream/src/stream_ext/filter_map_async.rs
+++ b/tokio-stream/src/stream_ext/filter_map_async.rs
@@ -52,11 +52,10 @@ where
 
         loop {
             if let Some(future) = me.future.as_mut().as_pin_mut() {
-                if let Some(item) = ready!(future.poll(cx)) {
-                    me.future.set(None);
+                let item = ready!(future.poll(cx));
+                me.future.set(None);
+                if let Some(item) = item {
                     return Poll::Ready(Some(item));
-                } else {
-                    me.future.set(None);
                 }
             }
 
--- a/tokio-stream/src/stream_ext/filter_map_async.rs
+++ b/tokio-stream/src/stream_ext/filter_map_async.rs
@@ -1,0 +1,75 @@
+use crate::Stream;
+
+use core::fmt;
+use core::future::Future;
+use core::pin::Pin;
+use core::task::{ready, Context, Poll};
+use pin_project_lite::pin_project;
+
+pin_project! {
+    /// Stream for the [`filter_map_async`](super::StreamExt::filter_map_async) method.
+    #[must_use = "streams do nothing unless polled"]
+    pub struct FilterMapAsync<St, Fut, F> {
+        #[pin]
+        stream: St,
+        #[pin]
+        future: Option<Fut>,
+        f: F,
+    }
+}
+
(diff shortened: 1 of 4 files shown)


tokio-stream/src/stream_ext/filter_map_async.rs
```diff
@@ -0,0 +1,75 @@
...
+        let mut me = self.project();
+
+        loop {
+            if let Some(future) = me.future.as_mut().as_pin_mut() {
+                if let Some(item) = ready!(future.poll(cx)) {
+                    me.future.set(None);
+                    return Poll::Ready(Some(item));
+                } else {
+                    me.future.set(None);
+                }
```
martin-g:
```suggestion
                me.future.set(None);
                if let Some(item) = ready!(future.poll(cx)) {
                    return Poll::Ready(Some(item));
                }
```

tokio-stream/src/stream_ext/filter_map_async.rs
```diff
@@ -0,0 +1,75 @@
...
+                Some(item) => {
+                    me.future.set(Some((me.f)(item)));
+                }
+                None => return Poll::Ready(None),
+            }
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        (0, self.stream.size_hint().1)
```
martin-g: See https://github.com/figsoda/tokio/blob/91d9df37420c39af90380ea7fb91b22636129ed5/tokio-stream/src/stream_ext/then.rs#L75
You will need to take into account the `future` too

tokio-stream/src/stream_ext/filter_map_async.rs
figsoda: This wouldn't work since `me.future` is borrowed by `future`. I also tried this, which hangs forever whenever the future is polled when not ready, since `me.future.set(None)` seems to set the contents of `poll` to `None` too. 

```suggestion
                let poll = future.poll(cx);
                me.future.set(None);
                if let Some(item) = ready!(poll) {
                    return Poll::Ready(Some(item));
                }
```

I've tweaked the doctest a little so it will catch this case.

tokio-stream/src/stream_ext/filter_map_async.rs
figsoda: nice catch

ADD-SP (commented):
> I also wanted to implement StreamExt::filter_async, but that seemed to require AsyncFnMut due to lifetime issues, which hasn't been in stable Rust for 6 months yet which is the MSRV policy. So I will hold on to that until then.

Yeah, the `StreamExt::filter_async` is tricky due to lifetime issue and it might be less useful before resolving this lifetime issue. I think `filter_map_async` is useful, but I prefer to hold off on implementing of `filter_async` as it is less useful at this stage.

ADD-SP (commented):
> The only test is the doctest, since I couldn't find any standalone tests for the surrounding functions. I have tested it for my use case, but it might still be helpful to have more tests in tree.

Is `tokio-stream/tests` a good place for a new test file?

tokio-stream/src/stream_ext/filter_map_async.rs
figsoda: i was overthinking it, pushed a simple solution that i overlooked

figsoda:
added some tests and fixed a few oversights

martin-g (approved):

tokio-stream/src/stream_ext/filter_map_async.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<T>> {
+        let mut me = self.project();
+
+        loop {
+            if let Some(future) = me.future.as_mut().as_pin_mut() {
+                let item = ready!(future.poll(cx));
+                me.future.set(None);
+                if let Some(item) = item {
+                    return Poll::Ready(Some(item));
```
Darksonn, line 58: So while we are polling the future, we do not poll the stream. It implies that something like `stream.buffer_unordered().filter_map_async(...)` is really dangerous because ongoing futures inside of the `buffer_unordered()` just pause out of nowhere.

On the other hand, this is a pre-existing problem with `stream.then()` too.

figsoda: Would the alternative be keeping a buffer of items, so the entire stream can be polled by awaiting on one `.next()`?

PR #7973, opened 2026-03-13
rt: shard the multi-thread inject queue to reduce remote spawn contention

alex:
The multi-threaded scheduler's inject queue was protected by a single global mutex (shared with idle coordination state). Every remote task spawn — any spawn from outside a worker thread — acquired this lock, serializing concurrent spawners and limiting throughput.

This change introduces `inject::Sharded`, which splits the inject queue into up to 8 independent shards, each an existing `Shared`/`Synced` pair with its own mutex and cache-line padding.

Design:
- Push: each thread is assigned a home shard on first push (via a global counter) and sticks with it. This keeps consecutive pushes from one thread cache-local while spreading distinct threads across distinct locks.
- Pop: workers rotate through shards starting at their own index, skipping empty shards via per-shard atomic length. pop_n drains from one shard at a time to keep critical sections bounded.
- Shard count: capped at 8 (and 1 under loom). Contention drops off steeply past a handful of shards, and is_empty()/len() scan all shards in the worker hot loop.
- is_closed: a single Release atomic set after all shards are closed, so the shutdown check stays lock-free.

Random shard selection via context::thread_rng_n (as used in #7757 for the blocking pool) was measured and found to be 20-33% slower on remote_spawn at 8+ threads. The inject workload is a tight loop of trivial pushes where producer-side cache locality dominates: with RNG, a hot thread bounces between shard cache lines on every push; with sticky assignment it stays hot on one mutex and list tail. RNG did win slightly (5-9%) on single-producer benchmarks where spreading tasks lets workers pop in parallel, but not enough to offset the regression at scale.

The inject state is removed from the global Synced mutex, which now only guards idle coordination. This also helps the single-threaded path since remote pushes no longer contend with worker park/unpark.

Results on remote_spawn benchmark (12,800 no-op tasks, N spawner threads, 64-core box):

  threads   before    after    improvement
  1         9.38 ms   7.33 ms  -22%
  2        14.94 ms   6.64 ms  -56%
  4        23.69 ms   5.34 ms  -77%
  8        34.81 ms   4.69 ms  -87%
  16       32.33 ms   4.54 ms  -86%
  32       30.37 ms   4.73 ms  -84%
  64       26.59 ms   5.34 ms  -80%

rt_multi_threaded benchmarks: spawn_many_local -8%, spawn_many_remote_idle -7%, yield_many -1%, rest neutral.

Developed in conjunction with Claude.


ADD-SP:
Is the inject queue still a FIFO queue?

alex:
Approximately -- each of the queue shards is FIFO, but nothing attempts to ensure ordering cross-shard.

tokio/src/runtime/scheduler/inject/sharded.rs
```diff
@@ -0,0 +1,293 @@
...
+
+    tokio_thread_local! {
+        /// Per-thread home shard for push operations.
+        ///
+        /// Each thread sticks to one shard for cache locality: consecutive
+        /// pushes from the same thread hit the same mutex and linked-list
+        /// tail. Distinct threads get distinct shards (modulo collisions)
+        /// via a global counter assigned on first use.
+        static PUSH_SHARD: Cell<usize> = const { Cell::new(UNASSIGNED) };
+    }
```
Darksonn: Would it be possible to place this in an existing thread local? Some systems have very limited number of thread locals available, so this is consuming valuable per-process resources.

Darksonn (commented):
In general this looks quite reasonable. Let's do it.

tokio/src/runtime/scheduler/inject/sharded.rs
```diff
@@ -0,0 +1,293 @@
...
+        for i in 0..num_shards {
+            let idx = (start + i) & self.shard_mask;
+            let shard = &*self.shards[idx];
+
+            // Fast path: skip empty shards without locking.
+            if shard.shared.is_empty() {
+                continue;
+            }
+
+            let mut synced = shard.synced.lock();
```
martin-g, line 186: I think the check done at https://github.com/tokio-rs/tokio/pull/7973/changes#diff-99033a57dffe06ad0f0dd632420d37a3eec3acdcd6baf04ddf52be06ff95c2adR232-R236 would be useful here too. 
To release the lock quickly if there is nothing to pop.

tokio/src/runtime/scheduler/inject/sharded.rs
```diff
@@ -0,0 +1,293 @@
...
+        // Close each shard under its own lock. After this loop no shard
+        // will accept a push.
+        let mut was_open = false;
+        for shard in self.shards.iter() {
+            let mut synced = shard.synced.lock();
+            was_open |= shard.shared.close(&mut synced);
+        }
+
+        // Publish the closed state for lock-free observers.
+        self.is_closed.store(true, Release);
```
martin-g: nit: this could be done conditionally:

```suggestion
        if was_open {
            self.is_closed.store(true, Release);
        }
```

tokio/src/runtime/scheduler/inject/sharded.rs
alex: I picked the existing Context thread local, hopefully that makes sense.

tokio/src/runtime/scheduler/inject/sharded.rs
alex: Done. (I'm skeptical this matters, but happy to implement!)

tokio/src/runtime/context.rs
```diff
@@ -60,6 +60,12 @@ struct Context {
     #[cfg(any(feature = "rt", feature = "macros"))]
     rng: Cell<Option<FastRand>>,
 
+    /// Home shard index for the sharded inject queue. External threads
+    /// pushing tasks are assigned a shard on first push and stick with it
+    /// for cache locality. `INJECT_SHARD_UNASSIGNED` means not yet assigned.
+    #[cfg(feature = "rt-multi-thread")]
+    inject_push_shard: Cell<usize>,
```
Darksonn, line 67: How does this interact with `block_in_place`? Let's say thread A is assigned shard 4, and then it invokes `block_in_place` and stops being a worker thread. One of the blocking threads become a worker thread to replace it. Then it might pick a different shard from 4, right? I guess this means we can end up with multiple workers using the same shard.

Darksonn: I guess this is the wrong question ... this is for when you push to a runtime from *outside* it, so you're not talking about your own shard.

alex: Yes, but that's true either way -- our max number of shards is 8, and you can have way more workers than that. A worker does not uniquely own its shard.

Darksonn: Ok, I guess it's okay then. Should we periodically pick a new shard? Perhaps after every 100 spawns? That way, if a program has two threads that continuously spawn a lot on the same shard, then eventually they pick a new shard and stop contending.

alex: Hmm, I suppose we could do that for the pathological case. 100 spawns might be too few though? If you set it too low you risk having a single thread bounce around and just dirty up everyone else's cache lines.

My preference would be to land this without the adaptive behavior and do that as a follow up -- even for the pathological case of two threads that happen to land on the same shard, this should still pareto dominate being unsharded.

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -1333,24 +1331,18 @@ impl Handle {
         }
     }
 
-    fn next_remote_task(&self) -> Option<Notified> {
+    fn next_remote_task(&self, hint: usize) -> Option<Notified> {
         if self.shared.inject.is_empty() {
             return None;
         }
```
Darksonn: Pop already has this as a fast path. I think there is no need to iterate the whole list twice.

tokio/src/runtime/context.rs
Darksonn: Ok, fair enough. But there's also another related concern of starvation. Imagine two queues A and B that both contain one job. Then repeatedly this happens:

* The active worker pops A.
* Another thread pushes to A.

In this scenario, I don't believe any new workers will be woken up, and the job from B will never be popped.

alex: Hmm, there should always be at least one worker per queue. So in this scenario you'd have also have a worker whose "home" queue is B. I didn't touch this part of the code, so I'm not an expert, but shouldn't that worker always look at the injector queue (and thus its home shard) at least from time to time?

Darksonn: Yes, there is another worker for which it is the home queue, but it's only woken up if there is work for it to consume.

https://github.com/tokio-rs/tokio/blob/ad8c59add6a1988d8c327fb3358beeeae3bbb5cd/tokio/src/runtime/scheduler/multi_thread/idle.rs#L51-L60

The way it ends up working right now is that the worker that is alive will pull multiple items from the inject queue, which results in another worker being woken up to perform work stealing. But with this sharding change, it may only get a single item from the inject queue, so it won't wake up another thread.

alex: Hmmm, I guess the question then is how did you end up with an item in B queue but that worker never woken up?

I'm ok implementing the occasional rehoming logic, I just want to make sure I have a clear picture of whether it's there to address a potential pathlogical performance case, or to fix a liveness issue -- I don't think it should be needed for correctness, so if there is a correctness bug I want to understand it on its own terms (and ideally write a test case).

Darksonn: I'm raising the concern as a liveness issue, which per [this guarantee](https://docs.rs/tokio/latest/tokio/runtime/index.html#detailed-runtime-behavior) makes it a correctness concern.

> how did you end up with an item in B queue but that worker never woken up?

The code for waking up a worker after pushing to the queue is here:

https://github.com/tokio-rs/tokio/blob/e5ab8fb14b9fe1191ebe419eb5f930433f2f5e22/tokio/src/runtime/scheduler/multi_thread/worker.rs#L1287-L1289

There is no relationship between which shard the item was pushed to, and which worker is woken up. All this code ensures that, after pushing an item, then if all workers are idle, then a worker is woken up. But it might not be the same worker as where the item was pushed, and if there is already a non-idle worker searching for work, no wakeup occurs.

Darksonn: See also https://github.com/tokio-rs/tokio/pull/8029

Darksonn: I've been thinking more about this, and I've come to the conclusion that it is okay as-is. As a perf matter, it could be beneficial to give `notify_parked_remote()` a hint about which worker it should prefer to wake up, but not a blocker.

Darksonn:
Please rebase or merge master to avoid conflicts with the LIFO slot changes.

Darksonn (approved):
LGTM

Darksonn:
Ah, looks like loom timed out?

tokio/src/runtime/scheduler/inject/sharded.rs
```diff
@@ -0,0 +1,282 @@
...
+            self.is_closed.store(true, Release);
+        }
+
+        was_open
+    }
+
+    /// Pushes a task into the queue.
+    ///
+    /// Selects a shard using the calling thread's home-shard index. Does
+    /// nothing if the selected shard is closed (which implies all shards
```
martin-g: > ... which implies all shards are closed ...

This is not true if close() is still running. It closes the shards one by one. So if the first shard is just closed and push() is called on it then there is no guarantee that the rest of the shards are closed too.

tokio/src/runtime/scheduler/inject/sharded.rs
```diff
@@ -0,0 +1,282 @@
...
+const MAX_SHARDS: usize = 1;
+
+#[cfg(not(loom))]
+const MAX_SHARDS: usize = 8;
+
+impl<T: 'static> Sharded<T> {
+    /// Creates a new sharded inject queue with a shard count derived
+    /// from the requested hint (rounded up to a power of two).
+    pub(crate) fn new(shard_hint: usize) -> Sharded<T> {
+        let num_shards = shard_hint.clamp(1, MAX_SHARDS).next_power_of_two();
```
martin-g, line 71: Maybe it should be documented that `MAX_SHARDS` must be a power of two.
If someone changes it to 6, for example, and `shard_hint` is bigger than 4 then `num_shards` will be `8` which would be bigger than `MAX_SHARDS`.

tokio/src/runtime/scheduler/inject/sharded.rs
```diff
@@ -0,0 +1,282 @@
...
+        was_open
+    }
+
+    /// Pushes a task into the queue.
+    ///
+    /// Selects a shard using the calling thread's home-shard index. Does
+    /// nothing if the selected shard is closed (which implies all shards
+    /// are closed, as `close` is the only path that sets the flag).
+    pub(crate) fn push(&self, task: task::Notified<T>) {
+        let idx = self.next_push_shard();
```
martin-g, line 144:
```suggestion
        if self.is_closed() {
            return;
        }
        
        let idx = self.next_push_shard();
```
Minor optimization to avoid acquiring the lock below.

tokio/src/runtime/scheduler/inject/sharded.rs
```diff
@@ -0,0 +1,282 @@
...
+        unsafe { shard.shared.push(&mut synced, task) };
+    }
+
+    /// Pushes a batch of tasks. The whole batch is placed in a single
+    /// shard to avoid taking multiple locks.
+    pub(crate) fn push_batch<I>(&self, iter: I)
+    where
+        I: Iterator<Item = task::Notified<T>>,
+    {
+        let idx = self.next_push_shard();
```
martin-g, line 158:
```suggestion
        if self.is_closed() {
            return;
        }
        
        let idx = self.next_push_shard();
```
Similar optimization as for `push()`

martin-g (approved):

tokio/src/runtime/scheduler/inject/sharded.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    /// Returns `true` if every shard reports empty.
+    pub(crate) fn is_empty(&self) -> bool {
+        for shard in self.shards.iter() {
+            if !shard.shared.is_empty() {
+                return false;
+            }
+        }
+        true
```
hawkw, line 111: take it or leave it, but this could also be written as:
```suggestion
       self.shards.iter().all(|shard| shard.is_empty())
```

tokio/src/runtime/context.rs
hawkw: I wonder if we can easily produce a test case for this scenario, perhaps by having one worker execute a  future that  just spawns a big pile of tasks in a loop without yielding? That's not a totally contrived scenario as you might imagine an accept loop or something under load where there's basically always new connections to handle without having to wait for a long period of time.

tokio/src/runtime/scheduler/inject/sharded.rs
alex: seems like this optimizes for the case where the queue is closed, which should be rare -- and which pessimizes normal operations. am I missing something?

alex:
Is there a recommended way to debug the loom tests timing? Looks like even in normal operations they take several hours :-( I'm assuming the likely cause of timeouts is accidentally making the state space too large? (I was hoping that limiting to 1 shard on loom would prevent that, but perhaps not)

Darksonn:
Limiting it to one shard is not necessarily enough. Any operation involved in concurrency (such as atomics or taking/releasing mutexes) that happens while other threads exists is a location where preemption can happen in loom, so as you introduce more mutex lock/unlocks during the part of the test with more than one thread, that increases the search space.

tokio/src/runtime/context.rs
Darksonn: The reason I'm no longer worried about it is that this scenario will cause all worker threads to wake up, and once all worker threads are alive, you can't have this kind of starvation.

alex:
Ah yeah. Same question though, what's the best way to debug it :-)

Darksonn:
If there's an actual bug that it's catching, I'll do it with printlns to understand the specific interleaving it fails on. I'm not sure when it's just too many interleavings. Probably the test has to be simplified.

alex:
Looks like the latest change was sufficient, hopefully it looks reasonable.

hawkw (changes requested):
Given that the previous change to use a sharded queue in `spawn_blocking` (#7757) introduced a regression that caused programs using `spawn_blocking` to hang (#8056), and that change ultimately had to be reverted (in #8057), I think we should be cautious about moving forwards with this.

In particular, I feel like --- as a substantial change to runtime internals that potentially affects all uses of the multi-threaded runtime, and which introduces a complex new concurrent data structure involving unsafe code --- this is the type of change which should really be introduced as an opt-in `tokio::runtime::Builder` setting which requires `tokio_unstable`. This way, we can allow users to start testing this in production without running the risk of introducing regressions which block users from picking up new tokio releases. We've taken this approach in the past for changes such as the [alternative timer implementation](https://docs.rs/tokio/1.52.0/tokio/runtime/struct.Builder.html#method.enable_alt_timer) (#7467) and [eager I/O driver handoff](https://docs.rs/tokio/1.52.0/tokio/runtime/struct.Builder.html#method.enable_eager_driver_handoff) (#8010)[^1], and I think it would be appropriate to do something similar here, as well.

[^1]: The latter of which is a _much_ smaller change!

alex:
On my TODO list to figure out how practice this is this weekend.

hawkw:
> On my TODO list to figure out how practice this is this weekend.

Great! If it's possible, I think that starting out as an opt-in experimental feature can be a useful way to get big improvements to runtime internals like this one merged faster and start trying them out in production

alex:
The "how [practical] this is" was specifically making this opt in.

alex:
Ok, spent some time this morning looking at a "pre refactor" (pre-factor?) to make it easy for us to support both the shared lock and sharded lock queues. Unfortunately, I'm extremely unhappy with the results -- the current diff is +340/-219 (just for the refactor), basically as big as this patch. And there's a bunch of annoying design pieces.

What I want with is an `enum Inject { Shared(SharedInject) }` where `SharedInject` has the `Mutex` and the `inject::Shared`. And then all the APIs from `SharedInject` are exposed on `Inject` by matching on all self-variants. When we go to added `Sharded`, that should be the only thing that needs to change

This leads to splitting off the Mutex that's currently covering both `idle` and `inject` (in `multi_threaded::Shared`) -- I think this is ok, but it's probably the most significant change.

The other obvious change is that `trace_multi_thread` takes a lock for each `pop`, rather than one lock for the whole time. I think this could probably be fixed with a new `pop_all()` (the current PR has this same design).

I also ended up with 2x implementations of `Pop`, one for where the `MutexGuard` is owned and another for where i's bothered.

All this to say: I think this can be done, the PR is much larger than I'd like. If anyone has any for a better architecture, let me know and I can mull on it. Otherwise I think I'll make https://github.com/tokio-rs/tokio/pull/8068 my priority and come back this once that's done.

alex:
(I'm also happy to put it up as a draft PR if it's of interest to anyone.)

hawkw:
So, very naively, I had kind of hoped we would be able to do this by making the number of shards a parameter that's provided when the sharded inject queue is constructed, and having the runtime builder either construct it with 1 or 8 shards depending on whether the sharded queue is enabled. But, this only works if the behavior of the queue with a single shard is more or less equivalent to a single queue. It sounds like this not actually be the case, though, because of splitting the mutex currently guarding both `idle` and `inject` into two separate mutices around the inject queue and idle worker state. This would, unfortunately, mean that once this is released, programs which do not enable the sharded inject queue still get different behavior relative to the current Tokio version, since these mutexen are separate...

I actually think that we should probably be at least a little concerned about separating the inject queue and idle state mutex. Off the top of my head, I can't think of any code that relies on the assumption that these two pieces of shared state are not modified concurrently, but I think we'll have to look closely to make sure that's the case.

alex:
Yeah, I think that's the right summary.

Maybe there's an even more incremental change of just using different mutexes for idle and inject in the existing shared structure? It's a tiny diff and should be the biggest semantic risk for the migration?

alex:
Going to close this, will open the replacement PR that does this incrementally (and off by default) later this evening. Thanks for the feedback!

Issue #7970, opened 2026-03-12
feat: opt-in file descriptor table pre-warming behind `tokio_unstable`


jlizen:
# feat: opt-in file descriptor table pre-warming behind `tokio_unstable`

Related: [tokio-rs/tokio#2702 (comment)](https://github.com/tokio-rs/tokio/issues/2702#issuecomment-664764992)

## Context

In #2702, @carllerche traced ~28ms tail latency spikes to `libc::socket` blocking while the kernel grows its FD table. The kernel uses RCU to synchronize the FD table in multi-threaded processes, so when the table needs to grow, the calling thread waits for all RCU readers to quiesce. He confirmed that pre-warming the table via `dup` eliminates the spike (p100 dropped from ~29ms to ~291µs).

I'm profiling a closed-source tokio service with [dial9-tokio-telemetry](https://github.com/dial9-rs/dial9-tokio-telemetry) and seeing latency spikes that I'm suspicious are the same pattern: multi-ms stalls on worker threads during connection bursts. Still collecting data to confirm the causal link.

I also dug around and found a few internal teams that had similar pre warming hacks via `dup` due to observed latency in their tokio runtimes.

## Why tokio specifically

The `socket()` call happens on runtime worker threads. When the RCU stall hits, it blocks the entire worker's event loop, not just one task. And tokio services tend to open many connections concurrently, which is what triggers table resizes.

## What I'd propose

A builder method behind `tokio_unstable`:

```rust
let rt = tokio::runtime::Builder::new_multi_thread()
    .prewarm_fd_table(1024)
    .enable_all()
    .build()?;
```

Implementation: `dup` a single FD up to the target, then close them all. The kernel FD table only grows, never shrinks during a process's lifetime, so the pre-warmed capacity is durable even after closing the duplicated FDs.

This is a mitigation, not a complete fix. It front-loads table growth before workers spawn, which eliminates the common case where the table is small at process start. If a burst at runtime pushes past the pre-warmed size, the RCU stall can still occur at the next growth boundary. But it removes the most predictable trigger and gives operators a knob to tune based on their expected FD usage.

For the runtime growth case, deeper mitigations could be explored as future scope: offloading `socket()` calls to a blocking thread so the RCU stall doesn't block the worker event loop, or adaptive pre-warming that tracks FD high-water marks and grows the table ahead of demand.

`tokio_unstable` makes sense here because the right value is workload-dependent, and this is defensive against OS behavior rather than core runtime functionality.

[...]

Issue #2702, opened 2020-07-26
rt: high tail latencies with threaded scheduler when under load


carllerche:
Given a TCP echo client that:

- Opens 200~1000 connections.
- Sends 8kb of data on each connection.
- Reads the data back.

Running this client on the threaded runtime results in significant tail latencies.

# Code

```rust
use tokio::io::{AsyncReadExt, AsyncWriteExt};
use tokio::net::TcpStream;
use std::net::SocketAddr;
use std::sync::Arc;
use std::time::Instant;

#[tokio::main(core_threads=2)]
async fn main() {
    const N: usize = 200;
    const STEP: usize = N/10;
    const BUFFER: usize = 8*1024;

    let addr: SocketAddr = "127.0.0.1:8080".parse().unwrap();
    let data = Arc::new(vec![10_u8; BUFFER]);

    let mut handles = Vec::with_capacity(N);
    let mut elapsed = Vec::with_capacity(N);

    for i in 0..N {
        if i % 10 == 9 {
            tokio::task::yield_now().await;
        }

        let data = data.clone();
        let now = Instant::now();

        handles.push(tokio::spawn(async move {
            let tts = now.elapsed();
            let mut buf = vec![10_u8; BUFFER];
            let mut socket = TcpStream::connect(addr).await.unwrap();

            socket.write_all(&data).await.unwrap();
            socket.read_exact(&mut buf).await.unwrap();

            assert_eq!(buf[..], data[..]);

            // now.elapsed()
            tts
        }));
    }

    for handle in handles.drain(..) {
        elapsed.push(handle.await.unwrap());
    }

    elapsed.sort();

    let mut i = STEP;

    while i <= N {
        println!("{}th = {:?}; tts", i, elapsed[i-1]);
        i += STEP;
    }
}
```

# Output

Running this against the `echo` example results in the following output:

```
20th = 241.19µs; tts
40th = 476.356µs; tts
60th = 697.187µs; tts
80th = 23.421715ms; tts
100th = 23.651342ms; tts
120th = 23.894434ms; tts
140th = 48.230702ms; tts
160th = 49.305577ms; tts
180th = 50.433882ms; tts
200th = 51.743111ms; tts
```

Compare this with `basic_scheduler`

```
20th = 112.65µs; tts
40th = 129.878µs; tts
60th = 145.256µs; tts
80th = 159.83µs; tts
100th = 176.598µs; tts
120th = 192.344µs; tts
140th = 209.232µs; tts
[...]

PR #7978, opened 2026-03-19
docs(runtime): document FD table pre-warming workaround

jlizen:
Related:#2702, #7970

## Motivation

On Linux, the kernel grows the per-process file descriptor table lazily. The growth operation uses RCU synchronization, which in a multi-threaded process can block the calling thread for tens of milliseconds while it waits for all RCU readers to quiesce. When this happens on a tokio worker thread (e.g. during a `socket()` syscall), the entire event loop stalls, not just the affected task.

This was first identified by @carllerche in #2702 (p100 dropped from ~29ms to ~291µs after pre-warming). Multiple teams have independently discovered the same `dup`/`close` workaround, but it isn't documented anywhere in tokio.

## Solution

Adds an example (`examples/prewarm-fd-table.rs`) demonstrating the workaround: use `fcntl(fd, F_DUPFD_CLOEXEC, target)` to force the kernel to expand the FD table before building the runtime. The kernel never shrinks the table, so the capacity persists after closing the duplicated FDs.

Also starts a `Performance tuning` section in the `runtime` docs, with a small note on fd prewarming.

### DIFF
--- a/benches/remote_spawn.rs
+++ b/benches/remote_spawn.rs
@@ -19,7 +19,10 @@ use tokio::runtime::{self, Runtime};
 /// Total number of tasks spawned across all threads per iteration.
 /// Must be divisible by the largest parallelism level (64).
 const TOTAL_TASKS: usize = 12_800;
-const _: () = assert!(TOTAL_TASKS % 64 == 0, "TOTAL_TASKS must be divisible by 64");
+const _: () = assert!(
+    TOTAL_TASKS.is_multiple_of(64),
+    "TOTAL_TASKS must be divisible by 64"
+);
 
 fn remote_spawn_contention(c: &mut Criterion) {
     let parallelism_levels = parallelism_levels();


tokio/src/runtime/mod.rs
```diff
@@ -373,6 +373,12 @@
 //! When a task is woken from a thread that is not a worker thread, then the
 //! task is placed in the global queue.
 //!
+//! On Linux, worker threads can also stall when a syscall like `socket()`
+//! triggers a file descriptor table resize. The kernel uses RCU synchronization
+//! for this, which can block for tens of milliseconds in multi-threaded
+//! processes. Pre-warming the FD table before building the runtime avoids this;
+//! see the [`prewarm-fd-table`] example.
```
Darksonn: The detailed runtime behavior section is not the place for this note. Please move it elsewhere, possibly to a new section.

examples/prewarm-fd-table.rs
```diff
@@ -0,0 +1,80 @@
...
+    // Use F_DUPFD_CLOEXEC to duplicate to a slot >= target. This forces the
+    // kernel to expand the FD table in a single syscall, without clobbering
+    // existing FDs. CLOEXEC prevents leaking to child processes.
+    let raw = unsafe {
+        libc::fcntl(
+            std::os::unix::io::AsRawFd::as_raw_fd(&dev_null),
+            libc::F_DUPFD_CLOEXEC,
+            target,
+        )
+    };
```
Darksonn, line 42: Please implement this example using safe methods from `nix` if possible.

Darksonn (commented):
Thanks!

examples/prewarm-fd-table.rs
Mark-Simulacrum: I think I've seen an implementation just based on `File::open` + `File::try_clone`, that would probably be even better than `nix`.

tokio/src/runtime/mod.rs
jlizen: There isn't anywhere else that is immediately obvious. I can start a `Performance tuning` section, which I imagine will collect other things over time?

Darksonn: Sounds good to me.

examples/prewarm-fd-table.rs
jlizen: I'll cut it over to `nix`. The issue with the `File::open/try_clone` approach @Mark-Simulacrum is that you can't target a specific FD slot, so you need N syscalls to expand to N slots.

(Glad to show both, also, if you think the perk of being stdlib-only warrants the downside)

jlizen: Actually, with nix, we still need unsafe for `from_raw_fd`.

We can do it safely with `rustix` though via `rustix::io::fcntl_dupfd_cloexec`.

Darksonn: Okay, then let's stick with `libc`.

With regards to `rustix`, I'm way of recommending crates that perform raw syscalls by default instead of calling into libc.

jlizen:
CI failures from netlify:
```
12:51:08 PM: Starting to prepare the repo for build
12:51:08 PM: Preparing Git Reference pull/7978/head
12:51:09 PM: Failed during stage 'preparing repo': Unable to access repository. The repository may have been deleted, the branch may not exist, or permissions may have changed: warning: unable to access 'tests-build/.gitattributes': Permission denied
error: unable to create file tests-build/Cargo.toml: Permission denied
error: unable to create file tests-build/README.md: Permission denied
fatal: cannot create directory at 'tests-build/src': Permission denied
: exit status 128
12:51:09 PM: Error checking out to pull/7978/head
12:51:09 PM: Failing build: Failed to prepare repo
```

(don't have perms to retry)

Darksonn:
You can push an empty commit to retry.

jlizen:
Same failure across retries. Might need a netlify cache clear.

examples/prewarm-fd-table.rs
```diff
@@ -0,0 +1,80 @@
...
+
+fn main() {
+    const FD_TARGET: u32 = 10_000;
+
+    println!("Pre-warming FD table to {FD_TARGET} entries...");
+    if let Err(e) = prewarm_fd_table(FD_TARGET) {
+        eprintln!("Warning: failed to pre-warm FD table: {e}");
+    } else {
+        println!("FD table pre-warmed successfully.");
+    }
```
martin-g:
```suggestion
    #[cfg(target_os = "linux")]
    {
      const FD_TARGET: u32 = 10_000;
  
      println!("Pre-warming FD table to {FD_TARGET} entries...");
      if let Err(e) = prewarm_fd_table(FD_TARGET) {
          eprintln!("Warning: failed to pre-warm FD table: {e}");
      } else {
          println!("FD table pre-warmed successfully.");
      }
    }
```
The println!()s won't make much sense for non-Linux

jlizen:
CI green now 👍

examples/prewarm-fd-table.rs
```diff
@@ -0,0 +1,80 @@
...
+    let dev_null = std::fs::File::open("/dev/null")?;
+
+    // Use F_DUPFD_CLOEXEC to duplicate to a slot >= target. This forces the
+    // kernel to expand the FD table in a single syscall, without clobbering
+    // existing FDs. CLOEXEC prevents leaking to child processes.
+    let raw = unsafe {
+        libc::fcntl(
+            std::os::unix::io::AsRawFd::as_raw_fd(&dev_null),
+            libc::F_DUPFD_CLOEXEC,
+            target,
```
martin-g, line 40: nit: The `target` is `c_int`, i.e. i32. Passing anything bigger than i32::MAX will lead to an overflow and a failure.

This is just an example, so there is no need to be perfect, but people will probably copy it verbatim.

Maybe change `target` to i32 and add an assertion that it is positive ?!

examples/prewarm-fd-table.rs
```diff
@@ -0,0 +1,80 @@
...
+//! blocking the entire event loop (not just one task).
+//!
+//! The workaround is to force the kernel to expand the FD table once per process
+//! (before any runtime starts), by duplicating an FD to a high slot and then
+//! closing it. The kernel never shrinks the FD table during a process's lifetime,
+//! so the capacity persists.
+//!
+//! This is most relevant for services that open many connections concurrently
+//! (e.g. HTTP servers, connection pools). The pre-warm target should be at least
+//! your expected peak FD count.
```
martin-g: Maybe also mention `RLIMIT_NOFILE` ?! 
Trying to set a bigger value will lead to a failure.

examples/prewarm-fd-table.rs
Mark-Simulacrum: On the other hand, the File approach is always safe, while this one requires that you pick an fd that isn't being used. I'm not sure there's any fully safe way to do it?

jlizen: i'll show both and add a note about the tradeoffs

jlizen: > this one requires that you pick an fd that isn't being used 

`F_DUPFD_CLOEXEC` will pick the lowest unallocated FD >= the target, so you don't need to strictly pick an unused one.  

But anyway it's nice to show a fully safe version with no dependencies besides stdlib.

ADD-SP (approved):
Looks good to me.

Shall we use this commit message？
```
examples: add FD table pre-warming example
```

jlizen:
That commit message seems fine to me. Appreciate the eyes, all!

jlizen:
> rustdoc link targets the GitHub master path — consider a docs.rs or intra-doc link so it does not rot across branches

Is there a good way to link to the example from within the rustdocs without naming the branch? AIUI intradoc / rustdoc links are not available for examples.

martin-g:
> Is there a good way to link to the example from within the rustdocs without naming the branch? AIUI intradoc / rustdoc links are not available for examples.

I think it is fine as it is.

jlizen:
Were there any blockers for merging, @ADD-SP ?

ADD-SP:
Let's go!

Issue #7979, opened 2026-03-20
Cancelled io-uring open futures leak fds


Darksonn:
We received the below report from Mufeed VH of Winfunc Research. Thank you @mufeedvh for reporting this! I've determined that this bug does not need to be handled under embargo, so I am posting it publicly on our bug tracker.

---

## Summary
An attacker who can trigger and cancel Tokio `io-uring` file opens in a Linux deployment can leak process file descriptors, leading to file-descriptor exhaustion and denial of service.

## Severity
LOW

CVSS v3.1: `AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:N/A:L` (3.7)

## Root Cause
Tokio's Linux `io-uring` filesystem path creates an `OpenAt` operation that only adopts the returned file descriptor if `Open::complete` runs. When the future is dropped after first poll, Tokio switches the operation lifecycle to `Cancelled`, but later completion handling discards the CQE result instead of adopting or closing a successful fd.

Relevant code:

- [`tokio/src/fs/open_options.rs#L520-L541`](https://github.com/tokio-rs/tokio/blob/ee04abe7970b769439e2fecd4b1dd8790e4234e1/tokio/src/fs/open_options.rs#L520-L541)
- [`tokio/src/io/uring/open.rs#L20-L25`](https://github.com/tokio-rs/tokio/blob/ee04abe7970b769439e2fecd4b1dd8790e4234e1/tokio/src/io/uring/open.rs#L20-L25)
- [`tokio/src/runtime/driver/op.rs#L77-L91`](https://github.com/tokio-rs/tokio/blob/ee04abe7970b769439e2fecd4b1dd8790e4234e1/tokio/src/runtime/driver/op.rs#L77-L91)
- [`tokio/src/runtime/io/driver/uring.rs#L75-L84`](https://github.com/tokio-rs/tokio/blob/ee04abe7970b769439e2fecd4b1dd8790e4234e1/tokio/src/runtime/io/driver/uring.rs#L75-L84)
- [`tokio/src/runtime/io/driver/uring.rs#L259-L279`](https://github.com/tokio-rs/tokio/blob/ee04abe7970b769439e2fecd4b1dd8790e4234e1/tokio/src/runtime/io/driver/uring.rs#L259-L279)

## Proof of Exploitability
[...]

PR #7983, opened 2026-03-21
rt: do not leak fd when cancelling io_uring open operation

mattiapitossi:
## Motivation

fixes: #7979 

## Solution

When we receive a CQ that contains a cancelled event (open op), we now close the related fd.

## result of PoC run after the fix

```sh
thread 'main' (1) panicked at src/main.rs:58:9:
expected many leaked fds, saw 1
before=7
after=8
leaked=1
```

### DIFF
--- a/tokio/Cargo.toml
+++ b/tokio/Cargo.toml
@@ -6,7 +6,7 @@ name = "tokio"
 #   - README.md
 # - Update CHANGELOG.md.
 # - Create "v1.x.y" git tag.
-version = "1.51.1"
+version = "1.51.0"
 edition = "2021"
 rust-version = "1.71"
 authors = ["Tokio Contributors <team@tokio.rs>"]
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Make sure you enable the full features of the tokio crate on Cargo.toml:
 
 ```toml
 [dependencies]
-tokio = { version = "1.51.1", features = ["full"] }
+tokio = { version = "1.51.0", features = ["full"] }
 ```
 Then, on your main.rs:
 
--- a/tokio/CHANGELOG.md
+++ b/tokio/CHANGELOG.md
@@ -1,14 +1,3 @@
-# 1.51.1 (April x, 2026)
-
-This release fixes an issue where cancelling a file opening operation
-could lead to a leak of file descriptors on Linux when using io-uring.
-
-### Unstable
-
-- runtime: (io-uring) ensure fds are closed when invoking cancel op ([#7983])
-
-[#7983]: https://github.com/tokio-rs/tokio/pull/7983
-
 # 1.51.0 (April 3rd, 2026)
 
 ### Added
--- a/tokio/README.md
+++ b/tokio/README.md
@@ -60,7 +60,7 @@ Make sure you enable the full features of the tokio crate on Cargo.toml:
 
 ```toml
 [dependencies]
-tokio = { version = "1.51.1", features = ["full"] }
+tokio = { version = "1.51.0", features = ["full"] }
 ```
 Then, on your main.rs:
 


tokio/src/runtime/io/driver/uring.rs
```diff
@@ -77,9 +79,15 @@ impl UringContext {
                     waker.wake_by_ref();
                     *ops.get_mut(idx).unwrap() = Lifecycle::Completed(cqe);
                 }
-                Some(Lifecycle::Cancelled(_)) => {
+                Some(Lifecycle::Cancelled(cancel_data)) => {
```
ADD-SP, line 82: nit: How about this?
```rust
Some(Lifecycle::Cancelled(CancelData::Open(_)))
```

*not a blocker*

tokio/src/runtime/io/driver/uring.rs
```diff
@@ -77,9 +79,15 @@ impl UringContext {
...
+                    if let CancelData::Open(_) = cancel_data {
+                        if let Ok(fd) = CqeResult::from(cqe).result {
+                            // SAFETY: the fd comes from the CQE result
+                            // of a cancelled event still to be closed
+                            unsafe { libc::close(fd as i32) };
```
ADD-SP: I prefer reling on standard library: `drop(std::os::fd::OwnedFd)`.

tokio/tests/fs_uring.rs
```diff
@@ -145,6 +146,53 @@ async fn cancel_op_future() {
     assert!(res.is_cancelled());
 }
 
+// see: https://github.com/tokio-rs/tokio/issues/7979
+#[tokio::test]
+async fn file_descriptors_are_closed_when_cancelling_open_op() {
+    let (_tmp_file, path): (Vec<NamedTempFile>, Vec<PathBuf>) = create_tmp_files(1);
+
+    let fd_count_before_access = fs::read_dir("/proc/self/fd").unwrap().count();
```
ADD-SP: `cargo test` is multi-threaded, so concurrent tests may affect `/proc/self/fd`, potentially leading to flaky tests. Do we have a better approach? (TBH, I don't have one yet).

*this is not an issue when using `cargo-nextest`*

Darksonn: We can move the test to its own file. Each file in the `tests/` directory is a separate test process.

tokio/tests/fs_uring.rs
```diff
@@ -145,6 +146,53 @@ async fn cancel_op_future() {
...
+        handle.abort();
+
+        let res = handle.await.unwrap_err();
+        assert!(res.is_cancelled());
+    }
+
+    let fd_count_after_cancel = fs::read_dir("/proc/self/fd").unwrap().count();
+    let leaked = fd_count_after_cancel.saturating_sub(fd_count_before_access);
+
+    assert!(leaked <= 64);
```
ADD-SP: We need a comment for this assertion to explain the magic number (64).

tokio/src/runtime/io/driver/uring.rs
mattiapitossi: Thanks, I prefer the version you suggested to avoid nested ifs, updated!

tokio/src/runtime/io/driver/uring.rs
```diff
@@ -77,9 +79,14 @@ impl UringContext {
                     waker.wake_by_ref();
                     *ops.get_mut(idx).unwrap() = Lifecycle::Completed(cqe);
                 }
-                Some(Lifecycle::Cancelled(_)) => {
+                Some(Lifecycle::Cancelled(CancelData::Open(_))) => {
```
martin-g: Now only canceled Open ops are handled.
But canceled Read/Write/... ops go to `Some(other) => panic!(...)`. Those should still do `ops.remove(idx);`

tokio/tests/fs_uring.rs
```diff
@@ -145,6 +146,53 @@ async fn cancel_op_future() {
...
+                    let mut opt = tokio::fs::OpenOptions::new();
+                    opt.read(true);
+                    opt
+                };
+
+                let fut = opt.open(&path[0]);
+
+                // If io_uring is enabled (and not falling back to the thread pool),
+                // the first poll should return Pending.
+                let _pending = Box::pin(fut).poll_unpin(cx);
```
martin-g:
```suggestion
                assert_pending!(Box::pin(fut).poll_unpin(cx));
```

tokio/src/runtime/io/driver/uring.rs
martin-g, line 169: Could FDs leak here too ?

Darksonn:
We can make a point release with this if you change the base to `tokio-1.50.x`

tokio/src/runtime/io/driver/uring.rs
ADD-SP: Emm, I didn't see the update, have you pushed this commit?

tokio/src/runtime/io/driver/uring.rs
mattiapitossi: Ouch, good catch! I reverted the commit

tokio/src/runtime/io/driver/uring.rs
mattiapitossi: Yes, sorry, actually I reverted it because as [pointed out](https://github.com/tokio-rs/tokio/pull/7983#discussion_r2973566612) by Martin this was excluding the cases where we remove the data from the slab when the operation is read/write. There is also the possibility to match this and another case below, but I prefer this version, wdyt?

tokio/tests/fs_uring.rs
mattiapitossi: Done in the new test file, thanks!

mattiapitossi: After I've added this check the CI started to fail, after some investigation I discovered that the other test didn't add the check intentionally because the first polling might return ready in some cases (see: https://github.com/tokio-rs/tokio/pull/7321/changes/da0175b7b5f3a46b79faa742db23d899453c9caf)

tokio/src/runtime/io/driver/uring.rs
ADD-SP: How about this?

```rust
match {
    Some(Lifecycle::Cancelled(CancelData::Open(_))) => {}
    Some(Lifecycle::Cancelled(_)) => {}
}
```

Darksonn: If both branches have shared logic, I would not use two separate branches like that.

mattiapitossi: Yes, this is the alternative I was referring to, but both share some logics (ops.remove(idx);) so you end up with:

```rust
match {
    Some(Lifecycle::Cancelled(CancelData::Open(_))) => {
      -- close fd
      ops.remove(idx);
    }
    Some(Lifecycle::Cancelled(_)) => {
      ops.remove(idx);
    }
}
```
And since we are in both cases in the cancelled lifecycle you could still have some other repeated common logic in the future. Probably better to keep this way, what do you think?

tokio/tests/fs_uring.rs
mattiapitossi: Added in the new file

tokio/src/runtime/io/driver/uring.rs
ADD-SP: not a blocker, just a style preference.

tokio/src/runtime/io/driver/uring.rs
ADD-SP: Not likely, the `tokio::fs::File` holds the `std::fs::File`, I think the FD will be closed automatically on drop.

ADD-SP: Interesting, @mattiapitossi will this leak FDs?

mattiapitossi: Unless I'm not considering something, I would expect that when the process finishes all the FDs will be automatically closed. The drop is just ensuring that every operations terminates, so I'm not expecting that we should also close the related fd.

mattiapitossi: There could be some situations though where the runtime gets dropped, but the process is still running so you could still have the leak in those cases (unless we handled all the cases in the dispatch_completion). So probably, for those cases, adding a close in the drop does not hurt?

Darksonn: We should close fds in runtime exit too. There are valid use-cases to create and destroy many runtimes throughout lifetime of single program.

mufeedvh:
The finder of this issue here.

I think this still needs changes: the fix covers `Cancelled(Open)` in `dispatch_completions()` / runtime drop, but it does **not** cover the case where the open CQE has already been stored as `Lifecycle::Completed(cqe)` and the future is then dropped before being re-polled.

That path still leaks because `Open::complete()` is the only place that adopts the fd, while `cancel_op()` currently does this for `Lifecycle::Completed(_)`:

```rust
Lifecycle::Completed(_) => {
    ops.remove(index);
}
```

So a successful completed `OpenAt` result can still be discarded without adopting or closing the fd.

I reproduced this on this PR branch by:
1. polling `open()` once to `Pending`,
2. letting it complete and get stored as `Completed(cqe)`,
3. dropping/aborting before the inner `open()` is polled again.

---

<details>
<summary>Runnable PoC for the completed-then-dropped-before-repoll path</summary>

```toml
[package]
name = "tokio_uring_completed_drop_poc"
version = "0.1.0"
edition = "2021"

[workspace]

[dependencies]
tokio = { path = "$REPO_ROOT/tokio", features = ["fs", "rt", "rt-multi-thread", "time", "io-uring", "sync"] }
```

```rs
use std::fs;
use std::future::Future;
use std::path::PathBuf;
use std::pin::Pin;
use std::time::Duration;

use tokio::sync::mpsc::{unbounded_channel, UnboundedSender};
use tokio::time::timeout;

/// Count currently-open fds in this process.
fn fd_count() -> usize {
    fs::read_dir("/proc/self/fd").unwrap().count()
}

/// Wrapper future used to hit the missing cleanup path.
///
/// First poll:
/// - polls the inner `tokio::fs::OpenOptions::open()` future once,
/// - expects `Pending` so we know we took the io_uring path,
/// - registers the task waker with Tokio's uring machinery.
///
/// Second poll:
/// - happens after the kernel completes the open and Tokio stores the CQE as
///   `Lifecycle::Completed(cqe)` and wakes the task,
/// - **intentionally does not poll the inner open future again**,
/// - stays pending forever so the task can be aborted.
///
/// Aborting the task here drops the inner `open()` future while Tokio still has
/// a completed CQE sitting in the slab. Today that goes through
/// `cancel_op(... Lifecycle::Completed(_))`, which removes the slab entry
/// without adopting or closing the returned fd.
struct PollOpenOnceThenNeverRepoll<F> {
    inner: Pin<Box<F>>,
    first_poll_tx: Option<UnboundedSender<()>>,
    second_poll_tx: Option<UnboundedSender<()>>,
    polled_once: bool,
}

impl<F: Future> Future for PollOpenOnceThenNeverRepoll<F> {
    type Output = ();

    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
        if !self.polled_once {
            let poll = self.inner.as_mut().poll(cx);
            assert!(
                matches!(poll, Poll::Pending),
                "open did not use io_uring path in this environment"
            );

            self.polled_once = true;
            self.first_poll_tx.take().unwrap().send(()).unwrap();
            return Poll::Pending;
        }

        // We were polled again after the inner open completed and woke the task.
        // Crucially, we do *not* re-poll the inner future here.
        if let Some(tx) = self.second_poll_tx.take() {
            tx.send(()).unwrap();
        }

        Poll::Pending
    }
}

async fn completed_then_dropped_before_repoll(path: PathBuf) {
    let (first_tx, mut first_rx) = unbounded_channel();
    let (second_tx, mut second_rx) = unbounded_channel();

    let handle = tokio::spawn(async move {
        let mut opt = tokio::fs::OpenOptions::new();
        opt.read(true);

        PollOpenOnceThenNeverRepoll {
            inner: Box::pin(opt.open(&path)),
            first_poll_tx: Some(first_tx),
            second_poll_tx: Some(second_tx),
            polled_once: false,
        }
        .await;
    });

    // Wait until the inner open has been polled once and registered with io_uring.
    first_rx.recv().await.unwrap();

    // Wait until Tokio wakes the task because the open completed. At this point
    // the CQE should already be stored as `Lifecycle::Completed(cqe)`.
    timeout(Duration::from_secs(2), second_rx.recv())
        .await
        .expect("timed out waiting for second poll")
        .expect("second poll channel closed unexpectedly");

    // Abort now, before the inner open future gets re-polled and consumes the CQE.
    handle.abort();
    let err = handle.await.unwrap_err();
    assert!(err.is_cancelled(), "task was not cancelled as expected");
}

fn main() {
    let rt = tokio::runtime::Builder::new_multi_thread()
        .worker_threads(1)
        .enable_all()
        .build()
        .unwrap();

    rt.block_on(async {
        let before = fd_count();
        let path = PathBuf::from("/etc/hosts");

        for _ in 0..128 {
            completed_then_dropped_before_repoll(path.clone()).await;
        }

        // Give completions a moment to settle before counting fds.

        let after = fd_count();
        let leaked = after.saturating_sub(before);

        println!("before={before}");
        println!("after={after}");
        println!("leaked={leaked}");

        // Vulnerable builds leak roughly one fd per iteration here.
        assert!(leaked >= 64, "expected many leaked fds, saw {leaked}");
    });
}
```

On this PR branch, this still produces a large positive leak count for me, e.g.:

```text
before=7
after=136
leaked=129
```

A complete fix should keep `after` close to `before` and should not satisfy `assert!(leaked >= 64)`.
</details>

---

So this still looks like a real fd-leak / DoS path.

I think `cancel_op()` also needs to handle `Lifecycle::Completed(cqe)` for `CancelData::Open(_)`, closing/adopting the fd before removing the slab entry. The new tests only cover “drop after first poll”, not “completed then dropped before re-poll”, which is the missing regression case.

tokio/src/runtime/io/driver/uring.rs
```diff
@@ -77,9 +79,16 @@ impl UringContext {
                     waker.wake_by_ref();
                     *ops.get_mut(idx).unwrap() = Lifecycle::Completed(cqe);
                 }
-                Some(Lifecycle::Cancelled(_)) => {
+                Some(Lifecycle::Cancelled(cancel_data)) => {
+                    if let CancelData::Open(_) = cancel_data {
```
Daksh14, line 83: We should not drop the `OwnedFd` here but we should drop it in the drop of the `Op` itself.

If any operation that opens an Fd has to drop it this way then it will become a big match here in the future for `CancelData` of each `Op`. Let's handle it in `cancel_op`

Daksh14:
@mattiapitossi Follow what @mufeedvh said. 

This `cancel_op` function is located here: https://github.com/tokio-rs/tokio/blob/master/tokio/src/runtime/io/driver/uring.rs#L259

This function gets called the drop glue of each `Op` io uring operation here: https://github.com/tokio-rs/tokio/blob/master/tokio/src/runtime/driver/op.rs#L86

So handle this case in the `cancel_op`

tokio/src/runtime/io/driver/uring.rs
Daksh14, line 77: @Darksonn @ADD-SP do you think the problem here might be that we're not calling `mem::replace` here so the Cancel data is just behind a get_mut right now. Inside `poll` of the `Op` `mem::replace` is called

In my batch completions PR https://github.com/tokio-rs/tokio/pull/7761 I actually have this change because it was actually causing weird effects for me too.

tokio/src/runtime/io/driver/uring.rs
mattiapitossi: I added the drop for the FD to handle the cases where we cancel the operation and the op it's already in completed status as you and @mufeedvh suggested. But for the other cases what do you have in mind? Probably I'm missing something here. Because for the completed status we are able to drop it, since we have the data from the CQE, but for the other cases we get the data to close the FDs in the dispatch_completions.

mattiapitossi:
@mufeedvh all the cases should now be covered, I added also a new test based on your PoC for the dropped and never re-polled case

tokio/Cargo.toml
Darksonn, line 1: Can you take this out? I want to do it in a separate PR, and the changelog also needs to be adjusted to include the udp socket error fix.

mattiapitossi: Done, I removed also the bumps in the README and reverted the changelog so we can do it in a separate PR

Darksonn:
I'll go ahead and merge this. @Daksh14 my understanding is that the code is correct in the current state where only `Open` exists, but that it can be improved for when other fd-owning ops are added and that it may be error-prone. Still, to unblock I'd like to merge this now.

PR #7984, opened 2026-03-22
benches: use `is_multiple_of` instead of manual modulo check in `remote_spawn`

ghost:
Fixes a clippy::manual_is_multiple_of warning in the remote_spawn benchmark.

\`TOTAL_TASKS % 64 == 0\` can be written as \`TOTAL_TASKS.is_multiple_of(64)\` which is more readable.


mattiapitossi:
Thanks!

PR #7986, opened 2026-03-24
metrics: add task schedule latency metric

robholt:
This new metric tracks the amount of time between when a task is scheduled and when it is polled, also known as queue delay. This duration is recorded in a histogram, just like the poll time metric.

## Motivation

Measuring the amount of time a unit of work spends waiting in a queue, also known as "queue delay" or "sojourn time", is an essential component of the [CoDel](https://en.wikipedia.org/wiki/CoDel) active queue management algorithm (and any algorithm based on it). CoDel is particularly popular because it requires no configuration or tuning to specific workloads, it maintains low latency and high throughput, and relies on a single parameter (queue delay) so it is relatively simple to implement.

For example, an HTTP server built on tokio could use this metric to power an overload protection layer that uses the CoDel algorithm to reject requests and prevent a latency spiral while still serving close to the maximum possible request volume, without the need for specific workload tuning.

There is some precedent for a task scheduler to expose this metric. The Go runtime [exposes this metric](https://pkg.go.dev/runtime/metrics#hdr-Supported_metrics) as `/sched/latencies:seconds`, and the Linux kernel's `schedstat` tracks process scheduling latency per-cpu and per-process.

## Solution

Measuring this requires keeping track of the time each task enters the queue in order to calculate the elapsed duration when the task is finally polled. I stored this time in the task header because it's hot data that's updated each time the task is scheduled. This does add some overhead, so I put the metric behind `tokio_unstable`, just like the poll time histogram metric.


tokio/src/runtime/builder.rs
```diff
@@ -1556,6 +1564,124 @@ impl Builder {
             self.metrics_poll_count_histogram.legacy_mut(|b|b.num_buckets = buckets);
             self
         }
+
+        /// Enables tracking the distribution of task schedule latencies. Task
+        /// schedule latency is the time between when a task is scheduled for
+        /// execution and when it is polled.
+        ///
+        /// Task schedule latencies are not instrumented by default as doing
+        /// so requires calling [`Instant::now()`] twice per task poll, which
```
```suggestion
        /// so requires calling [`Instant::now()`] when a task is scheduled and when it is polled, which
```

tokio/src/runtime/metrics/runtime.rs
```diff
@@ -571,6 +571,124 @@ impl RuntimeMetrics {
         pub fn blocking_queue_depth(&self) -> usize {
             self.handle.inner.blocking_queue_depth()
         }
+
+        /// Returns `true` if the runtime is tracking the distribution of task
+        /// schedule latencies.
+        ///
+        /// Task schedule latencies times are not instrumented by default as doing
+        /// so requires calling [`Instant::now()`] twice per task poll. The feature
```
martin-g:
```suggestion
        /// so requires calling [`Instant::now()`] when a task is scheduled and when it is polled. The feature
```

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -1271,6 +1274,14 @@ impl Worker {
 
 impl Handle {
     pub(super) fn schedule_task(&self, task: Notified, is_yield: bool) {
+        // SAFETY: There are no concurrent writes because tasks cannot be scheduled in
+        // multiple places concurrently. There are no concurrent reads because this field
+        // is only read when polling the task, which can only happen after it's scheduled.
+        #[cfg(tokio_unstable)]
+        unsafe {
+            task.set_scheduled_at(std::time::Instant::now());
```
martin-g: There is no check whether the new metric is enabled as advertised at https://github.com/tokio-rs/tokio/pull/7986/changes#diff-3f0d70e37008bd2c1c13853b6cff7b868a0b922eb242c618153b97550e4c8618R578-R579

tokio/src/runtime/metrics/batch.rs
```diff
@@ -95,6 +98,10 @@ impl MetricsBatch {
                             poll_started_at: now,
                         })
                 });
+                let schedule_latencies = worker_metrics
```
martin-g: This should also take into account `maybe_now` as `poll_timer` does above, no ?

tokio/src/runtime/scheduler/multi_thread/worker.rs
martin-g, line 722: Should `core.stats.start_poll(task.get_scheduled_at());` be called here too ?

tokio/tests/rt_unstable_metrics.rs
```diff
@@ -800,6 +800,53 @@ fn io_driver_ready_count() {
...
+            .enable_all()
+            .enable_metrics_schedule_latency_histogram()
+            .metrics_schedule_latency_histogram_configuration(HistogramConfiguration::linear(
+                Duration::from_millis(50),
+                3,
+            ))
+            .build()
+            .unwrap(),
+        tokio::runtime::Builder::new_multi_thread()
+            .enable_all()
```
martin-g, line 819:
```suggestion
            .worker_threads(2)
            .enable_all()
```
to guarantee that there are more than one workers even on constrained systems.
Same as https://github.com/robholt/tokio/blob/096b3a4a10095bdb5661c6e9faad4603fa228079/tokio/tests/rt_unstable_metrics.rs#L472C13-L472C31

tokio/src/runtime/metrics/batch.rs
```diff
@@ -206,15 +219,22 @@ impl MetricsBatch {
...
+            pub(crate) fn start_poll(&mut self, task_scheduled_at: Option<Instant>) {
                 self.poll_count += 1;
                 if let Some(poll_timer) = &mut self.poll_timer {
                     poll_timer.poll_started_at = Instant::now();
                 }
+                if let Some(task_scheduled_at) = task_scheduled_at {
+                    if let Some(schedule_latencies) = &mut self.schedule_latencies {
+                        let now = self.poll_timer.as_ref().map(|p| p.poll_started_at).unwrap_or_else(Instant::now);
+                        let elapsed = duration_as_u64(now - task_scheduled_at);
+                        schedule_latencies.measure(elapsed, 1);
```
martin-g: `Instant::now()` would fail on WASM.

```suggestion
                        if let Some(now) = self
                             .poll_timer
                             .as_ref()
                             .map(|p| p.poll_started_at)
                             .or_else(now)
                         {
                             let elapsed = duration_as_u64(now.saturating_duration_since(task_scheduled_at));
                             schedule_latencies.measure(elapsed, 1);
                         }
```

tokio/src/runtime/metrics/runtime.rs
```diff
@@ -571,6 +571,124 @@ impl RuntimeMetrics {
         pub fn blocking_queue_depth(&self) -> usize {
             self.handle.inner.blocking_queue_depth()
         }
+
+        /// Returns `true` if the runtime is tracking the distribution of task
+        /// schedule latencies.
+        ///
+        /// Task schedule latencies times are not instrumented by default as doing
```
martin-g:
```suggestion
        /// Task schedule latencies are not instrumented by default as doing
```

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -1271,6 +1274,14 @@ impl Worker {
...
+        }
```
martin-g: There is no check whether the new metric is enabled as advertised at https://github.com/tokio-rs/tokio/pull/7986/changes#diff-3f0d70e37008bd2c1c13853b6cff7b868a0b922eb242c618153b97550e4c8618R578-R579

```suggestion
        if self.shared.config.metrics_schedule_latency_histogram.is_some() {
            unsafe {
                task.set_scheduled_at(std::time::Instant::now());
            }
        }
```

tokio/src/runtime/builder.rs
```diff
@@ -1556,6 +1564,124 @@ impl Builder {
...
+        ///             .expect("configuration uses 488 buckets")
+        ///         )
+        ///     )
+        ///     .build()
+        ///     .unwrap();
+        /// # }
+        /// ```
+        ///
+        /// [`LogHistogram`]: crate::runtime::LogHistogram
+        pub fn metrics_schedule_latency_histogram_configuration(&mut self, configuration: HistogramConfiguration) -> &mut Self {
```
martin-g: nit: This method has no effect unless `enable_metrics_schedule_latency_histogram()` is also called. Worth emphasise it in the docs above ?

tokio/src/runtime/metrics/worker.rs
```diff
@@ -65,6 +65,9 @@ pub(crate) struct WorkerMetrics {
     #[cfg(tokio_unstable)]
     /// If `Some`, tracks the number of polls by duration range.
     pub(super) poll_count_histogram: Option<Histogram>,
+
```
martin-g, line 68: Please add some docs

tokio/src/runtime/task/core.rs
```diff
@@ -195,7 +195,10 @@ pub(crate) struct Header {
     pub(super) tracing_id: Option<tracing::Id>,
 
     /// The last time this task was scheduled. Used to measure schedule latency.
-    #[cfg(tokio_unstable)]
+    ///
+    /// Only enabled when the target supports 64-bit atomics because the metric
+    /// that uses this field also requires 64-bit atomics.
+    #[cfg(all(tokio_unstable, target_has_atomic = "64"))]
     pub(super) scheduled_at: UnsafeCell<Option<Instant>>,
```
Darksonn: How big is the task header?

tokio/src/runtime/scheduler/current_thread/mod.rs
```diff
@@ -675,12 +675,19 @@ impl Schedule for Arc<Handle> {
...
+            .config
+            .metrics_schedule_latency_histogram
+            .is_some()
+        {
+            // SAFETY: There are no concurrent writes because tasks cannot be scheduled in
+            // multiple places concurrently. There are no concurrent reads because this field
+            // is only read when polling the task, which can only happen after it's scheduled.
+            unsafe {
+                task.set_scheduled_at(std::time::Instant::now());
+            }
```
Darksonn: This can be a safe method on `Notified` instead. There's only one `Notified` reference per task, which makes it safe.

tokio/src/runtime/scheduler/multi_thread/worker.rs
robholt: I considered this, but it would also change the behavior of the poll time metric. I do think that would be more correct, but wanted to avoid changing too many things in one PR. I'm happy to make the change though if you think that's fine!

tokio/src/runtime/builder.rs
robholt: Fixed, thanks!

tokio/src/runtime/metrics/runtime.rs
robholt: Fixed, thanks!

tokio/src/runtime/metrics/batch.rs
robholt: It doesn't use the value of `maybe_now`, but I agree it does make sense to skip creating this on targets where `now` doesn't work. Fixed.

tokio/tests/rt_unstable_metrics.rs
robholt: Fixed, thanks!

tokio/src/runtime/metrics/runtime.rs
robholt: Fixed, thanks!

tokio/src/runtime/metrics/worker.rs
robholt: Whoops I missed this one, thanks!

tokio/src/runtime/metrics/batch.rs
robholt: I applied this change too. With the `maybe_now` change above it's somewhat redundant (because `schedule_latencies` will be `None` if `now()` is also `None`, but still makes the behavior more clear.

tokio/src/runtime/scheduler/multi_thread/worker.rs
robholt: Good catch, right there's no point in calling `Instant::now()` if the metric is not enabled. Fixed.

tokio/src/runtime/builder.rs
robholt: Good point, fixed!

tokio/src/runtime/task/core.rs
robholt: This field adds 16 bytes to the header, bringing it up to 48 bytes in total (56 if tracing is enabled too).

This is causing the `header_lte_cache_line` test (at the bottom of this file) to fail on 32-bit targets. Restricting the field to targets that support 64-bit atomics did not work because some 32-bit targets do support 64-bit atomics.

Maybe sticking this behind a feature (like tracing) would be best?

tokio/src/runtime/scheduler/current_thread/mod.rs
robholt: Makes sense, thanks! Updated

tokio/src/runtime/task/core.rs
Darksonn: I'm okay with changing that test to be `<= 64` to accomodate 32-bit targets, they are not that important anyway.

Regardless, the `Instant` struct is pretty large and space is at a premium in the task header. I don't think we want to dedicate 16 bytes just so we can have nanosecond precision even if it takes a billion years for it to get scheduled. Could we store a `u32` of milliseconds since the runtime was created, or something like that? I'm pretty sure we already store a sort of "epoch" instant in the `tokio::runtime::time` part of the driver.

robholt: That's a good idea, I updated the PR to use this "epoch" approach. I used a `u64` though because the struct is already 8-byte aligned on all targets because of the 64-bit `owner_id` field, so a `u32` just wastes 4 bytes. Using a `u64` also means we can keep the nanosecond precision, since `u64::MAX` in nanoseconds is equal to ~584 years.

That brings the size of the struct up to 40 bytes now.

And you're right, the time driver does track the startup time of the time driver. I opted not to use it however because the time driver is not enabled by default, so using it would mean the time driver must be enabled by users in order to use this schedule latency tracking.

tokio/src/runtime/task/mod.rs
```diff
@@ -247,6 +249,16 @@ impl<S> Notified<S> {
     pub(crate) fn task_meta<'meta>(&self) -> crate::runtime::TaskMeta<'meta> {
         self.0.task_meta()
     }
+
+    #[cfg(all(tokio_unstable, target_pointer_width = "64"))]
+    pub(crate) fn set_scheduled_at(&self, nanos: NonZeroU64) {
```
martin-g: `Notified` has only a setter for `scheduled_at`.
`LocalNotified` below has only a getter.
Why ?

tokio/src/runtime/builder.rs
```diff
@@ -1556,6 +1564,131 @@ impl Builder {
             self.metrics_poll_count_histogram.legacy_mut(|b|b.num_buckets = buckets);
             self
         }
+
+        /// Enables tracking the distribution of task schedule latencies. Task
+        /// schedule latency is the time between when a task is scheduled for
+        /// execution and when it is polled.
+        ///
+        /// **This feature is only supported on 64-bit targets.**
```
martin-g: Should this method be gated behind `#[cfg(target_pointer_width = "64")]` ?

tokio/src/runtime/builder.rs
```diff
@@ -133,6 +133,10 @@ pub struct Builder {
     /// Configures the task poll count histogram
     pub(super) metrics_poll_count_histogram: HistogramBuilder,
 
+    pub(super) metrics_schedule_latency_histogram_enable: bool,
```
```suggestion
    pub(super) metrics_schedule_latency_histogram_enabled: bool,
```

tokio/src/runtime/metrics/runtime.rs
```diff
@@ -571,6 +571,124 @@ impl RuntimeMetrics {
...
+        ///             let enabled = metrics.schedule_latency_histogram_enabled();
+        ///
+        ///             println!("Tracking task schedule latency distribution: {:?}", enabled);
+        ///         });
+        /// }
+        /// ```
+        ///
+        /// [`enable_metrics_schedule_latency_histogram()`]: crate::runtime::Builder::enable_metrics_schedule_latency_histogram
+        /// [`Instant::now()`]: std::time::Instant::now
+        pub fn schedule_latency_histogram_enabled(&self) -> bool {
```
martin-g: Should this method be gated behind the 64 bit check too ?

tokio/src/runtime/metrics/runtime.rs
```diff
@@ -571,6 +571,124 @@ impl RuntimeMetrics {
...
+        ///             for i in 0..buckets {
+        ///                 let range = metrics.schedule_latency_histogram_bucket_range(i);
+        ///                 println!("Histogram bucket {} range: {:?}", i, range);
+        ///             }
+        ///         });
+        /// }
+        /// ```
+        ///
+        /// [`metrics_schedule_latency_histogram_configuration()`]: crate::runtime::Builder::metrics_schedule_latency_histogram_configuration
+        pub fn schedule_latency_histogram_bucket_range(&self, bucket: usize) -> Range<Duration> {
```
martin-g:
```suggestion
        #[track_caller]
        pub fn schedule_latency_histogram_bucket_range(&self, bucket: usize) -> Range<Duration> {
```
as `schedule_latency_histogram_bucket_count()`

tokio/src/runtime/task/mod.rs
robholt: The `Notified` type is what's used when scheduling a task and `LocalNotified` is used when polling a task. If I add a getter to `Notified` it would be unused, so I'd also have to add a `#[allow(dead_code)]` annotation.

tokio/src/runtime/builder.rs
robholt: It could be, the only reason I didn't is for consistency with the other methods. For example, the existing poll time metrics only work on targets that support 64-bit atomics but the methods to configure it aren't gated behind `target_has_atomic = "64"`. I can gate this one though if that existing behavior isn't desirable.

tokio/src/runtime/metrics/runtime.rs
robholt: Similar to the other method, I didn't gate this one to maintain consistency with the existing poll time metrics.  `poll_time_histogram_enabled` is not gated behind `target_has_atomic = "64"` while `poll_time_histogram_bucket_count` is gated. I can change this one too though if the existing behavior isn't desirable.

tokio/src/runtime/task/mod.rs
```diff
@@ -247,6 +249,16 @@ impl<S> Notified<S> {
     pub(crate) fn task_meta<'meta>(&self) -> crate::runtime::TaskMeta<'meta> {
         self.0.task_meta()
```
rcoh, line 249: can we please expose this field in the task meta? That would really help for observability tools because it means that during `on_poll_start` we can unconditionally know what the scheduling delay was

tokio/src/runtime/scheduler/current_thread/mod.rs
```diff
@@ -97,6 +100,12 @@ struct Shared {
 
     /// This scheduler only has one worker.
     worker_metrics: WorkerMetrics,
+
+    /// Startup time of this scheduler.
+    ///
+    /// This instant is used as the basis of task `scheduled_at` measurements.
+    #[cfg(all(tokio_unstable, target_pointer_width = "64"))]
+    started_at: Instant,
```
Darksonn: This PR is using these cfgs all over the place. Can we please reduce them a bit? You can define a module that contains a `ScheduleLatencyInstant` which is either empty or contains an `Instant` depending on the configuration, and then code actually using it can stop having cfgs. Similarly for the u64 in the task header.

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -1271,6 +1287,22 @@ impl Worker {
...
+            .config
+            .metrics_schedule_latency_histogram
+            .is_some()
+        {
+            // SAFETY: `.max(1)` ensures the value can never be 0.
+            let scheduled_at = unsafe {
+                std::num::NonZeroU64::new_unchecked(
+                    self.shared.started_at.elapsed().as_nanos().max(1) as u64,
+                )
+            };
```
Darksonn: This use of unsafe is not necessary.
```suggestion
            let scheduled_at = NonZeroU64::new(self.shared.started_at.elapsed().as_nanos())
                .unwrap_or(NonZeroU64::MIN);
```

tokio/src/runtime/task/mod.rs
robholt: I think it would be best to do that in a follow-up PR. I agree it would be very useful, but it'll require some refactoring of the way `TaskMeta` is constructed.

tokio/src/runtime/scheduler/current_thread/mod.rs
robholt: Makes sense, the cfgs did get a bit out of control after the 64-bit restriction.

I've updated the PR to remove almost all of the cfgs and centralized the logic into a new `schedule_latency` module. The size of the task Header remains the same as the previous version on both 64-bit and 32-bit targets.

tokio/src/runtime/task/mod.rs
```diff
@@ -663,3 +677,99 @@ impl SpawnLocation {
         Self::from(Location::caller())
     }
 }
+
+// Task schedule latency is only tracked on 64-bit targets to avoid increasing the size
+// of the task Header beyond the bounds of one CPU cache line on 32-bit targets.
+#[cfg(all(tokio_unstable, target_pointer_width = "64"))]
+pub(crate) mod schedule_latency {
```
Darksonn: Please move these to a new file (or two) in the metrics directory.

tokio/src/runtime/task/mod.rs
```diff
@@ -663,3 +677,99 @@ impl SpawnLocation {
         Self::from(Location::caller())
     }
 }
+
+// Task schedule latency is only tracked on 64-bit targets to avoid increasing the size
+// of the task Header beyond the bounds of one CPU cache line on 32-bit targets.
+#[cfg(all(tokio_unstable, target_pointer_width = "64"))]
```
Darksonn: Please add a new feature flag to gate this behind. Otherwise anyone who turns on `--cfg tokio_unstable` is affected by the perf impact of this change.

tokio/src/runtime/metrics/batch.rs
```diff
@@ -53,6 +54,9 @@ pub(crate) struct MetricsBatch {
     #[cfg(tokio_unstable)]
     /// If `Some`, tracks poll times in nanoseconds
     poll_timer: Option<PollTimer>,
+
+    #[cfg(tokio_unstable)]
+    schedule_latencies: Option<HistogramBatch>,
```
Darksonn: I think I would prefer the following approach:

1. Everywhere in the code it's just protected by `#[cfg(feature = "schedule-latency")]` only.
2. In `lib.rs` you emit a `compile_error!` if you enable `schedule-latency` without `--cfg tokio_unstable`.

Right now the cfgs are pretty inconsistent. Sometimes you only require `tokio_unstable`. Sometimes you require `target_pointer_width = "64"`. Sometimes you require `target_has_atomic = "64"`. Please make these cfgs consistent.

robholt: With the original cfgs I was trying to align each part with the surrounding metrics code, but I do see your point that it was messy.

I've updated the code to only use `#[cfg(feature = "schedule-latency")]`, and had to add a few additional guards to `MetricsBatch` to account for the state where unstable metrics are enabled but `schedule-latency` isn't.

robholt:
@Darksonn I think I've addressed your comments, could you take another look?

Darksonn (approved):
Thanks.

PR #7991, opened 2026-03-27
sync: fix panic in `Chan::recv_many` when called with non-empty vector on closed channel

### LINKED ISSUE #7990: `mpsc::Chan::recv_many` panics due to faulty assertion when called with a non-empty buffer after receiver closed
**Version**
1.50.0 / `master` at `a8435c0`

**Platform**
`Linux antigravity 6.19.7-1-default #1 SMP PREEMPT_DYNAMIC Thu Mar 12 11:27:25 UTC 2026 (7f7ff04) x86_64 x86_64 x86_64 GNU/Linux`

**Description**

As [described in `#tokio-internals`](https://discord.com/channels/500028886025895936/1479686902339735583), `tokio::mpsc::Chan::recv_many` makes incorrect assumptions about the provided buffer when exiting through the post-loop `rx_fields.rx_closed && self.inner.semaphore.is_idle()` code path.

`assert!(buffer.is_empty());` asserts the incorrect precondition, it should `assert_eq!(buffer.len(), initial_length);` instead.

The following tests trigger the faulty assertion for bounded and unbounded channels:

```rust
#[should_panic]
#[maybe_tokio_test]
async fn recv_many_with_non_empty_buffer_bounded_rx_closed_and_idle() {
    let (_tx, mut rx) = mpsc::channel::<i32>(1);

    let mut buffer: Vec<i32> = vec![1];

    rx.close();

    // Panics, incorrectly expecting an empty buffer.
    rx.recv_many(&mut buffer, 1).await;
}

#[should_panic]
#[maybe_tokio_test]
async fn recv_many_with_non_empty_buffer_unbounded_rx_closed_and_idle() {
    let (_tx, mut rx) = mpsc::unbounded_channel::<i32>();

    let mut buffer: Vec<i32> = vec![1];

    rx.close();

    // Panics, incorrectly expecting an empty buffer.
    rx.recv_many(&mut buffer, 1).await;
}
```

I will provide a PR.

LeoniePhiline:
## Motivation

`Chan::recv_many` intends to assert that no slots
have been consumed when exiting with `Ready` via
the `rx_closed` code path.

Instead of asserting no items were added to the
buffer, it asserted buffer emptiness, incorrectly
making assumptions about the provided buffer.

When `recv_many` was called on an empty channel
with idle semaphore after the receiver was closed,
the method would panic.

## Solution

The branch coverage had been previously missing.

This changeset corrects the assertion
and adds tests covering the code path.

Fixes #7990.

### DIFF
--- a/tokio/src/sync/mpsc/chan.rs
+++ b/tokio/src/sync/mpsc/chan.rs
@@ -306,13 +306,11 @@ impl<T, S: Semaphore> Rx<T, S> {
                             return Ready(Some(value));
                         }
                         Some(Read::Closed) => {
-                            // TODO: This check may not be required as it most
-                            // likely can only return `true` at this point. A
-                            // channel is closed when all tx handles are
+                            // A channel is closed when all tx handles are
                             // dropped. Dropping a tx handle releases memory,
                             // which ensures that if dropping the tx handle is
                             // visible, then all messages sent are also visible.
-                            assert!(self.inner.semaphore.is_idle());
+                            debug_assert!(self.inner.semaphore.is_idle());
                             coop.made_progress();
                             return Ready(None);
                         }
@@ -380,13 +378,11 @@ impl<T, S: Semaphore> Rx<T, S> {
                                 if number_added > 0 {
                                     self.inner.semaphore.add_permits(number_added);
                                 }
-                                // TODO: This check may not be required as it most
-                                // likely can only return `true` at this point. A
-                                // channel is closed when all tx handles are
+                                // A channel is closed when all tx handles are
                                 // dropped. Dropping a tx handle releases memory,
                                 // which ensures that if dropping the tx handle is
                                 // visible, then all messages sent are also visible.
-                                assert!(self.inner.semaphore.is_idle());
+                                debug_assert!(self.inner.semaphore.is_idle());
                                 coop.made_progress();
                                 return Ready(number_added);
                             }
@@ -415,7 +411,7 @@ impl<T, S: Semaphore> Rx<T, S> {
             try_recv!();
 
             if rx_fields.rx_closed && self.inner.semaphore.is_idle() {
-                assert!(buffer.is_empty());
+                debug_assert_eq!(buffer.len(), initial_length);
                 coop.made_progress();
                 Ready(0usize)
             } else {
--- a/tokio/tests/sync_mpsc.rs
+++ b/tokio/tests/sync_mpsc.rs
@@ -358,6 +358,30 @@ async fn send_recv_many_unbounded_capacity() {
     assert_eq!(expected, buffer);
 }
 
+#[maybe_tokio_test]
+async fn recv_many_with_non_empty_buffer_bounded_rx_closed_and_idle() {
+    let (_tx, mut rx) = mpsc::channel::<i32>(1);
+
+    let mut buffer: Vec<i32> = vec![1];
+
+    rx.close();
+
+    assert_eq!(0, rx.recv_many(&mut buffer, 1).await);
+    assert_eq!(vec![1], buffer);
+}
+
+#[maybe_tokio_test]
+async fn recv_many_with_non_empty_buffer_unbounded_rx_closed_and_idle() {
+    let (_tx, mut rx) = mpsc::unbounded_channel::<i32>();
+
+    let mut buffer: Vec<i32> = vec![1];
+
+    rx.close();
+
+    assert_eq!(0, rx.recv_many(&mut buffer, 1).await);
+    assert_eq!(vec![1], buffer);
+}
+
 #[tokio::test]
 #[cfg(feature = "full")]
 async fn async_send_recv_unbounded() {


tokio/tests/sync_mpsc.rs
```diff
@@ -358,6 +358,30 @@ async fn send_recv_many_unbounded_capacity() {
...
+#[maybe_tokio_test]
+async fn recv_many_with_non_empty_buffer_bounded_rx_closed_and_idle() {
+    let (_tx, mut rx) = mpsc::channel::<i32>(1);
+
+    let mut buffer: Vec<i32> = vec![1];
+
+    rx.close();
+
+    // Does not panic, no longer expecting an empty buffer (#7990).
+    rx.recv_many(&mut buffer, 1).await;
```
Darksonn: I think that if someone sees this comment in one year, they will be confused about what you mean by "no longer". They don't know the context of this bug. Instead of this comment, can you add a comment above this test saying that it is a regression test for `<link to issue>`?

tokio/tests/sync_mpsc.rs
```diff
@@ -358,6 +358,30 @@ async fn send_recv_many_unbounded_capacity() {
...
+#[maybe_tokio_test]
+async fn recv_many_with_non_empty_buffer_unbounded_rx_closed_and_idle() {
+    let (_tx, mut rx) = mpsc::unbounded_channel::<i32>();
+
+    let mut buffer: Vec<i32> = vec![1];
+
+    rx.close();
+
+    // Does not panic, no longer expecting an empty buffer (#7990).
+    rx.recv_many(&mut buffer, 1).await;
```
Darksonn: Same comment here.

tokio/tests/sync_mpsc.rs
martin-g:
```suggestion
    let n = rx.recv_many(&mut buffer, 1).await;
    assert_eq!(n, 0);
    assert_eq!(buffer, &[1]);
```

tokio/tests/sync_mpsc.rs
martin-g:
```suggestion
    let n = rx.recv_many(&mut buffer, 1).await;
    assert_eq!(n, 0);
    assert_eq!(buffer, &[1]);
```

tokio/src/sync/mpsc/chan.rs
```diff
@@ -415,7 +415,7 @@ impl<T, S: Semaphore> Rx<T, S> {
             try_recv!();
 
             if rx_fields.rx_closed && self.inner.semaphore.is_idle() {
-                assert!(buffer.is_empty());
+                assert!(buffer.len() == initial_length);
```
martin-g:
```suggestion
                assert_eq!(buffer.len(), initial_length);
```

LeoniePhiline: I chose `assert!` because [it is a compiler intrinsic](https://doc.rust-lang.org/stable/src/core/macros/mod.rs.html#1689) while `assert_eq!` is [much more involved](https://doc.rust-lang.org/stable/src/core/macros/mod.rs.html#42).

tokio/tests/sync_mpsc.rs
LeoniePhiline: Intended as a minimal and focused regression test for a panic, should it really be extended to a full API test?

The change does make sense to me, but it makes the test lose focus. On the other hand, no other test covers this code branch, and the full expected behavior could just as well be tested in it.

In that case, though, documenting it as regression test would no longer be useful.

tokio/src/sync/mpsc/chan.rs
Darksonn: Using `assert_eq!` is better because it prints the values when they are not equal. But if you have perf concerns, we can make it into a `debug_assert_eq!`.

LeoniePhiline: Thanks. I would leave the decision up to you.

Performance is not a huge concern, as this is not a hot path, but of course the panic unwinding machinery will be part of the generated assembly.

Is seems useful to keep consistent with https://github.com/LeoniePhiline/tokio/blob/2222621764ca3b6fd70803874f76ccdc59885ca7/tokio/src/sync/mpsc/chan.rs#L389 and https://github.com/LeoniePhiline/tokio/blob/2222621764ca3b6fd70803874f76ccdc59885ca7/tokio/src/sync/mpsc/chan.rs#L315 at the other exit of the `pop` loop – should these be changed to `debug_assert`?

I myself would prefer debug assertions, as these are checking for bugs, not UB. And, to my mind, they are meant to verify the implementation in tests, not production code.

tokio/tests/sync_mpsc.rs
LeoniePhiline: > On the other hand, no other test covers this code branch, and the full expected behavior could just as well be tested in it.

I have adjusted the two tests, extending them to assert the expected behavior.

tokio/tests/sync_mpsc.rs
```diff
@@ -358,6 +358,30 @@ async fn send_recv_many_unbounded_capacity() {
...
+#[maybe_tokio_test]
+async fn recv_many_with_non_empty_buffer_bounded_rx_closed_and_idle() {
+    let (_tx, mut rx) = mpsc::channel::<i32>(1);
+
+    let mut buffer: Vec<i32> = vec![1];
+
+    rx.close();
+
+    assert_eq!(0, rx.recv_many(&mut buffer, 1).await);
+    assert_eq!(vec![1], buffer);
```
LeoniePhiline, line 370: I chose this assertion style over the [suggested diff](https://github.com/tokio-rs/tokio/pull/7991#discussion_r3006728233) for consistency with the other tests in the same file.

tokio/src/sync/mpsc/chan.rs
```diff
@@ -380,13 +378,11 @@ impl<T, S: Semaphore> Rx<T, S> {
...
                                 }
-                                // TODO: This check may not be required as it most
-                                // likely can only return `true` at this point. A
-                                // channel is closed when all tx handles are
+                                // A channel is closed when all tx handles are
                                 // dropped. Dropping a tx handle releases memory,
                                 // which ensures that if dropping the tx handle is
                                 // visible, then all messages sent are also visible.
-                                assert!(self.inner.semaphore.is_idle());
+                                debug_assert!(self.inner.semaphore.is_idle());
```
LeoniePhiline, line 385: Following up with https://github.com/tokio-rs/tokio/pull/7991#discussion_r3008939875, and after my review of the internals, I am proposing to demote these bug checks (not UB guards) to `cfg(debug)`.

This aligns these `pop` loop exit assertions with the `rx_closed` exit assertion.

tokio/tests/sync_mpsc.rs
LeoniePhiline: > If you add the test first, then I will need to merge with squash. Otherwise bisection will not work on the main branch if someone hits the first commit in a bisection.

Both commits are squashed and the commit message and PR description updated.

> I think that if someone sees this comment in one year, they will be confused about what you mean by "no longer". They don't know the context of this bug. Instead of this comment, can you add a comment above this test saying that it is a regression test for `<link to issue>`?

Following up on https://github.com/tokio-rs/tokio/pull/7991#discussion_r3006925548, I have extended the scope of the tests from panic-regression to behavior test. Therefore, the suggested comment seems no longer useful. The original comment is [removed](https://github.com/tokio-rs/tokio/compare/2222621764ca3b6fd70803874f76ccdc59885ca7..7dd98a368d9895c1add5cba91ac7b088088bb66b#diff-7ab060f084a0c7be80a9ec2b365a35ee383ed88892874fdb9024f9413340696cL369).

tokio/src/sync/mpsc/chan.rs
Darksonn: IMO such changes should happen in separate PR.

Darksonn:
I'm going to backport this to Tokio 1.47, so I'm changing the base of this PR now.

mattiapitossi (approved):

LeoniePhiline:
Thank you!

PR #7992, opened 2026-03-27
io: AioSource now employs IO Safety

asomers:
## Motivation

Use I/O Safety with POSIX AIO, and allow downstream consumers to use a version of the Nix crate past 0.29.0.

## Solution

Change `AioSource::register` to take a `BorrowedFd`.

### DIFF
--- a/tokio/src/io/bsd/poll_aio.rs
+++ b/tokio/src/io/bsd/poll_aio.rs
@@ -23,9 +23,9 @@ pub trait AioSource {
     ///
     /// # Safety
     ///
-    /// It's memory-safe safe, but not I/O safe.  If the file referenced by `kq` gets dropped, then
-    /// this source may end up notifying the wrong file.
-    #[deprecated(since = "1.51.0", note = "use register_borrowed instead")]
+    /// It's memory-safe, but not I/O safe.  If the file referenced by `kq` gets dropped, then this
+    /// source may end up notifying the wrong file.
+    #[deprecated(since = "1.52.0", note = "use register_borrowed instead")]
     fn register(&mut self, _kq: RawFd, _token: usize) {
         // This default implementation exists so new AioSource implementors that implement the
         // register_borrowed method can compile without the need to implement register.
@@ -35,7 +35,7 @@ pub trait AioSource {
     /// Registers this AIO event source with Tokio's reactor.
     fn register_borrowed(&mut self, kq: BorrowedFd<'_>, token: usize) {
         // This default implementation serves to provide backwards compatibility with AioSource
-        // implementors written before 1.51.0 that only implemented the unsafe `register` method.
+        // implementors written before 1.52.0 that only implemented the unsafe `register` method.
         #[allow(deprecated)]
         self.register(kq.as_raw_fd(), token)
     }


Darksonn:
This is a stable API. We cannot make breaking changes to it.

asomers:
> This is a stable API. We cannot make breaking changes to it.

If I can use crater to prove that ~~mio-aio~~ tokio-file (editted) is the only open source consumer, would that make it ok?

Darksonn:
Potentially, but I'm not a big fan of it.

asomers:
> Potentially, but I'm not a big fan of it.

Or I could leaf `AioSource::register`'s signature unchanged, but add a `register_borrowed` method with a default implementation, and deprecate `register`.  Would that be acceptable?

tokio/tests/io_async_fd.rs
```diff
@@ -99,7 +100,7 @@ impl Write for FileDescriptor {
     }
 }
 
-fn set_nonblocking(fd: RawFd) {
+fn set_nonblocking(fd: &OwnedFd) {
     use nix::fcntl::{OFlag, F_GETFL, F_SETFL};
 
     let flags = nix::fcntl::fcntl(fd, F_GETFL).expect("fcntl(F_GETFD)");
```
martin-g: Not caused by this PR but since the file is modified:

```suggestion
    let flags = nix::fcntl::fcntl(fd, F_GETFL).expect("fcntl(F_GETFL)");
```

tokio/tests/io_async_fd.rs
martin-g: Not caused by this PR but since the file is modified:

```suggestion
    nix::fcntl::fcntl(fd, F_SETFL(flags)).expect("fcntl(F_SETFL)");
```

Darksonn:
Adding a new method like you suggested is definitely okay.

tokio/src/io/bsd/poll_aio.rs
```diff
@@ -9,17 +9,36 @@ use mio::Token;
...
+/// [`AioSource::register_borrowed`].
 pub trait AioSource {
     /// Registers this AIO event source with Tokio's reactor.
-    fn register(&mut self, kq: RawFd, token: usize);
+    ///
+    /// # Safety
+    ///
+    /// It's memory-safe safe, but not I/O safe.  If the file referenced by `kq` gets dropped, then
+    /// this source may end up notifying the wrong file.
+    #[deprecated(since = "1.51.0", note = "use register_borrowed instead")]
```
martin-g:
```suggestion
    #[deprecated(since = "1.52.0", note = "use register_borrowed instead")]
```
1.51.0 has been released yesterday - https://github.com/tokio-rs/tokio/releases/tag/tokio-1.51.0

tokio/src/io/bsd/poll_aio.rs
```diff
@@ -9,17 +9,36 @@ use mio::Token;
...
-/// [`Aio`] object.
+/// [`Aio`] object.  Implementors must implement at least one of [`AioSource::register`] and
+/// [`AioSource::register_borrowed`].
 pub trait AioSource {
     /// Registers this AIO event source with Tokio's reactor.
-    fn register(&mut self, kq: RawFd, token: usize);
+    ///
+    /// # Safety
+    ///
+    /// It's memory-safe safe, but not I/O safe.  If the file referenced by `kq` gets dropped, then
```
martin-g:
```suggestion
    /// It's memory safe, but not I/O safe.  If the file referenced by `kq` gets dropped, then
```
nit: I am not a native speaker, so you can just ignore me here!

tokio/src/io/bsd/poll_aio.rs
```diff
@@ -9,17 +9,36 @@ use mio::Token;
...
+    fn register(&mut self, _kq: RawFd, _token: usize) {
+        // This default implementation exists so new AioSource implementors that implement the
+        // register_borrowed method can compile without the need to implement register.
+        unimplemented!("Use AioSource::register_borrowed instead")
+    }
+
+    /// Registers this AIO event source with Tokio's reactor.
+    fn register_borrowed(&mut self, kq: BorrowedFd<'_>, token: usize) {
+        // This default implementation serves to provide backwards compatibility with AioSource
+        // implementors written before 1.51.0 that only implemented the unsafe `register` method.
```
martin-g:
```suggestion
        // implementors written before 1.52.0 that only implemented the unsafe `register` method.
```

asomers:
I fixed those strings, @martin-g

martin-g (approved):

PR #7997, opened 2026-03-31
macros: improve overall macro hygiene

dybucc:
## Motivation

Multiple macro expansions were previously assuming the existence of some
standard library symbols to both be in-scope and referring to the right symbol,
and not only having the same identifier.

## Solution

This has now been refactored into using reexports from the `support` module
under the `macros` module. Some other macro invocations were also using absolute
standard library paths instead of calling into the afore mentioned module
through the special `$crate` metavariable. This has been changed, thus also
reexporting some other items in `support`.

### DIFF
--- a/tokio/src/macros/try_join.rs
+++ b/tokio/src/macros/try_join.rs
@@ -229,7 +229,7 @@ doc! {macro_rules! try_join {
                     if $crate::macros::support::Future::poll(fut.as_mut(), cx).is_pending() {
                         is_pending = true;
                     } else if fut.as_mut().output_mut().expect("expected completed future").is_err() {
-                        return $crate::macros::support::Poll::Ready(Err(fut.take_output().expect("expected completed future").err().unwrap()))
+                        return $crate::macros::support::Poll::Ready($crate::macros::support::Result::Err(fut.take_output().expect("expected completed future").err().unwrap()))
                     }
                 } else {
                     // Future skipped, one less future to skip in the next iteration
@@ -241,7 +241,7 @@ doc! {macro_rules! try_join {
             if is_pending {
                 $crate::macros::support::Poll::Pending
             } else {
-                $crate::macros::support::Poll::Ready(Ok(($({
+                $crate::macros::support::Poll::Ready($crate::macros::support::Result::Ok(($({
                     // Extract the future for this branch from the tuple.
                     let ( $($skip,)* fut, .. ) = &mut futures;
 
--- a/tokio/src/macros/select.rs
+++ b/tokio/src/macros/select.rs
@@ -659,7 +659,7 @@ doc! {macro_rules! select {
             $crate::macros::support::poll_fn(|cx| {
                 // Return `Pending` when the task budget is depleted since budget-aware futures
                 // are going to yield anyway and other futures will not cooperate.
-                ::std::task::ready!($crate::macros::support::poll_budget_available(cx));
+                $crate::macros::support::ready!($crate::macros::support::poll_budget_available(cx));
 
                 // Track if any branch returns pending. If no branch completes
                 // **or** returns pending, this implies that all branches are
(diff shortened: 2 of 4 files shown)


tokio/src/macros/try_join.rs
```diff
@@ -241,7 +241,7 @@ doc! {macro_rules! try_join {
             if is_pending {
                 $crate::macros::support::Poll::Pending
             } else {
-                $crate::macros::support::Poll::Ready(Ok(($({
+                $crate::macros::support::Poll::Ready(::std::result::Result::Ok(($({
```
Darksonn: Can you re-export it from `$crate::macros::support` instead? That's more reliable than using `::std`.

dybucc: Sorry, done.

tokio/src/macros/try_join.rs
Darksonn: Let's fix this one too, while we're at it.

martin-g: There are more in the other modules:
- https://github.com/dybucc/tokio/blob/2c22fc1973621893f07c43ebb5f49b8e03fabe1e/tokio/src/macros/addr_of.rs#L8
- https://github.com/dybucc/tokio/blob/2c22fc1973621893f07c43ebb5f49b8e03fabe1e/tokio/src/macros/select.rs#L702
- https://github.com/dybucc/tokio/blob/2c22fc1973621893f07c43ebb5f49b8e03fabe1e/tokio/src/macros/support.rs#L16
- https://github.com/dybucc/tokio/blob/2c22fc1973621893f07c43ebb5f49b8e03fabe1e/tokio/src/macros/support.rs#L24
- https://github.com/dybucc/tokio/blob/2c22fc1973621893f07c43ebb5f49b8e03fabe1e/tokio/src/macros/support.rs#L25
- https://github.com/dybucc/tokio/blob/2c22fc1973621893f07c43ebb5f49b8e03fabe1e/tokio/src/macros/trace.rs#L15-L21


I think the one in select.rs is intentionally left because it breaks some axum related crate.

martin-g: The improvements in select.rs for 1.50.0 broke the build of Apache DataFusion, but it was very easy to fix - https://github.com/apache/datafusion/pull/21000
I am in favour to improve the last entry in select.rs too!

dybucc: > There are more in the other modules:
> 
>     * https://github.com/dybucc/tokio/blob/2c22fc1973621893f07c43ebb5f49b8e03fabe1e/tokio/src/macros/addr_of.rs#L8
> 
>     * https://github.com/dybucc/tokio/blob/2c22fc1973621893f07c43ebb5f49b8e03fabe1e/tokio/src/macros/select.rs#L702
> 
>     * https://github.com/dybucc/tokio/blob/2c22fc1973621893f07c43ebb5f49b8e03fabe1e/tokio/src/macros/support.rs#L16
> 
>     * https://github.com/dybucc/tokio/blob/2c22fc1973621893f07c43ebb5f49b8e03fabe1e/tokio/src/macros/support.rs#L24
> 
>     * https://github.com/dybucc/tokio/blob/2c22fc1973621893f07c43ebb5f49b8e03fabe1e/tokio/src/macros/support.rs#L25
> 
>     * https://github.com/dybucc/tokio/blob/2c22fc1973621893f07c43ebb5f49b8e03fabe1e/tokio/src/macros/trace.rs#L15-L21
> 
> 
> I think the one in select.rs is intentionally left because it breaks some axum related crate.

@martin-g Done. Still, is the list item concerning `addr_of` really meant to use absolute paths? That one is part of a macro matcher and not a macro invocation, so it wouldn't make a lot of sense to force callers to use absolute paths, right?

I'm also not so sure about the `support.rs` ones, so I've not made changes there either, but I can't make a case against not using absolute paths/`$crate`-relative paths in this instance.

martin-g: @dybucc You are correct! 👍🏻

Issue #8000, opened 2026-04-01
`UdpSocket::recv` does not return socket errors


SRv6d:
**Version**
```
│   │   │   ├── tokio v1.50.0
│   │   │   │   └── tokio-macros v2.6.1 (proc-macro)
│   │   │   ├── tokio-util v0.7.18
│   │   │   │   └── tokio v1.50.0 (*)
│   │   ├── tokio v1.50.0 (*)
│   │   ├── tokio v1.50.0 (*)
│   ├── tokio v1.50.0 (*)
│   │   ├── tokio v1.50.0 (*)
│   │   ├── tokio-util v0.7.18 (*)
├── tokio v1.50.0 (*)
├── tokio-stream v0.1.18
│   └── tokio v1.50.0 (*)
├── tokio-util v0.7.18 (*)
│   │   ├── tokio v1.50.0 (*)
│   ├── tokio v1.50.0 (*)
│   ├── tokio-stream v0.1.18 (*)
│   ├── tokio v1.50.0 (*)
├── tokio v1.50.0 (*)
```

**Platform**
```
Linux 912834ca0917 6.17.8-orbstack-00308-g8f9c941121b1 #1 SMP PREEMPT Thu Nov 20 09:34:02 UTC 2025 aarch64 GNU/Linux
```

**Description**
`UdpSocket::recv` silently disregards any errors surfaced by the (Linux) kernel in `SO_ERROR`. This can cause sockets to hang instead of returning an appropriate error. For example, when calling `recv` after sending to a closed port.

[Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=675bb45ca4a03153783f3067798d7563)
```rust
use std::{io, net::SocketAddr};
use tokio::net::UdpSocket;

#[tokio::main]
async fn main() -> io::Result<()> {
    // Bind a server socket then immediately drop it to get a port we know is closed
    let server = UdpSocket::bind("127.0.0.1:0").await?;
    let target: SocketAddr = server.local_addr()?;
    drop(server);

    let sock = UdpSocket::bind("127.0.0.1:0").await?;
    sock.connect(target).await?;
    sock.send(b"ping").await?;

    let mut buf = [0; 8];
    let result = sock.recv(&mut buf).await;  // hangs infinitely
    println!("{:?}", result);

    Ok(())
}
```

[...]

PR #8001, opened 2026-04-01
Surface errors from `SO_ERROR` on `recv` for UDP sockets on Linux

SRv6d:
Fixes and adds a regression test for https://github.com/tokio-rs/tokio/issues/8000.

## Motivation

https://github.com/tokio-rs/tokio/issues/8000

## Solution

Adjusted `UdpSocket::recv` to wait for `Interest::READABLE | Interest::ERROR` instead of only `Interest::READABLE`. This ensures error-only readiness wakes the operation so the underlying socket error is surfaced correctly. The change is intentionally minimal and scoped to recv to avoid altering broader readiness behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Make sure you enable the full features of the tokio crate on Cargo.toml:
 
 ```toml
 [dependencies]
-tokio = { version = "1.50.0", features = ["full"] }
+tokio = { version = "1.51.0", features = ["full"] }
 ```
 Then, on your main.rs:
 
@@ -221,18 +221,18 @@ warrants a patch release with a fix for the bug, it will be backported and
 released as a new patch release for each LTS minor version. Our current LTS
 releases are:
 
- * `1.43.x` - LTS release until March 2026. (MSRV 1.70)
  * `1.47.x` - LTS release until September 2026. (MSRV 1.70)
+ * `1.51.x` - LTS release until March 2027. (MSRV 1.71)
 
 Each LTS release will continue to receive backported fixes for at least a year.
 If you wish to use a fixed minor release in your project, we recommend that you
 use an LTS release.
 
 To use a fixed minor version, you can specify the version with a tilde. For
-example, to specify that you wish to use the newest `1.43.x` patch release, you
+example, to specify that you wish to use the newest `1.47.x` patch release, you
 can use the following dependency specification:
 ```text
-tokio = { version = "~1.43", features = [...] }
+tokio = { version = "~1.47", features = [...] }
 ```
 
 ### Previous LTS releases
@@ -245,6 +245,7 @@ tokio = { version = "~1.43", features = [...] }
  * `1.32.x` - LTS release until September 2024.
  * `1.36.x` - LTS release until March 2025.
  * `1.38.x` - LTS release until July 2025.
+ * `1.43.x` - LTS release until March 2026.
 
 ## License
 
--- a/docs/contributing/how-to-specify-crates-dependencies-versions.md
+++ b/docs/contributing/how-to-specify-crates-dependencies-versions.md
@@ -1,7 +1,7 @@
 # How to specify crates dependencies versions
 
 Each crate (e.g., `tokio-util`, `tokio-stream`, etc.) should specify dependencies
-according to these rules:
+according to the following rules:
 
 1. The listed version should be the oldest version that the crate works with
 (e.g., if `tokio-util` works with `tokio` version `1.44` but not `1.43`, then
@@ -10,7 +10,7 @@ We don't require users to use the latest version unnecessarily.
 2. When a crate starts using a newer feature in a dependency, the version
 should be bumped to the version that introduced it.
 3. If a crate depends on an unreleased feature in a dependency, it may use
-`path = ` dependency to specify this. Since path dependencies must be removed
+`path =` dependency to specify this. Since path dependencies must be removed
 during the release of the crate, this ensures that it can't be released until
 the dependency has a new version.
 
@@ -20,12 +20,13 @@ Consider the following example from `tokio-stream`:
 [dependencies]
 futures-core = { version = "0.3.0" }
 pin-project-lite = "0.2.11"
-tokio = { path = "../tokio", features = ["sync"] }
+tokio = { version = "1.38.0", path = "../tokio", features = ["sync"] }
 ```
 
 In this case, local development of `tokio-stream` uses the local version
 of `tokio` via the `path` dependency. This means that it's currently not
-possible to release `tokio-stream`. Once a new version of `tokio` is
-released, the path dependency will be removed from `tokio-stream`.
+possible to release `tokio-stream`, and `tokio` should be released first.
+Once a new version of `tokio` is released (in this example the `1.38.0`),
+the path dependency should be removed.
 As mentioned before, this version should only be bumped when adding a new
 feature in the crate that relies on a newer version.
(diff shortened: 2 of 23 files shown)


tokio/tests/udp.rs
```diff
@@ -46,6 +47,26 @@ async fn send_recv_poll() -> std::io::Result<()> {
...
+    sender.connect(receiver_addr).await?;
+    sender.send(MSG).await?;
+
+    let mut recv_buf = [0u8; 32];
+    let err = time::timeout(Duration::from_secs(5), sender.recv(&mut recv_buf))
+        .await
+        .expect("timed out instead of returning error")
+        .unwrap_err();
+
+    assert_eq!(err.kind(), io::ErrorKind::ConnectionRefused);
```
martin-g: It seems Windows returns `ConnectionReset` here.

tokio/src/net/udp.rs
```diff
@@ -785,7 +785,7 @@ impl UdpSocket {
     pub async fn recv(&self, buf: &mut [u8]) -> io::Result<usize> {
         self.io
             .registration()
-            .async_io(Interest::READABLE, || self.io.recv(buf))
+            .async_io(Interest::READABLE | Interest::ERROR, || self.io.recv(buf))
```
martin-g, line 788: There are few more `recv_**` methods similar to this one:
- `read_buf` - https://github.com/tokio-rs/tokio/blob/6c2a446daa1368b98e449b622111a55329d357d0/tokio/src/net/udp.rs#L989
- `recv_buf_from` - https://github.com/tokio-rs/tokio/blob/6c2a446daa1368b98e449b622111a55329d357d0/tokio/src/net/udp.rs#L1120
- `recv_from` - https://github.com/tokio-rs/tokio/blob/6c2a446daa1368b98e449b622111a55329d357d0/tokio/src/net/udp.rs#L1318

tokio/tests/udp.rs
SRv6d: Yes, made the test accept both, we already do this in other test for similar reasons.

SRv6d:
> 1. https://github.com/tokio-rs/tokio/blob/6c2a446daa1368b98e449b622111a55329d357d0/tokio/src/io/interest.rs#L241
>     is an interesting comment
> 2. https://github.com/tokio-rs/tokio/blob/6c2a446daa1368b98e449b622111a55329d357d0/tokio/src/io/interest.rs#L258-L267
> 
> Not related to this PR but I think this needs improvement - it does not handle bitwise combinations.

That comment isn't really accurate, but I think I understand where it comes from. Linux is a bit of an outlier here in that for TCP sockets it seems to signal both `EPOLLIN` and `EPOLLERR` if there's data in `SO_ERROR`, but for UDP sockets it only fires `EPOLLERR`. This is why the Tokio `TcpStream` works correctly and surfaces errors regardless.
I haven't been able to test windows locally, but I have been able to run the regression tests I added as part of this MR on BSD and it passes without the fix, so it seems like Linux really is the outlier here.

Darksonn:
This concerns me a bit. I'm worried we're missing this in many other places.

@Thomasdezeeuw what do you think?

SRv6d:
This is only scoped to `recv` because that's the only path I looked into. That does not mean at all that we should limit it to just that, I just didn't want to get ahead of myself here. I suspect some other places might also be less trivial to fix.

Darksonn:
I don't necessarily think you have to fix all the cases. But it does worry me somewhat.

Darksonn:
What about `send` on an UDP socket? What happens if there's an error from sending - could it be delivered to the reader?

SRv6d:
> What about `send` on an UDP socket? What happens if there's an error from sending - could it be delivered to the reader?

The `send` path should be fine. I haven't tested it explicitly, but send for UDP essentially just means "I handed these bytes for sending to the kernel", there is no other guarantees whatsoever. The only valid error cases for `send` I can think of are `WouldBlock` or other IO issues with the kernel.

Thomasdezeeuw:
> This concerns me a bit. I'm worried we're missing this in many other places.
> 
> @Thomasdezeeuw what do you think?

I'm a little confused about `Interest::ERROR`. As @martin-g pointed out above, there is no `ERROR` interest in Mio because it's always set by all implementations. Falling back to setting `READABLE`: https://github.com/tokio-rs/tokio/blob/6c2a446daa1368b98e449b622111a55329d357d0/tokio/src/io/interest.rs#L240-L247 isn't always correct either, but that's another issue.

The checking of the Mio `Event` is done correctly here: https://github.com/tokio-rs/tokio/blob/43134f1e5784993eb4fb3863933d74ac9e28f598/tokio/src/io/ready.rs#L57-L99

Maybe the issue is in `ScheduledIo::wake`?
https://github.com/tokio-rs/tokio/blob/43134f1e5784993eb4fb3863933d74ac9e28f598/tokio/src/runtime/io/scheduled_io.rs#L238-L288

The check in `Ready:: satisfies` should probably always trigger for errors:
https://github.com/tokio-rs/tokio/blob/43134f1e5784993eb4fb3863933d74ac9e28f598/tokio/src/io/ready.rs#L268-L270

However I'm not sure if that will cause too many wake-ups.

Alternatively `Interest::ERROR` should probably be set for all calls to `async_io` (like this pr is doing) as otherwise they will get stuck on errors like reported in #8000.

tokio/src/net/udp.rs
Darksonn: I think we can merge this once these are fixed.

martin-g: What about the `peek` methods ? 
- peek
https://github.com/tokio-rs/tokio/blob/6c2a446daa1368b98e449b622111a55329d357d0/tokio/src/net/udp.rs#L1559
- peek_from
https://github.com/tokio-rs/tokio/blob/6c2a446daa1368b98e449b622111a55329d357d0/tokio/src/net/udp.rs#L1701
- peek_sender
https://github.com/tokio-rs/tokio/blob/6c2a446daa1368b98e449b622111a55329d357d0/tokio/src/net/udp.rs#L1818

SRv6d: Added peek and remaining recv functions.

Darksonn:
Thanks!

leshow:
I'm not sure if this is relevant but the docs for try_io and async_io say:

> This method is not intended to be used with combined interests. The closure should perform only one type of IO operation, so it should not require more than one ready state. This method may panic or sleep forever if it is called with a combined interest.


isn't passing `Interest::READABLE | Interest::ERROR` a "combined interest"? It seems like the comment means readable | writable but it's not entirely clear, maybe that line needs to be updated.

Darksonn:
You make a pretty good point. What the docs are actually trying to get at is that you should only perform a single syscall inside the closure, and those docs were written under the assumption that a combined interest only makes sense if you perform multiple syscalls, but that's not true for the error interest.

leshow:
OK, thanks for the clarification. I maintain a library that adds the recvmsg/recvmmsg/sendmsg/sendmmsg syscalls to a tokio udp socket so this bug existed in my codebase as well. I was thinking it might also be missing on UnixDatagram/UnixSocket as well, since they all function somewhat similarly. 

I'll write a test and submit a PR if the infinite hang is there as well

leshow:
Interestingly, I saw the infinite hang both with and without registering Interest for `UnixDatagram`

Darksonn:
Hmm. Do you mind opening a new bug with your observations?

Issue #7995, opened 2026-03-30
feat: a New runtime flavor for io-uring


Daksh14:
**Is your feature request related to a problem? Please describe.**

For io_uring a thread per core or ring per thread with work stealing is a needed flavor which will allow 
working with rings without any lock contention and synchronization primitives in between operations.

**Describe the solution you'd like**

Implement a new runtime flavor that spawns rings per each thread with work stealing. We need to look into the load balancing algorithm here @Noah-Kennedy .

Put it under `tokio_unstable` flag for now.

**Describe alternatives you've considered**

Using existing work stealing runtime forces us to use rings behind Mutexes increasing lock contention.

Ring structures benefits include lock free programming

**Additional context**

Azure's runtime https://github.com/Azure/kimojio-rs

PR #8010, opened 2026-04-03
runtime: optional eager I/O driver/timer handoff when polling tasks

hawkw:
## Motivation

As discussed in #4730, #6315, seanmonstar/reqwest#1907,
private-attribution/ipa#1120, oxidecomputer/omicron#9619, and many other
issues, it is possible for the runtime to end up in a situation where
the only unparked worker thread to begin executing a long-running or
permanently blocking task, preventing the time and I/O drivers from
making progress and waking other tasks, since no other worker is
unparked to steal the driver(s). This can result in increased latency
for I/O and timer notifications, or in some cases, can permanently
deadlock the runtime.

## Solution

This branch implements the solution I proposed four years ago in [this
comment][1]: if a worker has just returned from parking on the timer or
I/O driver and is about to poll a task from its run queue, it will
notify another parked worker to attempt to steal the driver. This should
reduce latency for dispatching I/O and timer notifications and reduce
the risk of deadlock. This is very similar to the approach taken in a
previous PR, #6251. The main difference here is that I've hidden this
behavior behind a runtime builder option,
`enable_eager_driver_handoff()`. I chose to do this because the
increased amount of cross-thread synchronization due to additional
worker thread wakeups has made an approach like this somewhat
controversial, and some users may prefer to avoid it despite the
potential benefits. Putting this behavior behind an unstable runtime
builder option allows those who want it to experiment with it, while not
committing to this approach as our permanent and canonical solution to
#6315. This was suggested by @carllerche in [this comment][2] on #6251.
I've also consolidated the wakeup triggered by a worker which had
previously parked on the timer or I/O driver with the wakeup from being
the last worker to `transition_from_searching`, so we only trigger a
single unpark, rather than a second potentially spurious wakeup if both
conditions are true.

Finally, I've added a test which reproduces a (slightly contrived)
deadlock, which is prevented by eager driver handoff.

[1]: https://github.com/tokio-rs/tokio/issues/4730#issuecomment-1147975074
[2]: https://github.com/tokio-rs/tokio/pull/6251#issuecomment-1917713257

### DIFF
--- a/tokio/tests/rt_unstable_eager_driver_handoff.rs
+++ b/tokio/tests/rt_unstable_eager_driver_handoff.rs
@@ -1,6 +1,9 @@
 // These tests only work on Unix platforms because they rely on Unix pipes
 // as a way of generating I/O events from within the same process.
-#![cfg(all(unix, feature = "full"))]
+//
+// Also, Miri doesn't like it when you leak a thread, which will happen in
+// the "deadlock" case below, so we skip these tests on Miri.
+#![cfg(all(not(miri), unix, feature = "full"))]
 
 use std::sync::mpsc::RecvTimeoutError;
 use std::time::Duration;


hawkw:
Oh I _really_ do not love that [my reproducer apparently cannot make the runtime deadlock on ARM](https://github.com/tokio-rs/tokio/actions/runs/23961933972/job/69893686870?pr=8010) --- that suggests some difference of behavior there which is _also_ disconcerting.

hawkw:
> Oh I _really_ do not love that [my reproducer apparently cannot make the runtime deadlock on ARM](https://github.com/tokio-rs/tokio/actions/runs/23961933972/job/69893686870?pr=8010) --- that suggests some difference of behavior there which is _also_ disconcerting.

I suspect this was due to the test's previous reliance on timers; if the test was running in a sufficiently slow environment (such as a QEMU VM on a GitHub Acions runner), the two timers could conceivably complete in the same poll of the timer wheel, resulting in it "not deadlocking". I've rewritten the test to use Unix pipes instead which should hopefully be less flaky.

ongardie-atomix:
FWIW as a random user, thank you! I ran into the problem this fixes when testing a transaction system. My test got stuck with one task hogging one thread, even though many idle threads were available. I knew I had an unyielding retry loop (which I should fix), but I continued to dig in to explain how one task could wedge the entire runtime. The scenario was a single task in an unyielding retry loop, but a timer was set that should have unblocked another task that would get this one unstuck.

At least in my case, this PR solves the problem:

- When I ran without `enable_eager_driver_handoff`, I saw a test failure on the 3rd attempt (and this was typical). 
- When I ran it with `enable_eager_driver_handoff`, I saw 100 successful test runs.

I think the PR makes sense conceptually and should make the runtime behavior more predictable for users. I can't speak to how it affects normal performance, however. I hope this can make its way to becoming easily enabled (without `tokio_unstable`) over time, and it seems like it might make a good default.

hawkw:
> I think the PR makes sense conceptually and should make the runtime behavior more predictable for users. I can't speak to how it affects normal performance, however. I hope this can make its way to becoming easily enabled (without `tokio_unstable`) over time, and it seems like it might make a good default.

I would certainly like to expose this behavior without the unstable flag, and possibly as a default eventually: personally, I also feel that this behavior is probably correct for a majority of applications, and I expect it will probably result in better overall performance in most cases. However, I wanted to be a bit conservative when introducing a change to the runtime's behavior --- even if we confirm this to be an improvement in most cases, we may want to change the interface for enabling it, for example. I'm also hoping there's a path to dropping the unstable flag after it's been out for a couple releases.

Darksonn:
I think if people try it out and confirm no serious perf regressions, we can just make this the default. It won't break any workflows.

It's a bit awkward taking into account that io-uring wants us to introduce per-thread IO contexts, which would bring back #6315 unless we make sure that threads have a way to check the ring owned by *other threads* for completions to avoid starvation. But probably this means that per-thread rings must be [guarded behind a new runtime flavor](https://github.com/tokio-rs/tokio/issues/7995) and the default Tokio configuration will just have a single shared ring.

hawkw:
> I think if people try it out and confirm no serious perf regressions, we can just make this the default. It won't break any workflows.

That sounds great, we could always end up changing the builder option to be `disable_eager_driver_handoff` for anyone who believes that this causes a performance regression for what they're doing, when we stabilize it.

PR #8028, opened 2026-04-09
runtime: use compare_exchange_weak() in worker queue

Darksonn:
I was looking at this code, and I realized there is no reason to use the strong CAS in these cases. If the CAS fails spuriously even though the old value is unchanged, we can just go around the loop and try again, instead of letting the `compare_exchange()` function's internal loop run again.


hawkw (approved):
your reasoning seems correct to me!

PR #8029, opened 2026-04-09
Overflow second half of tasks when local queue is filled instead of first half

Darksonn:
I realized that there are actually some edge case scenarios where [our fairness guarantees](https://docs.rs/tokio/latest/tokio/runtime/index.html#detailed-runtime-behavior) are violated. Specifically, this can occur in the scenario where a task is repeatedly moved back into the injection queue by `push_overflow()` due to another task spawning/notifying a lot of tasks.

To fix this, I'd like to suggest that `push_overflow()` will move the second half of tasks instead of the first half. Since the second half of the local queue only ever gets filled up when a local task notifies another task, this ensures that the tasks overflowed to the injection queue were recently notified.

This means that we can now actually prove the fairness guarantee by placing notified tasks in three buckets:

1. First half of local queue.
2. Injection queue.
3. Second half of local queue.

Whenever a task is newly notified, it can end up in any of the three buckets. However, all operations that happen from there on will move a task *down* to a lower bucket, or decrease the number of tasks in front of it in the current bucket. This ensures that tasks are eventually scheduled.

A few notes on why this works:

* When a worker steals tasks from another local queue, then its own local queue is empty and the first task is kept for immediate execution — only the other `n-1` tasks are added to the local queue. This means that work stealing always decreases the number of tasks in front of each stolen task by one. This implies that work stealing cannot move a task in the first half of one local queue into the second half of another local queue. It also implies that a task cannot be repeatedly stolen infinitely many times, as each time it is stolen, the number of tasks in front of it is reduced.
* When moving tasks from the injection queue to the local queue, all of them are placed in the first half of the local queue.
* When `push_overflow()` is triggered, then *as of this PR* only tasks from the second half of the local queue are moved to the injection queue.

Thus when a task is notified, it takes *at most* `(capacity of local queue) + (max length of injection queue) * (global queue interval)` operations before the task is polled. Since the fairness guarantee only applies when the total number of tasks is bounded, this means that the max length of the injection queue is also bounded, so the task will eventually be polled after a finite number of operations.

### DIFF
--- a/tokio/src/runtime/scheduler/multi_thread/queue.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/queue.rs
@@ -283,9 +283,7 @@ impl<T> Local<T> {
             "queue is not full; tail = {tail}; head = {head}"
         );
 
-        let prev = pack(head, head);
-
-        // Claim a bunch of tasks
+        // Claim all tasks.
         //
         // We are claiming the tasks **before** reading them out of the buffer.
         // This is safe because only the **current** thread is able to push new
@@ -298,15 +296,7 @@ impl<T> Local<T> {
         if self
             .inner
             .head
-            .compare_exchange(
-                prev,
-                pack(
-                    head.wrapping_add(NUM_TASKS_TAKEN),
-                    head.wrapping_add(NUM_TASKS_TAKEN),
-                ),
-                Release,
-                Relaxed,
-            )
+            .compare_exchange(pack(head, head), pack(tail, tail), Release, Relaxed)
             .is_err()
         {
             // We failed to claim the tasks, losing the race. Return out of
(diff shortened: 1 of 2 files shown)


carllerche:
It has been a while since I worked on this. I'm not really following. What is the mechamism by which the fairness isn't met? What are both halves of the queue (which is which)

Darksonn:
The scenario with a fairness issue is the case where a task gets moved from the injection queue to the local queue to the injection queue to the local queue to the injection queue to the local queue to the injection queue and so on. This PR ensures that once a task has moved from the injection queue to the local queue, it cannot be moved back to the injection queue.

The first half of the queue is the 128 slots closest to the head of the local queue (the end from which tasks are popped). The second half is the other 128 slots, which are the ones closest to the tail of the queue (the end from which tasks are pushed).

This PR makes it so that when `push_overflow()` is called because the local queue is full, it removes tasks from the "wrong" end of the queue. That is, unlike all other methods that remove tasks from the queue, it removes tasks from the tail of the queue, instead of from the head.

carllerche (approved):
I can't think of a reason why this wouldn't work or be a bad idea.

ADD-SP (approved):
Nice catch!

PR #8030, opened 2026-04-09
net: add 'try_io' function to 'unix::pipe' sender and receiver types

### LINKED ISSUE #7998: Pipes are missing a `try_io` method
( I wasn't sure which topic to file this issue under, it's either this one, or "support splicing pipes").

`pipe::Sender` / `pipe::Receiver` has a internally cached readiness state, that's only cleared when an IO attempt fails with `WouldBlock`. Therefore it's impossible to do IO using external functions, such as `rustix::pipe::splice`.

Add `pipe::{Sender, Receiver}::clear_readiness`.

I can probably recreate the `Sender` / `Receiver` every time I do a splice (i.e. `Receiver::from_owned_fd(recv.into_owned_fd())`), but that's pretty ridiculous.

Or, add an API for splicing pipes.

tglane:
## Motivation
#7998 mentioned, that the `try_io` method is missing on the `unix::pipe` types Sender and Receiver.

## Solution
I added the requested method to both `unix::pipe::{Sender | Receiver}` like it's implemented for other types like `TcpStream`. However, I found that it would not make much sense to have `try_io` take the `Interest` as a parameter because the Sender would only be interested in `Interest::WRITABLE` and the Receiver in `Interest::READABLE`. So I removed that parameter and hardcoded the corresponding interest.

Closes #7998

### DIFF
--- a/tokio/src/net/unix/pipe.rs
+++ b/tokio/src/net/unix/pipe.rs
@@ -722,6 +722,38 @@ impl Sender {
             .try_io(Interest::WRITABLE, || (&*self.io).write_vectored(buf))
     }
 
+    /// Tries to write from the socket using a user-provided IO operation.
+    ///
+    /// If the socket is ready, the provided closure is called. The closure
+    /// should attempt to perform IO operation on the socket by manually
+    /// calling the appropriate syscall. If the operation fails because the
+    /// socket is not actually ready, then the closure should return a
+    /// `WouldBlock` error and the readiness flag is cleared. The return value
+    /// of the closure is then returned by `try_io`.
+    ///
+    /// If the socket is not ready, then the closure is not called
+    /// and a `WouldBlock` error is returned.
+    ///
+    /// The closure should only return a `WouldBlock` error if it has performed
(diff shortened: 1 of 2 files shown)


tokio/src/net/unix/pipe.rs
```diff
@@ -722,6 +722,38 @@ impl Sender {
...
+    /// and a `WouldBlock` error is returned.
+    ///
+    /// The closure should only return a `WouldBlock` error if it has performed
+    /// an IO operation on the socket that failed due to the socket not being
+    /// ready. Returning a `WouldBlock` error in any other situation will
+    /// incorrectly clear the readiness flag, which can cause the socket to
+    /// behave incorrectly.
+    ///
+    /// The closure should not perform the IO operation using any of the methods
+    /// defined on the Tokio `TcpStream` type, as this will mess with the
```
Darksonn:
```suggestion
    /// defined on the Tokio `pipe::Sender` type, as this will mess with the
```

Darksonn (approved):
Thanks.

PR #8032, opened 2026-04-09
net: document pipe try_read*/try_write* readiness behavior

barry3406:
## Motivation

#7625 reports that calling `pipe::Receiver::try_read` immediately after creating the receiver returns `WouldBlock` even when data is already available in the pipe. The same applies to `try_read_vectored`, `try_read_buf`, and the `try_write*` methods on `pipe::Sender`.

This is intentional: `Registration::try_io` checks the cached readiness state and short-circuits with `WouldBlock` until the I/O driver has had a chance to deliver an event for the registration. The same model applies to every other Tokio I/O type. But the `pipe` docs do not currently call this out, and it's easy to assume `try_*` is a thin syscall wrapper — which led to the confusion in #7625 (and to my own first attempt at "fixing" it in this PR before @Darksonn and @Thomasdezeeuw kindly explained the actual model).

## Solution

Add a `# Notes` section to all five `try_*` methods on `pipe::Sender` and `pipe::Receiver` explaining that the runtime's I/O driver only delivers readiness events after control has been yielded back to it, and that calling `try_*` before any `.await` (such as `readable()` / `writable()`) will return `WouldBlock` even when the operation could otherwise succeed.

No code or behavior changes. The note follows the existing `# Notes` convention used in `udp.rs`, `tcp/stream.rs`, etc., and is positioned between `# Return` and `# Examples` to match `udp::UdpSocket::try_recv_from` which has the same shape.

Verified with the docs.rs-equivalent build from the [contributors guide](https://github.com/tokio-rs/tokio/blob/master/docs/contributing/pull-requests.md):

```
RUSTDOCFLAGS="--cfg docsrs --cfg tokio_unstable" RUSTFLAGS="--cfg docsrs --cfg tokio_unstable" cargo +nightly doc -p tokio --all-features --no-deps
```

No new warnings from `pipe.rs`, and the rendered HTML on both the `Receiver` and `Sender` pages contains all five new `# Notes` sections.


tokio/tests/net_unix_pipe.rs
```diff
@@ -544,3 +544,25 @@ async fn anon_pipe_into_blocking_fd() -> std::io::Result<()> {
...
+async fn anon_pipe_try_read_bypassing_mio() -> std::io::Result<()> {
+    use std::io::Write;
+
+    let (tx, rx) = pipe::pipe()?;
+
+    // Convert the sender to a blocking fd and write outside of tokio,
+    // so mio is never notified about the data.
+    let mut tx_file = File::from(tx.into_blocking_fd()?);
+    tx_file.write_all(b"hello")?;
+    drop(tx_file);
```
Darksonn: It is alarming to me that this code does not work. There's no such thing as "bypassing mio" since mio just gets notifications from the OS, and writing to the pipe really should result in a notification from the OS.

This points towards a deeper issue IMO. cc @Thomasdezeeuw

Darksonn: Actually thinking more about it, this is entirely expected behavior. If you insert a `yield_now()` so the IO driver has time to receive IO events, it will work even with a blocking fd. This is no different from `try_io()` on `TcpStream` or anywhere else.

If you want to bypass Tokio's readiness tracking, you can always invoke the syscall directly without using `try_io`.

Thomasdezeeuw: As @Darksonn has already mentioned you can't "bypass" Mio's readiness tracking, is done by the OS. What is happening here is the following:
1. tx_file.write_all writes the bytes to the buffer. But since the rx hasn't been registered, no readiness events are generated.
2. rx.try_read first check if we've got a readiness event before doing the actual read system call. Since the rx wasn't registered yet it doesn't know that the read is ready to be performed: https://github.com/tokio-rs/tokio/blob/4c7d8b2a14de81ace4d8e61cfaeb8eca26f6f68c/tokio/src/runtime/io/registration.rs#L183-L193.

I think if you change the rx.try_read to use `AsyncRead` it will work as expected. Or, as @Darksonn also mentioned, perform the read call directly bypassing Tokio's check if the I/O source is ready -- you'll have to deal with a potential `WouldBlock` error yourself though.

barry3406: Thanks for the explanation — I hadn't read `try_io` carefully. The short-circuit on empty readiness is intentional and consistent across every tokio IO type, so special-casing `pipe::Receiver` was wrong.

Just to confirm: the answer to #7625 is a `.readable().await` (or `yield_now().await`) before `try_read`, so the driver gets a chance to register and deliver the event?

If that's worth a docs note on `try_read*` / `try_write*`, I can repurpose this PR for that. Otherwise I'll close it.

Darksonn: If you want to write a docs note, that'd be great!

barry3406: Done — pushed `adef662`. Added a `# Notes` section to all five `try_*` methods on `pipe::Sender` / `pipe::Receiver`, matching the convention used by `udp::UdpSocket::try_recv_from` (heading between `# Return` and `# Examples`). Verified with the docs.rs build from the contributors guide; CI is green.

tokio/src/net/unix/pipe.rs
```diff
@@ -617,6 +617,13 @@ impl Sender {
     /// number of bytes written. If the pipe is not ready to write data,
     /// `Err(io::ErrorKind::WouldBlock)` is returned.
     ///
+    /// # Notes
+    ///
+    /// The runtime's I/O driver only delivers readiness events to a `Sender`
+    /// after control is yielded back to it. Calling this method before any
+    /// `.await` (such as [`writable`]) will return `WouldBlock` even when
+    /// the pipe has space available.
```
Darksonn: I would like to avoid "will" because that sounds like a guarantee. We don't guarantee that we will return `WouldBlock` because the IO driver could have been driven by another runtime thread, and in that case you might get readiness event from that other thread in parallel.

How about something along these lines?

> To avoid unnecessary syscalls, this will only attempt the write operation if the OS has informed Tokio that this pipe has become writable. Because of this `try_write()` may fail with a `WouldBlock` error if Tokio has not yet heard from the OS that this pipe has become writable.

Darksonn (approved):
Thanks.

Darksonn:
Ah, looks like the `WouldBlock` link doesn't work.

barry3406:
good catch, fixed in 3c06d4b7 — added [`WouldBlock`]: std::io::ErrorKind::WouldBlock reference definitions under each Notes section to match the convention in async_fd.rs. cargo doc --no-deps now resolves cleanly.

mattiapitossi:
Thanks!

PR #8039, opened 2026-04-11
ci: rename FreeBSD task names for clarity

xtqqczze:


ADD-SP (approved):
Thanks!

mattiapitossi:
At this point, I’m not totally sure that it’s still worth editing this file given that we will migrate to something else.

xtqqczze:
> At this point, I’m not totally sure that it’s still worth editing this file given that we will migrate to something else.

Due to speed issues, we probably want to delay migration to sometime nearer to 2026-06-01?

Darksonn:
Since we're removing this job soon, I do not think this is worth the churn involved in changing the list of required jobs. As for delaying the move from cirrus, I do not think we need to delay it. I would rather get it over with than wait.

PR #8040, opened 2026-04-12
Allow impl FnMut() in taskdumps instead of just fn()

Darksonn:
Follow-up to #8025 that removes the need for thread-locals.

cc @rcoh

### DIFF
--- a/tokio/src/runtime/task/trace/mod.rs
+++ b/tokio/src/runtime/task/trace/mod.rs
@@ -136,8 +136,8 @@ impl Context {
             }
         };
 
-        // SAFETY: This call can only access the trace_leaf_fn field, so it cannot
-        // break the trace frame linked list.
+        // SAFETY: This call can only access the trace_leaf_fn field, so it cannot break the trace
+        // frame linked list.
         unsafe { Self::try_with_current(inner) };
 
         ret
@@ -145,8 +145,8 @@ impl Context {
 
     /// Produces `true` if the current task is being traced; otherwise false.
     pub(crate) fn is_tracing() -> bool {
-        // SAFETY: This call can only access the trace_leaf_fn field, so it cannot
-        // break the trace frame linked list.
+        // SAFETY: This call can only access the trace_leaf_fn field, so it cannot break the trace
+        // frame linked list.
         unsafe { Self::try_with_current(|ctx| ctx.trace_leaf_fn.get().is_some()).unwrap_or(false) }
     }
 }
@@ -226,8 +226,8 @@ where
     // Even if this access fails, that's okay. In that case, we still call the closure without
     // actually performing any tracing.
     //
-    // SAFETY: This call can only access the trace_leaf_fn field, so it cannot
-    // break the trace frame linked list.
+    // SAFETY: This call can only access the trace_leaf_fn field, so it cannot break the trace
+    // frame linked list.
     unsafe {
         Context::try_with_current(|ctx| {
             old_trace_leaf_fn = ctx.trace_leaf_fn.replace(Some(trace_leaf_dyn));
--- a/tokio/src/runtime/task/trace/mod.rs
+++ b/tokio/src/runtime/task/trace/mod.rs
@@ -129,7 +129,7 @@ impl Context {
                 });
 
                 // SAFETY: The trace leaf fn is valid for the duration in which it's stored in the
-                // context. Furthermore, re-entrant calls are not possible since we store `None` for
+                // context. Furthermore, re-entrant calls are not possible because we store `None` for
                 // the duration in which we hold a mutable reference, so access is exclusive for that
                 // duration.
                 ret = Some(f(unsafe { trace_leaf_fn.as_mut() }));
--- a/tokio/src/runtime/task/trace/mod.rs
+++ b/tokio/src/runtime/task/trace/mod.rs
@@ -34,7 +34,7 @@ pub(crate) struct Context {
     /// The function that is invoked at each leaf future inside of Tokio
     ///
     /// For example, within tokio::time:sleep, sockets. etc.
-    trace_leaf_fn: Cell<Option<fn(&TraceMeta)>>,
+    trace_leaf_fn: Cell<Option<NonNull<dyn FnMut(&TraceMeta)>>>,
 }
 
 /// A [`Frame`] in an intrusive, doubly-linked tree of [`Frame`]s.
@@ -114,19 +114,38 @@ impl Context {
         }
     }
 
+    /// Calls the provided closure if we are being traced.
     fn try_with_current_trace_leaf_fn<F, R>(f: F) -> Option<R>
     where
-        F: FnOnce(&Cell<Option<fn(&TraceMeta)>>) -> R,
+        F: for<'a> FnOnce(&'a mut dyn FnMut(&TraceMeta)) -> R,
     {
+        let mut ret = None;
+
+        let inner = |context: &Context| {
+            if let Some(mut trace_leaf_fn) = context.trace_leaf_fn.get() {
+                context.trace_leaf_fn.set(None);
+                let _restore = defer(move || {
+                    context.trace_leaf_fn.set(Some(trace_leaf_fn));
+                });
+
+                // SAFETY: The trace leaf fn is valid for the duration in which it's stored in the
+                // context. Furthermore, re-entrant calls are not possible since we store `None` for
+                // the duration in which we hold a mutable reference, so access is exclusive for that
+                // duration.
+                ret = Some(f(unsafe { trace_leaf_fn.as_mut() }));
+            }
+        };
+
         // SAFETY: This call can only access the trace_leaf_fn field, so it cannot
         // break the trace frame linked list.
-        unsafe { Self::try_with_current(|context| f(&context.trace_leaf_fn)) }
+        unsafe { Self::try_with_current(inner) };
+
+        ret
     }
 
     /// Produces `true` if the current task is being traced; otherwise false.
     pub(crate) fn is_tracing() -> bool {
-        Self::try_with_current_trace_leaf_fn(|maybe_trace_leaf| maybe_trace_leaf.get().is_some())
-            .unwrap_or(false)
+        Self::try_with_current_trace_leaf_fn(|_| ()).is_some()
     }
 }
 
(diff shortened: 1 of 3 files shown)


rcoh (approved):
Clever solution to avoid both a TL and allocating a pointer

I have been thinking about it more — I think it will be more useful for the explicit capture trace if we can disable the re-wake.

The two ways that seem most natural are:
- putting a book in TraceMeta that can be mutated by the custom fn
- adding a return type

Do you prefer one?

Darksonn:
I'm not sure I understand the proposed solutions.

rcoh:
```rust
enum WakeOptions {
  /// Don't defer the wake. The caller MUST re-poll the future in normal mode to avoid lost wake
  DoNotWake,
  /// Defer the wake (equivalent to `task::yield_now().await`, this is the current behavior)
  #[default] 
  DeferWake
}

// option 1: return type of callback
pub fn trace_with<FN, FT, R>(f: FN, mut trace_leaf: FT) -> R
where
    FN: FnOnce() -> R,
    FT: FnMut(&TraceMeta) -> WakeOptions, // return WakeOptions to signal whether or not to schedule a wake
{

// option 2: inside TraceMeta:

struct TraceMeta {
   defer_wake: WakeOptions,
   ...
}

pub fn trace_with<FN, FT, R>(f: FN, mut trace_leaf: FT) -> R
where
    FN: FnOnce() -> R,
    FT: FnMut(&mut TraceMeta), // callback can mutate TraceMeta to signal behavior
{
```

Then on https://github.com/tokio-rs/tokio/pull/8040/changes#diff-b4283725d97f7155a412d4409e9942de696777c5be95306d87994ecc52132bfbR309 we would inspect the value to determine whether or not to schedule a wake

Darksonn:
Perhaps we should just pass the context to the function, and the tracer function can decide itself whether it wants to defer the waker or not.

rcoh:
Yeah that should work as long as there is a way to get access to it.

For the "wrapper future use case" we don't want it to wake anyway.

The only place where this is actually what you want is the "runtime wide dump" scenario and presumably we can't handle that separately inside of tokio.

So possibly the answer is "don't wake at all"

Darksonn:
Let's do that (in a separate PR).

rcoh:
I'll work on that now (stacked on this PR)

rcoh:
https://github.com/tokio-rs/tokio/pull/8043

tokio/src/runtime/task/trace/mod.rs
```diff
@@ -189,20 +209,46 @@ pub struct TraceMeta {
 /// assert!(count > 0);
 /// # }
 /// ```
-pub fn trace_with<F, R>(f: F, trace_leaf: fn(&TraceMeta)) -> R
+pub fn trace_with<FN, FT, R>(f: FN, mut trace_leaf: FT) -> R
```
martin-g, line 207: The function documentation needs to be updated too.

tokio/src/runtime/task/trace/mod.rs
```diff
@@ -114,19 +115,38 @@ impl Context {
...
+        unsafe { Self::try_with_current(inner) };
+
+        ret
     }
 
     /// Produces `true` if the current task is being traced; otherwise false.
     pub(crate) fn is_tracing() -> bool {
-        Self::try_with_current_trace_leaf_fn(|maybe_trace_leaf| maybe_trace_leaf.get().is_some())
-            .unwrap_or(false)
+        Self::try_with_current_trace_leaf_fn(|_| ()).is_some()
```
martin-g:
```suggestion
        unsafe {
            Self::try_with_current(|ctx| ctx.trace_leaf_fn.get().is_some())
                .unwrap_or(false)
        }
```
This is an alternative that avoids the temporary set to None + defer() just to discard ()

tokio/src/runtime/task/trace/mod.rs
```diff
@@ -34,7 +34,8 @@ pub(crate) struct Context {
     /// The function that is invoked at each leaf future inside of Tokio
     ///
     /// For example, within tokio::time:sleep, sockets. etc.
-    trace_leaf_fn: Cell<Option<fn(&TraceMeta)>>,
+    #[allow(clippy::type_complexity)]
+    trace_leaf_fn: Cell<Option<NonNull<dyn FnMut(&TraceMeta)>>>,
```
martin-g, line 38: This could be simplified a bit by introducing a typealias:
```
type TraceLeafFn = NonNull<dyn FnMut(&TraceMeta)>;
...
trace_leaf_fn: Cell<Option<TraceLeafFn>>,
```
and `#[allow(clippy::type_complexity)]` won't be needed.

tokio/src/runtime/task/trace/trace_impl.rs
```diff
@@ -2,73 +2,42 @@
...
             }
-            collector.push_backtrace(frames);
-            state.collector.set(Some(collector));
-        }
-    });
+
+            below_root
+        });
+    }
+    trace.push_backtrace(frames);
```
martin-g, line 42:
```suggestion
    
        trace.push_backtrace(frames);
    }
```
no need to add the empty `frames` if there is no root_addr

tokio/src/runtime/task/trace/mod.rs
hawkw: Hm, personally, I'm not convinced that makes it _clearer_ --- when I see the code with the type alias, I don't actually know what type of function a `TraceLeafFn` is, and have to go track down the definition. This way, when I see the field definition, I know it's a `FnMut(&TraceMeta)`, which I think is actually much more understandable than following Clippy's suggestion of hiding things behind aliases... 🤷‍♀️

tokio/src/runtime/task/trace/mod.rs
```diff
@@ -114,19 +115,38 @@ impl Context {
...
     fn try_with_current_trace_leaf_fn<F, R>(f: F) -> Option<R>
     where
-        F: FnOnce(&Cell<Option<fn(&TraceMeta)>>) -> R,
+        F: for<'a> FnOnce(&'a mut dyn FnMut(&TraceMeta)) -> R,
     {
+        let mut ret = None;
+
+        let inner = |context: &Context| {
+            if let Some(mut trace_leaf_fn) = context.trace_leaf_fn.get() {
+                context.trace_leaf_fn.set(None);
```
hawkw: nit, considered unimportant: isn't this equivalent to:
```suggestion
            if let Some(mut trace_leaf_fn) = context.trace_leaf_fn.replace(None) {
```

tokio/src/runtime/task/trace/mod.rs
```diff
@@ -114,19 +115,38 @@ impl Context {
...
+                let _restore = defer(move || {
+                    context.trace_leaf_fn.set(Some(trace_leaf_fn));
+                });
+
+                // SAFETY: The trace leaf fn is valid for the duration in which it's stored in the
+                // context. Furthermore, re-entrant calls are not possible since we store `None` for
```
hawkw: nit:
```suggestion
                // context. Furthermore, re-entrant calls are not possible, since we store `None` for
```

tokio/src/runtime/task/trace/mod.rs
```diff
@@ -189,20 +209,46 @@ pub struct TraceMeta {
...
+    // SAFETY: Pointer comes from reference, so not null.
+    let trace_leaf_dyn = unsafe { NonNull::new_unchecked(trace_leaf_dyn) };
+
+    let mut old_trace_leaf_fn = None;
+
+    // Even if this access fails, that's okay. In that case, we still call the closure without
+    // actually performing any tracing.
+    //
+    // SAFETY: This call can only access the trace_leaf_fn field, so it cannot
+    // break the trace frame linked list.
```
hawkw: nitpicky, sorry: these two comments are wrapped at different columns...
```suggestion
    // Even if this access fails, that's okay. In that case, we still call the 
    // closure without actually performing any tracing.
    //
    // SAFETY: This call can only access the trace_leaf_fn field, so it cannot
    // break the trace frame linked list.
```

tokio/src/runtime/task/trace/mod.rs
```diff
@@ -189,20 +209,46 @@ pub struct TraceMeta {
...
+    // SAFETY: The raw pointer is removed from the thread local before `trace_leaf` is dropped, so
+    // this transmute cannot lead to the violation of any lifetime requirements.
+    let trace_leaf_dyn = unsafe {
+        std::mem::transmute::<
+            *mut (dyn FnMut(&TraceMeta) + '_),
+            *mut (dyn FnMut(&TraceMeta) + 'static),
+        >(trace_leaf_dyn)
+    };
+    // SAFETY: Pointer comes from reference, so not null.
+    let trace_leaf_dyn = unsafe { NonNull::new_unchecked(trace_leaf_dyn) };
```
hawkw, line 222: am I correct that it is considered bad to perform the lifetime-extending transmute on the `&mut` reference and so it must be turned into a `*mut` pointer prior to doing so?

tokio/src/runtime/task/trace/trace_impl.rs
Darksonn: I don't understand what you are suggesting.

tokio/src/runtime/task/trace/mod.rs
Darksonn: Rewording to "because" instead of "since"

tokio/src/runtime/task/trace/mod.rs
Darksonn: Done

tokio/src/runtime/task/trace/mod.rs
Darksonn: I think it's clearer without this change.

tokio/src/runtime/task/trace/mod.rs
Darksonn: Done

tokio/src/runtime/task/trace/mod.rs
Darksonn: Done

tokio/src/runtime/task/trace/mod.rs
Darksonn: It would probably work either way, but I prefer doing so on a raw pointer.

tokio/src/runtime/task/trace/mod.rs
Darksonn: Done

tokio/src/runtime/task/trace/trace_impl.rs
martin-g: The suggestion moves `trace.push_backtrace(frames);` inside `if let Some(root_addr) = meta.root_addr {`.
My thinking is: do not add `frames` (an empty Vec) to `trace` if `meta.root` is `None`.

martin-g (approved):

tokio/src/runtime/task/trace/trace_impl.rs
Darksonn: That seems error-prone. If the code is changed to the future so it may not be empty, then we could accidentally forget to push them to the list. I don't think there's any benefit to this.

tokio/src/runtime/task/trace/mod.rs
hawkw: Fair enough, I think I would too --- just figured I'd ask.

hawkw (approved):

PR #8043, opened 2026-04-13
taskdump: skip double wake on `Trace::capture`/`Trace::trace_with`

rcoh:
## Motivation

> This is a _subtly_ breaking change but I think is very unlikely to break any actual code.

Previously, when `Trace::capture` was called, the task was always scheduled to be woken again. This is really problematic if you want to make a wrapper future that takes a TaskDump; When you capture a task dump you will immediately get polled again. If you capture a task dump on every poll, you have caused an infinite loop!

The alternative I tried was passing a no-op waker. This works _most_ of the time, but in a few edge cases can cause lost wakes.

This moves the rewake to only user that actually _wants_ to rewake the future, the `dump` path where we are trying to dump all futures.

## Solution

Move the deferred wake into `Runtime::dump`. Expose a `waker` accessor to enable this.

_Assisted by Claude Opus 4.7_

### DIFF
--- a/tokio/src/runtime/task/mod.rs
+++ b/tokio/src/runtime/task/mod.rs
@@ -509,6 +509,22 @@ impl<S: Schedule> LocalNotified<S> {
         mem::forget(self);
         raw.poll();
     }
+
+    /// Returns a new `Waker` for this task.
+    ///
+    /// The returned `Waker` is ref-counted independently of this
+    /// `LocalNotified`; it remains valid after the task is polled.
+    #[cfg(all(
+        tokio_unstable,
+        feature = "taskdump",
+        feature = "rt",
+        target_os = "linux",
+        any(target_arch = "aarch64", target_arch = "x86", target_arch = "x86_64")
+    ))]
+    pub(crate) fn waker(&self) -> std::task::Waker {
+        let header_ptr = self.task.raw.header_ptr();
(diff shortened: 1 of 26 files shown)


eliottreich:
There's a $35 bounty for Tokio async runtime improvements on [Code.TaskBounty](https://code.task-bounty.com/task/tokio-async-runtime-enhancement). This optimization would be a good fit.

Developers compete to implement it — you get working PRs with benchmarks to evaluate. Bounty: https://code.task-bounty.com/task/tokio-async-runtime-enhancement

tokio/src/runtime/task/mod.rs
```diff
@@ -509,6 +509,22 @@ impl<S: Schedule> LocalNotified<S> {
...
+    #[cfg(all(
+        tokio_unstable,
+        feature = "taskdump",
+        feature = "rt",
+        target_os = "linux",
+        any(target_arch = "aarch64", target_arch = "x86", target_arch = "x86_64")
+    ))]
+    pub(crate) fn waker(&self) -> std::task::Waker {
+        let header_ptr = self.task.raw.header_ptr();
+        waker::waker_ref::<S>(&header_ptr).clone()
```
Darksonn: By using the same return value as `waker_ref`, I think you can avoid this clone and refcount increment.

Darksonn (approved):
Thanks.

PR #8046, opened 2026-04-15
Adjust spawn_blocking() queue resize check

Darksonn:
Let's avoid wasting work on allocating a larger buffer many times just because a single item was popped while the lock was not held.

cc @alex

### DIFF
--- a/tokio/src/runtime/blocking/sharded_queue.rs
+++ b/tokio/src/runtime/blocking/sharded_queue.rs
@@ -61,18 +61,11 @@ impl Shard {
             let mut new_queue = VecDeque::with_capacity(new_cap);
 
             queue = self.queue.lock();
-            // If the queue is:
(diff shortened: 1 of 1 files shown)


mattiapitossi (approved):

alex:
Seems was reasonable. Was this found with profiling a real workload or just reviewing the code?

Darksonn:
Just reviewing the code.

afurm:
The new condition `new_queue.capacity() > queue.capacity()` triggers a resize even when the queue isn't full. If the original queue had room but `new_queue` was overallocated (e.g., from a prior shrink_to_fit), this allocates a larger buffer unnecessarily. Is that trade-off intentional given the comment about not wasting the allocation?

Darksonn:
I don't think we ever call `shrink_to_fit()` on the queue to begin with.

PR #8062, opened 2026-04-17
sync: fix underflow in mpsc channel `len()`

### LINKED ISSUE #8061: Panic in subtraction with overflow in tokio/src/sync/mpsc/list.rs
Since updating to Tokio 1.51.1 in Wasmtime's CI we've been seening spurious panics in debug mode on [this line of code](https://github.com/tokio-rs/tokio/blob/905c146aeda741ea2202f942a7c3a606dda13da5/tokio/src/sync/mpsc/list.rs#L278). Bisection originally showed https://github.com/tokio-rs/tokio/pull/7431 was at fault, but further reduction makes me think that it just made the scheduling to cause this bug more likely. I've reduced this down to:

```rust
use tokio_util::sync::PollSender;

#[tokio::main]
async fn main() {
    for _ in 0.. {
        let mut tasks = Vec::new();
        for _ in 0..100 {
            let (a, mut b) = tokio::sync::mpsc::channel::<u32>(1);
            let b = tokio::spawn(async move {
                b.is_empty();
                b.recv().await;
            });
            let a = tokio::spawn(async move {
                let mut a = PollSender::new(a);
                std::future::poll_fn(|cx| a.poll_reserve(cx)).await.unwrap();
                a.send_item(3).unwrap();
            });

            tasks.push(a);
            tasks.push(b);
        }
        for task in tasks {
            task.await.unwrap();
        }
    }
}
```

[runnable here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=a16d81a2a7a8d43a3f37b6bd44108e24)

which eventually panics with:

```
$ cargo run
   Compiling wat v0.1.0 (/home/alex/code/wat)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.19s
     Running `target/debug/wat`

thread 'tokio-runtime-worker' (1168048) panicked at /home/alex/code/tokio/tokio/src/sync/mpsc/list.rs:257:9:
attempt to subtract with overflow

thread 'main' (1167905) panicked at src/main.rs:23:24:
called `Result::unwrap()` on an `Err` value: JoinError::Panic(Id(2718217), "attempt to subtract with overflow", ...)
```

This takes ~5 seconds or so locally.

This panics on Tokio 1.37.0 (where `is_empty` was introduced) all the way through to the current 1.52.1.

Darksonn:
Fixes: #8061

### DIFF
--- a/tokio/src/sync/mpsc/list.rs
+++ b/tokio/src/sync/mpsc/list.rs
@@ -99,7 +99,6 @@ impl<T> Tx<T> {
         unsafe { block.as_ref().tx_close() }
     }
 
-
     fn find_block(&self, slot_index: usize) -> NonNull<Block<T>> {
         // The start index of the block that contains `index`.
         let start_index = block::start_index(slot_index);
@@ -248,18 +247,37 @@ impl<T> Rx<T> {
         self.len(tx) == 0
     }
 
-    fn has_value(&self, slot_index: usize) -> bool {
+    // Guaranteed to return true if `slot_index` is the fake message sent on channel close.
+    // Guaranteed to return false if `slot_index` is a fully sent message.
+    //
+    // For messages that are partially sent, may return either true or false.
+    fn is_maybe_closed(&self, tx: &Tx<T>, slot_index: usize) -> bool {
         let start_index = block::start_index(slot_index);
+
+        let tail = tx.block_tail.load(Acquire);
+        // SAFETY: Only the receiver frees blocks, so since we are the receiver, this will not be
+        // freed right now.
+        let tail_ref = unsafe { &*tail };
+        if tail_ref.is_at_index(start_index) {
+            return !tail_ref.has_value(slot_index);
+        }
+
+        // This method is optimized for checking whether the last value is present, so most of the
+        // time it is in `block_tail`. However, this isn't always the case since it's possible
+        // that the list was grown with an empty block, in which case `block_tail` points one block
+        // too far. To handle this case, we walk the list from the head.
         let mut block_ptr = Some(self.head);
 
         while let Some(block) = block_ptr {
+            // SAFETY: Only the receiver frees blocks, so since we are the receiver, this will not
+            // be freed right now.
             let block_ref = unsafe { block.as_ref() };
             if block_ref.is_at_index(start_index) {
-                return block_ref.has_value(slot_index);
+                return !block_ref.has_value(slot_index);
             }
             block_ptr = block_ref.load_next(Acquire);
         }
-        false
+        true
     }
 
     pub(crate) fn len(&self, tx: &Tx<T>) -> usize {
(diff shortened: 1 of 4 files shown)


tokio/src/sync/mpsc/list.rs
```diff
@@ -272,10 +268,24 @@ impl<T> Rx<T> {
     }
 
     pub(crate) fn len(&self, tx: &Tx<T>) -> usize {
-        // When all the senders are dropped, there will be a last block in the tail position,
-        // but it will be closed
         let tail_position = tx.tail_position.load(Acquire);
-        tail_position - self.index - (tx.is_closed() as usize)
+        let mut len = tail_position.wrapping_sub(self.index);
```
Darksonn, line 272: I realized that we should also be using wrapping arithmetic for the position indices. On 32-bit machines these indices can wrap around after a few billion messages.

tokio/src/sync/mpsc/list.rs
```diff
@@ -99,6 +99,11 @@ impl<T> Tx<T> {
         unsafe { block.as_ref().tx_close() }
     }
 
+    fn has_value(&self, slot_index: usize) -> bool {
+        let block = unsafe { self.find_block(slot_index).as_ref() };
```
mattiapitossi, line 103: Should we add a safety comment here?

tokio/src/sync/tests/loom_mpsc.rs
```diff
@@ -222,3 +222,50 @@ fn nonempty_after_send() {
...
+        drop(tx);
+
+        th1.join().unwrap();
+    });
+}
+
+#[test]
+fn len_accurate_during_close() {
+    loom::model(|| {
+        let (tx, rx) = mpsc::channel::<()>(2);
```
martin-g, line 259: This test would be more interesting if it sends [BLOCK_CAP](https://github.com/tokio-rs/tokio/blob/b010b5ddafb316e5d540de1736f5f7bfde42ec39/tokio/src/sync/mpsc/mod.rs#L146) messages before closing. This will test the edge case where `block_tail` advances to a new block during `close()`.

tokio/src/sync/mpsc/list.rs
```diff
@@ -272,10 +268,24 @@ impl<T> Rx<T> {
...
+            return 0;
+        }
+        // There are messages present in the queue. However, it's possible that the last message is
+        // a fake "closed" message that we do not wish to count. To avoid counting it, we do not
+        // count the last message if the ready bit is unset.
+        //
+        // Note that it is also possible for the ready bit to be unset on a normal message, but
+        // this happens only if that message is currently being sent *right now* in parallel on
+        // another thread. That is okay because it is optional to count messages that are currently
+        // being sent.
```
martin-g, line 284: Is it possible another thread to advance the `block_tail` between line 271 (`tail_position.load()`) and line 285 (`tx.has_value()`) ?
If this happens `tx.has_value()` may enter an infinite loop calling `block.grow()` in `find_block()`.

Darksonn: The loop in `find_block` can only run finitely many times because each iteration advances to the next block, and it can only move to the next block `offset/32` times.

tokio/src/sync/tests/loom_mpsc.rs
Darksonn: Good call. I don't actually think the code was right as-is!

mattiapitossi (approved):

PR #8066, opened 2026-04-18
io: split read/write ticks in ScheduledIo

### LINKED ISSUE #8054: Livelock in `PollEvented::poll_read()` which sometimes does not transition into Pending state despite nothing to read
**Version**
1.52.0

**Platform**
Windows 11, 64-bit.

**Description**
Repro app can be found here: https://github.com/sandersaares/whatiswrongwithyourdog-tokio-io-unhappiness

Summary: using Tokio I/O from a non-Tokio thread (attached via `Handle::enter()`) can occasionally cause a temporary livelock between the custom thread and the Tokio worker that is pumping I/O events.

Repro instructions with a simple Hyper HTTP client-server scenario:

1. Clone https://github.com/sandersaares/whatiswrongwithyourdog-tokio-io-unhappiness.git
1. `cargo build --release`
1. `.\target\release\server.exe`
1. (In another terminal) `while ($true) { .\target\release\client.exe }`

# Expected behavior

Every client invocation is fast (sub-second):

```text
total time taken: 4.7094ms
total time taken: 4.4425ms
total time taken: 3.6746ms
total time taken: 4.7404ms
total time taken: 4.0319ms
```

# Actual behavior

Sometimes (probabilistic, very timing-sensitive) it is not fast:

```text
total time taken: 2.7875637s
WARNING: total time exceeded 1 second
total time taken: 8.2463ms
total time taken: 7.3296ms
total time taken: 11.1569628s
WARNING: total time exceeded 1 second
```

Furthermore, during the "delay" both the `main()` thread and the Tokio worker are busy and spending a lot of CPU time in the server process.

# Investigation so far

It seems to be some sort of livelock. The behavior is very timing sensitive and I have never managed to reproduce with a profiler attached but ad-hoc instrumentation reveals the following behavior during the anomalous request handling:

* The central anomaly is that `tokio::io::poll_evented::PollEvented::poll_read()` does not return for this entire duration, instead looping constantly in the server's `main()` thread.
* Every iteration of the inner loop in `poll_read()` gets a `WouldBlock` error.
* For each iteration `SockState::update()` calls `self.cancel()` because `self.pending_evts` does not have `POLL_SEND` but `self.user_evts` does have `POLL_SEND`.
* `SockState::feed_event()` fires both "cancelled" and "write" events in roughly the same count as `poll_read()` iterations.
* `set_readiness()` is called in roughly the same count as `poll_read()` iterations.
* `clear_readiness()` is called and results in a tick mismatch in roughly the same count as `poll_read()` iterations.

My interpretation (take with grain of salt):

* The socket is really in a state where there is no data to read (`WouldBlock` is correct).
* The socket is writable, so re-registering the AFD interest will always raise a `write` event in `feed_event()` to signal that the socket is ready to support writes.
* Every re-registration of the read interest immediately raises the "writable" event because the re-registration also includes `POLL_SEND`.
* This write-event-triggered `set_readiness()` (from the Tokio worker processing I/O) increments the tick and invalidates the `clear_readiness()` that the read logic (on the `main()` thread) wants to perform, so the next iteration of `poll_read()` thinks it is ready to read again when it really should have gotten its read readiness cleared and go into a `Pending` state.
* `poll_read()` loops until something jiggles the timing enough to let `clear_readiness()` execute on the main thread before the competing `set_readiness()` increments the tick on the Tokio worker.

This is my first time looking into Tokio I/O internals, so I may be misinterpreting some behavior. Happy to debug/instrument further if you can tell me key points in the code to measure - it reproduces fairly reliably for me.

CrossEyedCat:
Separate read_tick and write_tick so write-only mio notifications do not invalidate clear_readiness for the read path. Fixes livelock when the reactor advances the shared tick between poll_read_ready and clear_readiness.

Fixes #8054

### Motivation
ScheduledIo used a single readiness generation counter (tick). The I/O driver incremented it on every mio event via set_readiness(Tick::Set, …), including write-only notifications.

A task can observe read readiness at tick T, then perform a read that returns WouldBlock and call clear_readiness with that snapshot. If another thread handles the reactor and processes a writable-only event before the clear runs, the shared tick becomes T+1. clear_readiness uses Tick::Clear(T), sees a tick mismatch, and does not clear the read bits. Read readiness stays set, so poll_read_ready keeps returning ready while read keeps returning WouldBlock — a busy-loop / livelock.

This was reported on Windows (e.g. Hyper HTTP/2 repro in #8054), but the tick logic is platform-agnostic.

### Solution
Replace the single 15-bit tick with 7-bit read_tick and 7-bit write_tick in the packed AtomicUsize (16 + 7 + 7 + 1 bit total so the layout still fits in usize on 32-bit targets, including wasm32).

On each mio event, merge_readiness_from_driver advances read_tick only if the event has read-side bits (readable / read closed / error [+ priority on Linux]), and write_tick only if it has write-side bits (writable / write closed).

ReadyEvent carries both ticks; clear_readiness checks only the tick(s) that correspond to the bits being cleared (via read/write tick masks).

### DIFF
--- a/tokio/CHANGELOG.md
+++ b/tokio/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 1.52.2 (unreleased)
+
+## Fixed
+
+- io: split read/write ticks in the I/O driver so write-only `mio` notifications no longer invalidate read-side `clear_readiness` ([#8054])
+
 # 1.52.1 (April 16th, 2026)
 
 ## Fixed
@@ -7,6 +13,7 @@
 [#7757]: https://github.com/tokio-rs/tokio/pull/7757
 [#8056]: https://github.com/tokio-rs/tokio/pull/8056
 [#8057]: https://github.com/tokio-rs/tokio/pull/8057
+[#8054]: https://github.com/tokio-rs/tokio/issues/8054
 
 # 1.52.0 (April 14th, 2026)
 
--- a/tokio/CHANGELOG.md
+++ b/tokio/CHANGELOG.md
@@ -1,9 +1,3 @@
-# 1.52.2 (unreleased)
-
-## Fixed
-
-- io: split read/write ticks in the I/O driver so write-only `mio` notifications no longer invalidate read-side `clear_readiness` ([#8054])
-
 # 1.52.1 (April 16th, 2026)
 
 ## Fixed
@@ -13,7 +7,6 @@
 [#7757]: https://github.com/tokio-rs/tokio/pull/7757
 [#8056]: https://github.com/tokio-rs/tokio/pull/8056
 [#8057]: https://github.com/tokio-rs/tokio/pull/8057
-[#8054]: https://github.com/tokio-rs/tokio/issues/8054
 
 # 1.52.0 (April 14th, 2026)
 
--- a/tokio/src/runtime/io/driver.rs
+++ b/tokio/src/runtime/io/driver.rs
@@ -101,10 +101,7 @@ pub(super) enum Direction {
 
 pub(super) enum Tick {
     /// Clear readiness bits; only the tick(s) for directions present in the mask are checked.
-    Clear {
-        read: Option<u8>,
-        write: Option<u8>,
-    },
+    Clear { read: Option<u8>, write: Option<u8> },
 }
 
 const TOKEN_WAKEUP: mio::Token = mio::Token(0);
--- a/tokio/src/runtime/io/driver.rs
+++ b/tokio/src/runtime/io/driver.rs
@@ -72,7 +72,10 @@ pub(crate) struct Handle {
 
 #[derive(Debug)]
 pub(crate) struct ReadyEvent {
-    pub(super) tick: u8,
+    /// Generation counter for read-side readiness (readable, read closed, error, …).
+    pub(super) read_tick: u8,
+    /// Generation counter for write-side readiness (writable, write closed).
+    pub(super) write_tick: u8,
     pub(crate) ready: Ready,
     pub(super) is_shutdown: bool,
 }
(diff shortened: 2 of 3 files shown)


tokio/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 1.52.2 (unreleased)
```
mattiapitossi: I would say we don't usually update the changelog unless we decide to release just this fix

CrossEyedCat: Thanks - removed the unreleased changelog section from this PR.

tokio/src/runtime/io/driver.rs
```diff
@@ -72,7 +72,10 @@ pub(crate) struct Handle {
 
 #[derive(Debug)]
 pub(crate) struct ReadyEvent {
-    pub(super) tick: u8,
+    /// Generation counter for read-side readiness (readable, read closed, error, …).
+    pub(super) read_tick: u8,
+    /// Generation counter for write-side readiness (writable, write closed).
+    pub(super) write_tick: u8,
```
mattiapitossi, line 78: Is it possible to verify that this solves the problem that was reported in the issue?

Darksonn: If this solves the problem, then that is surprising to me. @sandersaares can you please confirm?

sandersaares: I confirm that with this branch I can no longer reproduce the problem I reported.

sandersaares: Interestingly, even for the "not hanging" iterations, the durations in my repro app went down:

image

Issue #8067, opened 2026-04-18
Re-land sharded spawn blocking queue


alex:
https://github.com/tokio-rs/tokio/pull/7757 landed a sharded spawn blocking queue to improve scalability, however this was reverted due to causing hangs (see: https://github.com/tokio-rs/tokio/issues/8056).

I'm pretty sure the bug that caused this was due to handling of idle threads and whether a thread could be woken vs. needed to be spawned.

In any event, my proposed path to re-landing is:

- [ ] [Land regression tests for the bug](https://github.com/tokio-rs/tokio/pull/8068)
- [ ] Land refactor to enable...
- [ ] Land sharded queue in a disabled-by-default configuration so users can opt into it
- [ ] After we've had time to collect feedback/testing, make the sharded queue the default behavior

philjb:
I know you have the oxide/omicron example too but I will share how I saw the bug manifest. 

Our test infrastructure started a influxdb server and hit the http api twice as fast as possible to trigger concurrent execution of a python plugin. 

Here's the `spawn_blocking` call:

https://github.com/influxdata/influxdb/blob/b66fe858e3758b45c615e76e53cfedc73ef0c1c2/influxdb3_processing_engine/src/plugins.rs#L654-L655

Each http api request creates a future wrapping `process_request`, stuffing it into a `FuturesUnordered` that is polled on a loop {tokio::select!{}}.  The test plugin I was using had a facility to see if another instance (peer) of the plugin was running, if so, it is concurrent execution, if not, no. 

I added debugging information like this

```
    async fn process_request( ...
...
        let submit_time = std::time::Instant::now();
        let result = tokio::task::spawn_blocking(move || {
            let thread = std::thread::current();
            debug!(
                trigger = %trigger_name_closure,
                probe_id = %probe_id_closure,
                queue_wait_us = submit_time.elapsed().as_micros() as u64,
                thread_id = ?thread.id(),
                thread_name = thread.name().unwrap_or("unnamed"),
                "spawn_blocking: closure entered on blocking thread"
            );
            execute_request_trigger(
...
```

The server was running with 2 threads for async work in the runtime and the default 512 limit for spawned threads. 
I've summarized the debug logging below. One can see that the http requests came in at nearly the same time (29.722119) and pushed the futures to the unordered set (29.722503 & 29.722685) but one spawn_blocking waited for the first to finish before execution and both executed on the same thread `78`. The `queue_wait_us` value being of most significance. 

```
[...]

PR #8068, opened 2026-04-18
introduce a regression test for #8056

alex:
refs #8067

### DIFF
--- a/tokio/tests/rt_spawn_blocking.rs
+++ b/tokio/tests/rt_spawn_blocking.rs
@@ -36,16 +36,36 @@ fn issue_8056_regression_test() {
     // Regression test for the lost-spawn race in the blocking pool introduced
     // by #7757. See https://github.com/tokio-rs/tokio/issues/8056 for complete
     // details.
-
-    // The bug was a race condition in task spawning when deciding whether to
-    // spawn a new blocking worker or to wake an idle one.
+    //
+    // The bug was a race condition in `Spawner::spawn_task` - after pushing a
+    // task to the queue, it chose between "wake an idle worker" and "spawn a
+    // new worker", by reading the `num_idle_threads` `AtomicUsize`. The
+    // `num_idle_threads` counter was incremented by a worker before it calls
+    // `wait_for_task`, and was deremented after `Inner::run` after
+    // `wait_for_task` had already returned a `Task`.
+    //
+    // In that window, the counter indicates a worker is idle even though it
+    // has already claimed a notification and is about to run a task. In the
+    // the omnicron test case that reproduced this, this task was long lived,
+    // which would essential result in "under spawning" workers.
+    //
+    // This test attempts to reproduce that scenario by doing the following on
+    // a fresh runtime
+    // - Two awaited `spawn_blocking` calls to get pool workers cycling through
+    //   idle → notified → busy transitions.
+    // - `make_writer` spawns two closures that park in
+    //   `mpsc::Receiver::blocking_recv` — long-lived blocking tasks that turn
+    //    a stranded spawn into a real deadlock instead of a transient stall.
+    // - Two more rounds of "create a writer and immediately drop it" churn the
+    //   worker pool while those persistent tasks are still blocked, and then a
+    //   second persistent writer is created — this `spawn_blocking` is most
+    //   likely to trigger the bug, because the prior churn has left a worker
+    //   mid-transition with a stale `num_idle_threads`.
+    // - Finally, we finish all the tasks -- if any of them never got pulled
+    //   from the `spawn_blocking` queue then it hits the timeout.
 
     // Run multiple times to make hitting the race condition very likely.
     for _ in 0..512 {
-        // Fresh runtime per iteration, matching `#[tokio::test]`
-        // semantics in omicron. On the buggy code, a freshly-created
-        // blocking pool growing its first few workers is apparently
-        // where the stale-read race fires most reliably.
         let rt = Builder::new_current_thread().enable_all().build().unwrap();
 
         let completed = rt.block_on(async {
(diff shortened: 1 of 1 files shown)


tokio/tests/rt_spawn_blocking.rs
```diff
@@ -0,0 +1,91 @@
...
+            tx.send(()).await.unwrap();
+        }
+        drop(senders);
+        while let Some(res) = handles.join_next().await {
+            res.unwrap();
+        }
+    }
+
+    // Regression test for the lost-spawn race in the blocking pool introduced
+    // by #7757. See Issue #8056 for more details.
```
mattiapitossi: Probably adding direct links to the issue would be more convenient here?

alex:
(Failures on integration tests look to be a GH failure, if someone can retry them.)

mattiapitossi:
> (Failures on integration tests look to be a GH failure, if someone can retry them.)

Lately gh jobs are pretty unstable, I will re-run them

alex:
Thank you!

tokio/tests/rt_spawn_blocking.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    // Regression test for the lost-spawn race in the blocking pool introduced
+    // by #7757. See https://github.com/tokio-rs/tokio/issues/8056 for complete
+    // details.
+
+    // The bug was a race condition in task spawning when deciding whether to
+    // spawn a new blocking worker or to wake an idle one.
+
+    // Run multiple times to make hitting the race condition very likely.
+    for _ in 0..512 {
```
hawkw, line 68: how long does this test end up taking to run?

tokio/tests/rt_spawn_blocking.rs
```diff
@@ -0,0 +1,92 @@
...
+        // blocking pool growing its first few workers is apparently
+        // where the stale-read race fires most reliably.
+        let rt = Builder::new_current_thread().enable_all().build().unwrap();
+
+        let completed = rt.block_on(async {
+            // We use a timeout to turn a hang into a reliable (and fast)
+            // failure. On working code this needs like a millisecond, so any
+            // timeout will do.
+            tokio::time::timeout(Duration::from_secs(1), async {
+                // Test case is derived from the failing test in omnicron:
```
hawkw: nit:
```suggestion
                // Test case is derived from the failing test in omicron:
```

hawkw: Also, if you would like to reference our test as the source for the repro, it would be nice if this comment linked to the test and explained what the behavior in the test that triggered the bug actually was :)

tokio/tests/rt_spawn_blocking.rs
```diff
@@ -0,0 +1,92 @@
...
+            res.unwrap();
+        }
+    }
+
+    // Regression test for the lost-spawn race in the blocking pool introduced
+    // by #7757. See https://github.com/tokio-rs/tokio/issues/8056 for complete
+    // details.
+
+    // The bug was a race condition in task spawning when deciding whether to
+    // spawn a new blocking worker or to wake an idle one.
```
hawkw: I appreciate the commentary explaining the bug we're trying to reproduce here, but...it feels like there's something missing. The comments here don't actually explain _how exactly the race condition manifests_, they seem to just sort of dance around it. It would be nice if there was a slightly more concrete explanation of where the race occurs.

tokio/tests/rt_spawn_blocking.rs
hawkw, line 68: Since this is a race condition that we have to run multiple iterations in order to catch, I wonder if this might make sense to try and reproduce using `loom` (perhaps in addition to a normal test). It would be nice if we could use the model checker to _deterministically_ simulate the interleaving that reproduces the bug, rather than just trying to run some code a whole bunch of times and _hope_ one of them hits the race. It's hard to say for sure without a better understanding of the bug, but I wonder if `loom`'s deadlock detection could catch this...

tokio/tests/rt_spawn_blocking.rs
```diff
@@ -0,0 +1,92 @@
...
+                        res.unwrap();
+                    }
+                }
+
+                quick().await;
+                quick().await;
+                let (second_senders, second_handles) = make_writer();
+
+                drive_writer(persistent_senders, persistent_handles).await;
+                drive_writer(second_senders, second_handles).await;
```
hawkw, line 99: It would be really nice to get a better overview of what, exactly, the test is doing. Can we add a comment providing a high level overview of the specific behavior we are trying to produce here? At a glance, it's hard to assess what, exactly, is necessary to trigger the bug.

tokio/tests/rt_spawn_blocking.rs
alex:
```
running 1 test
test issue_8056_regression_test ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.18s
```

(macbook m1 max)

tokio/tests/rt_spawn_blocking.rs
alex: Sure, I can add more detail on the nature of the bug itself.

tokio/tests/rt_spawn_blocking.rs
alex: I originally thought loom would be a good fit, but this ended up being a quicker path for me to get a reproducer. Is there an existing loom test you think would be a good model for this type of test?

tokio/tests/rt_spawn_blocking.rs
alex: yup, will do!

tokio/tests/rt_spawn_blocking.rs
hawkw: cool, that's not bad!

alex:
Ok, took me a while, but there's now a description of the bug + how this test attempts to replicate it. Let me know if they make sense.

Darksonn:
Since this is a racy test that's not guaranteed to catch the issue every time, can you please try pushing the queue sharding code to see if the error triggers in CI as well. This will verify that the perf characteristics of the github instances do not adjust timings so that it does not occur.

alex:
Ok, cherry-picked the sharded queue on this. Assuming the failures look good, I can then rebase it back off.

Darksonn:
Looks good, thanks!

alex:
Ok, rebased that commit out

Darksonn (approved):
Thanks.

Issue #8065, opened 2026-04-18
Performance regression on high-QPS short-handler workloads since 1.51.0


kushudai:
## Summary

Since Tokio `1.51.0` ([#7431](https://github.com/tokio-rs/tokio/pull/7431), "runtime: steal tasks from the LIFO slot"), `schedule_local` on the multi-thread scheduler unconditionally calls `notify_parked_local()` whenever a parked worker exists, even when the task was just placed into the current worker's own empty LIFO slot. In `1.50.0` that case was a silent fast path: the worker kept the task and did not wake another worker.

On a high-QPS lookup service (microsecond handlers, no request-path I/O) this costs about +8.5% aggregate CPU going from `1.50.x` to `1.52.1` on an otherwise identical binary. `Builder::disable_lifo_slot()` does not appear to recover performance.  Currently, we suspect this is because #7431 removed the `should_notify` short-circuit from *both* arms of `schedule_local`.

## Environment

x86_64 Linux + Docker. Multi-thread runtime, 4 workers, `--cfg tokio_unstable` for `RuntimeMetrics`. Comparison is between two builds of the same service binary, one on `tokio = 1.50.x` (healthy), one on `tokio = 1.52.1` (regressed). New build call `.disable_lifo_slot()` on the `Builder` (we tried without it as well, same regression - this was an attempt to alleviate the problem). Tokio runtime metrics refreshed in-process every 5s, scraped every 30s via prometheus. All numbers below are per-worker, same datacenter, steady-state.

## Root cause: `schedule_local`

`schedule_local` in `tokio/src/runtime/scheduler/multi_thread/worker.rs`: compare [1.50.0 L1301–L1325](https://github.com/tokio-rs/tokio/blob/tokio-1.50.0/tokio/src/runtime/scheduler/multi_thread/worker.rs#L1301-L1325) against [1.51.0 L1301–L1319](https://github.com/tokio-rs/tokio/blob/tokio-1.51.0/tokio/src/runtime/scheduler/multi_thread/worker.rs#L1301-L1319). The `should_notify` local is gone; the "empty LIFO slot - stay quiet" fast path is removed in both arms, and the final guard collapses from `should_notify && core.park.is_some()` to just `core.park.is_some()` (~always true on a busy runtime). The PR's stated goal (making the LIFO slot stealable via the new `AtomicNotified` cell) is fine on its own; the regression is suspected to be a side-effect of how the surrounding control flow was refactored.

[...]

PR #8069, opened 2026-04-18
rt: don't notify when pushing task to empty LIFO slot

hawkw:
See #8065

This change un-does the increase in notifications described in #8065. Running the repro from https://github.com/tokio-rs/tokio/issues/8065#issuecomment-4272562127, I see voluntary context switches back to 1.50.0 counts (note the `cargo build` output saying that this is a local Tokio checkout):
```console
eliza@hekate ~/Code/tokio-8065 $ cargo build --release
   Compiling tokio v1.52.1 (/home/eliza/Code/tokio/tokio)
   Compiling tokio-8065 v0.1.0 (/home/eliza/Code/tokio-8065)
    Finished `release` profile [optimized] target(s) in 2.76s

eliza@hekate ~/Code/tokio-8065 $ nix run nixpkgs#time -- -v target/release/tokio-8065
	Command being timed: "target/release/tokio-8065"
	User time (seconds): 0.01
	System time (seconds): 0.00
	Percent of CPU this job got: 0%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:10.00
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 1984
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 146
	Voluntary context switches: 6563
	Involuntary context switches: 1
	Swaps: 0
	File system inputs: 0
	File system outputs: 0
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
``

### DIFF
--- a/tokio/src/runtime/scheduler/multi_thread/idle.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/idle.rs
@@ -150,6 +150,12 @@ impl Idle {
         lock.idle.sleepers.contains(&worker_id)
     }
 
+    /// Returns `true` if all other workers are currently parked.
+    pub(super) fn all_parked(&self) -> bool {
+        let state = State(self.state.fetch_add(0, SeqCst));
+        state.num_unparked() <= 1
+    }
+
     fn notify_should_wakeup(&self) -> bool {
         let state = State(self.state.fetch_add(0, SeqCst));
         state.num_searching() == 0 && state.num_unparked() < self.num_workers
--- a/tokio/src/runtime/scheduler/multi_thread/idle.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/idle.rs
@@ -23,8 +23,14 @@ pub(super) struct Synced {
     sleepers: Vec<usize>,
 }
 
-const UNPARK_SHIFT: usize = 16;
-const UNPARK_MASK: usize = !SEARCH_MASK;
+pub(super) struct TransitionToParked {
+    pub(super) is_last_searcher: bool,
+    pub(super) any_lifo: bool,
+}
+
+const UNPARK_SHIFT: usize = (usize::BITS as usize / 2) - 2;
+const ANY_LIFO: usize = 1 << (usize::BITS - 1);
+const UNPARK_MASK: usize = !(SEARCH_MASK | ANY_LIFO);
 const SEARCH_MASK: usize = (1 << UNPARK_SHIFT) - 1;
 
 #[derive(Copy, Clone)]
@@ -32,6 +38,10 @@ struct State(usize);
 
 impl Idle {
     pub(super) fn new(num_workers: usize) -> (Idle, Synced) {
+        assert!(
+            num_workers <= UNPARK_MASK,
+            "{num_workers} is too many workers (max is {UNPARK_MASK})"
+        );
         let init = State::new(num_workers);
 
         let idle = Idle {
@@ -88,7 +98,7 @@ impl Idle {
         shared: &Shared,
         worker: usize,
         is_searching: bool,
-    ) -> bool {
+    ) -> TransitionToParked {
         // Acquire the lock
         let mut lock = shared.synced.lock();
 
@@ -144,16 +154,23 @@ impl Idle {
         false
     }
 
+    pub(super) fn put_lifo(&self) -> bool {
+        State(self.state.fetch_or(ANY_LIFO, SeqCst)).any_lifo()
+    }
+
+    pub(super) fn clear_lifo(&self) {
+        self.state.fetch_and(!ANY_LIFO, SeqCst);
+    }
+
+    pub(super) fn should_attempt_lifo_steal(&self) -> bool {
+        let state = State(self.state.fetch_add(0, SeqCst));
+        state.any_lifo()
+    }
+
     /// Returns `true` if `worker_id` is contained in the sleep set.
     pub(super) fn is_parked(&self, shared: &Shared, worker_id: usize) -> bool {
         let lock = shared.synced.lock();
         lock.idle.sleepers.contains(&worker_id)
-    }
-
-    /// Returns `true` if all other workers are currently parked.
-    pub(super) fn all_parked(&self) -> bool {
-        let state = State(self.state.fetch_add(0, SeqCst));
-        state.num_unparked() <= 1
     }
 
     fn notify_should_wakeup(&self) -> bool {
@@ -192,7 +209,7 @@ impl State {
     /// Track a sleeping worker
     ///
     /// Returns `true` if this is the final searching worker.
-    fn dec_num_unparked(cell: &AtomicUsize, is_searching: bool) -> bool {
+    fn dec_num_unparked(cell: &AtomicUsize, is_searching: bool) -> TransitionToParked {
         let mut dec = 1 << UNPARK_SHIFT;
 
         if is_searching {
@@ -200,7 +217,11 @@ impl State {
         }
 
         let prev = State(cell.fetch_sub(dec, SeqCst));
-        is_searching && prev.num_searching() == 1
+        let is_last_searcher = is_searching && prev.num_searching() == 1;
+        TransitionToParked {
+            is_last_searcher,
+            any_lifo: prev.any_lifo(),
+        }
     }
 
     /// Number of workers currently searching
@@ -211,6 +232,10 @@ impl State {
     /// Number of workers currently unparked
     fn num_unparked(self) -> usize {
         (self.0 & UNPARK_MASK) >> UNPARK_SHIFT
+    }
+
+    fn any_lifo(self) -> bool {
+        self.0 & ANY_LIFO == ANY_LIFO
     }
 }
 
@@ -243,4 +268,12 @@ fn test_state() {
     let state = State::new(10);
     assert_eq!(10, state.num_unparked());
     assert_eq!(0, state.num_searching());
+}
+
+#[test]
+fn masks() {
+    println!("UNPARK_SHIFT = {UNPARK_SHIFT}");
+    println!("UNPARK_MASK  = {UNPARK_MASK:064b}");
+    println!("SEARCH_MASK  = {SEARCH_MASK:064b}");
+    println!("ANY_LIFO     = {ANY_LIFO:064b}");
 }
--- a/tokio/src/runtime/scheduler/multi_thread/queue.rs
+++ b/tokio/src/runtime/scheduler/multi_thread/queue.rs
@@ -440,6 +440,10 @@ impl<T> Steal<T> {
         self.len() == 0
     }
 
+    pub(crate) fn has_lifo(&self) -> bool {
+        self.0.lifo.is_some()
+    }
+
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by hawkw** [comment]
However, I'm not 100% sure that this change is correct, given https://github.com/tokio-rs/tokio/issues/8065#issuecomment-4272562127 (which is what originally prompted me to _remove_ the `should_notify` boolean and only notify a parked worker if the LIFO slot was not empty. In that comment, @Darksonn points out that the previous intent behind having `should_notify` was that we should be _always_ notifying a parked worker in any case where a task is available to steal, and that we only skipped doing so when pushing to an empty LIFO slot because the LIFO task was not stealable. Now, we are introducing a third state, in which a task is available to steal, but we haven't notified another worker. I'm not sure if this is correct off the top of my head. 

I worry that it would be possible to re-introduce the deadlock that can occur when a task is placed in the LIFO slot and the worker thread is immediately blocked by the task that notified it, because the blocked worker won't have any further reasons to unpark another worker while it is blocked. However, the `rt_threaded::lifo_stealable` test I added to reproduce such deadlocks does still pass after this change. I think that's because the test involves another task whose job is to keep the non-blocked worker from parking:
https://github.com/tokio-rs/tokio/blob/b010b5ddafb316e5d540de1736f5f7bfde42ec39/tokio/tests/rt_threaded.rs#L733-L743

I wonder if removing that task from the test makes it pass on 1.52.1 but hang on this branch. Let's find out..

**Comment by hawkw** [comment]
> I wonder if removing that task from the test makes it pass on 1.52.1 but hang on this branch. Let's find out..

Yup. If we remove `churn` from the test, like this:

```patch
diff --git a/tokio/tests/rt_threaded.rs b/tokio/tests/rt_threaded.rs
index 133ba4c3..ad00abaa 100644
--- a/tokio/tests/rt_threaded.rs
+++ b/tokio/tests/rt_threaded.rs
@@ -730,18 +730,6 @@ fn lifo_stealable() {
         .unwrap();

     rt.block_on(async {
-        // Keep the runtime busy so that the workers that might steal the
-        // blocked task don't all park themselves forever.
-        //
-        // Since this task will always be woken by whichever worker is holding
-        // the time driver, rather than a worker that's executing tasks, it
-        // shouldn't ever kick the victim task out of its worker's LIFO slot.
-        let churn = tokio::spawn(async move {
-            loop {
-                tokio::time::sleep(Duration::from_millis(4)).await;
-            }
-        });
-
         let victim_task_joined = tokio::spawn(async move {
             println!("[victim] task started");
             task_started_tx.send(()).unwrap();
@@ -791,7 +779,6 @@ fn lifo_stealable() {
         // Before possibly panicking, make sure that we wake up the blocker task
         // so that it doesn't stop the runtime from shutting down.
         block_thread_tx.send(()).unwrap();
-        churn.abort();
         result
             .expect("task in LIFO slot should complete within 30 seconds")
             .expect("task in LIFO slot should not panic");
```

...then the test still passes on `master` (b010b5ddafb316e5d540de1736f5f7bfde42ec39), but fails on this branch (b010b5ddafb316e5d540de1736f5f7bfde42ec39): 

```console
eliza@hekate ~/Code/tokio $ git co master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.

eliza@hekate ~/Code/tokio $ git rev-parse HEAD
b010b5ddafb316e5d540de1736f5f7bfde42ec39

eliza@hekate ~/Code/tokio $ cargo nextest run lifo_stealable
   Compiling examples v0.0.0 (/home/eliza/Code/tokio/examples)
   Compiling tokio-util v0.7.18 (/home/eliza/Code/tokio/tokio-util)
   Compiling tokio v1.52.1 (/home/eliza/Code/tokio/tokio)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 2.62s
────────────
 Nextest run ID 004702f1-2de3-42e8-b18c-69606f942b2f with nextest profile: default
    Starting 1 test across 235 binaries (1717 tests skipped)
        PASS [   0.003s] tokio::rt_threaded lifo_stealable
────────────
     Summary [   0.005s] 1 test run: 1 passed, 1717 skipped

eliza@hekate ~/Code/tokio $ git checkout eliza/should-notify-again
M	tokio/tests/rt_threaded.rs
Switched to branch 'eliza/should-notify-again'
Your branch is up to date with 'origin/eliza/should-notify-again'.

eliza@hekate ~/Code/tokio $ git rev-parse HEAD
bc62e2d94cb84d556af7d6a707405baa1058e832

eliza@hekate ~/Code/tokio $ cargo nextest run lifo_stealable
   Compiling tokio v1.52.1 (/home/eliza/Code/tokio/tokio)
   Compiling tokio-stream v0.1.18 (/home/eliza/Code/tokio/tokio-stream)
   Compiling tokio-util v0.7.18 (/home/eliza/Code/tokio/tokio-util)
   Compiling tests-integration v0.1.0 (/home/eliza/Code/tokio/tests-integration)
   Compiling tokio-test v0.4.5 (/home/eliza/Code/tokio/tokio-test)
   Compiling tokio-macros v2.7.0 (/home/eliza/Code/tokio/tokio-macros)
   Compiling stress-test v0.1.0 (/home/eliza/Code/tokio/stress-test)
   Compiling examples v0.0.0 (/home/eliza/Code/tokio/examples)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 3.77s
────────────
 Nextest run ID 67fc6e9f-99cd-48d7-bddd-582cce200425 with nextest profile: default
    Starting 1 test across 235 binaries (1717 tests skipped)
        FAIL [  30.020s] tokio::rt_threaded lifo_stealable
  stdout ───

    running 1 test
    [victim] task started
    [victim] task waiting for wakeup...
    [main] victim slot task start acked!
    [main] blocker task spawned
    [blocker] sending wakeup
    [blocker] blocking the worker thread...
    [main] result: Err(Elapsed(()))
    [blocker] done
    [victim] task running after wakeup
    test lifo_stealable ... FAILED

    failures:

    failures:
        lifo_stealable

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 28 filtered out; finished in 30.02s

  stderr ───

    thread 'lifo_stealable' (3775832) panicked at tokio/tests/rt_threaded.rs:783:14:
    task in LIFO slot should complete within 30 seconds: Elapsed(())
    stack backtrace:
       0: __rustc::rust_begin_unwind
                 at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/std/src/panicking.rs:689:5
       1: core::panicking::panic_fmt
                 at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/panicking.rs:80:14
       2: core::result::unwrap_failed
                 at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/result.rs:1867:5
       3: core::result::Result<T,E>::expect
                 at /home/eliza/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:1185:23
       4: rt_threaded::lifo_stealable::{{closure}}
                 at ./tests/rt_threaded.rs:783:14
       5: tokio::runtime::park::CachedParkThread::block_on::{{closure}}
                 at ./src/runtime/park.rs:284:71
       6: tokio::task::coop::with_budget
                 at ./src/task/coop/mod.rs:167:5
       7: tokio::task::coop::budget
                 at ./src/task/coop/mod.rs:133:5
       8: tokio::runtime::park::CachedParkThread::block_on
                 at ./src/runtime/park.rs:284:31
       9: tokio::runtime::context::blocking::BlockingRegionGuard::block_on
                 at ./src/runtime/context/blocking.rs:66:14
      10: tokio::runtime::scheduler::multi_thread::MultiThread::block_on::{{closure}}
                 at ./src/runtime/scheduler/multi_thread/mod.rs:92:22
      11: tokio::runtime::context::runtime::enter_runtime
                 at ./src/runtime/context/runtime.rs:65:16
      12: tokio::runtime::scheduler::multi_thread::MultiThread::block_on
                 at ./src/runtime/scheduler/multi_thread/mod.rs:91:9
      13: tokio::runtime::runtime::Runtime::block_on_inner
                 at ./src/runtime/runtime.rs:373:50
      14: tokio::runtime::runtime::Runtime::block_on
                 at ./src/runtime/runtime.rs:345:18
      15: rt_threaded::lifo_stealable
                 at ./tests/rt_threaded.rs:732:8
      16: rt_threaded::lifo_stealable::{{closure}}
                 at ./tests/rt_threaded.rs:701:20
      17: core::ops::function::FnOnce::call_once
                 at /home/eliza/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
      18: <fn() -> core::result::Result<(), alloc::string::String> as core::ops::function::FnOnce<()>>::call_once
                 at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/ops/function.rs:250:5

  Cancelling due to test failure:
────────────
     Summary [  30.023s] 1 test run: 0 passed, 1 failed, 1717 skipped
        FAIL [  30.020s] tokio::rt_threaded lifo_stealable
error: test run failed

eliza@hekate ~/Code/tokio $ 
```

Unfortunately, I think that this means we should *not* move forwards with putting back a `should_notify` boolean to skip notifying another worker when only the LIFO slot is occupied, because that results in us failing to prevent the deadlock in cases where there isn't another very active task to keep other workers from parking. Instead, I think we should remove `churn()` from the test so that it doesn't rely on another task stopping workers from parking.

Sadly, this means we have to accept the performance impact of additional notifications as part of the cost of preventing deadlocks here. Personally, I feel that this is appropriate, because I don't think we should consider a performance decrease from code that behaved incorrectly to be a "regression". 

In the long term, I think we might be able to reduce the amount of cross-thread notifications here by changing the logic for notifying a worker when pushing tasks to the queue so that it is less eager about  waking workers in general. At present we _always_ notify another worker if there are no workers currently in searching and _any_ workers are parked: https://github.com/tokio-rs/tokio/blob/b010b5ddafb316e5d540de1736f5f7bfde42ec39/tokio/src/runtime/scheduler/multi_thread/idle.rs#L153

We might consider changing the "lifo slot only" case to notify only if all other workers are parked, or something, so that we will notify here if it is necessary to prevent a deadlock, but not if there are no other currently searching workers. But I think we'll want to think carefully about that.

@Darksonn, what do you think?

Darksonn:
I think it is basically correct that you must issue these wakeups to avoid issues where a blocking task prevents execution of the task in the lifo slot.

We could perhaps consider entirely different solutions where a second thread wakes up periodically to check if the running worker thread is making progress. Of course that's a bigger runtime change.

hawkw:
> We could perhaps consider entirely different solutions where a second thread wakes up periodically to check if the running worker thread is making progress. Of course that's a bigger runtime change.

That is more or less what I was trying to implement here, though it's going to take some additional work.

PR #8075, opened 2026-04-20
Notify receivers in mpsc channel `OwnedPermit::release()` method

Darksonn:
The `OwnedPermit::release()` method is missing a check for whether the receiver should be woken up. This results in lost wakeups when combined with `Receiver::close()`.

This bug was discovered by asking Gemini to look for bugs in the `release` method of `tokio/src/sync/mpsc/bounded.rs`.

### DIFF
--- a/tokio/tests/sync_mpsc.rs
+++ b/tokio/tests/sync_mpsc.rs
@@ -788,6 +788,91 @@ async fn drop_permit_iterator_releases_permits() {
     }
 }
 
+#[test]
+fn dropping_last_permit_closes_channel() {
+    let (tx, mut rx) = mpsc::channel::<()>(100);
+
+    let permit = tx.try_reserve().unwrap();
+    rx.close();
+
+    let mut recv = tokio_test::task::spawn(rx.recv());
+    assert_pending!(recv.poll());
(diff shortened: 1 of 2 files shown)


Darksonn:
This probably requires a backport, so it should not be merged until base branch is updated.

tokio/tests/sync_mpsc.rs
```diff
@@ -788,6 +788,91 @@ async fn drop_permit_iterator_releases_permits() {
     }
 }
 
+#[test]
+fn dropping_last_permit_closes_channel() {
+    let (tx, mut rx) = mpsc::channel::<()>(100);
+
+    let permit = tx.try_reserve().unwrap();
+    rx.close();
```
martin-g, line 796: [tx.is_closed()](https://github.com/tokio-rs/tokio/blob/d1db1a93045d3be4e040b63a4654b70cd27fee2e/tokio/src/sync/mpsc/bounded.rs#L1068-L1070) returns `true` here.
The test name `dropping_last_permit_closes_channel` suggests that the channel is still open until line 800.

martin-g (approved):

PR #8078, opened 2026-04-20
Avoid illegal state in `FastRand`

Darksonn:
Since `tokio/src/util/rand/rt.rs` calls `from_pair()` with two random numbers, it is possible that `rt.rs` will create an rng with seed zero. However, the seed zero is not allowed because the Xorshift64+ algorithm produces zeroes indefinitely for this seed. The `from_seed()` method was also adjusted to check both integers for zero instead of just `two`.

This bug was discovered by asking Gemini to look for bugs in the `from_pair` method of `tokio/src/util/rand.rs`.

### DIFF
--- a/tokio/src/util/rand.rs
+++ b/tokio/src/util/rand.rs
@@ -39,7 +39,7 @@ impl RngSeed {
 
     fn from_u64(seed: u64) -> Self {
         let one = (seed >> 32) as u32;
-        let mut two = seed as u32;
+        let two = seed as u32;
 
         Self::from_pair(one, two)
     }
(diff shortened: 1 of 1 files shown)


ADD-SP (approved):
Is it possible to write a unit test using a fixed seed?

Darksonn:
I don't think a backport is necessary for this, since the illegal state only has 1/2^64 probability of happening. And since this does change the random numbers generated by some seeds, backporting to patch release seems better to avoid.

martin-g (approved):

PR #8080, opened 2026-04-21
Implemented io-uring `Op<Statx>` and applied to `read_uring` and `fs::try_exists`

asder8215:
## Motivation

This builds up on [vrtgs PR](https://github.com/tokio-rs/tokio/pull/7791) in their attempt of implementing `Statx` for io-uring and apply it to `fs::try_exists`. I used `libc` instead of `linux-raw-sys` for the `Statx` implementation, tried to preserve more of how the original code (prior to vrtgs' changes) looked in certain files. There are some inlining done for the metadata functions for `Statx` because I thought the body of the code in these functions were pretty small and didn't seem to involve generics (let me know if I should remove that). Some other review comments mentioned in review of vrtgs' PR is addressed such as safety comments, inlining for the `try_exists_uring`, avoiding using `CString` literals, etc.

To note, this is my first time contributing to Tokio, so let me know if there's anything that I'm doing incorrectly or if I missed something as per procedure here. Let me know if I should also squash some commits as well.

ccing @Daksh14 since I got to working on this PR per him sharing vrtgs' PR with me.

### DIFF
--- a/tokio/src/fs/open_options.rs
+++ b/tokio/src/fs/open_options.rs
@@ -6,7 +6,6 @@ use std::path::Path;
 cfg_io_uring! {
     mod uring_open_options;
     pub(crate) use uring_open_options::UringOpenOptions;
-    use crate::runtime::driver::op::Op;
 }
 
 #[cfg(test)]
@@ -518,8 +517,12 @@ impl OpenOptions {
     /// [`Other`]: std::io::ErrorKind::Other
     /// [`PermissionDenied`]: std::io::ErrorKind::PermissionDenied
     pub async fn open(&self, path: impl AsRef<Path>) -> io::Result<File> {
+        self.open_inner(path.as_ref()).await
+    }
+
+    async fn open_inner(&self, path: &Path) -> io::Result<File> {
         match &self.inner {
-            Kind::Std(opts) => Self::std_open(opts, path).await,
+            Kind::Std(opts) => Self::std_open(opts.clone(), path).await,
             #[cfg(all(
                 tokio_unstable,
                 feature = "io-uring",
@@ -528,6 +531,11 @@ impl OpenOptions {
                 target_os = "linux"
             ))]
             Kind::Uring(opts) => {
+                #[cfg(test)]
+                use super::mocks::MockFile as StdFile;
+                #[cfg(not(test))]
+                use std::fs::File as StdFile;
+
                 let handle = crate::runtime::Handle::current();
                 let driver_handle = handle.inner.driver().io();
 
@@ -538,18 +546,16 @@ impl OpenOptions {
                     Op::open(path.as_ref(), opts)?.await
                 } else {
                     let opts = opts.clone().into();
-                    Self::std_open(&opts, path).await
+                    Self::std_open(opts, path).await
                 }
             }
         }
     }
 
-    async fn std_open(opts: &StdOpenOptions, path: impl AsRef<Path>) -> io::Result<File> {
-        let path = path.as_ref().to_owned();
-        let opts = opts.clone();
-
-        let std = asyncify(move || opts.open(path)).await?;
-        Ok(File::from_std(std))
+    async fn std_open(opts: StdOpenOptions, path: &Path) -> io::Result<File> {
+        let path = path.to_owned();
+        let std = asyncify(move || opts.open(path).map(File::from_std)).await?;
+        Ok(std)
     }
 
     #[cfg(windows)]
--- a/tokio/src/fs/open_options.rs
+++ b/tokio/src/fs/open_options.rs
@@ -6,6 +6,7 @@ use std::path::Path;
 cfg_io_uring! {
     mod uring_open_options;
     pub(crate) use uring_open_options::UringOpenOptions;
+    use crate::runtime::driver::op::Op;
 }
 
 #[cfg(test)]
@@ -522,7 +523,7 @@ impl OpenOptions {
 
     async fn open_inner(&self, path: &Path) -> io::Result<File> {
         match &self.inner {
-            Kind::Std(opts) => Self::std_open(opts.clone(), path).await,
+            Kind::Std(opts) => Self::std_open(opts, path).await,
             #[cfg(all(
                 tokio_unstable,
                 feature = "io-uring",
@@ -531,11 +532,6 @@ impl OpenOptions {
                 target_os = "linux"
             ))]
             Kind::Uring(opts) => {
-                #[cfg(test)]
-                use super::mocks::MockFile as StdFile;
-                #[cfg(not(test))]
-                use std::fs::File as StdFile;
-
                 let handle = crate::runtime::Handle::current();
                 let driver_handle = handle.inner.driver().io();
 
@@ -543,19 +539,20 @@ impl OpenOptions {
                     .check_and_init(io_uring::opcode::OpenAt::CODE)
                     .await?
                 {
-                    Op::open(path.as_ref(), opts)?.await
+                    Op::open(path, opts)?.await
                 } else {
                     let opts = opts.clone().into();
-                    Self::std_open(opts, path).await
+                    Self::std_open(&opts, path).await
                 }
             }
         }
     }
 
-    async fn std_open(opts: StdOpenOptions, path: &Path) -> io::Result<File> {
+    async fn std_open(opts: &StdOpenOptions, path: &Path) -> io::Result<File> {
         let path = path.to_owned();
-        let std = asyncify(move || opts.open(path).map(File::from_std)).await?;
-        Ok(std)
+        let opts = opts.clone();
+
+        Ok(asyncify(move || opts.open(path)).await?.into())
     }
 
     #[cfg(windows)]
--- a/tokio/src/fs/read.rs
+++ b/tokio/src/fs/read.rs
@@ -68,7 +68,10 @@ pub async fn read(path: impl AsRef<Path>) -> io::Result<Vec<u8>> {
 
         let handle = crate::runtime::Handle::current();
         let driver_handle = handle.inner.driver().io();
-        if driver_handle.check_and_init()? {
+        if driver_handle
+            .check_and_init(io_uring::opcode::Read::CODE)
+            .await?
+        {
             return read_uring(path).await;
         }
     }
--- a/tokio/src/fs/read.rs
+++ b/tokio/src/fs/read.rs
@@ -54,7 +54,7 @@ use std::{io, path::Path};
 /// }
 /// ```
 pub async fn read(path: impl AsRef<Path>) -> io::Result<Vec<u8>> {
-    let path = path.as_ref().to_owned();
+    let path = path.as_ref();
 
     #[cfg(all(
         tokio_unstable,
@@ -68,13 +68,15 @@ pub async fn read(path: impl AsRef<Path>) -> io::Result<Vec<u8>> {
 
         let handle = crate::runtime::Handle::current();
         let driver_handle = handle.inner.driver().io();
-        if driver_handle
-            .check_and_init(io_uring::opcode::Read::CODE)
-            .await?
-        {
-            return read_uring(&path).await;
+        if driver_handle.check_and_init()? {
+            return read_uring(path).await;
         }
     }
 
+    read_spawn_blocking(path).await
+}
+
+async fn read_spawn_blocking(path: &Path) -> io::Result<Vec<u8>> {
+    let path = path.to_owned();
     asyncify(move || std::fs::read(path)).await
 }
--- a/tokio/src/fs/read.rs
+++ b/tokio/src/fs/read.rs
@@ -61,7 +61,16 @@ pub async fn read(path: impl AsRef<Path>) -> io::Result<Vec<u8>> {
         feature = "io-uring",
         feature = "rt",
         feature = "fs",
-        target_os = "linux"
+        // libc::statx is only supported on these platforms
+        // FIXME: Add musl target env when our minimum supported
+        // rust version is 1.93. To clarify, statx support is
+        // introduced to musl in 1.25 as mentioned officially here:
+        // https://musl.libc.org/releases.html.
+        // However, rustup target_env building for *-linux-musl
+        // uses 1.25 musl on all *-linux-musl platforms starting
+        // in 1.93 stable rust version.
+        // https://blog.rust-lang.org/2025/12/05/Updating-musl-1.2.5/
+        any(target_env = "gnu", target_os = "android")
     ))]
     {
         use crate::fs::read_uring;
--- a/tokio/src/fs/read_uring.rs
+++ b/tokio/src/fs/read_uring.rs
@@ -1,4 +1,4 @@
-use crate::fs::OpenOptions;
+use crate::fs::UringOpenOptions;
 use crate::runtime::driver::op::Op;
 
 use std::io;
@@ -17,15 +17,13 @@ const PROBE_SIZE_U32: u32 = PROBE_SIZE as u32;
 const MAX_READ_SIZE: usize = 64 * 1024 * 1024;
 
 pub(crate) async fn read_uring(path: &Path) -> io::Result<Vec<u8>> {
-    let file = OpenOptions::new().read(true).open(path).await?;
+    let fd = UringOpenOptions::new().read(true).open(path).await?;
 
-    // TODO: use io uring in the future to obtain metadata
-    let size_hint: Option<usize> = file.metadata().await.map(|m| m.len() as usize).ok();
+    let (size_hint, fd) = Op::metadata_fd(fd).await;
 
-    let fd: OwnedFd = file
-        .try_into_std()
-        .expect("unexpected in-flight operation detected")
-        .into();
+    let size_hint: Option<usize> = size_hint
+        .ok()
+        .map(|m| usize::try_from(m.len()).unwrap_or(usize::MAX));
 
     let mut buf = Vec::new();
 
--- a/tokio/src/fs/read_uring.rs
+++ b/tokio/src/fs/read_uring.rs
@@ -1,4 +1,4 @@
-use crate::fs::UringOpenOptions;
+use crate::fs::OpenOptions;
 use crate::runtime::driver::op::Op;
 
 use std::io;
@@ -17,13 +17,17 @@ const PROBE_SIZE_U32: u32 = PROBE_SIZE as u32;
 const MAX_READ_SIZE: usize = 64 * 1024 * 1024;
 
 pub(crate) async fn read_uring(path: &Path) -> io::Result<Vec<u8>> {
-    let fd = UringOpenOptions::new().read(true).open(path).await?;
+    let file = OpenOptions::new().read(true).open(path).await?;
 
-    let (size_hint, fd) = Op::metadata_fd(fd).await;
+    let size_hint = Op::file_metadata(&file)?
+        .await
+        .map(|m| m.len() as usize)
+        .ok();
 
-    let size_hint: Option<usize> = size_hint
-        .ok()
-        .map(|m| usize::try_from(m.len()).unwrap_or(usize::MAX));
+    let fd: OwnedFd = file
+        .try_into_std()
+        .expect("unexpected in-flight operation detected")
+        .into();
 
     let mut buf = Vec::new();
 
--- a/tokio/src/fs/read_uring.rs
+++ b/tokio/src/fs/read_uring.rs
@@ -1,11 +1,9 @@
-#[allow(unused)] // FIXME: remove when MSRV is 1.93 (due to statx on 1.25 musl)
 use crate::fs::OpenOptions;
 use crate::runtime::driver::op::Op;
 
 use std::io;
 use std::io::ErrorKind;
 use std::os::fd::OwnedFd;
-#[allow(unused)] // FIXME: remove when MSRV is 1.93 (due to statx on 1.25 musl)
 use std::path::Path;
 
 // this algorithm is inspired from rust std lib version 1.90.0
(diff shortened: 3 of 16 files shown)


tokio/src/io/uring/statx.rs
```diff
@@ -0,0 +1,152 @@
...
+        Ok(unsafe { Op::new(statx_op, Statx { path, buffer }) })
+    }
+
+    /// Retrieves the metadata information of the given path, following symlinks
+    /// if the path provided points to a symlink location.
+    #[inline]
+    pub(crate) fn metadata(path: &Path) -> io::Result<Op<Statx>> {
+        Op::statx(
+            path,
+            (libc::AT_STATX_SYNC_AS_STAT | libc::AT_SYMLINK_FOLLOW) as u32,
```
asder8215, line 83: out of curiosity, is it necessary to or the symlink flags with `libc::AT_STATX_SYNC_AS_STAT`? 
`libc::AT_STATX_SYNC_AS_STAT` = 0 from what I saw in `libc`, so this oring doesn't do anything.

asder8215:
Any reason why `libc` statx constants are not found for Linux 7.0?

image

tokio/src/fs/read.rs
```diff
@@ -69,12 +69,17 @@ pub async fn read(path: impl AsRef<Path>) -> io::Result<Vec<u8>> {
         let handle = crate::runtime::Handle::current();
         let driver_handle = handle.inner.driver().io();
         if driver_handle
-            .check_and_init(io_uring::opcode::Read::CODE)
+            .check_and_init(io_uring::opcode::OpenAt::CODE)
```
Daksh14: Interesting, why this change was needed?

asder8215: I think this was a mistake I made while rebasing vrtgs with tokio's master branch. I had to fix up some of the `.check_and_init` and copied and pasted the `io_uring::opcode::*` stuff. Didn't notice that this was `OpenAt` and not `Read`.

I'll fix this soon.

Darksonn:
The compilation error is probably related to whether the build is using gnu or musl libc. See https://github.com/rust-lang/rust/pull/154981/ with similar issue.

tokio/src/fs/try_exists.rs
```diff
@@ -23,6 +23,43 @@ use std::path::Path;
...
+        feature = "io-uring",
+        feature = "rt",
+        feature = "fs",
+        target_os = "linux"
+    ))]
+    {
+        let handle = crate::runtime::Handle::current();
+        let driver_handle = handle.inner.driver().io();
+        if driver_handle
+            .check_and_init(io_uring::opcode::OpenAt::CODE)
```
martin-g:
```suggestion
            .check_and_init(io_uring::opcode::Statx::CODE)
```

tokio/src/io/uring/statx.rs
```diff
@@ -0,0 +1,137 @@
...
+    pub(crate) fn file_metadata(file: &File) -> io::Result<Op<Statx>> {
+        let mut buffer = box_new_uninit::<statx>();
+        let empty_path = cstr(Path::new(""))?;
+
+        // io-uring was introduced in linux 5.1
+        // pass in an empty path instead of null to target the file descriptor
+        // status as specified by man:
+        // https://man7.org/linux/man-pages/man2/statx.2.html
+        let statx_op = opcode::Statx::new(
+            types::Fd(file.as_raw_fd()),
```
martin-g, line 124: Does something guarantee that the File won't be dropped during the io uring call ?
It this happens the kernel will try to use a closed fd or worse - a reused fd.
IMO it should use an OwnedFd here.

tokio/src/fs/open_options.rs
```diff
@@ -544,12 +548,12 @@ impl OpenOptions {
...
-    async fn std_open(opts: &StdOpenOptions, path: impl AsRef<Path>) -> io::Result<File> {
-        let path = path.as_ref().to_owned();
+    async fn std_open(opts: &StdOpenOptions, path: &Path) -> io::Result<File> {
+        let path = path.to_owned();
         let opts = opts.clone();
 
-        let std = asyncify(move || opts.open(path)).await?;
-        Ok(File::from_std(std))
+        let std = asyncify(move || opts.open(path).map(File::from_std)).await?;
+        Ok(std)
```
martin-g:
```suggestion
        asyncify(move || opts.open(path).map(File::from_std)).await
```
no need to unwrap+wrap
Also the variable name is confusing. It is not `std` - it is a tokio::fs::File

tokio/src/io/uring/utils.rs
```diff
@@ -29,3 +30,15 @@ impl UringFd for ArcFd {
...
+// TODO(MSRV 1.82): When bumping MSRV, switch to `Box::<T>::new_uninit()`.
+pub(crate) fn box_new_uninit<T>() -> Box<MaybeUninit<T>> {
+    // Box::<T>::new_uninit()
+    Box::new(MaybeUninit::uninit())
+}
+
+// TODO(MSRV 1.82): When bumping MSRV, switch to `Box::<MaybeUninit<T>>::assume_init()`.
+pub(crate) unsafe fn box_assume_init<T>(boxed: Box<MaybeUninit<T>>) -> Box<T> {
+    let raw = Box::into_raw(boxed);
+    unsafe { Box::from_raw(raw as *mut T) }
```
martin-g: Misses `Safety: ` comment

tokio/src/io/uring/statx.rs
```diff
@@ -0,0 +1,137 @@
...
+impl Cancellable for Statx {
+    fn cancel(self) -> CancelData {
+        CancelData::Statx(self)
+    }
+}
+
+impl Op<Statx> {
+    /// Submit a request to retrieve a file's status.
+    #[inline]
+    fn statx(path: &Path, flags: u32) -> io::Result<Op<Statx>> {
```
martin-g: What is the reason to use `u32` here ?
The flags are i32. And currently the callers cast i32 to u32, e.g. https://github.com/tokio-rs/tokio/pull/8080/changes#diff-245d7f607e2bab7a7358fa2126c82845bce7cb2a5725471f556424f6102ce456R75

tokio/src/io/uring/statx.rs
asder8215: I might need a bit more clarity to understand what you're seeing.

`Op::file_metadata` doesn't own `File`, it takes a reference to it. Wouldn't Rust's borrow checker prohibit this `file.as_raw_fd()` from being invalid since `&File` has to live for as long as the caller's `File`?

Moreover, I would assume that the statx syscall populates the buffer with the file metadata once, and doesn't need access to the fd/path after invocation (correct me if I'm wrong though)?

tokio/src/io/uring/statx.rs
asder8215: Was taking Daksh's advice from [vrtgs' PR](https://github.com/tokio-rs/tokio/pull/7791) and preserving the way vrtgs used u32:
> You can create an member function that takes in a `flags: u32, path: &Path` then you can have `fn statx_follow_symlinks` and `fn statx_no_symlinks` than have a vague `bool` argument which might confuse future readers and users.

Not sure if the `u32` was significant here. But I'll change it to `i32` in any case.

tokio/src/fs/open_options.rs
asder8215: This is also leftovers from vrtgs' PR. I'm okay with doing:

```rust
Ok(asyncify(move || opts.open(path)).await?.into())
```

If that's fine to you.

tokio/tests/fs_uring_statx.rs
```diff
@@ -0,0 +1,282 @@
...
+    let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
+
+    let handle = tokio::spawn(async move {
+        let fut = try_exists(path.clone());
+        tokio::pin!(fut);
+
+        poll_fn(move |_| {
+            // If io_uring is enabled (and not falling back to the thread pool),
+            // the first poll should return Pending.
+            assert_pending!(fut.as_mut().poll(&mut Context::from_waker(Waker::noop())));
```
mattiapitossi: This is not going to work because we run tests also on a kernel version <5.1 where io_uring is not available (see: https://github.com/tokio-rs/tokio/blob/e56ff72fe7574867c6c9bbd171afd8717897ea43/tokio/tests/fs_uring_cancel_open.rs#L43). 

I was thinking that we should skip some io_uring tests on 4.19 (or something else if someone has a better idea), to be able to assert that the poll actually returns pending.

mattiapitossi: I pushed a utility function for this issue, eventually you can now use the assert_pending (see: https://github.com/tokio-rs/tokio/pull/8134)

tokio/src/fs/try_exists.rs
```diff
@@ -23,6 +23,71 @@ use std::path::Path;
...
+            .await?
+        {
+            return try_exists_uring(path).await;
+        }
+    }
+
+    try_exists_spawn_blocking(path).await
+}
+
+cfg_io_uring! {
```
mattiapitossi, line 58: is this still needed given the cfg below?

asder8215: I forgot that the `cfg_io_uring` macro already gates this to `io_uring`,`rt`, `fs` features and `linux` platform.

I'll get rid of all the redundant feature gates here while keeping the `cfg_io_uring` macro. The `target_env` gate is required because a specific subset of users on the Linux platform can use the statx syscall.

tokio/tests/fs_uring_statx.rs
```diff
@@ -0,0 +1,301 @@
...
+}
+
+#[tokio::test]
+async fn stat_nonexistent_file() {
+    let path = tempdir().unwrap().path().join("nonexistent_path");
+    let exists = try_exists(path).await.unwrap();
+    assert!(!exists);
+}
+
+// Error is not produced on Linux 4.19 (Linux 7.1 it works)
```
mattiapitossi, line 152: Can we run this only on 7.1 with io_uring_supported?

asder8215: I'll try this out later today

tokio/src/io/uring/statx.rs
```diff
@@ -0,0 +1,154 @@
...
+                    path: empty_path,
+                    buffer,
+                },
+            )
+        })
+    }
+
+    /// Retrieves the metadata information of the given path without following symlinks.
+    #[inline]
+    #[allow(dead_code)]
```
mattiapitossi, line 154: Is there any reason to keep this given that's not used anywhere?

asder8215: I was going to keep this for later when we can support io-uring on `tokio::fs::symlink_metadata` (which unfortunately returns Rust's standard library `Metadata`). 

I'm currently working on implementing [`Metadata::from_statx`](https://github.com/rust-lang/rust/pull/156269) so that we can enable io-uring support on `tokio::fs::metadata`/`tokio::fs::symlink_metadata` once this gets stabilized.

I could remove this code for now, or put a comment above it to note that this needs `Metadata::from_statx` to be usable later down the line.

mattiapitossi: Yes please, let's add a TODO/comment so we know why it's not used for the moment

tokio/src/io/uring/statx.rs
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /// Retrieves the metadata information of the given path, following symlinks
+    /// if the path provided points to a symlink location.
+    #[inline]
+    pub(crate) fn metadata(path: &Path) -> io::Result<Op<Statx>> {
+        Op::statx(
+            path,
+            // we don't need to pass in AT_SYMLINK_FOLLOW here, it'll follow
+            // by default it seems
```
mattiapitossi: nit: any strong reason to keep this comment? Given that this is default behavior of statx I think it can be removed

Daksh14 (changes requested):
Can we add a test that 
1. Opens a file
2. retrivies metadata and statistics at time of sync flag using statx call
3. does some operations
4. Check if file is dropped and metadata, to check for any file leaks or statx leaks

Daksh14 (approved):
The statx fd leak test and fd uring statx test look good to me, its now wired in io-uring

I will rebase this PR over my io-uring batch PR

Daksh14 (changes requested):
Let others review the integrity of the tests

tokio/src/fs/read_uring.rs
```diff
@@ -14,28 +16,46 @@ const PROBE_SIZE_U32: u32 = PROBE_SIZE as u32;
...
 // SAFETY: cannot be higher than u32::MAX for safe cast
 // Set to read max 64 MiB at time
+#[allow(unused)] // FIXME: remove when MSRV is 1.93 (due to statx on 1.25 musl)
 const MAX_READ_SIZE: usize = 64 * 1024 * 1024;
 
-pub(crate) async fn read_uring(path: &Path) -> io::Result<Vec<u8>> {
-    let file = OpenOptions::new().read(true).open(path).await?;
-
-    // TODO: use io uring in the future to obtain metadata
-    let size_hint: Option<usize> = file.metadata().await.map(|m| m.len() as usize).ok();
```
Darksonn, line 23: Instead of making the entire `read_uring` function depend on musl, please apply this change *only* to this step.

tokio/src/fs/write.rs
```diff
@@ -48,49 +48,44 @@ pub async fn write(path: impl AsRef<Path>, contents: impl AsRef<[u8]>) -> io::Re
...
-    feature = "fs",
-    target_os = "linux"
-))]
-async fn write_uring(path: &Path, mut buf: OwnedBuf) -> io::Result<()> {
-    use crate::{fs::OpenOptions, runtime::driver::op::Op};
-    use std::os::fd::OwnedFd;
+cfg_io_uring! {
+    async fn write_uring(path: &Path, mut buf: OwnedBuf) -> io::Result<()> {
+        use crate::{fs::OpenOptions, runtime::driver::op::Op};
+        use std::os::fd::OwnedFd;
```
Darksonn: Please avoid unrelated changes.

tokio/src/io/uring/statx.rs
```diff
@@ -0,0 +1,154 @@
...
+    /// if the path provided points to a symlink location.
+    #[inline]
+    pub(crate) fn metadata(path: &Path) -> io::Result<Op<Statx>> {
+        Op::statx(path, libc::AT_STATX_SYNC_AS_STAT)
+    }
+
+    /// Retrieves the metadata information of the given file
+    pub(crate) fn file_metadata(file: &File) -> io::Result<Op<Statx>> {
+        let mut buffer = box_new_uninit::<statx>();
+        let empty_path = cstr(Path::new(""))?;
```
Darksonn: No need to allocate for this one.
```suggestion
        let empty_path: &'static CStr = c"";
```

tokio/src/io/uring/statx.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    fn complete(self, cqe: CqeResult) -> Self::Output {
+        // SAFETY: On success, we always receive 0, which should guarantee
+        // that the information about a file is stored inside the
+        // statx buffer. On failure, we'll receive an Error value,
+        // avoiding misuse of `.box_assume_init()`.
+        // Refer to man page description and return value:
+        // https://man7.org/linux/man-pages/man2/statx.2.html
+        cqe.result
+            .map(|_| Metadata(*unsafe { box_assume_init(self.buffer) }))
```
Darksonn: This `box_assume_init` helper isn't needed.
```suggestion
            .map(|_| Metadata(unsafe { *self.buffer.as_ptr() }))
```

tokio/src/io/uring/utils.rs
```diff
@@ -29,3 +30,23 @@ impl UringFd for ArcFd {
...
+///
+/// It is up to the caller to guarantee that the value really is in an initialized state.
+/// Calling this when the content is not yet fully initialized causes immediate undefined behavior.
+#[allow(unused)] // FIXME: remove when MSRV is 1.93 (due to statx on 1.25 musl)
+pub(crate) unsafe fn box_assume_init<T>(boxed: Box<MaybeUninit<T>>) -> Box<T> {
+    let raw = Box::into_raw(boxed);
+    // SAFETY: If the caller guarantees that the MaybeUninit is initialized, then
+    // costructing the box from a raw mut ptr of MaybeUninit<T> should be safe.
+    unsafe { Box::from_raw(raw as *mut T) }
+}
```
Darksonn: These are not needed. You can invoke `Box::new(MaybeUninit::new())` directly. If you want a FIXME, place it at the call-site.

tokio/tests/fs_uring_statx_fd_leak_test.rs
```diff
@@ -0,0 +1,166 @@
...
+                self.poll_senders[poll_count].send(()).unwrap();
+                self.poll_pending_counts += 1;
+            }
+            // The first two poll should return `Poll::Pending`` because it verifies
+            //  that io_uring is enabled and then executes the open operation in
+            // `tokio::fs::read`
+            else {
+                assert_pending!(self.inner.as_mut().poll(cx));
+                self.poll_pending_counts += 1;
+                self.poll_senders[poll_count].send(()).unwrap();
```
Daksh14: nit: can move `self.poll_pending_counts += 1;` and `self.poll_senders[poll_count].send(()).unwrap();` out of the if statement

asder8215:
Not sure what to make out of the failing check:

```
(node:1776) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
```

Don't think that's due to this PR.

tokio/src/fs/read_uring.rs
```diff
@@ -19,9 +19,26 @@ const MAX_READ_SIZE: usize = 64 * 1024 * 1024;
...
+        // rust version is 1.93. To clarify, statx support is
+        // introduced to musl in 1.25 as mentioned officially here:
+        // https://musl.libc.org/releases.html.
+        // However, rustup target_env building for *-linux-musl
+        // uses 1.25 musl on all *-linux-musl platforms starting
+        // in 1.93 stable rust version.
+        // https://blog.rust-lang.org/2025/12/05/Updating-musl-1.2.5/
+        any(target_env = "gnu", target_os = "android")
+    )]
+    let size_hint = Op::file_metadata(&file)?
```
Daksh14, line 37: This way we only use one `cfg` directive

```rust
let size_hint: Option<usize> = file.metadata().await.map(|m| m.len() as usize).ok();

#[cfg(any(
    // statx is supported on gnu and android.
    // musl support lands in rust 1.93 (bundled musl 1.2.5):
    // https://blog.rust-lang.org/2025/12/05/Updating-musl-1.2.5/
    target_env = "gnu",
    target_os = "android",
))]
let size_hint = Op::file_metadata(&file)?
    .await
    .map(|m| m.len() as usize)
    .ok();
```

tokio/src/fs/open_options.rs
```diff
@@ -518,6 +518,10 @@ impl OpenOptions {
     /// [`Other`]: std::io::ErrorKind::Other
     /// [`PermissionDenied`]: std::io::ErrorKind::PermissionDenied
     pub async fn open(&self, path: impl AsRef<Path>) -> io::Result<File> {
+        self.open_inner(path.as_ref()).await
+    }
+
+    async fn open_inner(&self, path: &Path) -> io::Result<File> {
```
Daksh14, line 524: why we needed to change here ?

asder8215: This was a change introduced by vrtgs originally, and I think the intention was to reduce monomorphization cost of `OpenOptions::open`, so instead of the compiler generating multiple implementations of `OpenOptions::open` (on `Path`, `String`, `PathBuf`, etc.) with all the code in `open_inner`, it will generate multiple implementations of `OpenOptions::open` on the wrapper function `self.open_inner(path.as_ref()).await` instead. It's supposed to reduce binary size and compile time as a result as well.

I can remove this change from this PR since this isn't particularly relevant to the Statx operation being introduced to `fs::read` and `fs::try_exists`, but I think this change is nice to make for the benefit of reducing the binary size of `tokio::fs` and compilation time. This inner function pattern exists a lot in the Rust standard fs library as well.

tokio/src/fs/read_uring.rs
asder8215: We'd have to apply `#[allow(dead_code)]` on top of the earlier `size_hint` if we do this on CI builds that enable io-uring features since it's going to complain about unused variables, no? Since the cfg gate macro or allow dead code macro consume 1 line still, I'm inclined to think that using a cfg directive might be better and let people know that this is still be used by platforms that don't have statx support.

tokio/src/fs/open_options.rs
Daksh14: I'm okay with keeping it @Darksonn @martin-g hopefully there are no downstream effects because of this change even though its internal

Darksonn: This particular change LGTM

tokio/src/io/uring/statx.rs
```diff
@@ -0,0 +1,154 @@
...
+impl Debug for Metadata {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let mut debug = f.debug_struct("Metadata");
+        debug.field("len", &self.len());
+        debug.finish_non_exhaustive()
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct Statx {
```
asder8215, line 48: @Darksonn should this struct be marked as `#[repr(C)]` since the path field here is being by the kernel? I know the `#[repr(C)]` removes dead code warning and is an indicator that this may be passing through an FFI boundary

Darksonn: No, that's not necessary. The kernel is given a pointer to the string heap allocation, not to the `Statx` struct.

You can name the field `_path` to suppress the warning.

asder8215:
Is there anything else in this PR that needs to be modified? Or does everything look good?

mattiapitossi (approved):

Daksh14 (approved):

PR #8081, opened 2026-04-22
macros: reduce noisy return type mismatch diagnostics

VinEckSie:
Avoid forcing a dyn Future cast when checking the generated async body.

This removes extra macro-generated E0271 diagnostics (a type mismatched an associated type of a trait) and leaves rustc's more direct type mismatch errors in place.

## Motivation

The macro currently uses an internal `dyn Future` cast to force early type checking
of the generated async body.

However, this introduces additional E0271 diagnostics (a type mismatched an associated type of a trait) related to async block
coercion, which can obscure the more relevant E0308 (expected type did not match the received type) return type mismatch errors.

In cases like #6306, this results in confusing error messages that point to the
async block rather than the actual issue (e.g. an extra semicolon causing `()`).

## Solution

Replace the `dyn Future` trait-object coercion with a simpler reference check:
    let _ = &body;

This still ensures the async body is type-checked, but avoids forcing a
trait-object cast, allowing rustc to emit clearer and more direct diagnostics
focused on the actual type mismatch.

### DIFF
--- a/tokio-macros/src/entry.rs
+++ b/tokio-macros/src/entry.rs
@@ -556,7 +556,7 @@ fn parse_knobs(mut input: ItemFn, is_test: bool, config: FinalConfig) -> TokenSt
             }
             _ => quote! {
                 if false {
-                    let _: &dyn ::core::future::Future<Output = #output_type> = &body;
+                    let _ = &body;
                 }
             },
         };
--- a/tests-build/tests/fail/macros_type_mismatch.stderr
+++ b/tests-build/tests/fail/macros_type_mismatch.stderr
@@ -1,14 +1,3 @@
-error[E0271]: expected `{async block@$DIR/tests/fail/macros_type_mismatch.rs:3:1: 3:15}` to be a future that resolves to `()`, but it resolves to `Result<(), _>`
- --> tests/fail/macros_type_mismatch.rs:3:1
-  |
-3 | #[tokio::main]
-  | ^^^^^^^^^^^^^^ expected `()`, found `Result<(), _>`
-  |
-  = note: expected unit type `()`
-                  found enum `Result<(), _>`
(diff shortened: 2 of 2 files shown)


VinEckSie:
I reran CI after updating the trybuild snapshots.

Locally (macOS, Rust 1.95.0, with `RUSTFLAGS="-Dwarnings"`), `tests-build::macros compile_fail_full` passes.

The Ubuntu CI job is still failing, so there may be a platform-specific difference in diagnostics.
Happy to investigate further if you have guidance on reproducing the CI behavior locally.

martin-g:
`cargo nextest run --workspace --features full,test-util` passes successfully here on Ubuntu 25.10, Rust 1.95.0

tokio-macros/src/entry.rs
```diff
@@ -556,7 +556,7 @@ fn parse_knobs(mut input: ItemFn, is_test: bool, config: FinalConfig) -> TokenSt
             }
             _ => quote! {
                 if false {
-                    let _: &dyn ::core::future::Future<Output = #output_type> = &body;
+                    let _ = &body;
                 }
```
Darksonn, line 560: The old code exists to check that the output type is correct. How does your updated code perform that check?

VinEckSie: Indeed, it does not preserve the previous output-type check. It only avoids the coercion that was producing the extra diagnostics.

I’ll revisit this and look for an approach that keeps the output type validation while reducing the noisy diagnostic.

VinEckSie: I tried a few approaches to keep the output-type check while reducing the extra diagnostics, but they either reintroduce noise or affect the error behavior.

This seems trickier than expected. I’m happy to continue exploring if you have a preferred direction.

Issue #8082, opened 2026-04-22
UDP receive errors aren't communicated through `try_*` methods


alexcrichton:
Locally I was in the process of testing the wasip2 and wasip3 targets to ensure that the Tokio test suite passes. I found, however, that the test added in https://github.com/tokio-rs/tokio/pull/8001, ignored on WASI targets, does not pass on wasip2 with all wasi-libc/wasmtime fixes in place. I debugged this to how the change in #8001 added the `ERROR` interest to `async` read-style methods, but it didn't apply the same change to `try_read_*` methods. For example this test currently fails with Tokio, although it's quite similar to the #8001 branch.

```rust
#[tokio::test]
async fn send_to_try_recv_closed_returns_err() -> std::io::Result<()> {
    let sender = UdpSocket::bind("127.0.0.1:0").await?;
    let receiver = UdpSocket::bind("127.0.0.1:0").await?;

    let receiver_addr = receiver.local_addr()?;
    drop(receiver);
    sender.connect(receiver_addr).await?;
    sender.send(MSG).await?;

    let interest = time::timeout(
        Duration::from_secs(5),
        sender.ready(Interest::READABLE | Interest::ERROR),
    )
    .await
    .expect("timed out instead of returning error")
    .unwrap();
    assert_eq!(interest, Ready::ERROR);

    let err = sender.try_recv(&mut [0u8; 32]).unwrap_err();
    assert_connection_refused_or_reset(&err);
    Ok(())
}
```

I assumed this would be a simple fix to add in `Interest::ERROR` [around here](https://github.com/tokio-rs/tokio/blob/6c03e03898d71eca976ee1ad8481cf112ae722ba/tokio/src/net/udp.rs#L892) in a similar manner to how #8001 was fixed. Unfortuantely though that doesn't actually resolve this test.

[...]

PR #8084, opened 2026-04-24
[net] Surface errors from SO_ERROR on try_recv for UDP sockets

martin-g:
## Motivation

Fixes and adds a regression test for #8082 

Same as https://github.com/tokio-rs/tokio/pull/8001 but for UdpSocket::try_recv/try_recv_buf methods.

## Solution

* Adjusted `UdpSocket::try_recv()` and `try_recv_buf()` to wait for `Interest::READABLE | Interest::ERROR` instead of only `Interest::READABLE`.
* Fixed `Interest::mask()` by delegating it to `Ready::from_interest()`


tokio/tests/udp.rs
```diff
@@ -86,6 +86,43 @@ async fn send_to_recv_closed_returns_err() -> std::io::Result<()> {
     Ok(())
 }
 
+#[tokio::test]
+#[cfg_attr(
+    target_os = "wasi",
+    ignore = "temporarily disabled for WASI pending https://github.com/WebAssembly/wasi-libc/pull/734"
```
mattiapitossi, line 110: Is this still true? I see they have released a fix

alexcrichton:
Could this also update other associated methods like `try_recv_buf_from`, `try_recv_from`, and maybe `try_peek_*` (I'm not sure if those are necessary)?

tokio/tests/udp.rs
alexcrichton: It is, yeah, we'll need this PR to go into Wasmtime itself before this test can pass for wasm guests.

tokio/src/io/interest.rs
```diff
@@ -256,14 +256,7 @@ impl Interest {
...
     pub(crate) fn mask(self) -> Ready {
-        match self {
-            Interest::READABLE => Ready::READABLE | Ready::READ_CLOSED,
-            Interest::WRITABLE => Ready::WRITABLE | Ready::WRITE_CLOSED,
-            #[cfg(any(target_os = "linux", target_os = "android"))]
-            Interest::PRIORITY => Ready::PRIORITY | Ready::READ_CLOSED,
-            Interest::ERROR => Ready::ERROR,
-            _ => Ready::EMPTY,
-        }
+        Ready::from_interest(self)
```
Darksonn: Why are you making this change?

Darksonn (commented):
Overall looks reasonable, but one question below.

tokio/src/io/interest.rs
martin-g: Because the previous version of the implementation did not support an Interest instance like `READABLE | ERROR`.

https://github.com/tokio-rs/tokio/issues/8082#issue-4311995092
```
Specifically in ready_event it uses [Interest::mask which doesn't handle 
READABLE | ERROR](https://github.com/tokio-rs/tokio/blob/6c03e03898d71eca976ee1ad8481cf112ae722ba/tokio/src/io/interest.rs#L258-L267) 
and instead only looks for readable interest. This causes try_io to never attempt
 I/O and unconditionally return "would block".
```

Darksonn: Hmm. But `Interest::READABLE` should still be mapped to `READABLE | READ_CLOSED`, right?

martin-g: Yes, it is https://github.com/tokio-rs/tokio/blob/02ff0833e00adeb8cd451420f586f968deb70f78/tokio/src/io/ready.rs#L241-L244

Darksonn: Okay perfect. Could we just delete `mask` and change callers to use `from_interest`?

Or another question is ... should `READABLE` also imply `ERROR`?

martin-g: > Okay perfect. Could we just delete `mask` and change callers to use `from_interest`?

Done!

> Or another question is ... should `READABLE` also imply `ERROR`?

I am not sure whether this is always valid.

martin-g:
Hm. The new `poll_recv_closed_returns_err()` test seems to hang at CI! But it passes quickly here. I'm investigating!

mattiapitossi:
@martin-g probably related to https://www.githubstatus.com/incidents/1qzncbrxcsy4?

martin-g:
> @martin-g probably related to https://www.githubstatus.com/incidents/1qzncbrxcsy4?

No. `High wait time` here means to wait more time until a VM is started for the CI run.

In this case the VM is assigned and the tests are running but this one particular test case does not receive the ConnectionRefused error

ricochet:
I dug into why `poll_recv_closed_returns_err` hangs on CI but passes locally. I think there's another change needed.

`ScheduledIo::wake()` only wakes the reserved reader/writer slots on a readable/writable bit:
```rust
// check for AsyncRead slot
if ready.is_readable() {
    if let Some(waker) = waiters.reader.take() { ... }
}
```

`recv().await` path is fine because it uses the list waiters, which wake via ready.satisfies(interest) and therefore catch `Ready::ERROR`. But a poller parked in the reader slot is never woken by an error-only event.

Ready::ERROR with no readable bit is what happens when Linux/epoll delivers EPOLLERR. On macOS/kqueue the error tends to surface as readable, so the reader slot does get woken.

The fix should basically be:
```rust
  if ready.is_readable() || ready.is_error() { /* wake reader */ }
  if ready.is_writable() || ready.is_error() { /* wake writer */ }
```

Aside: I was testing this because with wasmtime 45, the wasmtime-side fix referenced earlier is now landed. The tests no longer need a wasi ignore.

PR #8088, opened 2026-04-27
miri tests: fstat is supported now

RalfJung:
`fstat` on non-file-backed FDs works now in Miri so we can enable two more tests here.


Darksonn (approved):
Thanks.

mattiapitossi (approved):

martin-g (approved):

PR #8092, opened 2026-04-28
Make LIFO slot stealing opt-in

### LINKED ISSUE #8065: Performance regression on high-QPS short-handler workloads since 1.51.0

Darksonn:
We made it possible to steal the LIFO slot in #7431, but this causes a performance regression #8065. Thus, make it opt-in.

The feature is split into two options:

* Can stealing happen at all?
* Are workers woken up eagerly to check whether stealing should be carried out?

Both options are needed to make the runtime fully tolerate slow tasks, but I split them up as the first option can probably be enabled by default without adverse performance impact. But for now, both are off by default.

Fixes: #8065

### DIFF
--- a/tokio/src/runtime/builder.rs
+++ b/tokio/src/runtime/builder.rs
@@ -130,7 +130,7 @@ pub struct Builder {
     pub(super) disable_lifo_slot: bool,
 
     /// When true, the LIFO slot can be work stolen.
-    enable_lifo_stealing: bool,
+    enable_lifo_slot_stealing: bool,
 
     /// Whether or not to enable eager hand-off for tasks in the LIFO slot.
     enable_eager_lifo_handoff: bool,
@@ -342,7 +342,7 @@ impl Builder {
             metrics_poll_count_histogram: HistogramBuilder::default(),
 
             disable_lifo_slot: false,
-            enable_lifo_stealing: false,
+            enable_lifo_slot_stealing: false,
             enable_eager_lifo_handoff: false,
 
             timer_flavor: TimerFlavor::Traditional,
@@ -1362,6 +1362,12 @@ impl Builder {
         /// See the docs for [`disable_lifo_slot`] for more details on what the
         /// LIFO slot is.
         ///
+        /// # Panics
+        ///
+        /// This option requires the LIFO slot to be enabled. If this option is
+        /// set together with [`disable_lifo_slot`], then building the runtime
+        /// will fail with a panic.
+        ///
         /// # Unstable
         ///
         /// This configuration option is unstable because we are considering
@@ -1371,7 +1377,7 @@ impl Builder {
         /// [`disable_lifo_slot`]: Self::disable_lifo_slot()
         /// [`enable_eager_lifo_handoff`]: Self::enable_eager_lifo_handoff()
         pub fn enable_lifo_slot_stealing(&mut self) -> &mut Self {
-            self.enable_lifo_stealing = true;
+            self.enable_lifo_slot_stealing = true;
             self
         }
 
@@ -1387,6 +1393,12 @@ impl Builder {
         /// blocks and the LIFO slot needs to be stolen. This can hurt the
         /// performance of some workloads. See [tokio-rs/tokio#8065] for
         /// details.
+        ///
+        /// # Panics
+        ///
+        /// This option requires that LIFO slot stealing is enabled. If this
+        /// option is enabled without [`enable_lifo_slot_stealing`], then this
+        /// results in a panic when building the runtime.
         ///
         /// # Unstable
         ///
@@ -1765,7 +1777,7 @@ impl Builder {
                 #[cfg(tokio_unstable)]
                 unhandled_panic: self.unhandled_panic.clone(),
                 disable_lifo_slot: self.disable_lifo_slot,
-                enable_lifo_stealing: self.enable_lifo_stealing,
+                enable_lifo_slot_stealing: self.enable_lifo_slot_stealing,
                 enable_eager_lifo_handoff: self.enable_eager_lifo_handoff,
                 // This setting never makes sense for a current thread runtime,
                 // as it only configures how the I/O driver is stolen across
@@ -1933,6 +1945,15 @@ cfg_rt_multi_thread! {
             let seed_generator_1 = self.seed_generator.next_generator();
             let seed_generator_2 = self.seed_generator.next_generator();
 
+            if self.disable_lifo_slot {
+                assert!(!self.enable_lifo_slot_stealing, "Enabling LIFO slot stealing does not make sense when LIFO slot is disabled");
+                assert!(!self.enable_eager_lifo_handoff, "Enabling eager LIFO handoff does not make sense when LIFO slot is disabled");
+            }
+
+            if self.enable_eager_lifo_handoff {
+                assert!(self.enable_lifo_slot_stealing, "Enabling eager LIFO handoff requires LIFO slot stealing to be enabled");
+            }
+
             let (scheduler, handle, launch) = MultiThread::new(
                 worker_threads,
                 driver,
@@ -1953,7 +1974,7 @@ cfg_rt_multi_thread! {
                     #[cfg(tokio_unstable)]
                     unhandled_panic: self.unhandled_panic.clone(),
                     disable_lifo_slot: self.disable_lifo_slot,
-                    enable_lifo_stealing: self.enable_lifo_stealing,
+                    enable_lifo_slot_stealing: self.enable_lifo_slot_stealing,
                     enable_eager_lifo_handoff: self.enable_eager_lifo_handoff,
                     enable_eager_driver_handoff: self.enable_eager_driver_handoff,
                     seed_generator: seed_generator_1,
--- a/tokio/src/runtime/builder.rs
+++ b/tokio/src/runtime/builder.rs
@@ -129,6 +129,12 @@ pub struct Builder {
     /// This option should only be exposed as unstable.
     pub(super) disable_lifo_slot: bool,
 
+    /// When true, the LIFO slot can be work stolen.
+    enable_lifo_stealing: bool,
+
+    /// Whether or not to enable eager hand-off for tasks in the LIFO slot.
+    enable_eager_lifo_handoff: bool,
+
     /// Specify a random number generator seed to provide deterministic results
     pub(super) seed_generator: RngSeedGenerator,
 
@@ -336,6 +342,8 @@ impl Builder {
             metrics_poll_count_histogram: HistogramBuilder::default(),
 
             disable_lifo_slot: false,
+            enable_lifo_stealing: false,
+            enable_eager_lifo_handoff: false,
 
             timer_flavor: TimerFlavor::Traditional,
 
@@ -1306,15 +1314,16 @@ impl Builder {
         /// This configuration option will disable this heuristic resulting in
         /// all scheduled tasks being pushed into the worker-local queue. This
         /// was intended as a workaround for the LIFO slot not being stealable.
-        /// As of Tokio 1.51, tasks can be stolen from the LIFO slot. In a
-        /// future version, this option may be deprecated.
+        /// As of Tokio 1.51, tasks can be stolen from the LIFO slot when
+        /// [`enable_lifo_slot_stealing`] is enabled. In a future version, this
+        /// option may be deprecated.
         ///
         /// # Unstable
         ///
         /// This configuration option was considered a workaround for the LIFO
-        /// slot not being stealable. Since this is no longer the case, we will
-        /// revisit whether or not this option is necessary. See
-        /// issue [tokio-rs/tokio#4941].
+        /// slot not being stealable. Since LIFO slot stealing can now be
+        /// enabled, we will revisit whether or not this option is necessary.
+        /// See issue [tokio-rs/tokio#4941].
         ///
         /// # Examples
         ///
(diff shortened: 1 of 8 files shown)


Darksonn:
There aren't any changes on master that can't be part of a 1.52.2 release, so merging into master is fine.

tokio/src/runtime/time/wheel/mod.rs
```diff
@@ -50,9 +50,13 @@ pub(super) const MAX_DURATION: u64 = (1 << (6 * NUM_LEVELS)) - 1;
...
     /// Creates a new timing wheel.
     pub(crate) fn new() -> Wheel {
+        let mut levels = Vec::with_capacity(NUM_LEVELS);
+        for i in 0..NUM_LEVELS {
+            levels.push(Level::new(i));
+        }
         Wheel {
             elapsed: 0,
-            levels: Box::new(array::from_fn(Level::new)),
+            levels: levels.into_boxed_slice().try_into().unwrap(),
```
Darksonn, line 59: I changed this because I was getting some stack overflows in the runtime constructor. Note that these arrays are very large.

tokio/src/runtime/builder.rs
mattiapitossi, line 2018: Probably we want to add the two new options also here

Darksonn: Hrm, it's already missing a lot of other options.

tokio/src/runtime/builder.rs
```diff
@@ -129,6 +129,12 @@ pub struct Builder {
     /// This option should only be exposed as unstable.
     pub(super) disable_lifo_slot: bool,
 
+    /// When true, the LIFO slot can be work stolen.
+    enable_lifo_stealing: bool,
```
```suggestion
    enable_lifo_slot_stealing: bool,
```
to be consistent with the old options, e.g. `disable_lifo_slot` and the method - `enable_lifo_slot_stealing()`

tokio/src/runtime/builder.rs
```diff
@@ -129,6 +129,12 @@ pub struct Builder {
...
+
+    /// Whether or not to enable eager hand-off for tasks in the LIFO slot.
+    enable_eager_lifo_handoff: bool,
```
martin-g, line 136:
```suggestion
    enable_eager_lifo_slot_handoff: bool,
```

tokio/src/runtime/scheduler/multi_thread/queue.rs
```diff
@@ -436,15 +437,18 @@ impl<T> Steal<T> {
 
     /// Return true if the queue is empty,
```
martin-g: The docs need update

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -1147,10 +1148,12 @@ impl Core {
                 continue;
             }
 
-            let target = &worker.handle.shared.remotes[i];
-            if let Some(task) = target
-                .steal
-                .steal_into(&mut self.run_queue, &mut self.stats)
+            let target = &shared.remotes[i];
+            let steal_lifo = shared.config.enable_lifo_stealing;
```
martin-g: nit: could be moved out of the loop

tokio/src/runtime/builder.rs
```diff
@@ -1332,11 +1341,65 @@ impl Builder {
...
+        /// this option, workers will steal the LIFO slot task when the queue is
+        /// otherwise empty.
+        ///
+        /// Note that the LIFO slot is only stolen if another worker thread
+        /// attempts work stealing while the LIFO slot is non-empty, but the
+        /// runtime will not wake up other workers to steal the task unless the
+        /// [`enable_eager_lifo_handoff`] option is also enabled.
+        ///
+        /// See the docs for [`disable_lifo_slot`] for more details on what the
+        /// LIFO slot is.
```
martin-g, line 1363: Add some info how the new option(s) depend on the value of `disable_lifo_slot` ?!
Currently it is possible to do: `builder.disable_lifo_slot().enable_lifo_slot_stealing()`.

Or even add `debug_assert!(!self.disable_lifo_slot, "Enabling LIFO slot stealing does not make sense when LIFO slot is disabled");`

tokio/src/runtime/tests/loom_multi_thread/queue.rs
```diff
@@ -21,7 +21,7 @@ fn basic() {
             let mut n = 0;
```
martin-g, line 21: All loom tests use `lifo_slot = false`.
Would a new test with `lifo_slot = true` be useful ?

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -1349,23 +1352,29 @@ impl Handle {
...
             if let Some(prev) = core.run_queue.push_lifo(task) {
                 // There was a previous task in the LIFO slot which needs
                 // to be pushed to the back of the run queue.
                 core.run_queue
                     .push_back_or_overflow(prev, self, &mut core.stats);
+                true
+            } else {
+                // We pushed to the LIFO slot, and it was empty, so only notify if eager lifo
+                // handoff is enabled.
+                self.shared.config.enable_eager_lifo_handoff
```
martin-g, line 1370:
```suggestion
                self.shared.config.enable_lifo_stealing && self.shared.config.enable_eager_lifo_handoff
```
no need to notify if stealing is not enabled, no ?

tokio/src/runtime/scheduler/multi_thread/worker/taskdump.rs
```diff
@@ -67,8 +67,8 @@ impl Shared {
 
         for remote in self.remotes.iter() {
             let steal = &remote.steal;
-            while !steal.is_empty() {
-                if let Some(task) = steal.steal_into(&mut local, &mut stats) {
+            while steal.can_steal(true) {
+                if let Some(task) = steal.steal_into(&mut local, &mut stats, true) {
                     local.push_back([task].into_iter());
                 }
```
martin-g, line 73:
```suggestion
                } else {
                    break;
                }
```
queue.rs, line 462 could return None early

tokio/src/runtime/builder.rs
mattiapitossi: Ok maybe we could add all missing fields in another PR

tokio/src/runtime/scheduler/multi_thread/worker/taskdump.rs
Darksonn: This is not quite obvious. It's possible for `can_steal` to return true but `steal_into` fail if another thread is concurrently performing work stealing here. As it is now, this code is a spinlock until that's fixed, which may be a potential problem, but I don't want to change this behavior in this PR.

tokio/src/runtime/scheduler/multi_thread/worker.rs
Darksonn: I'm changing builder to panic in this scenario, so it's not necessary to check both here.

tokio/src/runtime/tests/loom_multi_thread/queue.rs
Darksonn: None of the existing tests push to the lifo slot, so I don't think it's useful for them to pass `true` here.

tokio/src/runtime/builder.rs
Darksonn: Updating docs and making these panic.

tokio/src/runtime/scheduler/multi_thread/worker.rs
Darksonn: I think the code is easier to read with it inside the loop. I don't care about perf impact. Optimizer can move it outside if it wants to.

tokio/src/runtime/tests/loom_multi_thread/queue.rs
martin-g: I meant to add a **new** test that covers this path.

Darksonn: Done

tokio/src/runtime/config.rs
```diff
@@ -43,6 +43,12 @@ pub(crate) struct Config {
     /// future version.
```
mattiapitossi: Probably this could be "..when the LIFO slot stealing will be enabled by default" ?

Darksonn: Done

mattiapitossi (approved):
Looks good to me

carllerche (changes requested):
I would recommend against specific options for stealing from LIFO etc.. and have a general "wake up more aggressively" option that is used as a basic toggle for the runtime to decide if it should bias for waking up or not.

Darksonn:
I agree but I don't want to do that in a patch release.

carllerche:
Feel free to dismiss my review if it doesn't apply here

Noah-Kennedy (approved):

Noah-Kennedy:
The implementation looks sound, have approved.

I do somewhat question adding options in a patch but I will not block on that.

Darksonn:
I don't feel that strongly about this, so I'm not going to dismiss your review. I filed #8100 to revert instead.

Darksonn:
I assume Eliza will pick this up when re-enabling this, so closing.

PR #8094, opened 2026-04-29
feat: add AsyncReadExt::read_interruptible()

vojtechkral:
Adds `AsyncReadExt::read_interruptible()`

## Motivation

I find myself sometimes in a need of stopping async reader(s) based on a future for graceful shutdown.

I'm already using this code in a project so I thought might as well offer that for upstreaming, in case others find it useful. (If not that's ok too.)

## Solution

The usual, a method on the `AsyncReadExt` trait returning a wrapper.


Darksonn:
Please use [`run_until_cancelled`][1] or `tokio::select!` instead.

[1]: https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html#method.run_until_cancelled

PR #8095, opened 2026-04-29
sync: reset Chan::rx_waker in chan::Rx's Drop impl

TheJokr:
## Motivation

I recently fixed a memory leak in an application where tokio's RawTask storage was being kept alive by a leaked Waker. The task itself was polling an `mpsc::Receiver` before then being aborted. During cleanup, all references to the RawTask were dropped, except for the one stored in `mpsc::chan::Chan::rx_waker`. While the `Receiver` was dropped as part of the task's future, one of the channel's `Sender`s was leaked outside the task. This meant the `Chan` was never dropped and its `rx_waker` contained the leaked Waker.

I fixed the leak by properly cleaning up the `Sender`, but I also think keeping `rx_waker` around in this case is unnecessary. Once `chan::Rx` is dropped, it can't be polled anymore, so waking up the registered task will always be spurious. (In my case above, `Chan` itself would still be leaked, of course.)

## Solution

Reset `rx_waker` explicitly in `impl Drop for Rx` using `AtomicWaker::take_waker` and add a regression test to illustrate the problem.

### DIFF
--- a/tokio/src/sync/mpsc/chan.rs
+++ b/tokio/src/sync/mpsc/chan.rs
@@ -517,6 +517,11 @@ impl<T, S: Semaphore> Drop for Rx<T, S> {
                 sem: &self.inner.semaphore,
             };
 
+            // When Rx is dropped, there is nothing for a task to poll anymore.
+            // This means we can drop our waker to potentially free up resources.
+            // Do so before draining the channel where panics may occur.
+            self.inner.rx_waker.take_waker();
+
             guard.drain();
         });
     }
--- a/tokio/tests/sync_mpsc.rs
+++ b/tokio/tests/sync_mpsc.rs
@@ -1533,4 +1533,38 @@ fn drop_all_elements_during_panic() {
     // `mpsc::Chan`'s drop is called, freeing the `Block` memory allocation.
 }
 
+/// A task's wakers keep its memory allocation alive. This test ensures
+/// `mpsc::channel` releases the rx waker immediately upon drop to potentially
+/// free up resources.
+#[test]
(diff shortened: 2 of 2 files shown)


martin-g (approved):

TheJokr:
@martin-g I see you approved this. Can we merge the PR, or is there something else that needs to be done first?

tokio/src/sync/mpsc/chan.rs
```diff
@@ -519,6 +519,10 @@ impl<T, S: Semaphore> Drop for Rx<T, S> {
 
             guard.drain();
```
Darksonn, line 525: What if dropping one of the values panics here? Please add a test for that.

TheJokr: Good point, I've moved the `take_waker()` call up before the channel is drained. I'm not sure this scenario can happen in practice though. If the waker was set, but the channel also contained an item, then either:
1. The waker was set first and should have been woken when the item was added, or
2. The item was added first and no waker should have been registered in the first place

As far as I can tell, `AtomicWaker` ensures that 1) would still result in the waker being woken immediately after being registered.

There is a test for the general case of items having a panicking Drop impl already in `tests/sync_mpsc.rs`.

Darksonn: Now we have the reverse problem if dropping the waker panics. Can you place it after the creation of `Guard` but before the call to `drain`?

TheJokr: Thanks, done

Darksonn (approved):
Thanks.

PR #8090, opened 2026-04-27
tokio-stream: Implement FusedStream for Fuse

stormshield-fabs:
## Motivation

Fix https://github.com/tokio-rs/tokio/issues/8089

## Solution

Implement `futures_core::FusedStream` for `Fuse`, making it interoperable with code that expects this bound to be satisfied.

PR #8096, opened 2026-04-29
tokio-stream: implement FusedStream for stream adaptors

Rachit2323:
Implements `futures_core::FusedStream` for 10 stream adaptor types in
tokio-stream, improving interoperability with code that depends on this
trait bound.

Follows the precedent set by #8090 (Fuse) and #7854 (IntervalStream).

## Adaptors covered

- `Map`, `Filter`, `FilterMap`, `Skip`, `SkipWhile` — delegate `is_terminated()` to inner stream
- `Take` — terminated when `remaining == 0`
- `TakeWhile` — terminated when internal `done` flag is set
- `Then` — terminated when inner stream is done and no future is in-flight
- `Chain`, `Merge` — terminated when both inner streams are terminated

## Intentionally excluded

`MapWhile` — its current `poll_next` does not track when the closure
returns `None` early (before the inner stream is exhausted), so a
correct `is_terminated()` is not possible without a separate semantic
change to the adaptor itself.

## Tests

Added 21 targeted tests in `tokio-stream/tests/stream_fused.rs`
covering `is_terminated()` transitions for each adaptor.

### DIFF
--- a/tokio-stream/src/stream_ext/filter.rs
+++ b/tokio-stream/src/stream_ext/filter.rs
@@ -60,8 +60,8 @@ where
 
 impl<St, F> FusedStream for Filter<St, F>
 where
-    Self: Stream,
     St: FusedStream,
+    F: FnMut(&St::Item) -> bool,
 {
     fn is_terminated(&self) -> bool {
         self.stream.is_terminated()
--- a/tokio-stream/src/stream_ext/filter.rs
+++ b/tokio-stream/src/stream_ext/filter.rs
@@ -60,8 +60,8 @@ where
 
 impl<St, F> FusedStream for Filter<St, F>
 where
+    Self: Stream,
     St: FusedStream,
-    F: FnMut(&St::Item) -> bool,
 {
     fn is_terminated(&self) -> bool {
         self.stream.is_terminated()
--- a/tokio-stream/src/stream_ext/filter.rs
+++ b/tokio-stream/src/stream_ext/filter.rs
@@ -3,6 +3,7 @@ use crate::Stream;
 use core::fmt;
 use core::pin::Pin;
 use core::task::{ready, Context, Poll};
+use futures_core::FusedStream;
 use pin_project_lite::pin_project;
 
 pin_project! {
@@ -54,5 +55,15 @@ where
 
     fn size_hint(&self) -> (usize, Option<usize>) {
         (0, self.stream.size_hint().1) // can't know a lower bound, due to the predicate
+    }
+}
+
+impl<St, F> FusedStream for Filter<St, F>
+where
+    St: FusedStream,
+    F: FnMut(&St::Item) -> bool,
+{
+    fn is_terminated(&self) -> bool {
+        self.stream.is_terminated()
     }
 }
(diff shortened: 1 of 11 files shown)


tokio-stream/tests/stream_fused.rs
```diff
@@ -0,0 +1,188 @@
...
+
+#[tokio::test]
+async fn take_terminated_when_remaining_zero() {
+    let mut stream = tokio_stream::iter(vec![1, 2]).take(2);
+    assert_eq!(stream.next().await, Some(1));
+    assert!(!stream.is_terminated());
+    assert_eq!(stream.next().await, Some(2));
+    // remaining hits 0 after getting the second item
+    assert_eq!(stream.next().await, None);
+    assert!(stream.is_terminated());
```
martin-g:
```suggestion
    assert!(stream.is_terminated());
    assert_eq!(stream.next().await, None);
```
swapping the assertions, so that it does not look like `is_terminated() == true` depends on the last call to `.next()`

martin-g (approved):

tokio-stream/src/stream_ext/filter.rs
```diff
@@ -56,3 +57,13 @@ where
         (0, self.stream.size_hint().1) // can't know a lower bound, due to the predicate
     }
 }
+
+impl<St, F> FusedStream for Filter<St, F>
+where
+    St: FusedStream,
+    F: FnMut(&St::Item) -> bool,
```
vilgotf, line 64: The `F` bound is not necessary. This applies to other adaptors as well.

vilgotf: Never mind, I forgot that `Stream` is a sub-trait of `FusedStream`. Adding `Self: Stream` breaks from `std` convention and looks a bit odd imo

/Repeated in a new thread for visibility./

tokio-stream/src/stream_ext/filter.rs
```diff
@@ -56,3 +57,13 @@ where
         (0, self.stream.size_hint().1) // can't know a lower bound, due to the predicate
     }
 }
+
+impl<St, F> FusedStream for Filter<St, F>
+where
+    Self: Stream,
```
vilgotf: Disregard my previous comment, I forgot that `Stream` is a sub-trait of `FusedStream`. Adding `Self: Stream` breaks from `std` convention and looks a bit odd imo.

tokio-stream/src/stream_ext/filter_map.rs
```diff
@@ -56,3 +57,13 @@ where
         (0, self.stream.size_hint().1) // can't know a lower bound, due to the predicate
     }
 }
+
+impl<St, F> FusedStream for FilterMap<St, F>
+where
+    Self: Stream,
```
vilgotf: Your ai left this `Self` bound (as well as for `Map`)…

Rachit2323: My bad , I removed Self: Stream from both Map and FilterMap as suggested. I believe we still need to keep F: FnMut(St::Item) -> Option<T> since it’s required by the Stream implementation. 
Correct ?

vilgotf: Yes

mattiapitossi (approved):

mattiapitossi:
Thanks for the PR! it seems that now there are some network hiccups in the CI, so it's unrelated to the CL. I'll merge when those are resolved

PR #8097, opened 2026-04-29
[chore] functional slice building

vilgotf:
Noticed some redundant procedural code.


martin-g (commented):
What is the reason at some places to use Range::map+collect and at others repeat_with+take ?

mattiapitossi (approved):

vilgotf:
> What is the reason at some places to use Range::map+collect and at others repeat_with+take ?

It's whether or not indices matter: `FnMut({Int}) -> T` vs `FnMut() -> T`.

martin-g (approved):

PR #8099, opened 2026-05-01
task: add `JoinMap::try_join_next`

jgrund:
## Motivation

`tokio_util::task::JoinMap` is a key-tracking wrapper around `JoinSet`. The
underlying `JoinSet` exposes both `join_next` and `try_join_next`, but
`JoinMap` only forwards the async `join_next`. That asymmetry forces callers
who need to drain completed tasks from a synchronous context — e.g. a
non-async `tick()` helper, or a non-async branch of a `select!`-driven event
loop — to pull in `futures-util` solely to call
`FutureExt::now_or_never(self.map.join_next())`.

There's no reason the wrapper should be more restrictive than the primitive
it wraps.

## Solution

Add `JoinMap::try_join_next`, mirroring the existing `join_next`
implementation but delegating to `JoinSet::try_join_next_with_id`. The
id↔key bookkeeping (looping past already-removed entries via
`remove_by_id`) is identical to the async path.

This intentionally does not add a `try_join_next_with_id` variant —
`JoinMap` does not currently expose `join_next_with_id` either, so adding
only the sync `_with_id` form would introduce an asymmetry on the JoinMap
surface. Both can be added together as a follow-up if a use case appears.

- Method documented with a rustdoc example.
- New unit tests in `tokio-util/tests/task_join_map.rs` cover:
  - `None` when the map is empty.
  - `None` when no spawned task has completed yet.
  - `Some((key, Ok(value)))` for a successfully-completed task.
  - `Some((key, Err(JoinError)))` for an aborted task.
  - Subsequent calls advance through additional completed tasks.

### DIFF
--- a/tokio-util/src/task/join_map.rs
+++ b/tokio-util/src/task/join_map.rs
@@ -468,6 +468,60 @@ where
         }
     }
 
+    /// Tries to join one of the tasks in the map that has completed and
+    /// returns its output, along with the key corresponding to that task.
+    ///
+    /// Returns `None` if there are no completed tasks, or if the map is empty.
+    ///
+    /// # Returns
+    ///
+    /// This function returns:
+    ///
+    ///  * `Some((key, Ok(value)))` if one of the tasks in this `JoinMap` has
+    ///    completed. The `value` is the return value of that task, and `key`
+    ///    is the key associated with the task.
+    ///  * `Some((key, Err(err)))` if one of the tasks in this `JoinMap` has
+    ///    panicked or been aborted. `key` is the key associated with the task
+    ///    that panicked or was aborted.
+    ///  * `None` if there are no completed tasks ready to be joined, or the
+    ///    `JoinMap` is empty.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use tokio_util::task::JoinMap;
+    ///
+    /// # #[tokio::main(flavor = "current_thread")]
+    /// # async fn main() {
+    /// let mut map = JoinMap::new();
+    /// map.spawn("answer", async { 42 });
+    ///
+    /// let (key, res) = loop {
+    ///     if let Some(joined) = map.try_join_next() {
+    ///         break joined;
+    ///     }
+    ///     tokio::task::yield_now().await;
+    /// };
+    ///
+    /// assert_eq!(key, "answer");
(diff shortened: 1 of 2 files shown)


tokio-util/src/task/join_map.rs
```diff
@@ -468,6 +468,60 @@ where
...
+            let (res, id) = match self.tasks.try_join_next_with_id()? {
+                Ok((id, output)) => (Ok(output), id),
+                Err(e) => {
+                    let id = e.id();
+                    (Err(e), id)
+                }
+            };
+            if let Some(key) = self.remove_by_id(id) {
+                break Some((key, res));
+            }
```
Darksonn, line 521: This check fails when the task with the given id has been replaced by a new task with that id (see `fn insert`). Please include a test that triggers this case and goes around the loop.

jgrund: Added `try_join_next_skips_replaced_task` in 46a12033 — spawns task A, replaces key 1 with task B, then asserts a single `try_join_next` call returns B. Verified the test fails when the loop is removed.

tokio-util/tests/task_join_map.rs
```diff
@@ -358,6 +358,131 @@ async fn abort_all() {
...
+
+    let (tx2, rx2) = oneshot::channel::<()>();
+    map.spawn(1, async {
+        let _ = rx2.await;
+        22
+    });
+    tx2.send(()).unwrap();
+    tokio::task::yield_now().await;
+
+    let (key, res) = map.try_join_next().unwrap();
```
martin-g, line 478: Wrap in a loop ?! To make the test more stable.
The other tests do it and the method docstring recommends it.

tokio-util/src/task/join_map.rs
martin-g:
```suggestion
    ///  * `Some((key, Err(err)))` if one of the tasks in this `JoinMap` has
```

tokio-util/tests/task_join_map.rs
jgrund: This test is to assert that we don't yield overwritten tasks. If we wrap in a loop and wait till we have `Some` then we will miss the case this test is trying to assert.

martin-g (approved):

jgrund:
Hi, can this be merged?

PR #8101, opened 2026-05-01
Fix CI on tokio-1.47.x

Darksonn:
Backport of #8041

PR #8103, opened 2026-05-02
ci: install valgrind via apt-get instead of install-action

Darksonn:
Avoids 'unable to contact snap store' errors on GitHub Actions runners.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -241,7 +241,7 @@ jobs:
             toolchain: 1.82
 
       - name: Install Valgrind
-        uses: taiki-e/install-action@valgrind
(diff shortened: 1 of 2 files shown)


Darksonn:
I saw https://github.com/taiki-e/install-action/commit/40088a0f3b939f11215a144aa0b25f96ddc00cda

I guess this is a temporary incident? I saw it yesterday also, though, so it's been going on for a while.

Darksonn:
Regardless, it seems harmless to install valgrind this way, so let's just keep it. cc @taiki-e 

Closing because this will land on master when https://github.com/tokio-rs/tokio/pull/8101 merges into master.

PR #8107, opened 2026-05-02
time: defer waker clone on spurious poll

vilgotf:
The alternative Timer eagerly cloned the incoming waker instead of
checking whether it is equal to the current waker.

### DIFF
--- a/tokio/src/runtime/time_alt/entry.rs
+++ b/tokio/src/runtime/time_alt/entry.rs
@@ -231,11 +231,13 @@ impl Handle {
     pub(crate) fn register_waker(&self, waker: &Waker) {
         let mut lock = self.entry.state.lock();
         if !lock.cancelled && !lock.woken_up {
+            // PANIC: no intermediary state is possible should the user-controllable `Waker`
+            // panic on `Clone` or `Drop`.
             let maybe_old_waker = match &lock.waker {
                 Some(current_waker) if current_waker.will_wake(waker) => None,
                 _ => lock.waker.replace(waker.clone()),
             };
-            // unlock before calling waker
+            // unlock before dropping waker
             drop(lock);
             drop(maybe_old_waker);
         }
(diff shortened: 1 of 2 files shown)


tokio/src/runtime/time_alt/entry.rs
```diff
@@ -228,10 +228,13 @@ impl Handle {
...
     }
 
-    pub(crate) fn register_waker(&self, waker: Waker) {
+    pub(crate) fn register_waker(&self, waker: &Waker) {
         let mut lock = self.entry.state.lock();
         if !lock.cancelled && !lock.woken_up {
-            let maybe_old_waker = lock.waker.replace(waker);
+            let maybe_old_waker = match &lock.waker {
+                Some(current_waker) if current_waker.will_wake(waker) => None,
+                _ => lock.waker.replace(waker.clone()),
```
ADD-SP, line 238: The `waker.clone()` is a tricky part. The `Clone` might be implemented by user code, so it might panic, this is the reason I clone the waker unconditionally outside of the `register_waker`.

So there is a trade-off I made before: avoding possible panic (in the tokio runtime) from user code at the expense of the performance.

What do you think? Maybe there is a better idea? Shall we add a comment here?

Darksonn: I don't believe you have made any changes yet, so a panic does not leave it in an inconsistent state as far as I can tell.

ADD-SP: Yes, the state is not changed, but panic can poison the Mutex, and handling a poisoned lock is always tricky, so I try to avoid it.

Darksonn: If you use the mutex re-export from `crate::loom` then there is no poisoning.

vilgotf: I don't think the panic matters, for the reasons stated above, but documenting possible user-panic locations and why they are safe (à la `SAFETY:`) seems like good practice.

ADD-SP (approved):
Thanks!

PR #8108, opened 2026-05-02
Route nto io-sock variants (nto71_iosock, nto80) to impl_noproc for u…

skarlsson:
…cred

QNX 8.0 ships only the io-sock network stack, which dropped both LOCAL_PEEREID and the unpcbid struct that the existing impl_netbsd implementation of get_peer_cred relies on. The same applies to QNX 7.1 with the optional io-sock stack. Peer credentials are simply not available on io-sock — the existing impl_noproc fallback (espidf, vita, hurd) is the right destination.

Splitting the four target_os = "nto" cfg sites (mod gate × 2, pub(crate) use dispatch × 2) keeps the legacy io-pkt envs (nto70, nto71) on impl_netbsd and routes io-sock envs (nto71_iosock, nto80) to impl_noproc.

### DIFF
--- a/tokio/src/net/unix/ucred.rs
+++ b/tokio/src/net/unix/ucred.rs
@@ -41,7 +41,13 @@ impl UCred {
 ))]
 pub(crate) use self::impl_linux::get_peer_cred;
 
-#[cfg(any(target_os = "netbsd", target_os = "nto"))]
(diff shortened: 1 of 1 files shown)


tokio/src/net/unix/ucred.rs
Darksonn, line 1: Are these platforms something that could be at least build tested in CI?

Darksonn (approved):

PR #8109, opened 2026-05-03
stream: implement `Peekable::size_hint`

Rachit2323:
## Motivation                                                                                            
                                                                                                           
`Peekable<T>` buffers one item in `self.peek` when `.peek()` is called,
  but had no `size_hint` implementation. This caused it to fall back to                                    
  the default `(0, None)`, which means `collect()` always pre-allocates                                    
  with capacity 0 and resizes on every push — even when the exact count                                    
  is known.  
  
 ## Solution                                           

  Add a `size_hint` impl that adds 1 to both the lower and upper bounds
  when `self.peek` is occupied. Uses `saturating_add` for the lower bound                                  
  and `checked_add` for the upper bound to handle overflow safely.
  The approach mirrors `std::iter::Peekable::size_hint`.                                                   
                                                        
  Added `tokio-stream/tests/stream_peekable.rs` with tests covering:                                       
  normal consumption, peek then consume, empty stream, double peek,
  and overflow safety.


tokio-stream/tests/stream_peekable.rs
```diff
@@ -0,0 +1,94 @@
...
+    assert_eq!(s.size_hint(), (0, Some(0)));
+}
+
+#[tokio::test]
+async fn peek_returns_correct_item() {
+    let mut s = stream::iter(vec![10, 20, 30]).peekable();
+
+    assert_eq!(s.peek().await, Some(&10));
+    assert_eq!(s.peek().await, Some(&10)); // second peek returns same item
+    assert_eq!(s.next().await, Some(10));  // next() gives same item
```
martin-g:
```suggestion
    assert_eq!(s.next().await, Some(10)); // next() gives same item
```

tokio-stream/tests/stream_peekable.rs
```diff
@@ -0,0 +1,94 @@
...
+    assert_eq!(s.next().await, Some(1));
+    assert_eq!(s.size_hint(), (2, Some(2)));
+
+    // peek again
+    let _ = s.peek().await;
+    assert_eq!(s.size_hint(), (2, Some(2)));
+
+    s.next().await;
+    assert_eq!(s.size_hint(), (1, Some(1)));
+
```
martin-g, line 36:
```suggestion
    // peek the last item
    let _ = s.peek().await;
    assert_eq!(s.size_hint(), (1, Some(1)));
```

tokio-stream/src/stream_ext/peekable.rs
```diff
@@ -47,4 +47,15 @@ impl<T: Stream> Stream for Peekable<T> {
...
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let peek_len = if self.peek.is_some() { 1 } else { 0 };
+        let (lo, hi) = self.stream.size_hint();
+        let lo = lo.saturating_add(peek_len);
+        let hi = match hi {
+            Some(x) => x.checked_add(peek_len),
+            None => None,
+        };
```
martin-g:
```suggestion
        let hi = hi.and_then(|x| x.checked_add(peek_len));
```

tokio-stream/tests/stream_peekable.rs
```diff
@@ -0,0 +1,94 @@
...
+    let mut s = MaxHint(true).peekable();
+
+    // before peek: delegates directly to inner
+    assert_eq!(s.size_hint(), (usize::MAX, Some(usize::MAX)));
+
+    // after peek: peek_len=1, checked_add(1) on usize::MAX must give None not panic
+    let _ = s.peek().await;
+    assert_eq!(s.size_hint(), (usize::MAX, None));
+}
```
martin-g, line 98:
```suggestion
}

#[tokio::test]
async fn size_hint_unbounded_upper() {
    // A stream that reports unknown upper bound
    struct Unbounded;
    impl Stream for Unbounded {
        type Item = u32;
        fn poll_next(self: std::pin::Pin<&mut Self>, _cx: &mut std::task::Context<'_>) -> std::task::Poll<Option<u32>> {
            std::task::Poll::Ready(Some(42))
        }
        fn size_hint(&self) -> (usize, Option<usize>) {
            (1, None)
        }
    }
    let mut s = Unbounded.peekable();
    assert_eq!(s.size_hint(), (1, None));
    let _ = s.peek().await;
    assert_eq!(s.size_hint(), (2, None)); // still unbounded after peek
}
```

martin-g (approved):

Rachit2323:
Friendly ping, anything left to do here?

Darksonn:
I'll rename this since it isn't a bugfix - the old value is not wrong just inefficient. Otherwise LGTM, merging now.

PR #8110, opened 2026-05-03
sync: implement `PartialEq` and `Eq` for `CancellationToken`

### LINKED ISSUE #8087: Provide a way to check if CancellationToken are the same
I want to check if the `CancellationToken` are the same one in the following code to catch a bug:

```rust
impl<'a> Drop for ApiGuard<'a> {
    fn drop(&mut self) {
        let s = match self
            .sessions
            .lock()
            .unwrap()
            .remove(&self.account_id)
            .unwrap()
        {
            Session::Game(_) => unreachable!(),
            Session::Api(v) => v,
        };

        assert_eq!(s, self.active); // This line does not compile.

        self.active.cancel();
    }
}
```

Provide method similar to `Arc::ptr_eq`.

Wrap `CancellationToken` with `Arc` and use `Arc::ptr_eq` to check. The problem with this is it create extra allocation.

tglane:
## Motivation
#8087 requests for a way to check if two `CancellationToken`s are equal/the same.

## Solution
Implementations for `PartialEq` and `Eq` were added to `CancellationToken` to check, whether two tokens point to the same tree node of the cancellation tree. With this two tokens are considered equal if cancelling one will always also cancel the other and vice versa.

Closes #8087

### DIFF
--- a/tokio-util/src/sync/cancellation_token.rs
+++ b/tokio-util/src/sync/cancellation_token.rs
@@ -122,6 +122,19 @@ impl Clone for CancellationToken {
     }
 }
 
+impl PartialEq for CancellationToken {
+    /// Checks if two tokens are equal in terms of their cancellation operation.
+    ///
+    /// Two tokens are considered equal if cancelling one will always also cancel the other and vice
+    /// versa. This is only true for cloned tokens and not for tokens in a parent-child
+    /// relationship.
+    fn eq(&self, other: &CancellationToken) -> bool {
+        Arc::ptr_eq(&self.inner, &other.inner)
+    }
+}
+
+impl Eq for CancellationToken {}
+
 impl Drop for CancellationToken {
     fn drop(&mut self) {
         tree_node::decrease_handle_refcount(&self.inner);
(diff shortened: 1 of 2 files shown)


tokio-util/src/sync/cancellation_token.rs
```diff
@@ -122,6 +122,18 @@ impl Clone for CancellationToken {
     }
 }
 
+impl PartialEq for CancellationToken {
+    /// Checks if two tokens are equal in terms of their cancellation operation.
+    ///
+    /// Two tokens are considered equal if cancelling one will always also cancel the other and vice
+    /// versa.
```
Darksonn: It would be nice if this section mentioned that parent/child relationships return false.

Darksonn (approved):

martin-g (approved):
LGTM!

Do we also want to add impl for `Hash` too ?

Darksonn:
I mean, sure, adding hash seems fine.

tglane:
Thanks for adding the impl for `Hash` @martin-g

Darksonn (approved):
Thanks.

PR #8119, opened 2026-05-04
sync: document memory ordering guarantees for Semaphore

cong-or:
I was looking at #7435 — the semaphore implementation already uses careful                                                                                                                 Acquire/Release/AcqRel orderings throughout batch_semaphore.rs, but none of                                                                                                                 
that is visible to users reading the public docs.                                                                                                                                           
                                                                                                                                                                                              
  This adds a "Memory ordering" section to the Semaphore rustdoc covering the                                                                                                                 
  guarantees @Darksonn outlined in the issue:

  - Successful acquisition is AcqRel
  - Releasing permits (drop, add_permits, forget_permits) and close are AcqRel
  - Failed attempts and available_permits/is_closed are Acquire loads

  Also included a plain-English summary: writes before a release are visible
  after a subsequent acquire. Felt worth spelling out since not everyone thinks
  in terms of atomic orderings.

  Single-file change, 30 lines of doc comments, no code changes

### DIFF
--- a/tokio/src/sync/semaphore.rs
+++ b/tokio/src/sync/semaphore.rs
@@ -28,23 +28,22 @@ use std::sync::Arc;
 ///
 /// # Memory ordering
 ///
-/// Successfully acquiring a permit (via [`acquire`], [`acquire_many`],
-/// [`try_acquire`], [`try_acquire_many`], or their `_owned` variants) performs
-/// an `AcqRel` operation on the semaphore's internal state. Similarly,
-/// releasing permits (by dropping a [`SemaphorePermit`] or
-/// [`OwnedSemaphorePermit`], or by calling [`add_permits`] or
-/// [`forget_permits`]) and closing the semaphore (via [`close`]) are also
-/// `AcqRel` operations.
+/// If a task writes some data and then releases a permit, any task that later
+/// acquires a permit is guaranteed to see that data. This makes it safe to use
+/// a semaphore to hand data off between tasks through shared state.
+///
+/// Stated more precisely in terms of atomic memory orderings: acquiring a
+/// permit (via [`acquire`], [`acquire_many`], [`try_acquire`],
+/// [`try_acquire_many`], or their `_owned` variants), releasing permits (by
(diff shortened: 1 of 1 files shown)


tokio/src/sync/semaphore.rs
```diff
@@ -26,6 +26,36 @@ use std::sync::Arc;
...
 ///
+/// # Memory ordering
+///
+/// Successfully acquiring a permit (via [`acquire`], [`acquire_many`],
+/// [`try_acquire`], [`try_acquire_many`], or their `_owned` variants) performs
+/// an `AcqRel` operation on the semaphore's internal state. Similarly,
+/// releasing permits (by dropping a [`SemaphorePermit`] or
+/// [`OwnedSemaphorePermit`], or by calling [`add_permits`] or
+/// [`forget_permits`]) and closing the semaphore (via [`close`]) are also
+/// `AcqRel` operations.
```
Darksonn: Instead of saying that it performs that operation on its internal state, I would much rather say that it has the same synchronization guarantees as an atomic. We don't want to promise that we're using an atomic internally (and in fact, there's a mutex internally, so it's not even true).

Or maybe it's better to just to say that these operations (that is, the `AcqRel` operations) will all be ordered and that each operation synchyronizes-with all previous operations.

Darksonn (commented):
Hey I'm sorry I missed this when it was originally posted. I have a few review comments below, but if you're no longer interested in working on it that's also okay and I'll do the fixups myself on top of your change.

tokio/src/sync/semaphore.rs
cong-or: @Darksonn  Good catch, thanks! Agreed that "internal state" wrongly implied there's an atomic inside.

Reworded it to lead with the plain-English idea first, data written before a release is visible to whoever acquires next; and kept the precise wording below: acquire/release/close are AcqRel, totally ordered, each synchronizing-with all prior ones, just like AcqRel on a single atomic. Happy to adjust the wording if you'd prefer.

Darksonn (approved):
Thanks.

PR #8076, opened 2026-04-20
Require that an `RwLock` has `max_readers != 0`

Darksonn:
If an `RwLock` is created with max readers equal to zero, then write locks do not ensure mutual exclusion because they acquire 0 permits. This can be abused to trigger UB.
```rs
let v = RwLock::with_max_readers(vec![1], 0);
let mut w1 = v.write().await;
let w2 = v.write().await;
let one = &w2[0];
*w1 = Vec::new();
println!("{one}"); // UAF
```
This bug was discovered by asking Gemini to look for bugs in the `with_max_readers` method of `tokio/src/sync/rwlock.rs`.

PR #8122, opened 2026-05-07
Prepare Tokio v1.47.5

Darksonn:
# 1.47.5 (May 7th, 2026)

### Fixed

* sync: fix underflow in mpsc channel `len()` ([#8062])
* sync: notify receivers in mpsc `OwnedPermit::release()` method ([#8075])
* sync: require that an `RwLock` has `max_readers != 0` ([#8076])
* sync: return `Empty` from `try_recv()` when mpsc is closed with outstanding permits ([#8074])

[#8062]: https://github.com/tokio-rs/tokio/pull/8062
[#8074]: https://github.com/tokio-rs/tokio/pull/8074
[#8075]: https://github.com/tokio-rs/tokio/pull/8075
[#8076]: https://github.com/tokio-rs/tokio/pull/8076

### DIFF
--- a/tokio/Cargo.toml
+++ b/tokio/Cargo.toml
@@ -6,7 +6,7 @@ name = "tokio"
 #   - README.md
 # - Update CHANGELOG.md.
 # - Create "v1.x.y" git tag.
-version = "1.47.4"
+version = "1.47.5"
 edition = "2021"
 rust-version = "1.70"
 authors = ["Tokio Contributors <team@tokio.rs>"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1050,9 +1050,9 @@ checksum = "69cdb34c158ceb288df11e18b4bd39de994f6657d83847bdffdbd7f346754b0f"
 
 [[package]]
 name = "rand"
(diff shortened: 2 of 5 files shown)


tokio/Cargo.toml
martin-g, line 4: https://github.com/tokio-rs/tokio/pull/8122/changes#diff-f4bfdbfbf2e4c7b2036d9a38e3e99cf403fb50ce4f319c12552bc02e06e4fee7R89 still has a `path` dependency for `tokio-macros`

mattiapitossi: This was prior to the path fix, I think removing the path will create the ambiguity between packages?

mattiapitossi (approved):

tokio/Cargo.toml
Darksonn: Yeah, the path dep situation was a bit broken at the time.

martin-g: I thought that was improved only in more recent versions, not in the 1.47.x branch.
If this step is no more needed then maybe remove line 4 ?!

PR #8124, opened 2026-05-07
time: consolidate mutex locks on spurious poll

vilgotf:
The match guard (`is_woken_up`) and fallback (`register_waker`) both acquired the mutex separately.


martin-g (approved):

ADD-SP (approved):
Thanks!

PR #8126, opened 2026-05-08
tracing: remove unnecessary span clone

vilgotf:
Span cloning is left as is in `BatchSemaphore` due to its `unsafe` usage (though it should be possible to change this too).


Darksonn (approved):
Thanks! Sorry about the delay for this one.

PR #8128, opened 2026-05-08
time: clean up Instant overflow prevention

vilgotf:
The `Instant::far_future` ctor unnecessarily called `Instant::now` (again). Shrinking `Duration` before calling `Instant::add` achieves the same result more succinctly.

Also fixed `timeout_at` using the wrong `Location` for its delay.

### DIFF
--- a/tokio/src/time/interval.rs
+++ b/tokio/src/time/interval.rs
@@ -1,4 +1,4 @@
-use crate::time::{sleep_until, Duration, Instant, Sleep};
+use crate::time::{safe_delay, sleep_until, Duration, Instant, Sleep};
 use crate::util::trace;
 
 use std::future::{poll_fn, Future};
@@ -139,7 +139,7 @@ fn internal_interval_at(
 
     Interval {
         delay,
-        period,
+        period: safe_delay(period),
         missed_tick_behavior: MissedTickBehavior::default(),
         #[cfg(all(tokio_unstable, feature = "tracing"))]
         resource_span,
@@ -479,9 +479,7 @@ impl Interval {
             self.missed_tick_behavior
                 .next_timeout(timeout, now, self.period)
         } else {
-            timeout
-                .checked_add(self.period)
-                .unwrap_or_else(Instant::far_future)
+            timeout + self.period
         };
 
         // When we arrive here, the internal delay returned `Poll::Ready`.
--- a/tokio/src/time/interval.rs
+++ b/tokio/src/time/interval.rs
@@ -351,7 +351,7 @@ impl MissedTickBehavior {
                             "too much time has elapsed since the interval was supposed to tick",
                         ),
                 );
-                now + period - offset
+                now + (period - offset)
             }
         }
     }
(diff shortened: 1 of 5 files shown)


vilgotf:
Supersedes #7193

tokio/src/time/sleep.rs
```diff
@@ -121,12 +121,8 @@ pub fn sleep_until(deadline: Instant) -> Sleep {
 #[cfg_attr(docsrs, doc(alias = "wait"))]
 #[track_caller]
 pub fn sleep(duration: Duration) -> Sleep {
-    let location = trace::caller_location();
-
-    match Instant::now().checked_add(duration) {
-        Some(deadline) => Sleep::new_timeout(deadline, location),
-        None => Sleep::new_timeout(Instant::far_future(), location),
-    }
```
Darksonn, line 129: I think I would prefer to keep the `checked_add()` logic, to always perform the addition correctly when at all possible.

vilgotf: The previous code has no "correct" addition, even though it looks like it. This becomes clear when you inline `Instant::far_future()`:

```rust
const SAFE_DELAY: Duration = Duration::from_secs(86400 * 365 * 30);
let deadline = Instant::now()
    .checked_add(duration)
    .unwrap_or_else(|| Instant::now().checked_add(SAFE_DELAY).unwrap());
```

This PR changes the above to:

```rust
const SAFE_DELAY: Duration = Duration::from_secs(86400 * 365 * 30);
let deadline = Instant::now().checked_add(duration.min(SAFE_DELAY)).unwrap();
```

Darksonn:
Hey, sorry for getting away for this for so long. I'm going through the list of PRs now. I think this looks reasonable, but unfortunately it looks to be out-of-date now. Do you want to rebase it?

vilgotf:
Rebased. Note that this is PR employs a different solution to #8385 and therefore undoes some of its changes.

tokio/src/time/interval.rs
```diff
@@ -661,7 +650,7 @@ mod tests {
             MissedTickBehavior::Delay,
             MissedTickBehavior::Skip,
         ] {
-            let next = behavior.next_timeout(timeout, now, Duration::MAX);
+            let next = behavior.next_timeout(timeout, now, safe_delay(Duration::MAX));
```
vilgotf: Thus unit test (with the changes made to this file) looks like it's already covered by the integration tests. I would like to remove it, if you agree

tokio/src/time/interval.rs
```diff
@@ -357,7 +351,7 @@ impl MissedTickBehavior {
                             "too much time has elapsed since the interval was supposed to tick",
                         ),
                 );
-                saturating_add(now, period - offset)
+                now + period - offset
```
Darksonn:
```suggestion
                now + (period - offset)
```

tokio/src/time/interval.rs
Darksonn, line 388: Here's a thought. If we made this into an `Option<Pin<Box<Sleep>>>`, then we can handle overflow by simply storing `None` and treating that as an infinite duration. In that case, we do not pay the cost of registering a timer at all.

tokio/src/time/interval.rs
Darksonn: I do generally prefer integration tests, so if there's an existing test that covers this, that's fine.

tokio/src/time/interval.rs
vilgotf: That is unfortunately not possible because timers are observable through the `reset*` methods. Basically it's the same issue as #8137/#8198, where a task waiting on `Interval::tick` must be woken on another task calling `Interval::reset`.

I applied the suggestion to `Timeout`, however.

Darksonn: I see. Then we'd have to adjust `Sleep` itself to get access to its waker which seems more tricky. Though not impossible.

vilgotf: I'll leave that for another PR, if you don't mind.

Darksonn (approved):
Thanks.

PR #1800, opened 2019-11-20
time: Eagerly bind delays to timer

kleimkuhler:
## Motivation

Similar to #1666, it is no longer necessary to lazily register delays with the
executions default timer. All delays are expected to be created from within a
runtime, and should panic if not done so.

## Solution

`tokio::time` now assumes there to be a `CURRENT_TIMER` set when creating a
delay; this can be assumed if called within a tokio runtime. If there is no
current timer, the application will panic with a "no current timer" message.

## Follow-up

Similar to #1666, `HandlePriv` can probably be removed, but this mainly prepares
for 0.2 API changes. Because it is not in the public API, this can be done in a
following change.

Signed-off-by: Kevin Leimkuhler <kleimkuhler@icloud.com>

PR #8132, opened 2026-05-08
time: move lazy-registration state into Sleep

vilgotf:
## Motivation

Timers are a complex implementation detail of the Sleep future and should therefore be simple. Moving the lazy-registration state into Sleep reduces Timer responsibilities and correctly synchronizes Tracing event emission.

## Solution

Moves the lazy-registration state from Timer into Sleep. Timer initialization becomes a 2-step process (à la the `pin_init` crate): `Timer::init(self: Pin<&mut Self>)` __must__ be called after construction.

### DIFF
--- a/tokio/src/runtime/mod.rs
+++ b/tokio/src/runtime/mod.rs
@@ -416,7 +416,7 @@ cfg_process_driver! {
     mod process;
 }
 
-#[cfg_attr(not(feature = "time"), allow(dead_code))]
+#[allow(dead_code)]
 #[derive(Debug, Copy, Clone, PartialEq)]
 pub(crate) enum TimerFlavor {
     Traditional,
--- a/tokio/src/runtime/mod.rs
+++ b/tokio/src/runtime/mod.rs
@@ -514,6 +514,14 @@ cfg_time! {
                 }
             }
         }
+
+        pub(crate) fn handle(&self) -> &scheduler::Handle {
+            match self {
+                Timer::Traditional(entry) => &entry.driver,
+                #[cfg(all(tokio_unstable, feature = "rt-multi-thread"))]
+                Timer::Alternative(entry) => &entry.driver,
+            }
+        }
     }
 }
 
--- a/tokio/src/runtime/mod.rs
+++ b/tokio/src/runtime/mod.rs
@@ -514,14 +514,6 @@ cfg_time! {
                 }
             }
         }
-
-        pub(crate) fn handle(&self) -> &scheduler::Handle {
-            match self {
-                Timer::Traditional(entry) => &entry.driver,
-                #[cfg(all(tokio_unstable, feature = "rt-multi-thread"))]
-                Timer::Alternative(entry) => &entry.driver,
-            }
-        }
     }
 }
 
--- a/tokio/src/runtime/mod.rs
+++ b/tokio/src/runtime/mod.rs
@@ -430,8 +430,6 @@ cfg_time! {
     #[cfg(all(tokio_unstable, feature = "rt-multi-thread"))]
     pub(crate) mod time_alt;
 
-    use crate::time::Instant;
-
     use std::task::{Context, Poll};
     use std::pin::Pin;
 
@@ -446,8 +444,7 @@ cfg_time! {
     impl Timer {
         #[cfg_attr(not(all(tokio_unstable, feature = "rt-multi-thread")), allow(unused_variables))]
         #[track_caller]
-        pub(crate) fn new(deadline: Instant) -> Self {
-            let handle = scheduler::Handle::current();
+        pub(crate) fn new(handle: scheduler::Handle, deadline: u64) -> Self {
             match handle.timer_flavor() {
                 TimerFlavor::Traditional => {
                     Timer::Traditional(time::TimerEntry::new(handle))
@@ -459,7 +456,7 @@ cfg_time! {
             }
         }
 
-        pub(crate) fn init(self: Pin<&mut Self>, deadline: Instant) {
+        pub(crate) fn init(self: Pin<&mut Self>, deadline: u64) {
             // Safety: we never move the inner entries.
             let this = unsafe { self.get_unchecked_mut() };
             match this {
@@ -480,15 +477,8 @@ cfg_time! {
             }
         }
 
-        pub(crate) fn flavor(self: Pin<&Self>) -> TimerFlavor {
-            match self.get_ref() {
-                Timer::Traditional(_) => TimerFlavor::Traditional,
-                #[cfg(all(tokio_unstable, feature = "rt-multi-thread"))]
-                Timer::Alternative(_) => TimerFlavor::Alternative,
-            }
-        }
-
-        pub(crate) fn reset(self: Pin<&mut Self>, deadline: Instant) {
+        #[cfg_attr(not(all(tokio_unstable, feature = "rt-multi-thread")), allow(unused_variables))]
+        pub(crate) fn reset(self: Pin<&mut Self>, handle: scheduler::Handle, deadline: u64) {
             // Safety: we never move the inner entries.
             let this = unsafe { self.get_unchecked_mut() };
             match this {
@@ -496,8 +486,11 @@ cfg_time! {
                 Timer::Traditional(entry) => unsafe {
                     Pin::new_unchecked(entry).reset(deadline)
                 }
+                // Safety: we never move the inner entries.
                 #[cfg(all(tokio_unstable, feature = "rt-multi-thread"))]
-                Timer::Alternative(_) => panic!("not implemented yet"),
+                Timer::Alternative(entry) => unsafe {
+                    Pin::new_unchecked(entry).set(time_alt::Timer::new(handle, deadline))
+                },
             }
         }
 
--- a/tokio/src/runtime/mod.rs
+++ b/tokio/src/runtime/mod.rs
@@ -430,6 +430,8 @@ cfg_time! {
     #[cfg(all(tokio_unstable, feature = "rt-multi-thread"))]
     pub(crate) mod time_alt;
 
+    use crate::time::Instant;
+
     use std::task::{Context, Poll};
     use std::pin::Pin;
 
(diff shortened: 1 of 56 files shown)


tokio/src/time/sleep.rs
```diff
@@ -251,18 +252,10 @@ impl Sleep {
         deadline: Instant,
         location: Option<&'static Location<'static>>,
     ) -> Sleep {
-        use crate::runtime::scheduler;
-        let handle = scheduler::Handle::current();
-        let entry = Timer::new(handle, deadline);
+        // Panic if the time driver is not enabled (backwards compat)
+        _ = scheduler::Handle::current().driver().time();
```
vilgotf: It is unclear to me whether the documentation for calling `sleep` outside a runtime with `time` enabled is _guaranteed_ to panic or a helpful description of the current behaviour. I left it as is, but it would be nice to drop.

tokio/src/time/sleep.rs
```diff
@@ -349,74 +345,44 @@ impl Sleep {
...
-            }
-            #[cfg(all(tokio_unstable, feature = "rt-multi-thread"))]
-            crate::runtime::TimerFlavor::Alternative => {
-                let handle = me.entry.as_ref().scheduler_handle().clone();
-                me.entry.set(Timer::new(handle, deadline));
-            }
-        }
+        let handle = scheduler::Handle::current();
+        let time_source = handle.driver().time().time_source();
+        let deadline = time_source.deadline_to_tick(deadline);
```
vilgotf, line 349: I originally added an early return here if the deadline is already elapsed, but that is a breaking change (it broke some test in `DelayQueue`)

edit: re-landed in #8137

tokio/src/runtime/mod.rs
```diff
@@ -416,7 +416,7 @@ cfg_process_driver! {
     mod process;
 }
 
-#[cfg_attr(not(feature = "time"), allow(dead_code))]
+#[allow(dead_code)]
```
vilgotf, line 419: This inexplicably shows up as dead code in CI

tokio/src/time/sleep.rs
```diff
@@ -349,74 +349,50 @@ impl Sleep {
...
 
-            let duration = {
-                let clock = me.entry.as_ref().clock();
-                let time_source = me.entry.as_ref().driver().time_source();
-                let now = time_source.now(clock);
-                let deadline_tick = time_source.deadline_to_tick(deadline);
-                deadline_tick.saturating_sub(now)
-            };
+            let _resource_enter = this.inner.ctx.resource_span.enter();
+            let _async_op_enter = this.inner.ctx.async_op_span.enter();
```
ADD-SP: This reverses the order of `tracing::trace_span!` and `this.inner.ctx.*_span.enter()`? Is this intentional?

tokio/tests/rt_metrics.rs
```diff
@@ -131,6 +131,7 @@ fn worker_park_count() {
     let rt = current_thread();
     let metrics = rt.metrics();
     rt.block_on(async {
+        time::pause();
```
ADD-SP: Emm, why do we need this? Is it because this fixes the test failure?

vilgotf: My understanding is that the deadline elapses between the calls to `sleep()` and `poll` otherwise (when run with miri)

tokio/src/time/sleep.rs
vilgotf: I personally prefer this ordering, but it's somewhat out of scope of this PR. I searched through Tokio for precedence, and `Sleep` is the only resource with span updates

ADD-SP (commented):
Could you resolve the merge conflict?

tokio/tests/rt_metrics.rs
ADD-SP: So it is flaky?

ADD-SP (commented):
Overall looks good to me, just a minor question.

tokio/tests/rt_metrics.rs
vilgotf: No. The timer is now "more" lazy, so the above is not possible with the old code. (See the early `Ready` return in `poll`).

Edit: actually, the early return is not sound as `is_elapsed` would return false

tokio/tests/time_sleep.rs
```diff
@@ -10,15 +10,19 @@ use futures::task::noop_waker_ref;
...
-    time::sleep_until(now).await;
-    assert_elapsed!(now, ms(1));
+    let sleep = time::sleep_until(now);
+
+    tokio::pin!(sleep);
+
+    assert!(!sleep.is_elapsed());
+
+    sleep.as_mut().await;
+    assert_elapsed!(now, ms(0));
```
vilgotf, line 24: w/ `start_paused`, there's no race to call `time::pause`, so `Instant::now` is exactly 0. Internally, the sleep deadline is equal to the wheel's `elapsed` field, which causes the timer fire immediately.

tokio/src/time/sleep.rs
vilgotf: This behaviour was added in https://github.com/tokio-rs/tokio/pull/1800 as part of a clean up. I think it's okay to remove (but not this late in this PR)

tokio/tests/time_sleep.rs
ADD-SP: Just a question:

Are these two assertions flaky during your development process?

```rust
    assert_elapsed!(now, ms(0));
```

```rust
    assert_elapsed!(now, ms(1));
```

It looks like a potential flaky test, not sure if you have already observed it.

tokio/tests/tracing_time.rs
```diff
@@ -40,31 +46,35 @@ async fn test_sleep_creates_span() {
...
-        .drop_span(async_op_poll_span)
+        .exit(sleep_span.clone())
         .drop_span(sleep_span)
         .run_with_handle();
 
     {
         let _guard = tracing::subscriber::set_default(subscriber);
 
-        _ = tokio::time::sleep(Duration::from_millis(7));
+        tokio::time::sleep(Duration::from_millis(7)).await;
```
ADD-SP, line 77: This change looks interesting, it looks like the old test never poll the `Sleep` future. Does it mean the old test never be executed?

ADD-SP (commented):
Sorry for the late review, just two questions for tests.

tokio/tests/time_sleep.rs
vilgotf: The first one is not possible due to time being paused on creation. The second one is in practice not possible (I think). In general, we should migrate from `time::pause` to `start_paused = true`

tokio/tests/tracing_time.rs
vilgotf: The old test (correctly) asserted that _creation_ emitted an event. However, the event is now deferred so the test has to now drive the future.

tokio/src/time/sleep.rs
```diff
@@ -349,74 +349,47 @@ impl Sleep {
...
```
ADD-SP: Emm, if I understand correctly, we `.enter()` twice times for both `resource_span` and `async_op_span`. Any reason for this? Did I miss something?

tokio/tests/time_sleep.rs
ADD-SP: > In general, we should migrate from time::pause to start_paused = true

I agree.

> The first one is not possible due to time being paused on creation. 

You are right.

tokio/src/time/sleep.rs
vilgotf: oops! good catch

ADD-SP (commented):
Overall looks good to me. Could you sync changes from the base branch? And then we can merge it after passing the CI.

vilgotf:
Actually, I introduced a bug in `Sleep::reset` when called from another runtime.

Timer ticks are relative to the runtime's `start_time`, so timers must also be extended from the same "tick-space". `Sleep::reset`, however, resolved ticks using the _current_ runtime instead of the existing timer's runtime (which might have different `start_time`s!).

I think this also deserves a regression test.

---

My plan is to revert timers from operating on `u64` back to `Instant`. Timers must already store a runtime handle for `Drop` cancellation, so we can reuse that handle for tick resolution. This prevents "tick-space" issues at the type level.

tokio/src/time/sleep.rs
```diff
@@ -300,7 +288,11 @@ impl Sleep {
         #[cfg(not(all(tokio_unstable, feature = "tracing")))]
         let inner = Inner {};
 
-        Sleep { inner, entry }
+        Sleep {
+            deadline,
+            inner,
+            timer: None,
```
ADD-SP, line 295: This changes the timing we save the runtime handle. Previous, we save the handle when creating the timer. In this PR, we save the handle during the `poll`. This is a potential breaking change.

vilgotf: It is possible, but more complicated, to retain the handle on creation. However, I eventually wish to defer handle acquisition to `poll`/`reset` (such that `block_on(sleep())` works without an intermediate async-block).

I don't know Tokio's stance on Hyrum's law, I figured it would be okay to try this change and see whether someone is actually affected.

edit: the diff is not too horrible with some `Clone` calls sprinkled in

ADD-SP: I hope we can do the right thing especially for cleaner implementation. However, the Hyrum's law is unforgiving. There is an example <https://github.com/tokio-rs/tokio/pull/5280> where a tiny changes causes a breaking change.

So unfortunately, we have to be conservative about potential breaking changes.

tokio/tests/time_rt.rs
```diff
@@ -164,3 +166,31 @@ fn timeout_value() {
...
+    let mut sleep = std::pin::pin!(sleep(Duration::from_millis(1)));
+    assert_pending!(sleep.as_mut().poll(&mut noop_context()));
+
+    let deadline = sleep.deadline();
+    rt.block_on(async { sleep.as_mut().reset(deadline + Duration::from_millis(1)) });
+
+    let now = Instant::now();
+    rt_past.block_on(sleep);
+    assert_eq!(now.elapsed(), Duration::from_millis(2));
+}
```
ADD-SP, line 196: TODO: add comments to explain this test, it is hard for me to understand it within one minutes.

Not a merge blocker.

ADD-SP (approved):
Thanks!

PR #8133, opened 2026-05-08
chore: replace `RunResult` with `ControlFlow`

vilgotf:


tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -249,11 +250,6 @@ pub(crate) struct Context {
 /// Starts the workers
 pub(crate) struct Launch(Vec<Arc<Worker>>);
 
-/// Running a task may consume the core. If the core is still available when
-/// running the task completes, it is returned. Otherwise, the worker will need
-/// to stop processing.
```
hawkw, line 254: I think it's unfortunate this documentation was removed. What do you think of adding a comment explaining this on the `run` function?

mattiapitossi:
it seems there’s a spurious failure in FreeBSD for a test that uses tracing mock:

```
thread 'test_oneshot_creates_span' (182990) panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tracing-mock-0.1.0-beta.1/src/subscriber.rs:1051:36:
  called `Result::unwrap()` on an `Err` value: PoisonError { .. } 
```

https://github.com/tokio-rs/tokio/actions/runs/25592508695/attempts/1?pr=8133

I’m trying to run it again

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -558,7 +552,7 @@ fn run(worker: Arc<Worker>) {
 }
 
 impl Context {
-    fn run(&self, mut core: Box<Core>) -> RunResult {
+    fn run(&self, mut core: Box<Core>) -> ControlFlow<(), Box<Core>> {
```
Darksonn: We can use an empty enum here to indicate that the continue case is never used.

vilgotf:
Rebased

tokio/src/runtime/scheduler/multi_thread/worker.rs
```diff
@@ -560,9 +557,7 @@ fn run(worker: Arc<Worker>) {
         context::set_scheduler(&cx, || {
             let cx = cx.expect_multi_thread();
 
-            // This should always be an error. It only returns a `Result` to support
-            // using `?` to short circuit.
-            assert!(cx.run(core).is_err());
+            _ = cx.run(core);
```
Darksonn: Honestly, I would be ok with just dropping the `ControlFlow` return value from `run` entirely. We only make use of the `?` operator in two locations, so we can do a match on the control flow enum in those two places inside `run`, which I think is readable enough.

Darksonn (approved):
Thanks.

PR #8134, opened 2026-05-09
fs(tests): improve io_uring tests

mattiapitossi:
## Motivation

Currently we are running all fs tests in a Kernel version where io_uring is not supported to check if the fallback mechanism works. This comes with the limitation that each io_uring test cannot perform some checks (e.g., assert_pending). 

I don't think it's necessary to check the fallback mechanism logic in every tests, so I was thinking to not run some tests where io_uring is not supported (or probably run it anyway and skip the assert pending check depending on the Kernel).

See also: https://github.com/tokio-rs/tokio/pull/8080#issue-4300800008

## Solution

Introduce a check in some tests to see if io_uring is supported, otherwise skip the test. I was thinking that another solution would be probably to skip the tests if the kernel is <5.1, but probably the check if io_uring is supported it's better since it shows the intent and it's not hidden behind a filter in the CI.

### DIFF
--- a/.github/workflows/uring-kernel-version-test.yml
+++ b/.github/workflows/uring-kernel-version-test.yml
@@ -40,6 +40,8 @@ jobs:
           make defconfig
           make -j$(nproc)
 
+      # We are running tests also on a Kernel version that does not support io_uring
+      # to check if the fallback mechanism works
       - name: Generate test binaries with io_uring enabled
         run: |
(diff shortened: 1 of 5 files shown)


Darksonn (approved):
The oldest supported Linux LTS is 5.10. I think it's fine for us to not support io-uring on unsupported Linux versions.

PR #8135, opened 2026-05-09
refactor the spawn blocking queue to make adding a new sharded implementation easy

alex:
refs #8067

This is easiest to review with with whitespace-only changes hidden (`?w=1` in github, `-w` in `git diff`).

### DIFF
--- a/tokio/src/runtime/blocking/pool.rs
+++ b/tokio/src/runtime/blocking/pool.rs
@@ -74,11 +74,8 @@ impl SpawnerMetrics {
 }
 
 struct Inner {
-    /// State shared between worker threads.
-    shared: Mutex<Shared>,
-
-    /// Pool threads wait on this.
-    condvar: Condvar,
+    /// Queue + notification implementation.
+    impl_: InnerImpl,
 
     /// Spawned threads use this name.
     thread_name: ThreadNameFn,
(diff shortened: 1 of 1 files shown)


tokio/src/runtime/blocking/pool.rs
```diff
@@ -493,105 +492,202 @@ cfg_unstable_metrics! {
...
+        on_no_idle: F,
+    ) -> Result<(), SpawnError>
+    where
+        F: FnOnce(&mut Shared) -> Result<(), SpawnError>,
+    {
+        match self {
+            InnerImpl::Locked(l) => {
+                let mut shared = l.inner.lock();
+
+                if shared.shared.shutdown {
```
martin-g: nit: The naming is becoming confusing :-/ 
`shared.shared` (and `inner.inner`) suggests that something is over-abstracted or they just need more descriptive names.

tokio/src/runtime/blocking/pool.rs
```diff
@@ -493,105 +492,202 @@ cfg_unstable_metrics! {
...
-        let mut is_counted_idle;
+    /// Run a worker thread's main loop.
+    fn run_worker(
+        &self,
+        metrics: &SpawnerMetrics,
+        keep_alive: Duration,
+        worker_thread_id: usize,
+    ) -> Option<thread::JoinHandle<()>> {
+        match self {
+            InnerImpl::Locked(l) => {
```
martin-g: Maybe extract the logic below to `LockedImpl::run_worker(...)` ?!
Adding later more InnerImpl variants will be simpler and keep the complexity of `run_worker()` (line 546) low.

It will make this PR's diff bigger but I think it's worth it.

tokio/src/runtime/blocking/pool.rs
```diff
@@ -74,11 +74,8 @@ impl SpawnerMetrics {
 }
 
 struct Inner {
-    /// State shared between worker threads.
-    shared: Mutex<Shared>,
-
-    /// Pool threads wait on this.
-    condvar: Condvar,
+    /// Queue + notification implementation.
+    impl_: InnerImpl,
```
martin-g: nit: Why the trailing `_` ? Because it collides with a keyword.
Maybe `inner_impl` ?!

martin-g (approved):

tokio/src/runtime/blocking/pool.rs
alex: Attempted a rename to improve this -- I didn't see any obvious places to remove any indirection (but please shout if you see one)

alex:
Sorry it took me a few days to respond to the feedback

tokio/src/runtime/blocking/pool.rs
```diff
@@ -102,9 +99,35 @@ struct Inner {
...
+    mutex: Mutex<LockedInner>,
+    condvar: Condvar,
+}
+
+struct LockedInner {
     queue: VecDeque<Task>,
     num_notify: u32,
+    /// Thread-management state. Split apart so that it's re-usable in a
+    /// sharded queue implementation.
+    shared: Shared,
```
Darksonn: Usually the field called "shared" is outside of the mutex, since stuff inside the mutex is what we have exclusive, not shared, access to. What's this field for?

Darksonn (commented):
Overall looks ok to me.

tokio/src/runtime/blocking/pool.rs
alex: renamed to thread_mgmt_state

alex:
For clarity: Is this waiting on me, or reviewing bandwidth? (I think it's the latter, but want to make sure no one is waiting on me!)

alex:
@martin-g If this looks good to you, do you mind merging? (Hoping to get back to the sharded impl)

Darksonn:
Sorry I didn't realize this hadn't been merged.

alex:
thanks much!

PR #2290, opened 2020-03-02
Notify task when resetting a Delay to a time in the past

thomaswhiteway:
If a Delay has been polled then the task that polled it may be waiting
for a notification.  If the delay gets reset to a time in the past, then
it immediately becomes elapsed, so should notify the relevant task.

## Motivation

Under certain conditions `DelayQueue` may update the current deadline to be in the past.  The delay would immediately transition to elapsed, but wouldn't notify the appropriate task so the scheduler would never re-poll it.  

## Solution

When resetting the time on a `Delay`, if the `Entry` is transitioning into `ELAPSED` state call the waker to wake any registered task.

PR #8137, opened 2026-05-10
time: skip fired Timer registration

vilgotf:
## Motivation

Timer firing and registration is expensive. Adding an elapsed Sleep state enables skipping fired timer registration as well as reduce Timer complexity.

## Solution

Add an elapsed Sleep state. Make Timer creation and extension fallible for past deadlines. Make Timer firing a terminal state (don't reregister fired timers, but create new ones) and clean up dead code.

### Compatibility Notes

- Calling `Sleep::reset` with an elapsed deadline no longer spuriously wakes the previously registered waker.
- Calling `Sleep::poll` for an elapsed deadline no longer spuriously wakes the waker.
- Calling `Sleep::poll` for an elapsed deadline no longer spuriously returns `Poll::Pending` if it previously returned `Poll::Ready` (i.e. it is Fused).

### DIFF
--- a/tokio/src/time/sleep.rs
+++ b/tokio/src/time/sleep.rs
@@ -227,8 +227,15 @@ pin_project! {
         driver: scheduler::Handle,
         inner: Inner,
         #[pin]
-        timer: Option<Timer>,
+        state: SleepState,
     }
+}
+
+#[derive(Debug)]
+enum SleepState {
+    Active { timer: Timer },
+    Elapsed,
+    Inactive,
 }
 
 cfg_trace! {
@@ -292,7 +299,7 @@ impl Sleep {
             deadline,
             driver: handle,
             inner,
-            timer: None,
+            state: SleepState::Inactive,
         }
     }
 
@@ -309,7 +316,11 @@ impl Sleep {
     ///
     /// A `Sleep` instance is elapsed when the requested duration has elapsed.
     pub fn is_elapsed(&self) -> bool {
-        self.timer.as_ref().is_some_and(Timer::is_elapsed)
+        match &self.state {
+            SleepState::Active { timer } => timer.is_elapsed(),
+            SleepState::Elapsed => true,
+            SleepState::Inactive => false,
+        }
     }
 
     /// Resets the `Sleep` instance to a new deadline.
@@ -369,12 +380,12 @@ impl Sleep {
             );
         }
 
-        match this.timer.as_mut().as_pin_mut() {
+        match this.state.as_mut().as_timer() {
             Some(timer) => timer.reset(handle.clone(), deadline),
             None => {
                 let timer = Timer::new(handle.clone(), deadline);
-                this.timer.set(Some(timer));
-                this.timer.as_pin_mut().unwrap().init(deadline);
+                this.state.set(SleepState::Active { timer });
+                this.state.as_timer().unwrap().init(deadline);
             }
         }
     }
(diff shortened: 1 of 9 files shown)


Darksonn:
This is a very large PR and most of the change is in one big commit, so it's going to be quite difficult to review. I recommend splitting this change up into smaller changes that are individually easier to review.

vilgotf:
This is a stacked PR and is not really meant for review until after its dependent PR is merged. I will mark this as a draft for now.

> I recommend splitting this change up into smaller changes that are individually easier to review.

I tried splitting the commit, but it is not really possible. Still, I managed to reduce the diff. Most of the deletions are from unused code

tokio/src/time/sleep.rs
```diff
@@ -369,13 +380,21 @@ impl Sleep {
...
+                Ok(()) => return,
+                Err(recovered) => handle = recovered,
+            }
+        }
+
+        if let Ok(timer) = Timer::new(handle, deadline) {
+            this.state.set(SleepState::Active { timer });
+            let timer = this.state.as_mut().as_timer().unwrap();
+            if timer.init(deadline).is_err() {
+                this.state.set(SleepState::Elapsed);
```
ADD-SP, line 394: Shall we wake up the `Waker` we registered in earlier? Is it possible for a task who is already `.await` this `Sleep`?

Since invoking either `.reset` or `.await` requires `Pin<&mut Self>`, we can guarantee that there is no concurrent call to `.reset` and `.await` against the same `Sleep` instance.

Am I correct? If so, the changes for `tokio-util/tests/time_delay_queue.rs` looks safe and not a breaking change.

vilgotf: I was confused about this. Not waking the previous waker is definitely a behavioural change, and as you point out, it can be kept, but I'm not sure what the original intention was.

---

Okay, I reviewed the code once again. The original code comment explicitly states that the wake is required:

> Wake the waker; this is needed because we might reset _after_ a poll, and otherwise the task won't be awoken to poll again.

My understanding is that the original author was worried about users calling `reset`, _not_ followed by `poll`. I think the aforementioned scenario is already broken, but in the spirit of not trying to break users, I'll concede that someone may depend on it.

edit: it was added (ad341b3bd9447b7aa39ea0124435af475d142d57) to fix a broken DelayQueue test, but it is no longer seems to be necessary.

edit2: it's worth noting that the alternative timer does not wake the previous waker. So _if_ we say that _not_ waking the waker is a breaking change, then the alternative timer (in its current state) is also a breaking change. Should we "fix" the alternative timer as well then? Or is that rather an argument for this breaking change being acceptable?

edit3: ad341b3bd9447b7aa39ea0124435af475d142d57 actually restored behaviour introduced in https://github.com/tokio-rs/tokio/pull/2217 and https://github.com/tokio-rs/tokio/pull/2290. It remains that the alt-timer never implemented this, so I don't think it is necessary.

tokio/src/time/sleep.rs
```diff
@@ -445,25 +469,26 @@ impl Sleep {
...
-    // - Shutdown: No timer has been setup, which is a misuse error.
-    //
-    // Both cases are extremely rare, and pretty accurately fit into
-    // "logic errors", so we just panic in this case. A user couldn't
-    // really do much better if we passed the error onwards.
-    fn poll(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Self::Output> {
+    fn poll(self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Self::Output> {
+        if matches!(self.state, SleepState::Elapsed) {
+            return Poll::Ready(());
+        }
```
vilgotf, line 475: TODO: Store the current waker (#8198), probably by adding a `waker: Option<Waker>` field like `DelayQueue`.

tokio/src/time/sleep.rs
ADD-SP: Thanks for detailed investigation! Let's restore the original behavior to unblock this PR. We can discuss this topic in the issue.

PR #8138, opened 2026-05-12
tokio-util: make CancellationToken::is_cancelled lock-free

dflemstr:
## Motivation

Previously, `is_cancelled` acquired the per-node mutex on every call. Under contention with `cancel`, `child_token`, or token-drop operations, this could block on a futex and produce multi-millisecond stalls - fatal for real-time or latency-sensitive callers (see #7775).

## Solution

Replace `Inner::is_cancelled` with an `AtomicBool` directly on `TreeNode`, making the flag the single source of truth and `is_cancelled()` lock-free.

Memory ordering is the subtle part: `WaitForCancellationFuture::poll` reads `is_cancelled` *before* `cancel` writes it, and relies on its `Notified` waker registration happening-before `notify_waiters`. A plain `load(Acquire)`/`store(Release)` pair only orders one direction and would permit a lost wakeup. To get bidirectional ordering, reads use `compare_exchange(false, false, AcqRel, Acquire)` and writes use `swap(true, AcqRel)` - both RMWs so they form a release-acquire pair from either side. Reads performed under the node's mutex use `Relaxed` since the mutex already provides the ordering.

Tests cover:
  - publication (data written before `cancel` is visible after `is_cancelled` returns true, single-threaded and under loom)
  - concurrent readers racing against `cancel` and `child_token`
  - the lost-wakeup race scenario, both as a non-loom stress loop and as a direct loom interleaving check

Refs: #7775

### DIFF
--- a/tokio-util/src/loom.rs
+++ b/tokio-util/src/loom.rs
@@ -3,7 +3,7 @@
 
 pub(crate) mod sync {
     #[cfg(all(test, loom))]
-    pub(crate) use loom::sync::{Arc, Mutex, MutexGuard};
+    pub(crate) use loom::sync::{atomic, Arc, Mutex, MutexGuard};
     #[cfg(not(all(test, loom)))]
-    pub(crate) use std::sync::{Arc, Mutex, MutexGuard};
+    pub(crate) use std::sync::{atomic, Arc, Mutex, MutexGuard};
 }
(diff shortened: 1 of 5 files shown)


dflemstr:
PR description is now slightly out of date as I made the `is_cancelled` just do a single load but `is_cancelled_with_sync` (name to be discussed) exhibits the behavior initially described in the PR description.

PR #8140, opened 2026-05-12
add WakeList collection interfaces

vilgotf:
Also fixes an issue where the lock is needlessly re-acquired if exactly 32 wakers are collected by scanning the remainder before releasing the lock.

This change leaves `batch_semaphore` as-is due to its complexity.


Darksonn:
I think I prefer to keep this code as a loop. I do not think iterators help in this particular case.

vilgotf:
> I think I prefer to keep this code as a loop. I do not think iterators help in this particular case.

Can I make you reconsider or should I close this PR? I find iterators much easier to read and reason about; the nested loops were particularly hairy in my option.

PR #8141, opened 2026-05-13
codec: use libc::memchr for LinesCodec delimiter scan

RyanJamesStewart:
## Summary (v2)

Updated after review. The previous revision pulled in the `memchr` crate; per @Darksonn's feedback that's not happening, and my claim that `memchr` was already transitive via `tokio`/`bytes` was wrong (`cargo tree -p tokio-util -e normal` confirms it's not in any normal dep tree). Apologies for the bad call.

This revision mirrors the approach in `tokio/src/util/memchr.rs`:

- New `tokio-util/src/util/memchr.rs` — `libc::memchr` on unix when the optional `libc` feature is enabled, scalar `iter::position` fallback otherwise.
- `codec` feature now enables `libc`, so existing users of `LinesCodec` on unix get the accelerated scan with no extra opt-in; non-unix targets (windows, wasm) keep the existing scalar path.
- `libc` is added as `optional = true` under `[target.'cfg(unix)'.dependencies]`. No new crate dep beyond `libc` (which `tokio`/`tokio-util` already use elsewhere with the same versioning pattern).
- **`AnyDelimiterCodec` is reverted to master** and dropped from this PR. The multi-needle case doesn't map onto the single-needle libc shim, and is a separate question.

## Benchmarks

Decoding a `BytesMut` of fixed-length lines through `LinesCodec::decode` (the full happy path — find `\n`, `split_to`, `str::from_utf8`, allocate the `String`):

| line length | `iter().position` | `libc::memchr` | speedup |
|---|---|---|---|
| 16 B  | 21.4 ns/line | 19.9 ns/line | 1.07× |
| 40 B  | 27.2 ns/line | 21.3 ns/line | 1.27× |
| 120 B | 63.8 ns/line | 26.6 ns/line | 2.40× |
| 512 B | 196.0 ns/line | 49.9 ns/line | **3.93×** |

The scan itself (no allocation) is ~35× faster on a 1 MiB delimiter-free buffer — the DoS-surface case the codec docs warn about.

## Behavior / tests

Behavior is unchanged — the `Decoder` contract (partial reads, `decode` vs `decode_eof`, returning `Ok(None)` to request more bytes, the discard-on-overlength state machine, the `next_index` resume across calls) holds exactly.

- `cargo test -p tokio-util --features full` — all green (codec / framed / framed_read / framed_write / length_delimited suites + doctests).
- `cargo clippy -p tokio-util --features full --all-targets` — clean.
- `cargo build -p tokio-util --no-default-features --features codec --target wasm32-wasip1` — clean (exercises the scalar fallback path).
- `cargo fmt --check` — clean.

## Dep tree

- `cargo tree -p tokio-util -e normal` (no features) — no `libc`.
- `cargo tree -p tokio-util -e normal --features full` — `libc` present (also pulled by `tokio/net`).
- `cargo tree -p tokio-util -e normal --no-default-features --features codec` — `libc` present on unix only.

## MSRV

`libc` 0.2.168 (matching `tokio`'s pin) is below the workspace MSRV; no change.

### DIFF
--- a/tokio-util/Cargo.toml
+++ b/tokio-util/Cargo.toml
@@ -25,7 +25,7 @@ full = ["codec", "compat", "io-util", "time", "net", "rt", "join-map"]
 
 net = ["tokio/net"]
 compat = ["futures-io"]
-codec = []
+codec = ["libc"]
 time = ["tokio/time", "slab"]
 io = []
 io-util = ["io", "tokio/rt", "tokio/io-util"]
@@ -45,6 +45,9 @@ pin-project-lite = "0.2.11"
 slab = { version = "0.4.4", optional = true } # Backs `DelayQueue`
 tracing = { version = "0.1.29", default-features = false, features = ["std"], optional = true }
 hashbrown = { version = "0.15.0", default-features = false, optional = true }
+
+[target.'cfg(unix)'.dependencies]
+libc = { version = "0.2.168", optional = true } # Backs the LinesCodec delimiter scan via libc::memchr
 
 [dev-dependencies]
 tokio = { version = "1.0.0", features = ["full"] }
--- a/tokio-util/src/codec/lines_codec.rs
+++ b/tokio-util/src/codec/lines_codec.rs
@@ -115,9 +115,7 @@ impl Decoder for LinesCodec {
             // there's no max_length set, we'll read to the end of the buffer.
             let read_to = cmp::min(self.max_length.saturating_add(1), buf.len());
 
-            let newline_offset = buf[self.next_index..read_to]
-                .iter()
-                .position(|b| *b == b'\n');
+            let newline_offset = crate::util::memchr::memchr(b'\n', &buf[self.next_index..read_to]);
 
             match (self.is_discarding, newline_offset) {
                 (true, Some(offset)) => {
--- a/tokio-util/src/util/memchr.rs
+++ b/tokio-util/src/util/memchr.rs
@@ -1,0 +1,102 @@
+//! Search for a byte in a byte array using libc.
+//!
+//! When nothing pulls in libc, then just use a trivial implementation. Note
+//! that we only depend on libc on unix.
+
+#[cfg(not(all(unix, feature = "libc")))]
+fn memchr_inner(needle: u8, haystack: &[u8]) -> Option<usize> {
+    haystack.iter().position(|val| needle == *val)
+}
+
+#[cfg(all(unix, feature = "libc"))]
+fn memchr_inner(needle: u8, haystack: &[u8]) -> Option<usize> {
+    let start = haystack.as_ptr();
+
+    // SAFETY: `start` is valid for `haystack.len()` bytes.
+    let ptr = (unsafe { libc::memchr(start.cast(), needle as _, haystack.len()) })
+        .cast::<u8>()
+        .cast_const();
+
+    if ptr.is_null() {
+        None
+    } else {
+        // SAFETY: `ptr` will always be in bounds, since libc guarantees that the ptr will either
+        //          be to an element inside the array or the ptr will be null
+        //          since the ptr is in bounds the offset must also always be non null
+        //          and there can't be more than isize::MAX elements inside an array
+        //          as rust guarantees that the maximum number of bytes a allocation
+        //          may occupy is isize::MAX
+        unsafe {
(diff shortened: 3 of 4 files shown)


tokio-util/src/codec/any_delimiter_codec.rs
```diff
@@ -141,9 +141,13 @@ impl Decoder for AnyDelimiterCodec {
...
             let read_to = cmp::min(self.max_length.saturating_add(1), buf.len());
 
-            let new_chunk_offset = buf[self.next_index..read_to]
-                .iter()
-                .position(|b| self.seek_delimiters.contains(b));
+            let haystack = &buf[self.next_index..read_to];
+            let new_chunk_offset = match self.seek_delimiters.as_slice() {
+                [d0] => memchr::memchr(*d0, haystack),
+                [d0, d1] => memchr::memchr2(*d0, *d1, haystack),
+                [d0, d1, d2] => memchr::memchr3(*d0, *d1, *d2, haystack),
```
martin-g:
```suggestion
                [d0, d1, d2] => memchr::memchr3(*d0, *d1, *d2, haystack),
                [d0, d1, d2, d3] => match (memchr::memchr3(*d0, *d1, *d2, haystack), memchr::memchr(*d3, haystack)) {
                    (Some(i1), Some(i2)) => Some(cmp::min(i1, i2)),
                    (i1, i2) => i1.or(i2),
                },
```
Does it makes sense to add an arm for 4 needles too ?
The main (?) example uses 4: https://github.com/tokio-rs/tokio/blob/c4019c236e385248f0a118fd06a096b5572d969a/tokio-util/src/codec/any_delimiter_codec.rs#L27

martin-g (approved):

tokio-util/Cargo.toml
```diff
@@ -37,6 +37,7 @@ __docs_rs = ["futures-util"]
 [dependencies]
 tokio = { version = "1.47.0", features = ["sync"] }
 bytes = "1.5.0"
+memchr = { version = "2.5.0", optional = true } # Backs the codec delimiter scans
```
Darksonn: We are not adding this dependency. You can use `libc::memchr` similar to `tokio/src/util/memchr.rs` if you want to do this.

Darksonn (changes requested):

tokio-util/Cargo.toml
Darksonn: You claim that the dependency is already in the dependency tree via tokio and bytes, but this is just false.

RyanJamesStewart:
Pushed v2. Summary of changes from review:

On the dep policy — agreed, dropped the `memchr` crate. The new revision uses a small `tokio-util/src/util/memchr.rs` shim mirroring `tokio/src/util/memchr.rs`: `libc::memchr` on unix gated on an optional `libc` feature, scalar `iter::position` fallback otherwise. The `codec` feature pulls in `libc`, so existing `LinesCodec` users on unix get the fast path automatically with no API change. Also — the original PR body's claim that `memchr` was transitive via `tokio`/`bytes` was wrong (confirmed via `cargo tree -p tokio-util -e normal`); apologies for that.

On the multi-needle suggestion — since the `memchr` crate is no longer in the picture, the `memchr2`/`memchr3` + 4-needle dispatch doesn't apply here. `AnyDelimiterCodec` is reverted to master and dropped from this PR; if the multi-needle case is worth pursuing it'd be a separate change with a hand-rolled multi-byte scan.

Diff is now narrowed to ~106 lines (most of that is the shim file): one source change in `LinesCodec`, the shim, and the `Cargo.toml` wiring. All tests + clippy + fmt green locally; verified building for `wasm32-wasip1` to exercise the scalar fallback. PR body has full benchmarks and dep-tree confirmation.

RyanJamesStewart:
v2 is pushed, addressing the review. The memchr crate dependency is dropped; LinesCodec now uses a libc::memchr shim with a scalar fallback under tokio-util/src/util/memchr.rs, mirroring tokio/src/util/memchr.rs as suggested in review. libc is optional and gated to cfg(unix). The review also correctly flagged that the crate isn't transitively present via tokio/bytes; cargo tree -p tokio-util -e normal confirms it, and I've corrected the PR body. AnyDelimiterCodec is reverted to master and dropped from this PR: the multi-needle case doesn't map onto a single-needle libc::memchr shim and is a separate question. Re-requesting review. Note that martin-g's earlier approval predates v2 and is against the old diff.

Darksonn (approved):
Thanks.

PR #8142, opened 2026-05-13
sync: remove useless conversion in oneshot Receiver::poll

RyanJamesStewart:
## Summary

`Inner::poll_recv` returns `Poll<Result<T, RecvError>>` and `Receiver`'s `Future::Output` is `Result<T, RecvError>`, so the `.map_err(Into::into)` in `Receiver::poll` is an identity conversion. Newer `clippy` flags it as `clippy::useless_conversion`. This removes it from both `#[cfg]` arms — no behavior change.

Verified `cargo clippy -p tokio --features full --lib` is clean afterwards (and `RUSTFLAGS="--cfg tokio_unstable" cargo clippy -p tokio --features full,tracing --lib` for the other arm), and `cargo test -p tokio --test sync_oneshot` passes.


martin-g (approved):

mattiapitossi (approved):
Thanks!

PR #8144, opened 2026-05-14
net: add `SocketAddr` methods to Unix sockets

elomatreb:
## Motivation

Add methods to `UnixStream` and `UnixListener` to use them with `SocketAddr` (the tokio wrapper type) directly, instead of having to go through a `Path` interface. This makes handling connection code in applications that can deal with both path-based sockets and the linux-only abstract socket namespace more convenient.

## Solution

- Add a `connect_addr` method to `UnixStream`
- Add a `bind_addr` method to `UnixListener`

Both methods were implemented by moving the relevant code out from the path-based existing implementation and making the existing implementation use it as well.

The test was adapted from the existing path-based test as well.

## Open Questions

- Should I add documentation tests? The existing methods didn't have them either.
- Semi-related: The path-based methods have a heuristic for dealing with abstract sockets. This has caught me off guard before, but clearly documenting that is probably better in a separate MR?

### DIFF
--- a/tokio/src/net/unix/listener.rs
+++ b/tokio/src/net/unix/listener.rs
@@ -90,7 +90,23 @@ impl UnixListener {
         #[cfg(not(any(target_os = "linux", target_os = "android")))]
         let addr = StdSocketAddr::from_pathname(path)?;
 
-        let listener = mio::net::UnixListener::bind_addr(&addr)?;
+        let addr = SocketAddr::from(addr);
+        UnixListener::bind_addr(&addr)
+    }
+
+    /// Creates a new `UnixListener` bound to the specified address.
+    ///
+    /// # Panics
+    ///
+    /// This function panics if it is not called from within a runtime with
+    /// IO enabled.
+    ///
+    /// The runtime is usually set implicitly when this function is called
+    /// from a future driven by a tokio runtime, otherwise runtime can be set
+    /// explicitly with [`Runtime::enter`](crate::runtime::Runtime::enter) function.
+    #[track_caller]
+    pub fn bind_addr(socket_addr: &SocketAddr) -> io::Result<UnixListener> {
+        let listener = mio::net::UnixListener::bind_addr(&socket_addr.0)?;
         let io = PollEvented::new(listener)?;
         Ok(UnixListener { io })
     }
--- a/tokio/src/net/unix/stream.rs
+++ b/tokio/src/net/unix/stream.rs
@@ -86,7 +86,17 @@ impl UnixStream {
         #[cfg(not(any(target_os = "linux", target_os = "android")))]
         let addr = StdSocketAddr::from_pathname(path)?;
 
-        let stream = mio::net::UnixStream::connect_addr(&addr)?;
+        let addr = SocketAddr::from(addr);
+        UnixStream::connect_addr(&addr).await
+    }
+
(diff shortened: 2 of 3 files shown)


elomatreb:
The failed macOS check is caused by https://github.com/actions/runner-images/issues/14097 I think.

mattiapitossi:
> The failed macOS check is caused by [actions/runner-images#14097](https://github.com/actions/runner-images/issues/14097) I think.

I've opened https://github.com/tokio-rs/tokio/pull/8145, this should fix the problem

elomatreb:
@mattiapitossi thanks, that helped.

tokio/src/net/unix/stream.rs
```diff
@@ -86,7 +86,17 @@ impl UnixStream {
...
-        let stream = mio::net::UnixStream::connect_addr(&addr)?;
+        let addr = SocketAddr::from(addr);
+        UnixStream::connect_addr(&addr).await
+    }
+
+    /// Connects to the socket named by `socket_addr`.
+    ///
+    /// This function will create a new Unix socket and connect to the address
+    /// specified, associating the returned stream with the default event
+    /// loop's handle.
```
martin-g, line 106: `UnixStream::new()` calls `PollEvented::new()`, which calls `Registration::new_with_interest_and_handle()` that also needs IO enabled.
Please add the same `# Panics` as for `bind_addr()` above

martin-g (approved):

tokio/src/net/unix/stream.rs
elomatreb: I mirrored the respective existing methods, and `UnixStream::connect` does not have a panics section. Should I add it to both?

tokio/src/net/unix/listener.rs
```diff
@@ -90,7 +90,23 @@ impl UnixListener {
...
+    /// # Panics
+    ///
+    /// This function panics if it is not called from within a runtime with
+    /// IO enabled.
+    ///
+    /// The runtime is usually set implicitly when this function is called
+    /// from a future driven by a tokio runtime, otherwise runtime can be set
+    /// explicitly with [`Runtime::enter`](crate::runtime::Runtime::enter) function.
+    #[track_caller]
+    pub fn bind_addr(socket_addr: &SocketAddr) -> io::Result<UnixListener> {
```
Darksonn, line 108: Why is this `&SocketAddr` rather than just `SocketAddr`? Isn't this type copy?

elomatreb: This is `tokio::net::unix::SocketAddr`, which contains a buffer internally (essentially a `PathBuf`), so it's not `Copy`. This also matches the signature from std: https://doc.rust-lang.org/std/os/unix/net/struct.UnixStream.html#method.connect_addr (for the same reason)

elomatreb: Actually comparing it to `PathBuf` is misleading because it's not actually backed by a heap allocation, but the underlying type is still not `Copy`.

Darksonn: Ok, that's fine then.

tokio/src/net/unix/stream.rs
martin-g: Yes, please!

Darksonn (approved):
Thanks.

PR #8145, opened 2026-05-14
ci: fix macos runner

mattiapitossi:
We started to observe failures in macOS runners, it seems that disabling the cache-bin option for rust-cache fixes the issue (this is a temporary fix that will be reverted once the upstream fixes the issue). see: https://github.com/Swatinem/rust-cache/issues/341


avantgardnerio (approved):
Nice, ty!

martin-g (approved):

PR #8146, opened 2026-05-15
time: fix Sleep livelock when eager combinator starves current_thread timer driver

### LINKED ISSUE #7883: Timers should make fewer assumptions about poller yielding back to the runtime
### Summary

Tokio's timers exhibit behavior that can cause livelocks with combinators that don't yield.

While this is technically allowed behavior per the futures API, it is a livelock trap, and arguably should be revised. Infinite loops are also allowed behavior, and triggering a loop of wake-ups is far removed from what a sane runtime should do.

At the same time, it may make sense to discuss with the Rust async team (more on this below). This topic is niche, but it does affect correctness in a critical regard.

### What's going on

I discovered this while developing a futures collector crate. Failing to yield to the runtime can cause Tokio timers to spin forever, waking themselves up and returning Pending repeatedly. The collector polls futures insofar as they generate wake-up notifications, but every time a Sleep timer is polled, it generates another wake-up.

The Tokio timer is assuming that by yielding repeatedly (waking and returning Pending), the collector will eventually yield back to the Tokio runtime. So the Sleep timer spits out wake-ups until this happens. The livelock happens because the collector is not mandated (at all) to yield back to the top-level runtime, and since it received.

In other words, non-cooperative intermediate pollers break the timer assumption that yielding will eventually push back control to the top-level runtime call, which could be as high-level as main().

Interestingly, I think other futures collectors have run into [the same trap](https://github.com/conradludgate/futures-buffered/commit/f878029657831dbd063ff01866aaf1a41fdbd839). This is a link to an old commit from `futures-buffered`, and the details of the commit (and the change it makes) strongly suggest that the exact same thing livelock. I would love to get @conradludgate to confirm based on their memory that this was indeed what happened.

Futures collectors already stand out as a somewhat curious construction in the Rust async story. Of course, they're very cute and useful tools and we couldn't be without them. They happen to raise this problem based on the way they are typically implemented, but it's also possible that simpler combinators could reproduce this issue (see next section).

### Extent of the Problem and its Ingredients

I have only observed this problem with futures collectors, and I am most certain of its existence in that regard.

However, the same livelock may be possible even with small combinators like `join!` or `select!`, depending on how they are implemented. I haven't checked the details, as there are possible many in the ecosystem. But a simple implementation that loops would theoretically trigger the same livelock with Tokio timers. It's the act of eagerly polling woken futures that causes the livelock.

```
// pseudo-code for a select! combinator

fn poll_combinator() {
    loop {
        if !is_woken() {
            break Poll::Pending;
        }
        set_waker(cx.waker());

        let poll1 = self.fut1.poll();
        if poll1.is_ready() {
            return poll1;
        }
        let poll2 = self.fut2.poll();
        if poll2.is_ready() {
           return poll2;
        }
    }
}
```

I don't think there are many implementations of this sort. On the other hand however, it requires only a couple ingredients based on my analysis:

- The combinator tracks whether its inner futures were woken
- The combinator checks woken status and polls those futures in a loop, inside its own poll() implementation

Such a combinator is not necessarily behaving wrongly. It's merely being eager about polling its inner futures, and persisting in doing so whenever they generate wake-ups. Besides being contractually correct, there are legitimate use cases for such eager polling. Sometimes the implementation is simpler. And for a deeply nested stack of futures and async functions, it may even be more performant to eagerly poll.

There is a dialectic between polling and yielding. Tokio wants the combinator to yield all the way back to the top-level runtime entry point, and the Tokio timers assume such behavior and rely on it. Combinators may or may not uphold the timers' assumption; some of them poll their futures again before returning to the caller.

### Should this be solved, and how?

First of all, we're diving into Rust's async model and what it means to poll and yield. It makes sense to follow Rust project guidance, and perhaps consult the Rust WG async team. It's also definitely worth citing some existing discussion on this topic. So, I'm linking the following issue, which is doubly relevant here.

https://github.com/rust-lang/futures-rs/issues/2053

Note that the above issue tracks down to behavior in `FuturesUnordered` that was identified and addressed in a PR. However, the original issue (and its solution) were framed in terms of performance: the collector should yield back to the runtime, in order to avoid starvation. That's all fine and well, and also correct. Nobody would want a general-purpose FuturesUnordered with a poor cooperation policy. However, little was discussed in the way of *correctness*; task starvation and livelock are not the same, and in the latter case, no progress is being made.

This distinction matters because it impacts the correctness question for Tokio timers. Tokio timers can issue infinite wake-ups even when no progress is being made. This quirk in their implementation is what causes *livelock*, as distinct from mere starvation (and poor performance). However, such a livelock is only possible when the combinator keeps polling woken futures, because such eager polling also assumes that progress *is* being made.

Either way, somebody should change their code's behavior. Either the Tokio timers are mistaken for assuming that yielding will eventually bubble up to the main runtime, or combinators and collectors are mistaken if they assume that polling woken futures is a valid way of making progress.

### Arguments in favor of changing Tokio timer behavior

I know that the alternative Timer is currently under development. It may be worth incorporating this topic onto the roadmap of the alternative Timer, at least as a question, but that's up to the maintainers. Based on my testing, the alternative timer exhibits the same behavior as the existing Timer.

Changing how timers complete themselves would remove a potential source of livelock with arbitrary user code. I'm biased here, but it probably also happened to @conradludgate , and it's not unimaginable for bespoke combinators in large codebases to trigger the same livelock. Would you want to catch such a livelock in production? Definitely not. The surprising nature of this behavior alone (is it not strange for Timers to act this way? Am I crazy?) suggests that Tokio should change how timers operate.

Relatedly, I noticed that the performance of Tokio timers is rather weak compared to the smol runtime. Smol timers don't exhibit this behavior, and I wonder if performance could be improved by making the Sleep task able to complete itself (rather than having the runtime switch it to completed). The current design means that the call stack has to rewind all the way to `fn main()` just for timers to be completed and descend back down the stack again.

### Counter arguments

**Objection 1**: Calling `wake_by_ref` at any time  is allowed behavior (spurious wake-ups). Therefore, it's fine for Tokio timers to trigger wake-ups whenever they wish, even if they're not making progress while issuing these wake-ups.

My problem with this objection is not that I disagree, rather that it sends a strange signal to the rest of the system. A wake-up requires that at least one poll happen afterward, but when that poll happens is otherwise at the discretion of the poller. Tokio's timers are saying _"poll me in the future"_ but what they want to say is _"poll me in the future, but pssst don't poll me immedately in a loop, otherwise nothing happens"_. There's no such way for the latter to be expressed in the API, and what this ultimately means is that Tokio timers are falsely communicating progress even where no progress is happening.

So while calling `wake_by_ref` at any time is allowed behavior, the assumption that Tokio timers make (that a yield will eventually happen) doesn't have to be upheld.

**Objection 2**: Inter-task dependencies can cause the same deadlock trap in the following scenario: a task outside the combinator depends on a task inside the combinator, and the task on the inside spams wake-ups + Pending until the outside task gets run.

This objection only works if the task inside the combinator issues wake-ups repeatedly as a means of "escaping" the combinator. This is really strange of an assumption to make, and wake-ups have no such property or guarantee. Rather than being an objection, I think this situation strengthens the argument for changing Tokio timers. The timers are essentially analogous to the task inside the combinator, while the timer wheel (that completes them) is analogous to the task outside the combinator.

**Objection 3**: Rust's safety model permits livelock.

Yes, and it also permits deadlock. This has nothing to do with memory safety, but that doesn't mean we shouldn't strive to eliminate livelocks and deadlocks through better crate design.

### Conclusion

I'm clearly leaning more toward Tokio fixing this, but I'm not a maintainer, and I'll leave it to them. I think that Tokio timers should make fewer assumptions, and that one less livelock trap in a big runtime is a win for some poor developer plodding away at 3 AM in 2029.

Futures collectors probably want to yield every now and then, if only for fairness' sake. However, such yielding is not mandated anywhere by the Rust async model. It would be a good idea to get the input of the Rust async working group, and while this issue relates to past issues, this is the first time (I believe?) that its *correctness* is being discussed.

AsthaMishra:
Fixes #7883

## Root cause

In a `current_thread` runtime the timer driver only runs inside
`park_internal` — i.e., only when the active task returns `Poll::Pending`
back to the runtime. An eager combinator that detects synchronous
wakeups and re-polls inner futures without yielding prevents
`park_internal` from ever running, which starves the timer driver.

`Sleep` is therefore never told its deadline has passed, even though
real time has moved on. The cooperative budget mechanism doesn't save
it: once the budget is exhausted `poll_proceed` defers the waker via
`context::defer`, but because the task never returns `Poll::Pending`,
the deferred wake never fires and the driver still doesn't run.

## Fix

Add a wall-clock check at the top of `Sleep::poll_elapsed`, before the
coop `poll_proceed` call. If the deadline has already passed on the
real clock, `Sleep` completes immediately without requiring the timer
driver to have processed the timeout:

```rust
if me.entry.deadline() <= crate::time::Instant::now() {
    return Poll::Ready(Ok(()));
}
```

Also guards the wall-clock shortcut behind `clock::is_paused()` so that
time-paused tests (`time::pause()`) continue to advance the mock clock
via the timer driver as expected.

### DIFF
--- a/tokio/src/time/sleep.rs
+++ b/tokio/src/time/sleep.rs
@@ -419,6 +419,20 @@ impl Sleep {
 
         ready!(crate::trace::trace_leaf());
 
+        // If the deadline has already passed according to the wall clock, complete
+        // immediately. This prevents starvation of the timer driver when an eager
+        // combinator loops without yielding to the runtime (issue #7883): coop
+        // budget exhaustion defers the wake but the task never returns Pending, so
+        // the timer driver never runs. Checking the clock here lets Sleep
+        // self-complete in that scenario.
+        //
+        // When time is paused (test-util), skip this shortcut: the timer driver is
+        // responsible for advancing the mock clock, and bypassing it breaks tests
+        // that assert on how much mock time has elapsed.
+        if !crate::time::clock::is_paused() && me.entry.deadline() <= crate::time::Instant::now() {
+            return Poll::Ready(Ok(()));
+        }
+
         // Keep track of task budget
         #[cfg(all(tokio_unstable, feature = "tracing"))]
         let coop = ready!(trace_poll_op!(
--- a/tokio/src/time/sleep.rs
+++ b/tokio/src/time/sleep.rs
@@ -419,20 +419,6 @@ impl Sleep {
 
         ready!(crate::trace::trace_leaf());
 
-        // If the deadline has already passed according to the wall clock, complete
-        // immediately. This prevents starvation of the timer driver when an eager
-        // combinator loops without yielding to the runtime (issue #7883): coop
-        // budget exhaustion defers the wake but the task never returns Pending, so
-        // the timer driver never runs. Checking the clock here lets Sleep
-        // self-complete in that scenario.
-        //
-        // When time is paused (test-util), skip this shortcut: the timer driver is
-        // responsible for advancing the mock clock, and bypassing it breaks tests
-        // that assert on how much mock time has elapsed.
-        if !crate::time::clock::is_paused() && me.entry.deadline() <= crate::time::Instant::now() {
-            return Poll::Ready(Ok(()));
-        }
-
         // Keep track of task budget
         #[cfg(all(tokio_unstable, feature = "tracing"))]
         let coop = ready!(trace_poll_op!(
--- a/tokio-util/tests/time_delay_queue.rs
+++ b/tokio-util/tests/time_delay_queue.rs
@@ -545,6 +545,9 @@ async fn reset_later_after_slot_starts() {
     // that slot, but doesn't know when, it must wake immediately to advance
     // the wheel.
     queue.reset_at(&foo, now + ms(120));
+    // The wake is deferred (via context::defer) so we must yield to the
+    // runtime once to flush the deferred waker list before checking.
+    tokio::task::yield_now().await;
     assert!(queue.is_woken());
 
     assert_pending!(poll!(queue));
@@ -607,6 +610,9 @@ async fn reset_earlier_after_slot_starts() {
     // that slot, but doesn't know when, it must wake immediately to advance
     // the wheel.
     queue.reset_at(&foo, now + ms(120));
+    // The wake is deferred (via context::defer) so we must yield to the
+    // runtime once to flush the deferred waker list before checking.
+    tokio::task::yield_now().await;
     assert!(queue.is_woken());
 
     assert_pending!(poll!(queue));
--- a/tokio/src/runtime/time/mod.rs
+++ b/tokio/src/runtime/time/mod.rs
@@ -23,6 +23,7 @@ use super::time_alt;
 
 use crate::loom::sync::atomic::{AtomicBool, Ordering};
 use crate::loom::sync::Mutex;
+#[cfg(feature = "rt")]
 use crate::runtime::context;
 use crate::runtime::driver::{self, IoHandle, IoStack};
 use crate::time::error::Error;
@@ -449,7 +450,10 @@ impl Handle {
         // would cause eager combinators to re-poll Sleep without ever yielding
         // to the runtime, preventing the timer driver from advancing.
         if let Some(waker) = waker {
+            #[cfg(feature = "rt")]
             context::defer(&waker);
+            #[cfg(not(feature = "rt"))]
+            waker.wake();
         }
     }
 
--- a/tokio/src/runtime/time/mod.rs
+++ b/tokio/src/runtime/time/mod.rs
@@ -23,6 +23,7 @@ use super::time_alt;
 
 use crate::loom::sync::atomic::{AtomicBool, Ordering};
 use crate::loom::sync::Mutex;
+use crate::runtime::context;
 use crate::runtime::driver::{self, IoHandle, IoStack};
 use crate::time::error::Error;
 use crate::time::{Clock, Duration};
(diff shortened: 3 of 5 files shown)


tokio/src/time/sleep.rs
```diff
@@ -419,6 +419,20 @@ impl Sleep {
...
+        // budget exhaustion defers the wake but the task never returns Pending, so
+        // the timer driver never runs. Checking the clock here lets Sleep
+        // self-complete in that scenario.
+        //
+        // When time is paused (test-util), skip this shortcut: the timer driver is
+        // responsible for advancing the mock clock, and bypassing it breaks tests
+        // that assert on how much mock time has elapsed.
+        if !crate::time::clock::is_paused() && me.entry.deadline() <= crate::time::Instant::now() {
+            return Poll::Ready(Ok(()));
+        }
```
Darksonn: My intended fix for #7883 is that the timers should defer the wakeup using Tokio's deferred wakeup startegy. Not that we should check the time every single time.

tokio/src/time/sleep.rs
```diff
@@ -419,6 +419,20 @@ impl Sleep {
...
+        // immediately. This prevents starvation of the timer driver when an eager
+        // combinator loops without yielding to the runtime (issue #7883): coop
+        // budget exhaustion defers the wake but the task never returns Pending, so
+        // the timer driver never runs. Checking the clock here lets Sleep
+        // self-complete in that scenario.
+        //
+        // When time is paused (test-util), skip this shortcut: the timer driver is
+        // responsible for advancing the mock clock, and bypassing it breaks tests
+        // that assert on how much mock time has elapsed.
+        if !crate::time::clock::is_paused() && me.entry.deadline() <= crate::time::Instant::now() {
```
vilgotf: Having previously benchmarked tokio timers, I happen to know that `Instant::now` is about 50 times as expensive as polling the timer (~200 ns vs ~3.5 µs). This might be acceptable, but I just wanted to throw this out for context

tokio/src/time/sleep.rs
AsthaMishra: Done — replaced waker.wake() with context::defer(&waker) in the InsertError::Elapsed branch of reregister, matching the same deferred wakeup strategy used by yield_now(). Added a regression test (sleep_not_starved_by_eager_combinator) that resets a sleep to an elapsed deadline inside an eager combinator and verifies it completes in ≤10 polls. Two tokio-util delay_queue tests that checked is_woken() immediately after reset_at() also needed a yield_now().await to flush the deferred waker list first.

tokio/src/time/sleep.rs
AsthaMishra: Noted

PR #8148, opened 2026-05-17
Relax Lines::get_ref receiver to &self

cuishuang:
## Motivation

Lines::get_ref only returns a shared reference to the underlying reader and does not mutate internal state, so it does not need a mutable receiver. 

Changing the receiver from &mut self to &self also makes it consistent with other Tokio IO wrapper get_ref methods such as BufReader, BufWriter, BufStream, Take, and Chain.


Darksonn:
This is a breaking change.

PR #8151, opened 2026-05-18
rt: replace current-thread scheduler VecDeque with LinkedList

lthiery:
This PR changes `current_thread` runtime `Core.tasks` from a `VecDeque` to an intrusive linked list, `LocalQueue<Arc<Handle>>`. This avoids dynamic allocations in `Core.task`.

I ran `benches/rt_current_thread.rs` on an AMD Ryzen 9 5950X:
* spawn_many_local: 2-3% faster, noise?
* spawn_many_remote_idle: no change
* spawn_many_remote_busy: over 20% faster

Worth noting, if you blow up the number of spawns from 1k to 10k, `spawn_many_remote_busy` ends up being  15% slower with this change. I would guess at this point, the VecDeque ends up being more cache friendly but this also feels a bit like a degenerate situation at this point and it's probably better to take the win at <1k tasks?

### DIFF
--- a/tokio/src/runtime/scheduler/current_thread/mod.rs
+++ b/tokio/src/runtime/scheduler/current_thread/mod.rs
@@ -3,7 +3,8 @@ use crate::loom::sync::Arc;
 use crate::runtime::driver::{self, Driver};
 use crate::runtime::scheduler::{self, Defer, Inject};
 use crate::runtime::task::{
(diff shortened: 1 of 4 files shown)


ADD-SP (commented):
Thanks for your contribution!

Since tokio never shrinks the capacity of `VecDeque`, it should have better performance in general after warming up (assuming the queue length doesn't grow up infinitely).

I prefer keeping the `VecDeque` for better performance (after warming up) and simple implementation.

lthiery:
Fair enough!

Issue #6634, opened 2024-06-13
Interval panics with Duration::MAX


nazar-pc:
**Version**
1.37.0

**Platform**
Ubuntu x86-64 with 6.9.4 kernel

**Description**
Tokio timers do not handle overflow gracefully, I hit this multiple times already so decided to finally report it upstream.

I tried this code:

```rust
#[tokio::main]
async fn main() {
    tokio::time::interval(std::time::Duration::MAX).tick().await;
}
```

I expected to see this happen: effectively timer is idle indefinitely.

Instead, this happened:
```
thread 'main' panicked at library/std/src/time.rs:417:33:
overflow when adding duration to instant
```

`Instant + Duration` should be saturating to prevent this issue.

wathenjiang:
Hi. I believe it has be fixed in `1.38.0`. Please see https://github.com/tokio-rs/tokio/releases/tag/tokio-1.38.0 and https://github.com/tokio-rs/tokio/pull/6487.

nazar-pc:
Indeed, what a coincidence, thanks!

wathenjiang:
But there is still a possibility that 1.38.0 may panic, it eventually fixed in https://github.com/tokio-rs/tokio/pull/6612.

PR #8152, opened 2026-05-18
Don't panic when using an interval with max Duration

marshallpierce:
Currently, `MissedTickBehavior::next_timeout` panics when `period` is `Duration::MAX`. (Using such a large duration is convenient when you want to conditionally disable an interval.)

## Motivation

Nothing should panic when using `Duration::MAX`.

## Solution

Use `checked_add` with `Instant::far_future` as a fallback instead of `ops::Add`, as this fits with another case that uses `far_future` instead of `std::time::Instant::saturating_add`.

Side note 1: consider adding `saturating_add` to `Instant`, since this idiom is likely to be more common, especially given side note 2.

Side note 2: consider removing `ops::*` impls for `Instant` as there are likely other panics lurking. In my current project we have done this with our own Instant equivalent, and require all time manipulation to handle edge cases explicitly. With some `saturating_*` and perhaps `add_or_panic(duration, "explanation of why it will never panic")`, the same could probably be done here. Given the [history](https://github.com/tokio-rs/tokio/issues/6634) of previous similar issues, I think a more blanket approach to resolving that bug class is warranted.

### DIFF
--- a/tokio/src/time/interval.rs
+++ b/tokio/src/time/interval.rs
@@ -340,10 +340,12 @@ impl MissedTickBehavior {
     /// If a tick is missed, this method is called to determine when the next tick should happen.
     fn next_timeout(&self, timeout: Instant, now: Instant, period: Duration) -> Instant {
         match self {
-            Self::Burst => timeout + period,
-            Self::Delay => now + period,
+            Self::Burst => timeout
+                .checked_add(period)
+                .unwrap_or_else(Instant::far_future),
+            Self::Delay => now.checked_add(period).unwrap_or_else(Instant::far_future),
             Self::Skip => {
-                now + period
+                now.checked_add(period).unwrap_or_else(Instant::far_future)
                     - Duration::from_nanos(
                         ((now - timeout).as_nanos() % period.as_nanos())
                             .try_into()
--- a/tokio/tests/time_interval.rs
+++ b/tokio/tests/time_interval.rs
@@ -352,6 +352,37 @@ async fn reset_at_bigger_than_interval() {
     check_interval_poll!(i, start, 1701);
 }
 
+#[tokio::test(start_paused = true)]
+async fn interval_missed_tick_doesnt_panic_with_max_duration() {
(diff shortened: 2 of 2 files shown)


tokio/src/time/interval.rs
martin-g, line 527: This may need the same improvement

martin-g, line 592: This one too

tokio/tests/time_interval.rs
```diff
@@ -352,6 +352,29 @@ async fn reset_at_bigger_than_interval() {
...
+    for behavior in [
+        MissedTickBehavior::Delay,
+        MissedTickBehavior::Burst,
+        MissedTickBehavior::Skip,
+    ] {
+        let mut long_interval = time::interval(Duration::MAX);
+        long_interval.set_missed_tick_behavior(behavior);
+
+        // cause a missed tick
+        time::advance(ms(300)).await;
```
martin-g: It would be interesting here to advance with a period longer than `far_future`, i.e. bigger than 30 years. The `Skip` strategy does some subtractions which may lead to an underflow ?!

tokio/src/time/interval.rs
vilgotf, line 480: This may need the same improvement

Refs: #7193, #8128

marshallpierce: Take a look at https://github.com/tokio-rs/tokio/pull/8157 to see the continuation of this concept to the point of removing _all_ panicking operations.

tokio/tests/time_interval.rs
marshallpierce: Done.

tokio/src/time/interval.rs
marshallpierce: If you're inclined to move forward with this partial approach I can address these other cases, but if you're thinking the holistic solution in #8157 is the way to go I'll not bother...

vilgotf:
This bug was resolved in #8385

Darksonn:
Thanks for the PR

PR #8155, opened 2026-05-20
ci(freebsd): set copyback to false

abh1nav10:
Sets `copyback` to false on FreeBSD `vmactions` jobs.


martin-g:
> Sets `copyback` to false on FreeBSD `vmactions` jobs.

In case anyone else wonders what this setting does:
`When using rsync or scp, you can define copyback: false to not copy files back from the VM in to the host.`
https://github.com/vmactions/freebsd-vm#2-share-code

martin-g (approved):

mattiapitossi:
Thanks!

PR #8156, opened 2026-05-20
net: enable Miri tests for TCP socket

WhySoBad:
## Motivation

Because Miri didn't have TCP socket support, the tokio tests which relied on `TcpStream`s and `TcpListener`s had to be ignored when being run with Miri. However, over the past months TCP socket support was added to Miri, so these tests no longer need to be ignored when being run with Miri. 

Some tests were using very small buffers or many iterations which caused them to be very slow with Miri. For the tests I considered important I've changed the buffer size/iteration count, the other tests I've annotated with `#[cfg_attr(miri, ignore)]`. 

A few tests using TCP sockets currently don't work with Miri, but there are open issues/PRs for those:
- `readv` can leak memory: https://github.com/rust-lang/miri/pull/5054
- Host I/O events are only processed when entering the scheduler: https://github.com/rust-lang/miri/issues/5047

I'll open new PRs here once those issues are fixed in Miri and landed in a nightly rust version.

### DIFF
--- a/tokio/tests/tcp_echo.rs
+++ b/tokio/tests/tcp_echo.rs
@@ -1,13 +1,11 @@
 #![warn(rust_2018_idioms)]
 // WASIp1 doesn't support bind
-// No `socket` on miri.
 #![cfg(all(
     feature = "net",
     feature = "macros",
     feature = "rt",
     feature = "io-util",
     not(all(target_os = "wasi", target_env = "p1")),
-    not(miri)
 ))]
 
 use tokio::io::{self, AsyncReadExt, AsyncWriteExt};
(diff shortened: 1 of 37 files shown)


tokio/tests/tcp_echo.rs
```diff
@@ -17,25 +15,25 @@ use tokio_test::assert_ok;
 
 #[tokio::test]
 async fn echo_server() {
-    const ITER: usize = 1024;
+    const BYTES: &[u8] = b"foo bar baz";
+    const ITER: usize = 32;
```
ADD-SP, line 22: We lowered the `ITER` from 1000 to 32, is is because the Miri tests are slow?

WhySoBad: Yes exactly. We could also use `cfg` attributes to only change the iteration count for Miri.

Darksonn: Let's do the cfg.

ADD-SP (approved):
Thanks!

PR #8157, opened 2026-05-20
Remove panicking operations on Instant

marshallpierce:
Since there have been numerous times where Instant + Duration or similar has caused overflow/panics, I removed all panicking operations.

This is a continuation of the discussion in #8152.

## Motivation

Rather than pondering a hypothetical world without panicking operations on Instant in #8152, I made this to be a concrete instance of that.

## Solution

Test code uses `checked_*(...).unwrap()`, and main code uses saturating
ops, or expect() where that made more sense. In a couple of instances, logic is rearranged to better fit this new world order, but no behavior was (intentionally) changed other than saturating instead of panicking.

### DIFF
--- a/tokio-stream/Cargo.toml
+++ b/tokio-stream/Cargo.toml
@@ -38,14 +38,14 @@ signal = ["tokio/signal"]
 [dependencies]
 futures-core = { version = "0.3.0" }
 pin-project-lite = "0.2.11"
-tokio = { version = "1.38.0", features = ["sync"] }
+tokio = { version = "1.38.0", features = ["sync"], path = "../tokio" }
 tokio-util = { version = "0.7.0", optional = true }
 
 [dev-dependencies]
-tokio = { version = "1.38.0", features = ["full", "test-util"] }
+tokio = { version = "1.38.0", features = ["full", "test-util"], path = "../tokio" }
 async-stream = "0.3"
 parking_lot = "0.12.0"
-tokio-test = "0.4"
+tokio-test = { version = "0.4", path = "../tokio-test" }
 futures = { version = "0.3", default-features = false }
 
 [package.metadata.docs.rs]
--- a/tokio-util/src/time/delay_queue.rs
+++ b/tokio-util/src/time/delay_queue.rs
@@ -418,7 +418,7 @@ struct Data<T> {
     /// instant.
     inner: T,
 
-    /// The instant at which the item is returned.
+    /// The offset in ms at which the item is returned.
     when: u64,
 
     /// Set to true when stored in the `expired` queue
@@ -505,10 +505,6 @@ impl<T> DelayQueue<T> {
     ///
     /// See [type] level documentation for more details.
     ///
-    /// # Panics
-    ///
-    /// This function panics if `when` is too far in the future.
-    ///
     /// # Examples
     ///
     /// Basic usage
@@ -521,7 +517,7 @@ impl<T> DelayQueue<T> {
     /// # async fn main() {
     /// let mut delay_queue = DelayQueue::new();
     /// let key = delay_queue.insert_at(
-    ///     "foo", Instant::now() + Duration::from_secs(5));
+    ///     "foo", Instant::now().checked_add(Duration::from_secs(5)).unwrap());
     ///
     /// // Remove the entry
     /// let item = delay_queue.remove(&key);
@@ -565,7 +561,7 @@ impl<T> DelayQueue<T> {
                 waker.wake();
             }
 
-            let delay_time = self.start + Duration::from_millis(when);
+            let delay_time = self.start.saturating_add(Duration::from_millis(when));
             if let Some(ref mut delay) = &mut self.delay {
                 delay.as_mut().reset(delay_time);
             } else {
@@ -598,7 +594,7 @@ impl<T> DelayQueue<T> {
             Expired {
                 key,
                 data: data.inner,
-                deadline: self.start + Duration::from_millis(data.when),
+                deadline: self.start.saturating_add(Duration::from_millis(data.when)),
             }
         }))
     }
@@ -655,7 +651,7 @@ impl<T> DelayQueue<T> {
     /// [type]: #
     #[track_caller]
     pub fn insert(&mut self, value: T, timeout: Duration) -> Key {
-        self.insert_at(value, Instant::now() + timeout)
+        self.insert_at(value, Instant::now().saturating_add(timeout))
     }
 
     #[track_caller]
@@ -704,7 +700,8 @@ impl<T> DelayQueue<T> {
     /// ```
     #[track_caller]
     pub fn deadline(&self, key: &Key) -> Instant {
-        self.start + Duration::from_millis(self.slab[*key].when)
+        self.start
+            .saturating_add(Duration::from_millis(self.slab[*key].when))
     }
 
     /// Removes the key from the expired queue or the timer wheel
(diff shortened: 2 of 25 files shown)


marshallpierce:
This PR is in a catch-22: I can't add new behavior in `tokio` and use it from `tokio-*` because the clippy CI job removes the manifest `patch`. However, if I _don't_ use the new features (e.g. `saturating_add`), then _building_ fails because `cargo build` _does_` use the `patch`.

Darksonn:
Hm, the clippy job isn't supposed to compile the `tokio` crate.

Darksonn:
You have to add a path dependency in the Cargo.toml
See docs/contributing/how-to-specify-crates-dependencies-versions.md

marshallpierce:
Ah, I see, thanks. I thought the patch in the root Cargo.toml would do it.

tokio-stream/Cargo.toml
```diff
@@ -38,14 +38,14 @@ signal = ["tokio/signal"]
 [dependencies]
 futures-core = { version = "0.3.0" }
 pin-project-lite = "0.2.11"
-tokio = { version = "1.38.0", features = ["sync"] }
+tokio = { version = "1.38.0", features = ["sync"], path = "../tokio" }
```
Darksonn, line 41: What are you changing in the core tokio crate that requires this dependency?

tokio/src/time/instant.rs
```diff
@@ -162,42 +177,6 @@ impl From<Instant> for std::time::Instant {
     }
 }
 
-impl ops::Add<Duration> for Instant {
```
martin-g, line 165: Tokio is very strict about backwards compatibility.
These impls will have to stay

tokio/src/runtime/time/source.rs
```diff
@@ -16,7 +16,10 @@ impl TimeSource {
 
     pub(crate) fn deadline_to_tick(&self, t: Instant) -> u64 {
         // Round up to the end of a ms
-        self.instant_to_tick(t + Duration::from_nanos(999_999))
+        self.instant_to_tick(
+            t.checked_add(Duration::from_nanos(999_999))
```
martin-g: Is there a drawback of using `saturating_add()` here ?

tokio/src/time/interval.rs
```diff
@@ -475,7 +486,11 @@ impl Interval {
         // However, if a tick took excessively long and we are now behind,
         // schedule the next tick according to how the user specified with
         // `MissedTickBehavior`
-        let next = if now > timeout + Duration::from_millis(5) {
+        let next = if now
+            > timeout
+                .checked_add(Duration::from_millis(5))
```
martin-g: What if `timeout` is 3ms before `far_future` ?
Wouldn't it be safer to use `saturating_add()` instead ?

tokio/src/time/interval.rs
martin-g:
```suggestion
    /// This is equivalent to calling `reset_at(Instant::now().saturating_add(period))`.
```

martin-g:
```suggestion
    /// This is equivalent to calling `reset_at(Instant::now().saturating_add(after))`.
```

tokio-util/src/time/delay_queue.rs
```diff
@@ -624,11 +620,6 @@ impl<T> DelayQueue<T> {
     ///
     /// See [type] level documentation for more details.
     ///
-    /// # Panics
-    ///
-    /// This function panics if `timeout` is greater than the maximum
-    /// duration supported by the timer in the current `Runtime`.
```
martin-g, line 630: This is still valid.
See https://github.com/tokio-rs/tokio/blob/39d0cb9b99e64f0e49fe572ea6550bbb25626b91/tokio-util/src/time/wheel/mod.rs#L48

tokio/src/time/instant.rs
```diff
@@ -148,6 +156,13 @@ impl Instant {
     pub fn checked_sub(&self, duration: Duration) -> Option<Instant> {
         self.std.checked_sub(duration).map(Instant::from_std)
     }
+
+    /// Returns `self` advanced by `duration`, unless that would overflow the underlying
+    /// representation used by `Instant`, in which case an `Instant` decades in the future is
+    /// returned.
```
martin-g: `tokio::time::Instant::from_std()` allows to create an instance that is beyond `far_future` (i.e. more than 30 years in the future). In that case saturating_add() will actually return a result that is in the past.
Worth documenting it.

tokio/src/time/instant.rs
Darksonn: cc @obi1kenobi any ideas why this didn't trigger a semver check failure?

marshallpierce: Understood. Perhaps a way to do this would be to have a `pub(crate) struct CheckedInstant` that is used for internal calculations, and converted to `tokio::time::Instant` on public API boundaries?

tokio-stream/Cargo.toml
marshallpierce: I kept getting issues in the `tokio-` crates in the clippy CI check where different versions of tokio were used, leading to type mismatches. After a few rounds of not getting it right, I sprinkled `path` in a number of places and that did the trick. With some time to investigate, a few of these can probably be pulled back.

tokio/src/time/instant.rs
Darksonn: I think ideally you just remove the uses of panicking operations and do not change the type.

tokio-util/src/time/delay_queue.rs
marshallpierce: Restored.

tokio/src/time/interval.rs
marshallpierce: Fair point, fixed.

tokio/src/runtime/time/source.rs
marshallpierce: I suppose not. It's _implausible_ but I guess if we're aiming for safety we might as well go all the way.

tokio/src/time/instant.rs
marshallpierce: Now, we find the precise max Instant once at startup, so we don't have to be vague in the comments.

tokio/src/time/instant.rs
marshallpierce: Now the panicking ops are back.

tokio/tests/time_sleep.rs
```diff
@@ -265,6 +279,7 @@ async fn reset_after_firing() {
 
 #[tokio::test]
 async fn exactly_max() {
+    panic!("TODO: this sleeps forever...");
```
marshallpierce, line 282: What should be done with this and the test below? It's not clear what the intent of the tests are.

Darksonn: This test verifies that Tokio's auto-advance feature works correctly with paused time even if the duration is very large.

marshallpierce: I've been taking a peek at why these tests are broken by these changes, and so far what I've figured out is that CoreGuard::block_on busy-loops forever, whereas if I make Instant::max be now + 10 years, it only loops a handful of times. I'll keep digging next week, but if anyone has a suggestion of where to look, that would hasten the search.

Darksonn: Sorry I don't know.

Darksonn:
Please remove the draft marker when this is ready for review.

PR #8159, opened 2026-05-20
io: advance partially written buffers correctly in `write_all_vectored`

### LINKED ISSUE #8158: tokio-util write_all_vectored can drop data after partial vectored writes
### Description

`tokio_util::io::write_all_vectored` can advance the provided `IoSlice` list incorrectly after a partial `poll_write_vectored` result. The local `advance_slices` helper removes fully written buffers, then re-slices the new first buffer with `&first[..left]`. At that point, `left` is the number of bytes already consumed from that buffer, so the helper keeps the written prefix instead of the unwritten suffix.

### Example

For buffers `[b"ab", b"cd"]`, if `poll_write_vectored` returns `2`, the helper removes the first buffer and then replaces `b"cd"` with `b"cd"[..0]`. The future then sees no non-empty buffers and returns `Ok(())`, silently dropping `b"cd"`.

For buffers `[b"hello", b"world"]`, if `poll_write_vectored` returns `3`, the helper resumes from `b"hel"` instead of `b"lo"`, so subsequent writes can duplicate already-written bytes and skip unwritten bytes.

### Expected behavior

After a partial vectored write, `write_all_vectored` should continue from the unwritten suffix of the current buffer, and should continue with the next buffer when the write count lands exactly on a buffer boundary.

### Suggested fix

Change the helper to retain `&first[left..]`, and add regression coverage for a partial write inside a buffer and an exact buffer-boundary write.

fallintoplace:
Fixes #8158.

## Summary

- Correct `advance_slices` to retain the unwritten suffix after a partial vectored write.
- Add regression coverage for a partial write inside a buffer.
- Add regression coverage for a partial write that lands exactly on a buffer boundary.

## Root cause

The local `advance_slices` helper removes fully written buffers, then adjusts the new first `IoSlice`. At that point, `left` is the number of bytes already consumed from that first remaining buffer, so the helper should keep `&first[left..]`. Keeping `&first[..left]` can duplicate already-written bytes or drop unwritten bytes.

## Tests

- `cargo fmt --check`
- `cargo test -p tokio-util --test io_write_all_vectored --features full`
- `cargo test -p tokio-util --features full`
- `cargo clippy -p tokio-util --test io_write_all_vectored --features full -- -D warnings`

### DIFF
--- a/tokio-util/src/io/write_all_vectored.rs
+++ b/tokio-util/src/io/write_all_vectored.rs
@@ -147,7 +147,7 @@ fn advance_slices<'a>(bufs: &mut &mut [IoSlice<'a>], n: usize) {
 
     *bufs = &mut std::mem::take(bufs)[remove..];
     if let Some(first) = bufs.first_mut() {
-        let buf = &first[..left];
+        let buf = &first[left..];
         // necessary due to limitating in the borrow checker,
         // when tokio MSRV reaches 1.81.0 this entire function
         // can be replaced with `IoSlice::advance_slices`
--- a/tokio-util/tests/io_write_all_vectored.rs
+++ b/tokio-util/tests/io_write_all_vectored.rs
@@ -10,6 +10,55 @@ use std::io::IoSlice;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
+struct PartialVectoredWriter {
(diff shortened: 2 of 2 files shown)


ADD-SP (approved):
Thanks!

PR #8160, opened 2026-05-20
io: fix write_all_vectored dropping data after partial writes

### LINKED ISSUE #8158: tokio-util write_all_vectored can drop data after partial vectored writes

LeSingh1:
## Motivation

`tokio_util::io::write_all_vectored` can silently drop or duplicate bytes after a partial `poll_write_vectored` result, because the local `advance_slices` helper re-slices the first remaining buffer with `&first[..left]` instead of `&first[left..]`. At that point, `left` is the number of bytes already consumed from that buffer, so the helper keeps the written prefix rather than the unwritten suffix.

For `[b"ab", b"cd"]` with a write count of 2, the helper removes the first buffer and then replaces `b"cd"` with `b"cd"[..0]`. The future sees no non-empty buffers and returns `Ok(())`, silently dropping `b"cd"`. For `[b"hello", b"world"]` with a write count of 3, the helper resumes from `b"hel"` instead of `b"lo"`, duplicating already-written bytes and skipping unwritten bytes.

## Solution

Change the helper to slice the first remaining buffer as `&first[left..]`, matching the behavior of `std::io::IoSlice::advance_slices` (which will replace this helper once the MSRV reaches 1.81.0).

The new `write_all_vectored_partial_writes` test exercises three scenarios that all fail on the current code: an exact buffer-boundary write that drops the next buffer, a partial write inside a buffer that duplicates bytes, and a partial write that spans into the second buffer.

Closes #8158.

### DIFF
--- a/tokio-util/src/io/write_all_vectored.rs
+++ b/tokio-util/src/io/write_all_vectored.rs
@@ -147,7 +147,7 @@ fn advance_slices<'a>(bufs: &mut &mut [IoSlice<'a>], n: usize) {
 
     *bufs = &mut std::mem::take(bufs)[remove..];
     if let Some(first) = bufs.first_mut() {
-        let buf = &first[..left];
+        let buf = &first[left..];
         // necessary due to limitating in the borrow checker,
         // when tokio MSRV reaches 1.81.0 this entire function
         // can be replaced with `IoSlice::advance_slices`
--- a/tokio-util/tests/io_write_all_vectored.rs
+++ b/tokio-util/tests/io_write_all_vectored.rs
@@ -140,3 +140,85 @@ async fn write_all_vectored_with_empty_slice() {
     write_all_vectored(&mut wr, buf).await.unwrap();
     assert_eq!(&wr.buf[..], b"hello");
 }
+
+#[tokio::test]
+async fn write_all_vectored_partial_writes() {
+    // Writer that reports writing only `limit` bytes per call, exercising the
+    // partial write path inside `write_all_vectored`. This is the path that
+    // previously dropped or duplicated bytes when `advance_slices` advanced the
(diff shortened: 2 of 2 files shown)


mattiapitossi:
Thanks for the contribution! The original reported of this issue has already opened a [PR](https://github.com/tokio-rs/tokio/pull/8159) addressing this, so we'll continue the discussion there

LeSingh1:
Thanks @mattiapitossi, makes sense to close in favour of #8159 since the reporter opened it first. Closing.

PR #8161, opened 2026-05-21
Async scoped tasks proposal

RedKinda:
## Motivation

Have an equivalent of `std::thread::scope` in tokio. This PR is a followup on https://github.com/tokio-rs/tokio/issues/3162 and provides a potential solution for an ergonomic API. From [this blog](https://without.boats/blog/the-scoped-task-trilemma/) this PR would fall into the `Concurrency + Parallelizability` category, disallowing direct borrowing. It is additionally different both from `async_scoped` because it provides a safe API, and also from `tokio_scoped` which blocks the worker thread.

## Solution

The main thing this PR does is take ownership of shared resources and store them inside the tokio runtime. The scoped task then receives two parameters, `Pin<&mut R>` of the user provided resource, and `Scope` which enables spawning scoped tasks. Then, once the scope closes (and enforces all scoped tasks are finished), the resource ownership is sent back to the user through the JoinHandle. This ownership passing is done due to the fact that runtime might be dropped in the middle of scoped operation, in which case the tasks holding the references will be dropped first, and the actual data will be dropped later, to prevent use-after-frees.

Example use:
```rs
    let data = vec!["1", "2", "3", "4"];

    let scope_handle = spawn_scoped(data, |scope, data_inner| Box::pin(async move {
        let mut tasks = vec![];
        for i in data_inner.iter() {
            let task = scope.spawn(async {
                i.parse::<i32>().unwrap()
            }).await;

            tasks.push(task);
        }

        let mut sum = 0;
        
        for task in tasks {
            sum += task.await.unwrap();
        }

        sum
    }));
```

Known issues:
- `scope.spawn` is async, because it needs to immediately do `rwlock.read().await`. Not sure if this is resolvable, maybe a while loop of `try_read`s is good enough.
- `tokio::spawn_scoped` takes a `FnOnce -> Pin<Box<dyn Future>>` - there has got to be a way to not need this and use generics instead, but I couldn't figure it out.
- possibly inappropriate use of Pin - I am not 100% sure if the Pin is required to be used at all when giving user access to the resource
- more tests - there are some tests (passing) but given this touches a bunch of unsafe and lifetimes it would be very beneficial to include more tests, probably also with loom
- missing documentation, add example use

I am aware this topic has been quite controversial and every solution has drawbacks, this PR is a potential solution for a good enough safe API to do safe scopes, and there might be more issues that I am not aware of. I did consider making this an external crate, but given it changes tokio runtime structs, it was not possible to do in a very nice way.

### DIFF
--- a/tokio/Cargo.toml
+++ b/tokio/Cargo.toml
@@ -41,6 +41,7 @@ full = [
   "time",
 ]
 
+async-scope = ["rt"]
 fs = []
 io-util = ["bytes"]
 # stdin, stdout, stderr
--- a/tokio/src/macros/cfg.rs
+++ b/tokio/src/macros/cfg.rs
@@ -746,3 +746,16 @@ macro_rules! cfg_io_uring {
         )*
     };
 }
+
+macro_rules! cfg_async_scope {
+    ($($item:item)*) => {
+        $(
+            #[cfg(all(
+                tokio_unstable,
+                feature = "async-scope",
+                feature = "rt",
+            ))]
+            $item
+        )*
+    };
+}
--- a/tokio/src/runtime/driver.rs
+++ b/tokio/src/runtime/driver.rs
@@ -29,6 +29,8 @@ pub(crate) struct Handle {
     /// Time driver handle
     pub(crate) time: TimeHandle,
 
+    pub(crate) scope: ScopeDriver,
+
     /// Source of `Instant::now()`
     #[cfg_attr(not(all(feature = "time", feature = "test-util")), allow(dead_code))]
     pub(crate) clock: Clock,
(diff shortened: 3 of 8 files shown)


Darksonn:
But you can already do this, right?
```rs
let data = ...;
    // data moved here
    let ret = use_data(&mut data).await;
    // now move data back
    (data, ret)
});
let (data, ret) = jh.await;
```

RedKinda:
The idea is that the data provided as the spawn_scoped parameter can be shared between scoped tasks as an immutable ref without any synchronization/Arc-s 

Also realized current impl allows for unsoundness with the way lifetimes are defined, will work on that.

Darksonn:
My gut reaction is that this is out of scope and belongs in a third-party library. If Rust ever adds support that lets us do real scoped tasks, then we can add that, but for now, I don't really want to add it.

carllerche:
Yes, I agree that this should be explored out of tree. If there are any missing hooks / primitives to support building this, we can talk about adding those.

Darksonn:
I'm going to close this, but regardless thanks for the PR!

PR #8168, opened 2026-05-24
docs(time): clarify Sleep behavior when polled after completion

### LINKED ISSUE #8073: Is `tokio::time::Sleep` fused?
Is the `Future` implementation of the `tokio::time::Sleep` type fused? I.e., does it return `Poll::Ready(())` again if polled again after completion? Or is that just an implementation detail and not guaranteed?

The reason I'm asking is because the docs for the `select` macro have some examples with `select` branches matching on `&mut sleep` in a loop. For example, this one:

```rust
let sleep = time::sleep(Duration::from_millis(50));
tokio::pin!(sleep);

loop {
    tokio::select! {
        _ = &mut sleep => {
            println!("operation timed out");
            break;
        }
        _ = some_async_work() => {
            println!("operation completed");
        }
    }
}
```

There is no issue here because of the `break` in the branch. But imagine that you want to change this to a _"things are taking longer than usual"_ message and no longer abort the async work. You might think that removing the `break` is enough to achieve this. However, if `Sleep` is not guaranteed to be fused, this might lead to a panic in future tokio versions. So you need a proper `if` guard (which is difficult to write correctly) or you need to bring the `futures` crate in and use its `fuse` method.

Either way, I think it would be good to be explicit in the docs about this. Either warn about fuse semantics in the docs for the `select` macro or document that the `Sleep` type is safe to poll again after completion.

zrbecker:
## Motivation

Closes #8073. The behavior of polling a `Sleep` after it has resolved is not currently documented, leaving users uncertain whether reusing the same `Sleep` in a `select!` loop is safe.

## Solution

Add a "Polling after completion" section to the `Sleep` docs stating that polling a completed `Sleep` does not panic and that the timer remains elapsed until `reset` is called. Add a regression test in `tokio/tests/time_sleep.rs`, and mirror it in `tokio/tests/time_alt.rs` so both the default and alternative timer implementations are covered.

### DIFF
--- a/tokio/src/time/sleep.rs
+++ b/tokio/src/time/sleep.rs
@@ -217,12 +217,8 @@ pin_project! {
     ///
     /// # Polling after completion
     ///
-    /// `Sleep` is safe to poll after completion. Polling a completed
-    /// `Sleep` does not panic, and the timer remains elapsed until you
-    /// [`reset`](Sleep::reset) it, so a completed `Sleep` may remain as
-    /// a branch of a [`select!`] loop. A given `poll` may still return
-    /// `Poll::Pending` for runtime reasons such as cooperative
-    /// scheduling, but this does not change the timer's state.
+    /// `Sleep` is safe to poll after completion. Polling a completed `Sleep` does
+    /// not panic and is idempotent, but may spuriously return `Poll::Pending`.
     ///
     /// [`select!`]: ../macro.select.html
     /// [`tokio::pin!`]: ../macro.pin.html
--- a/tokio/src/time/sleep.rs
+++ b/tokio/src/time/sleep.rs
@@ -136,6 +136,9 @@ pin_project! {
     /// use it with [`select!`] or by calling `poll`, you have to pin it first.
     /// If you use it with `.await`, this does not apply.
     ///
+    /// `Sleep` is safe to poll after completion. Polling a completed `Sleep` does
+    /// not panic and is idempotent, but may spuriously return `Poll::Pending`.
+    ///
     /// # Examples
     ///
     /// Wait 100ms and print "100 ms have elapsed".
@@ -214,11 +217,6 @@ pin_project! {
     ///     }
     /// }
     /// ```
-    ///
-    /// # Polling after completion
-    ///
-    /// `Sleep` is safe to poll after completion. Polling a completed `Sleep` does
-    /// not panic and is idempotent, but may spuriously return `Poll::Pending`.
     ///
     /// [`select!`]: ../macro.select.html
     /// [`tokio::pin!`]: ../macro.pin.html
--- a/tokio/src/time/sleep.rs
+++ b/tokio/src/time/sleep.rs
@@ -215,6 +215,15 @@ pin_project! {
     /// }
     /// ```
     ///
+    /// # Polling after completion
+    ///
+    /// `Sleep` is safe to poll after completion. Polling a completed
+    /// `Sleep` does not panic, and the timer remains elapsed until you
+    /// [`reset`](Sleep::reset) it, so a completed `Sleep` may remain as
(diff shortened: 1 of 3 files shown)


tokio/src/time/sleep.rs
```diff
@@ -215,6 +215,15 @@ pin_project! {
...
     /// ```
     ///
+    /// # Polling after completion
+    ///
+    /// `Sleep` is safe to poll after completion. Polling a completed
+    /// `Sleep` does not panic, and the timer remains elapsed until you
+    /// [`reset`](Sleep::reset) it, so a completed `Sleep` may remain as
+    /// a branch of a [`select!`] loop. A given `poll` may still return
+    /// `Poll::Pending` for runtime reasons such as cooperative
+    /// scheduling, but this does not change the timer's state.
```
vilgotf: This section should be above the example section (which is usually the last section). This PR should also coordinate with #8137 (which makes Sleep Fused).

I enjoy technical writing and have an alternative suggestion, but please push back if it is too dense.

```suggestion
    /// # Polling after completion
    ///
    /// `Sleep` is safe to poll after completion. Polling a completed `Sleep` does
    /// not panic and is idempotent, but may spuriously return `Poll::Pending`.
```

## Notes

> [...], and the timer remains elapsed until you `reset` it [...]

Exhaustively listing non-changing properties (currently `deadline` & `elapsed`) can be more concisely expressed as the future being idempotent. Interactions with `reset` are better put under its own method (which I think is already there, but could be improved).

> [...], so a completed `Sleep` may remain as a branch of a `select!` loop.

Exemplifying about `select` is not bad per se, but is not something that's usually mentioned.

> A given `poll` may still return `Poll::Pending` for runtime reasons such as cooperative scheduling, but this does not change the timer's state.

Internal details are easily made out-of-date. I don't think it's that useful to know exactly why `Pending` may be returned (e.g. atomic ordering reasons)

tokio/src/time/sleep.rs
```diff
@@ -215,6 +215,11 @@ pin_project! {
     /// }
     /// ```
     ///
+    /// # Polling after completion
```
vilgotf: I think you forgot to move this above `# Examples`. I think the header is not necessary. The `Unpin` paragraph does not use a header, for example.

zrbecker: Thanks, just accepted the suggested edit via Github's UI and pulled onto my machine to do the move. Should be updated now.

tokio/src/time/sleep.rs
zrbecker: Moved above `# Examples` and adopted your wording. Agreed on all three notes.

On #8137 coordination, happy to defer to whatever fits your timeline. Should we (1) Land this PR with the hedged "may spuriously return Pending" wording, and #8137 drops the hedge, (2) Hold this PR until #8137 lands, and update the wording here?

vilgotf: Option 1 is likely the better option. I foresee my PR requiring a few rounds of reviews before landing.

tokio/src/time/sleep.rs
```diff
@@ -136,6 +136,9 @@ pin_project! {
     /// use it with [`select!`] or by calling `poll`, you have to pin it first.
     /// If you use it with `.await`, this does not apply.
     ///
+    /// `Sleep` is safe to poll after completion. Polling a completed `Sleep` does
+    /// not panic and is idempotent, but may spuriously return `Poll::Pending`.
```
Darksonn, line 140: I think it would be useful to clarify that if it spuriously returns `Pending`, then a wakeup will actually be emitted. Perhaps we could say something along the lines of "but may return `Poll::Pending` if [the coop budget](https://docs.rs/tokio/latest/tokio/task/coop/index.html) or other similar Tokio features apply".

PR #8169, opened 2026-05-25
time: ensure timers stay in the same runtime after `.reset()`

ADD-SP:
## Motivation

It is possible that a timer is `.reset()`ed in a runtime that is different from the runtime that the timer is created in. However, currently, there is no compile-time guarantee to keep this invariant.

Fix the regression introduced by <https://github.com/tokio-rs/tokio/pull/8132>.

## Solution

* Check if we are in the same runtime during the timer registration.
* Add a regression test.

### DIFF
--- a/tokio/src/runtime/time_alt/timer.rs
+++ b/tokio/src/runtime/time_alt/timer.rs
@@ -29,7 +29,7 @@ impl Timer {
     #[track_caller]
     pub(crate) fn new(handle: scheduler::Handle, deadline: Instant) -> Self {
         let tick = deadline_to_tick(&handle, deadline);
-        let entry = with_current_temp_local_context(|ctx| match ctx {
+        let entry = with_current_temp_local_context(&handle, |ctx| match ctx {
             Some(TempLocalContext::Running { registration_queue }) => {
                 let entry = EntryHandle::new(tick);
                 unsafe { registration_queue.push_front(entry.clone()) }
@@ -57,7 +57,7 @@ impl Timer {
     }
 }
 
-fn with_current_temp_local_context<F, R>(f: F) -> R
+fn with_current_temp_local_context<F, R>(sched_hdl: &scheduler::Handle, f: F) -> R
 where
     F: FnOnce(Option<TempLocalContext<'_>>) -> R,
 {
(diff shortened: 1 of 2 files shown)


ADD-SP:
cc @vilgotf

vilgotf (approved):
Right, this is needed again.

tokio/src/runtime/time_alt/timer.rs
```diff
@@ -69,8 +69,38 @@ where
...
+        // so we need to check it at runtime.
+        let is_same_rt = context::with_current(|cur_sched_hdl| {
+            use crate::runtime::scheduler::Handle;
+
+            match (sched_hdl, cur_sched_hdl) {
+                (Handle::CurrentThread(_), _) => {
+                    panic!("alternative timer is not supported in the current-thread runtime")
+                }
+                (_, Handle::CurrentThread(_)) => {
+                    panic!("alternative timer is not supported in the current-thread runtime")
```
martin-g: Does this need to panic ?

```suggestion
                    false
```

Returning `false` seems to be enough for a test similar to the added one below:

```rust
#[test]
/// It is possible that a timer is created in one runtime,
/// but `.reset()` is called in a different runtime.
/// In this case, the timer should be registered in the original runtime.
fn reset_should_stay_on_same_runtime2() {
    let rt1 = tokio::runtime::Builder::new_multi_thread()
        .worker_threads(1)
        .enable_alt_timer()
        .build()
        .unwrap();

    let rt2 = tokio::runtime::Builder::new_current_thread()
        .build()
        .unwrap();

    // Register the timer into the local timer wheel of `rt1`.
    //
    // We cannot use bare `rt1.block_on` as the local timer wheel of `rt1`
    // is only accessible from the worker threads of `rt1`,
    // but `rt1.block_on` runs the future on the current thread.
    // So we need to use `rt1.spawn` to run the future on the worker thread.
    let sleep = rt1
        .block_on(
            #[allow(clippy::async_yields_async)]
            rt1.spawn(async {
                // 1 hour is long enough to make sure the timer is not fired before we call `reset()`.
                tokio::time::sleep(Duration::from_secs(3600))
            }),
        )
        .unwrap();
    let mut sleep = Box::pin(sleep);
    assert_pending!(sleep.as_mut().poll(&mut noop_context()));

    // reset the timer created from `rt1` in `rt2`,
    // which should register the timer into the local timer wheel of `rt1`.
    let sleep = rt2
        .block_on({
            #[allow(clippy::async_yields_async)]
            rt2.spawn(async move {
                sleep
                    .as_mut()
                    .reset(Instant::now() + Duration::from_secs(3600));
                sleep
            })
        })
        .unwrap();

    // drop `rt1` to fire all timers registered in `rt1`,
    // including the timer we just reset.
    drop(rt1);

    // If this assertion fails, it means the timer is not registered in `rt1`.
    // This can happen if the timer is registered in `rt2` instead of `rt1`,
    assert!(sleep.now_or_never().is_some());
}
```

Note: I've changed only `let rt2 = tokio::runtime::Builder::new_current_thread().build().unwrap();` from the original test! 
If the suggestion is going to be accepted then the test name and the comments should be updated. If not, then please a test that should panic.

ADD-SP: The alternative timer is only available for multi-thread runtime, so the `panic!` path is actually `unreachable!`. I don't think we need a regress test for it. But probably we should use `unreachable!` here instead of `panic!`.

ADD-SP: Changed to `unreachable!`: <https://github.com/tokio-rs/tokio/pull/8169/commits/c7d963e973358f06ab9e713df761fa84dfdb9aef>

ADD-SP: Ah, did you mean the second match arm? If so, I think you are right (thanks for the confusing GitHub UI).

Added in <https://github.com/tokio-rs/tokio/pull/8169/commits/ada80941551fa3c269b177653440086f99e23fa6>.

martin-g (approved):

